Issue #240, opened 2014-11-24
Tag Helpers: Map PascalCase type & property names to HTML element & attribute dashed equivalents


DamianEdwards:
By default (convention), Tag Helper type and property names should map to their HTML element attribute equivalents such that pascal-cased names are converted to dashed-names automatically. This can of course be overridden with the `HtmlElementName` and `HtmlAttributeName` attributes.

e.g.
```
"public class MyTagHelper : TagHelper" maps to "<my-tag-helper>"
"public string MyProperty { get; set; }" maps to "my-property='blah'"
```

NTaylorMullen:
2ad828f5bc4b04dcc76403d0c969a70713b31be5
8c47f6a67ecd3da91635c251770a6b872ee33f0b

PR #262, opened 2014-12-23
Tag Helpers: Change tag and attribute names to be kebab-cased.

NTaylorMullen:
- Lower kebab-cased casing is the HTML convention.
- Modified tests to account for new tag and attribute names.
- Also added additional tests to validate new lower snake casing.

#240

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -97,5 +106,22 @@ private static bool IsValidProperty(PropertyInfo property)
                    property.SetMethod != null &&
                    property.SetMethod.IsPublic;
         }
+
+        /// <summary>
+        /// Converts from pascal/camel case to lower kebab-case.
+        /// </summary>
+        /// <example>
+        /// SomeThing => some-thing
+        /// capsONInside => caps-ON-Inside
```
NTaylorMullen: Fail comment is all, it should all be lowercased.

Issue #129, opened 2014-09-08
Tag Helpers: Ignore '@' symbols in Tag Helper attributes if not string or character.


NTaylorMullen:
The typical flow for a Razor developer is to write `@Foo` when they want to switch from HTML to C# code.  Now that we're dealing with HTML attributes that could potentially be projected into a C# context we need to consider the impact.

Do we ignore `@` symbols in C# projected HTML attributes?

AKA:
```C#
public class CurrentAgeTagHelper : TagHelper
{
    public int Age { get; set; }
    ...
}
```

```
@{
    myAge = "3";
}

<currentAge age="@myAge" />
<currentAge age="myAge" /> @* This has equivalent output to the line above *@
<currentAge age="3" /> @* This has equivalent output to the line above *@
```

dougbu:
will fix together with https://github.com/aspnet/Mvc/issues/1253

dougbu:
Fixed with 26afdbd88959c94c17755f1ae7ce4add4fbe8489 through de4cafa8cdbe357ca3c4a34b084e3cf09c0db231

PR #268, opened 2015-01-12
Handle complex expressions for bound non-string tag helper attributes

dougbu:
- #129 and aspnet/Mvc#1253
- add new `CSharpTagHelperAttributeValueVisitor` that writes the raw expression
- update existing tests to include new line pragmas and updated line mappings
- add tests of embedded `@(...)` and mix that with normal expressions

nits:
- cleanup long lines in `CSharpTagHelperCodeRenderer`
- ease debugging of `LineMapping` values
- improve assertion failures about code mapping miss-matches
- add `GENERATE_BASELINES` reminder to test project.json


src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <summary>
+        /// Writes code for the given <paramref name="chunk"/>.
+        /// </summary>
+        /// <param name="chunk">The <see cref="TemplateChunk"/> to render.</param>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown unconditionally to inform user of unexpected template e.g. <c>@&lt;p&gt;paragraph@&lt;/p&gt;</c>.
+        /// </exception>
+        protected override void Visit(TemplateChunk chunk)
+        {
+            throw new InvalidOperationException(RazorResources.ParseError_InlineMarkup_Blocks_Cannot_Be_Nested);
```
dougbu: - plan to create new resources or update the wording of these (if unused) in next iteration
- @NTaylorMullen is another error notification mechanism available this late in processing?

NTaylorMullen: Not that we expose. We'll need to flow the shared `ParserErrorSink` through, this will be the first bit of CodeGeneration that generates errors in "expected" cases.

dougbu: :+1: for now I'm going to just `throw`.  later we can plumb `ParserErrorSink` through.

another possibility for the future would be to update `TagHelperBlockBuilder` (which already has access to the `ParserErrorSink`) to prevent these `Chunk` or corresponding `SyntaxTreeNode` types within a bound non-`string` attribute value.

dougbu:
Updated - added resources and removed a few unused ones

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
NTaylorMullen: This can't go in while it's throwing sadly, it'll crash VS in the editor.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+    /// <remarks>
+    /// Since attribute value is not written out as HTML, does not emit instrumentation. Further this
+    /// <see cref="CodeVisitor{CSharpCodeWriter}"/> writes identical code at design- and runtime.
+    /// </remarks>
+    public class CSharpTagHelperAttributeValueVisitor : CodeVisitor<CSharpCodeWriter>
+    {
+        private const string DefaultPseudoMethod = "default";
+
+        private string _attributeTypeName;
+        private bool firstChild;
```
NTaylorMullen: `_firstChild`

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <remarks>
+        /// <para>
+        /// Write a line pragma and the code for all children. Writes <c>default(property type)</c> if
+        /// <paramref name="chunk"/> has no children. Pads the first code block to align with source.
+        /// </para>
+        /// <para>Also tracks code mappings for all children.</para>
+        /// </remarks>
+        protected override void Visit(ChunkBlock chunk)
+        {
+            var currentIndent = Writer.CurrentIndent;
```
NTaylorMullen: Why capture the indent? The `CSharpLineMappingWriter` below takes care of indent management within its scope.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+        // Tracks the code mapping and writes code for a leaf node in the attribute value Chunk tree.
+        private void RenderCode(string code, Span association)
+        {
+            if (firstChild)
+            {
+                firstChild = false;
+
+                // Indent to match placement within line whether or not line starts with whitespace. Since last action
+                // was to write out line pragma with a final newline, Indent() call always indents.
+                Writer.Indent(association.Start.CharacterIndex);
```
NTaylorMullen: Still confused as to why you're unable to use the `CSharpPaddingBuilder`.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// </para>
+        /// <para>Also tracks code mappings for all children.</para>
+        /// </remarks>
+        protected override void Visit(ChunkBlock chunk)
+        {
+            var currentIndent = Writer.CurrentIndent;
+            try
+            {
+                // Entire attribute value is an expression. Write line number pragma for it.
+                using (var mappingWriter = new CSharpLineMappingWriter(Writer, chunk.Start, Context.SourceFile))
```
NTaylorMullen: Add a comment stating that this wont actually create a design time line mapping for the code below.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <param name="chunk">The <see cref="TemplateChunk"/> to render.</param>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown unconditionally to inform user of unexpected template e.g. <c>@&lt;p&gt;paragraph@&lt;/p&gt;</c>.
+        /// </exception>
+        protected override void Visit(TemplateChunk chunk)
+        {
+            throw new InvalidOperationException(
+                RazorResources.FormatTagHelpers_InlineMarkupBlocks_NotSupported_InAttributes(_attributeTypeName));
+        }
+
```
NTaylorMullen: Need to add two more invalid cases:

1. `SectionChunk`
1. `SetLayoutChunk` - you can also change this to be `topLevel: true` instead of special casing here, should have been done already.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: done

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+        // Tracks the code mapping and writes code for a leaf node in the attribute value Chunk tree.
+        private void RenderCode(string code, Span association)
```
NTaylorMullen: Don't need the whole association, just need the `SourceLocation`, aka `association.Start`

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: removed which also allowed me to clean up the `try` block

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: I'll try again but it previously seemed to consistently return an empty string in tested scenarios...

test/Microsoft.AspNet.Razor.Test/project.json
```diff
@@ -18,5 +18,8 @@
                 "System.Reflection.TypeExtensions": "4.0.0-beta-*"
             }
         }
+    },
+    "compilationOptions": {
+        "define": [ "__RemoveThisBitTo__GENERATE_BASELINES" ]
```
NTaylorMullen: hahaha :+1:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: will see if I can make it clear in a short comment that `RenderCode` captures line mappings for leaf `Chunk`s rather than doing one here.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: see methods starting at lines 131 and 144 in current iteration.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -0,0 +1,193 @@
...
+                // Entire attribute value is an expression. Write line number pragma for it.
+                using (var mappingWriter = new CSharpLineMappingWriter(Writer, chunk.Start, Context.SourceFile))
+                {
+                    firstChild = true;
+                    Accept(chunk.Children);
+
+                    if (firstChild)
+                    {
+                        // Attribute value was empty.
+                        Writer
```
NTaylorMullen: I'm not sure this is the write solution for an empty attribute because we dont want it to compile if they haven't provided anything. I'd say just do nothing, that way it doesn't compile and you'll get a reasonable C# error as to why it's not compiling. Also, doing `default(foo)` doesn't work when you have attribute types such as `Foo<T>` because the attribute type name comes through as `Foo'T`.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: need the `Span` when using `CSharpPaddingBuilder` :smiling_imp:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
NTaylorMullen: Goodness, need to get my eyes checked. :+1:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
NTaylorMullen: Hahah yes you do if you're able to use it :smile:

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpTagHelperAttributeValueVisitor.cs
dougbu: good point about generics.  removed the special case, which means `_attributeTypeName` is now used only in error messages.

dougbu:
@NTaylorMullen rebased and extensively updated

Third commit mentions responses to CR comments (I'll remove that part before pushing):
- do not `throw` in `CSharpTagHelperAttributeValueVisitor`; use `ParserErrorSink` instead
- add error if bound non-`string` attribute value is empty (don't handle)
- move `#line` pragma to surround entire bound attribute value assignment
 - need non-`null` `ChunkBlock.Association` value for this to work
- clean up some whitespace special-cases needed for previous approach

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -230,48 +230,67 @@ private void RenderTagHelpersCreation(TagHelperChunk chunk)
...
+                                attributeValueChunk.Association.Start,
+                                _context.SourceFile))
                             {
+                                // Place the assignment LHS to align RHS with original attribute value's indentation.
+                                // Unfortunately originalIndent is incorrect if original line contains tabs. Unable to
+                                // use a CSharpPaddingBuilder because the Association has no Previous node; lost the
+                                // original Span sequence when the parse tree was rewritten.
+                                var originalIndent = attributeValueChunk.Start.CharacterIndex;
+                                var generatedLength = valueAccessor.Length + " = ".Length;
+                                var newIndent = originalIndent - generatedLength;
```
NTaylorMullen: You need to padd by less so when a breakpoint is hit the yellow highlight evenly surrounds the value.

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
```diff
@@ -66,6 +66,7 @@ public override void GenerateStartBlockCode(Block target, CodeGeneratorContext c
 
                 attributes[attribute.Key] = new ChunkBlock
                 {
+                    Association = first?.Association,
```
NTaylorMullen: Why not `Association = target` ?

src/Microsoft.AspNet.Razor/ParserResults.cs
```diff
@@ -15,15 +17,22 @@ public class ParserResults
...
+        /// The <see cref="ParserErrorSink"/> used to collect <see cref="RazorError"/>s encountered when parsing the
+        /// current Razor document.
+        /// </param>
         public ParserResults([NotNull] Block document,
                              [NotNull] IEnumerable<TagHelperDescriptor> tagHelperDescriptors,
-                             [NotNull] IList<RazorError> parserErrors)
-            : this(parserErrors == null || parserErrors.Count == 0, document, tagHelperDescriptors, parserErrors)
+                             [NotNull] ParserErrorSink errorSink)
+            : this(
+                  errorSink.Errors == null || errorSink.Errors.Count() == 0,
```
NTaylorMullen: Same comment from other location.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: took my inspiration from `CSharpPaddingBuilder` and it doesn't reduce the padding as you describe.  could you point me to where we do less?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeValueCodeRenderer.cs
```diff
@@ -16,15 +16,23 @@ public class TagHelperAttributeValueCodeRenderer
...
+        /// </param>
         /// <param name="writer">The <see cref="CSharpCodeWriter"/> that's used to write code.</param>
         /// <param name="context">A <see cref="CodeGeneratorContext"/> instance that contains information about
         /// the current code generation process.</param>
-        /// <param name="renderAttributeValue"><see cref="Action"/> that renders the raw value of the HTML attribute.</param>
+        /// <param name="renderAttributeValue">
+        /// <see cref="Action"/> that renders the raw value of the HTML attribute.
+        /// </param>
+        /// <param name="complexValue">
+        /// Indicates whether or not the source attribute value contains more than simple text.
```
NTaylorMullen: Give an example.

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
dougbu: because I didn't notice `target` was a `Span`.  I'll see if that changes anything....

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Talked to pranav he said he did it here: `RenderRuntimeExpressionBlockChunkWithContentSpan`

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
NTaylorMullen: It's a block but Association just needs to be a Syntaxtreenode

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/ParserResults.cs
dougbu: :ok: 

raising the visibility of that commit comment:
> errors won't be `null`. Might be being too protective here.
Also:
```
!errorSink.Errors.Any()
```

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: I'm missing something.  that method calculates the generated length of the code written before an expression and passes that to an `CSharpPaddingBuilder` instance.  in turn, that instance does the calculation I have here.

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
dougbu: OIC  `target` won't work because it's the whole HTML element / `TagHelperBlock` and starts prior to the attribute value.  think of cases like
``` html
  <tag-helper
    some-attribute="some-value"
    another-attribute="another-value" />
```

(suspect I can change VS options to do the above by default :smiley_cat:)

src/Microsoft.AspNet.Razor/ParserResults.cs
```diff
@@ -31,40 +40,50 @@ public class ParserResults
...
                                 [NotNull] IEnumerable<TagHelperDescriptor> tagHelperDescriptors,
-                                [NotNull] IList<RazorError> errors)
+                                [NotNull] ParserErrorSink errorSink)
         {
             Success = success;
             Document = document;
             TagHelperDescriptors = tagHelperDescriptors;
-            ParserErrors = errors ?? new List<RazorError>();
+            ErrorSink = errorSink;
+            ParserErrors = errorSink.Errors?.ToList() ?? new List<RazorError>();
```
dougbu: @NTaylorMullen I'll simplify this code too.  either `(errorSink.Errors as IList<RazorError>) ?? errorSink.Errors.ToList()` (because the second part will never execute today) or just `errorSink.Errors.ToList()`.

but why is `ParserErrors` an `IList<RazorError>` and not `IEnumerable<RazorError>`?  no apparent reason to make `ParserErrors` writeable.  the collection is only read twice, including the `Any()` call above (replacing the `Count()` call).  the underlying collection is a `List<RazorError>` anyhow.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeValueCodeRenderer.cs
dougbu: :ok:

dougbu:
@NTaylorMullen updated per the actionable comments.  a couple of items remain open above.

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
NTaylorMullen: I c I c, dur definitely makes sense now that I think about it. It's good as is :+1:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: @pranavkm handled centering the line projections, see if you can bug him real quick for a walk through on what he did. For some reason I can't recall :cry:

NTaylorMullen:
:shipit: after talking to pranav and handling the centering bits.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: @NTaylorMullen @pranavkm and I agree the best-effort code I have here mimics what he did in `RenderRuntimeExpressionBlockChunkWithContentSpan`.  If we want to centre things, that would be a separate bug and it would require calculation of where the statement or expression ends.

Issue #260, opened 2014-12-18
Tag Helper code generation produces unused variables


pranavkm:
Code generation for Tag helpers produces ValueBuffer fields that might not be used.
E.g.

> The field 'ASPV_Views_Account_ConfirmEmail_cshtml.__tagHelperStringValueBuffer' is assigned but its value is never used

 If the application has WarningAsErrors on if you have precompilation enabled, these warnings prevent the application from launching. Once https://github.com/aspnet/Mvc/pull/1677 goes in, this would cause runtime view compilation failures (since the flag gets passed along).

NTaylorMullen:
2d2c2cdf0fb448ad8e44affe56d2156b10be0f79

PR #269, opened 2015-01-13
Add warning disable for unused variable warnings for TagHelper fields.

NTaylorMullen:
- Added a 0414 warning disable/restore around the __tagHelperStringValueBuffer since it's the only TagHelper utility field that "may" never be used.
- Regenerated baselines for TagHelper test files.

#260


pranavkm:
:shipit:

Issue #187, opened 2014-10-10
Support disabling tag helper processing at the element level


NTaylorMullen:
There should be the ability to disable a specific element instance (in a CSHTML file) from being processed as a Tag Helper. e.g. consider that the `input` tag is mapped to the `InputTagHelper` class but I want a specific input element instance to be client-side only:

```
@addtaghelper "InputTagHelper"
@addtaghelper "DivTagHelper"

<input />  @* This will be processed as a tag helper *@

<!input /> @* This will *not* be processed as a tag helper and the '!' will be removed from the output HTML *@

<!div>     @* This will *not* be processed as a tag helper and the '!' will be removed from the output HTML *@
    Hello there
</!div>
```

If a prefix is also declared for the page with the *same* value, an exception should be thrown at parse time to indicate that the combination is invalid, e.g.:

```
@addtaghelper "InputTagHelper", prefix: "!"

@* Explodes at parse time *@
```

NTaylorMullen:
Note: We've designed out the "only remove the `!` if an element is a `TagHelper`.

NTaylorMullen:
94230a5a145d11b3772cd54a1ae17b036ac326ff
cf3d0492721cd61b262babfa3cc9b9efbf9c21aa

PR #276, opened 2015-01-23
Add TagHelper parse level opt-out character '!'. 

NTaylorMullen:
- Added the ability to opt-out of TagHelper parsing by adding a '!' to the beginning of a tag name.
- Modified parsing logic to allow bangs in tags.
- Added a new EscapedTagHelperCodeGenerator to strip '!' from escaped TagHelper output.
- Added extensive unit tests.
- Added functional test to validate output for runtime and design time.

#187

/cc @dougbu @pranavkm

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
+++ b/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
     associated with the data types.
-
+    
     Example:
-
+    
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-
-    There are any number of "resheader" rows that contain simple
+                
+    There are any number of "resheader" rows that contain simple 
     name/value pairs.
-
-    Each data row contains a name, and value. The row also contains a
-    type or mimetype. Type corresponds to a .NET class that support
-    text/value conversion through the TypeConverter architecture.
-    Classes that don't support this are serialized and stored with the
+    
+    Each data row contains a name, and value. The row also contains a 
+    type or mimetype. Type corresponds to a .NET class that support 
+    text/value conversion through the TypeConverter architecture. 
+    Classes that don't support this are serialized and stored with the 
     mimetype set.
-
-    The mimetype is used for serialized objects, and tells the
-    ResXResourceReader how to depersist the object. This is currently not
+    
+    The mimetype is used for serialized objects, and tells the 
+    ResXResourceReader how to depersist the object. This is currently not 
     extensible. For a given mimetype the value must be set accordingly:
-
-    Note - application/x-microsoft.net.object.binary.base64 is the format
-    that the ResXResourceWriter will generate, however the reader can
+    
+    Note - application/x-microsoft.net.object.binary.base64 is the format 
+    that the ResXResourceWriter will generate, however the reader can 
     read any of the formats listed below.
-
+    
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-
+    
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with
+    value   : The object must be serialized with 
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array
+    value   : The object must be serialized into a byte array 
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
@@ -129,13 +129,16 @@
   <data name="ScopeManager_EndCannotBeCalledWithoutACallToBegin" xml:space="preserve">
     <value>Must call '{2}.{1}' before calling '{2}.{0}'.</value>
   </data>
-  <data name="HtmlElementNameAttribute_AdditionalTagsCannotContainNull" xml:space="preserve">
-    <value>Parameter {0} must not contain null tag names.</value>
+  <data name="HtmlElementNameAttribute_ElementNameCannotBeNullOrWhitespace" xml:space="preserve">
+    <value>Tag name cannot be null or whitespace.</value>
   </data>
   <data name="ArgumentCannotBeNullOrEmpty" xml:space="preserve">
     <value>The value cannot be null or empty.</value>
   </data>
   <data name="TagHelperDescriptorResolver_EncounteredUnexpectedError" xml:space="preserve">
     <value>Encountered an unexpected error when attempting to resolve tag helper directive '{0}' with value '{1}'. Error: {2}</value>
+  </data>
+  <data name="HtmlElementNameAttribute_InvalidElementName" xml:space="preserve">
+    <value>Tag helpers cannot target element name '{0}' because it contains a '{1}' character.</value>
   </data>
 </root>
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
@@ -19,6 +19,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="tag">The HTML tag name for the <see cref="TagHelper"/> to target.</param>
         public HtmlElementNameAttribute([NotNull] string tag)
         {
+            ValidateTagName(tag, nameof(tag));
+
             Tags = new[] { tag };
         }
 
@@ -29,12 +31,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="additionalTags">Additional HTML tag names for the <see cref="TagHelper"/> to target.</param>
         public HtmlElementNameAttribute([NotNull] string tag, [NotNull] params string[] additionalTags)
         {
-            if (additionalTags.Contains(null))
+            ValidateTagName(tag, nameof(tag));
+
+            foreach (var tagName in additionalTags)
             {
-                throw new ArgumentNullException(
-                    nameof(additionalTags),
-                    Resources.FormatHtmlElementNameAttribute_AdditionalTagsCannotContainNull(nameof(additionalTags)));
-            };
+                ValidateTagName(tagName, nameof(additionalTags));
+            }
 
             var allTags = new List<string>(additionalTags);
             allTags.Add(tag);
@@ -45,6 +47,23 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// An <see cref="IEnumerable{string}"/> of tag names for the <see cref="TagHelper"/> to target.
         /// </summary>
-        public IEnumerable<string> Tags { get; private set; }
+        public IEnumerable<string> Tags { get; }
+
+        private static void ValidateTagName(string tagName, string parameterName)
+        {
+            if (string.IsNullOrWhiteSpace(tagName))
+            {
+                throw new ArgumentException(
+                    Resources.HtmlElementNameAttribute_ElementNameCannotBeNullOrWhitespace,
+                    parameterName);
+            }
+
+            if (tagName.Contains('!'))
+            {
+                throw new ArgumentException(
+                    Resources.FormatHtmlElementNameAttribute_InvalidElementName(tagName, '!'),
+                    parameterName);
+            }
+        }
     }
 }
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -133,7 +133,9 @@ namespace Microsoft.AspNet.Razor.Parser
                 IDisposable tagBlockWrapper = null;
                 try
                 {
-                    if (!EndOfFile && !AtSpecialTag)
+                    var atSpecialTag = AtSpecialTag;
+
+                    if (!EndOfFile && !atSpecialTag)
                     {
                         // Start a Block tag.  This is used to wrap things like <p> or <a class="btn"> etc.
                         tagBlockWrapper = Context.StartBlock(BlockType.Tag);
@@ -157,7 +159,7 @@ namespace Microsoft.AspNet.Razor.Parser
                         }
                         else
                         {
-                            complete = AfterTagStart(tagStart, tags, tagBlockWrapper);
+                            complete = AfterTagStart(tagStart, tags, atSpecialTag, tagBlockWrapper);
                         }
                     }
 
@@ -187,6 +189,7 @@ namespace Microsoft.AspNet.Razor.Parser
 
         private bool AfterTagStart(SourceLocation tagStart,
                                    Stack<Tuple<HtmlSymbol, SourceLocation>> tags,
+                                   bool atSpecialTag,
                                    IDisposable tagBlockWrapper)
         {
             if (!EndOfFile)
@@ -197,9 +200,16 @@ namespace Microsoft.AspNet.Razor.Parser
                         // End Tag
                         return EndTag(tagStart, tags, tagBlockWrapper);
                     case HtmlSymbolType.Bang:
-                        // Comment
-                        Accept(_bufferedOpenAngle);
-                        return BangTag();
+                        // Comment, CDATA, DOCTYPE, or a parser-escaped HTML tag.
+                        if (atSpecialTag)
+                        {
+                            Accept(_bufferedOpenAngle);
+                            return BangTag();
+                        }
+                        else
+                        {
+                            goto default;
+                        }
                     case HtmlSymbolType.QuestionMark:
                         // XML PI
                         Accept(_bufferedOpenAngle);
@@ -275,30 +285,48 @@ namespace Microsoft.AspNet.Razor.Parser
         {
             // Accept "/" and move next
             Assert(HtmlSymbolType.ForwardSlash);
-            var solidus = CurrentSymbol;
+            var forwardSlash = CurrentSymbol;
             if (!NextToken())
             {
                 Accept(_bufferedOpenAngle);
-                Accept(solidus);
+                Accept(forwardSlash);
                 return false;
             }
             else
             {
-                var tagName = String.Empty;
-                if (At(HtmlSymbolType.Text))
+                var tagName = string.Empty;
+                HtmlSymbol bangSymbol = null;
+
+                if (At(HtmlSymbolType.Bang))
+                {
+                    bangSymbol = CurrentSymbol;
+
+                    var nextSymbol = Lookahead(count: 1);
+
+                    if (nextSymbol != null && nextSymbol.Type == HtmlSymbolType.Text)
+                    {
+                        tagName = "!" + nextSymbol.Content;
+                    }
+                }
+                else if (At(HtmlSymbolType.Text))
                 {
                     tagName = CurrentSymbol.Content;
                 }
+
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    String.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
+                    // Note tagName may contain a '!' escape character. This ensures </!text> doesn't match here. 
+                    // </!text> tags are treated like any other escaped HTML end tag.
+                    string.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
                     matched)
                 {
-                    return EndTextTag(solidus, tagBlockWrapper);
+                    return EndTextTag(forwardSlash, tagBlockWrapper);
                 }
                 Accept(_bufferedOpenAngle);
-                Accept(solidus);
+                Accept(forwardSlash);
+
+                OptionalBangEscape();
 
                 AcceptUntil(HtmlSymbolType.CloseAngle);
 
@@ -347,14 +375,22 @@ namespace Microsoft.AspNet.Razor.Parser
             return seenCloseAngle;
         }
 
-        // Special tags include <! and <? tags
+        // Special tags include <!--, <!DOCTYPE, <![CDATA and <? tags
         private bool AtSpecialTag
         {
             get
             {
-                return (At(HtmlSymbolType.OpenAngle) &&
-                       (NextIs(HtmlSymbolType.Bang) ||
-                        NextIs(HtmlSymbolType.QuestionMark)));
+                if (At(HtmlSymbolType.OpenAngle))
+                {
+                    if (NextIs(HtmlSymbolType.Bang))
+                    {
+                        return !IsBangEscape(lookahead: 1);
+                    }
+
+                    return NextIs(HtmlSymbolType.QuestionMark);
+                }
+
+                return false;
             }
         }
 
@@ -651,20 +687,41 @@ namespace Microsoft.AspNet.Razor.Parser
 
         private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable tagBlockWrapper)
         {
-            // If we're at text, it's the name, otherwise the name is ""
-            HtmlSymbol tagName;
-            if (At(HtmlSymbolType.Text))
+            HtmlSymbol bangSymbol = null;
+            HtmlSymbol potentialTagNameSymbol;
+
+            if (At(HtmlSymbolType.Bang))
             {
-                tagName = CurrentSymbol;
+                bangSymbol = CurrentSymbol;
+
+                potentialTagNameSymbol = Lookahead(count: 1);
             }
             else
             {
-                tagName = new HtmlSymbol(CurrentLocation, String.Empty, HtmlSymbolType.Unknown);
+                potentialTagNameSymbol = CurrentSymbol;
+            }
+
+            HtmlSymbol tagName;
+
+            if (potentialTagNameSymbol == null || potentialTagNameSymbol.Type != HtmlSymbolType.Text)
+            {
+                tagName = new HtmlSymbol(potentialTagNameSymbol.Start, string.Empty, HtmlSymbolType.Unknown);
+            }
+            else if (bangSymbol != null)
+            {
+                tagName = new HtmlSymbol(bangSymbol.Start, "!" + potentialTagNameSymbol.Content, HtmlSymbolType.Text);
+            }
+            else
+            {
+                tagName = potentialTagNameSymbol;
             }
 
             Tuple<HtmlSymbol, SourceLocation> tag = Tuple.Create(tagName, _lastTagStart);
 
-            if (tags.Count == 0 && String.Equals(tag.Item1.Content, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase))
+            if (tags.Count == 0 &&
+                // Note tagName may contain a '!' escape character. This ensures <!text> doesn't match here. 
+                // <!text> tags are treated like any other escaped HTML start tag.
+                string.Equals(tag.Item1.Content, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase))
             {
                 Output(SpanKind.Markup);
                 Span.CodeGenerator = SpanCodeGenerator.Null;
@@ -709,7 +766,9 @@ namespace Microsoft.AspNet.Razor.Parser
 
                 return true;
             }
+
             Accept(_bufferedOpenAngle);
+            OptionalBangEscape();
             Optional(HtmlSymbolType.Text);
             return RestOfTag(tag, tags, tagBlockWrapper);
         }
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Razor.Generator;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Tokenizer.Symbols;
 
@@ -42,58 +43,69 @@ namespace Microsoft.AspNet.Razor.Parser
             {
                 if (NextIs(HtmlSymbolType.Bang))
                 {
-                    AcceptAndMoveNext(); // Accept '<'
-                    BangTag();
+                    // Checking to see if we meet the conditions of a special '!' tag: <!DOCTYPE, <![CDATA[, <!--.
+                    if (!IsBangEscape(lookahead: 1))
+                    {
+                        AcceptAndMoveNext(); // Accept '<'
+                        BangTag();
+                        return;
+                    }
+
+                    // We should behave like a normal tag that has a parser escape, fall through to the normal 
+                    // tag logic.
                 }
                 else if (NextIs(HtmlSymbolType.QuestionMark))
                 {
                     AcceptAndMoveNext(); // Accept '<'
                     XmlPI();
+                    return;
+                }
+
+                Output(SpanKind.Markup);
+
+                // Start tag block
+                var tagBlock = Context.StartBlock(BlockType.Tag);
+
+                AcceptAndMoveNext(); // Accept '<'
+
+                if (!At(HtmlSymbolType.ForwardSlash))
+                {
+                    OptionalBangEscape();
+
+                    // Parsing a start tag
+                    var scriptTag = At(HtmlSymbolType.Text) &&
+                                    string.Equals(CurrentSymbol.Content, "script", StringComparison.OrdinalIgnoreCase);
+                    Optional(HtmlSymbolType.Text);
+                    TagContent(); // Parse the tag, don't care about the content
+                    Optional(HtmlSymbolType.ForwardSlash);
+                    Optional(HtmlSymbolType.CloseAngle);
+
+                    if (scriptTag)
+                    {
+                        Output(SpanKind.Markup);
+                        tagBlock.Dispose();
+
+                        SkipToEndScriptAndParseCode();
+                        return;
+                    }
                 }
                 else
                 {
-                    Output(SpanKind.Markup);
+                    // Parsing an end tag
+                    // This section can accept things like: '</p  >' or '</p>' etc.
+                    Optional(HtmlSymbolType.ForwardSlash);
 
-                    // Start tag block
-                    var tagBlock = Context.StartBlock(BlockType.Tag);
-
-                    AcceptAndMoveNext(); // Accept '<'
-
-                    if (!At(HtmlSymbolType.ForwardSlash))
-                    {
-                        // Parsing a start tag
-                        var scriptTag = At(HtmlSymbolType.Text) &&
-                                        string.Equals(CurrentSymbol.Content, "script", StringComparison.OrdinalIgnoreCase);
-                        Optional(HtmlSymbolType.Text);
-                        TagContent(); // Parse the tag, don't care about the content
-                        Optional(HtmlSymbolType.ForwardSlash);
-                        Optional(HtmlSymbolType.CloseAngle);
-
-                        if (scriptTag)
-                        {
-                            Output(SpanKind.Markup);
-                            tagBlock.Dispose();
-
-                            SkipToEndScriptAndParseCode();
-                            return;
-                        }
-                    }
-                    else
-                    {
-                        // Parsing an end tag
-                        // This section can accept things like: '</p  >' or '</p>' etc.
-                        Optional(HtmlSymbolType.ForwardSlash);
-                        // Whitespace here is invalid (according to the spec)
-                        Optional(HtmlSymbolType.Text);
-                        AcceptAll(HtmlSymbolType.WhiteSpace);
-                        Optional(HtmlSymbolType.CloseAngle);
-                    }
-
-                    Output(SpanKind.Markup);
-
-                    // End tag block
-                    tagBlock.Dispose();
+                    // Whitespace here is invalid (according to the spec)
+                    OptionalBangEscape();
+                    Optional(HtmlSymbolType.Text);
+                    AcceptAll(HtmlSymbolType.WhiteSpace);
+                    Optional(HtmlSymbolType.CloseAngle);
                 }
+
+                Output(SpanKind.Markup);
+
+                // End tag block
+                tagBlock.Dispose();
             }
         }
     }
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
@@ -184,5 +184,38 @@ namespace Microsoft.AspNet.Razor.Parser
             Initialize(Span);
             NextToken();
         }
+
+        private bool IsBangEscape(int lookahead)
+        {
+            var potentialBang = Lookahead(lookahead);
+
+            if (potentialBang != null &&
+                potentialBang.Type == HtmlSymbolType.Bang)
+            {
+                var afterBang = Lookahead(lookahead + 1);
+
+                return afterBang != null &&
+                    afterBang.Type == HtmlSymbolType.Text &&
+                    !string.Equals(afterBang.Content, "DOCTYPE", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+
+        private void OptionalBangEscape()
+        {
+            if (IsBangEscape(lookahead: 0))
+            {
+                Output(SpanKind.Markup);
+
+                // Accept the parser escape character '!'.
+                Assert(HtmlSymbolType.Bang);
+                AcceptAndMoveNext();
+
+                // Setup the metacode span that we will be outputing.
+                Span.CodeGenerator = SpanCodeGenerator.Null;
+                Output(SpanKind.MetaCode, AcceptedCharacters.None);
+            }
+        }
     }
 }
--- a/src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Text;
@@ -73,6 +74,43 @@ namespace Microsoft.AspNet.Razor.Parser
             {
                 SpanConfig(span);
             }
+        }
+
+        protected TSymbol Lookahead(int count)
+        {
+            if (count < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
+            }
+            else if (count == 0)
+            {
+                return CurrentSymbol;
+            }
+
+            // We add 1 in order to store the current symbol.
+            var symbols = new TSymbol[count + 1];
+            var currentSymbol = CurrentSymbol;
+
+            symbols[0] = currentSymbol;
+
+            // We need to look forward "count" many times.
+            for (var i = 1; i <= count; i++)
+            {
+                NextToken();
+                symbols[i] = CurrentSymbol;
+            }
+
+            // Restore Tokenizer's location to where it was pointing before the look-ahead.
+            for (var i = count; i >= 0; i--)
+            {
+                PutBack(symbols[i]);
+            }
+
+            // The PutBacks above will set CurrentSymbol to null. EnsureCurrent will set our CurrentSymbol to the 
+            // next symbol.
+            EnsureCurrent();
+
+            return symbols[count];
         }
 
         protected internal bool NextToken()
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Xunit;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class HtmlElementNameAttributeTest
+    {
+        public static TheoryData InvalidTagNameData
+        {
+            get
+            {
+                var invalidTagNameError =
+                    "Tag helpers cannot target element name '{0}' because it contains a '{1}' character.";
+                var nullOrWhitespaceTagNameError =
+                    "Tag name cannot be null or whitespace.";
+
+                // tagName, expectedExceptionMessage
+                return new TheoryData<string, string>
+                {
+                    { "!", string.Format(invalidTagNameError, "!", "!") },
+                    { "hello!", string.Format(invalidTagNameError, "hello!", "!") },
+                    { "!hello", string.Format(invalidTagNameError, "!hello", "!") },
+                    { "he!lo", string.Format(invalidTagNameError, "he!lo", "!") },
+                    { "!he!lo!", string.Format(invalidTagNameError, "!he!lo!", "!") },
+                    { string.Empty, nullOrWhitespaceTagNameError },
+                    { Environment.NewLine, nullOrWhitespaceTagNameError },
+                    { "\t", nullOrWhitespaceTagNameError },
+                    { " \t ", nullOrWhitespaceTagNameError },
+                    { " ", nullOrWhitespaceTagNameError },
+                    { Environment.NewLine + " ", nullOrWhitespaceTagNameError },
+                    { null, nullOrWhitespaceTagNameError },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidTagNameData))]
+        public void SingleArgumentConstructor_ThrowsOnInvalidTagNames(
+            string tagName, 
+            string expectedExceptionMessage)
+        {
+            // Arrange
+            expectedExceptionMessage += Environment.NewLine + "Parameter name: tag";
+
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(
+                "tag", 
+                () => new HtmlElementNameAttribute(tagName));
+            Assert.Equal(exception.Message, expectedExceptionMessage);
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidTagNameData))]
+        public void MultipleArgumentConstructor_ThrowsOnInvalidTagNames(
+            string tagName, 
+            string expectedExceptionMessage)
+        {
+            // Arrange
+            expectedExceptionMessage += Environment.NewLine + "Parameter name: additionalTags";
+
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(
+                "additionalTags", 
+                () => new HtmlElementNameAttribute("p", "div", "span", tagName));
+            Assert.Equal(exception.Message, expectedExceptionMessage);
+        }
+    }
+}
--- a/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
@@ -19,6 +19,1642 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData OptOut_WithAttributeTextTagData
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var blockFactory = new BlockFactory(factory);
+                var errorFormatNormalUnclosed =
+                    "The \"{0}\" element was not closed.  All elements must be either self-closing or have a " +
+                    "matching end tag.";
+                var errorMatchingBrace =
+                    "The code block is missing a closing \"}\" character.  Make sure you have a matching \"}\" " +
+                    "character for all the \"{\" characters within this block, and that none of the \"}\" " +
+                    "characters are being interpreted as markup.";
+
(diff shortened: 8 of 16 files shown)


src/Microsoft.AspNet.Razor/Generator/TagHelpers/EscapedTagHelperCodeGenerator.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Diagnostics;
+using Microsoft.AspNet.Razor.Parser.SyntaxTree;
+
+namespace Microsoft.AspNet.Razor.Generator
+{
+    /// <summary>
+    /// A <see cref="BlockCodeGenerator"/> that is responsible for generating valid 
+    /// <see cref="Compiler.LiteralChunk"/>s from escaped tag helpers.
+    /// </summary>
+    public class EscapedTagHelperCodeGenerator : SpanCodeGenerator
```
dougbu: not a big fan of this name.  makes it seem like we go down a different generation code path than any other element that doesn't match a tag helper descriptor (albeit after correcting the tag name)?  but this class really does a quick fixup of the generated tree to correct the tag name.  suggest `RemoveBangCodeGenerator`.

more generally, why does the escaped tag name last this long?  couldn't you leave the `'!'` character out of the original `Span`, perhaps with an `Association` back to a `Span` that includes the `'!'` as well.

dougbu: oops, meant `RemoveBangLiteralCodeGenerator

src/Microsoft.AspNet.Razor/Generator/TagHelpers/EscapedTagHelperCodeGenerator.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /// <summary>
+        /// Creates a <see cref="Compiler.LiteralChunk"/> for the given <paramref name="target"/>.
+        /// </summary>
+        /// <param name="target">
+        /// The <see cref="Span"/> responsible for this <see cref="EscapedTagHelperCodeGenerator"/>.
+        /// </param>
+        /// <param name="context">A <see cref="CodeGeneratorContext"/> instance that contains information about
+        /// the current code generation process.</param>
+        /// <remarks>Removes the first tag helper escape character '!' from the generated
```
dougbu: `'!'` -> `(<c>'!'</c>)` (it's code and parenthetical)

src/Microsoft.AspNet.Razor/Generator/TagHelpers/EscapedTagHelperCodeGenerator.cs
```diff
@@ -0,0 +1,39 @@
...
+        public override void GenerateCode(Span target, CodeGeneratorContext context)
+        {
+            var literalText = target.Content;
+            var bangIndex = literalText.IndexOf('!');
+
+            // The requirement for this code generator is for output to be escaped by '!'; therefore there should 
+            // always be a '!' in the literalText.
+            Debug.Assert(bangIndex != -1);
+
+            literalText = literalText.Remove(bangIndex, 1);
```
dougbu: isn't the tag name always a separate `Span` from the preceding `<` or `</` ?  if yes, simplify this method:
``` c#
Debug.Assert(!string.IsNullOrEmpty(literalText));
Debug.Assert(literalText.StartsWith("!", StringComparison.Ordinal);
literalText = literalText.SubString(1);
```

note additional guard against an NRE

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -197,6 +197,23 @@ private void TagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags)
...
+                        {
+                            using (tagBlockWrapper = Context.StartBlock(BlockType.Tag))
+                            {
+                                var complete = StartTag(tags, tagBlockWrapper);
+
+                                // If the tag was completed this will ensure the '>' is contained in the tag block.
+                                Output(SpanKind.Markup);
+
+                                return complete;
+                            }
```
dougbu: ouch.  why can't this branch `goto` the `default` branch instead?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -197,6 +197,23 @@ private void TagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags)
                         // End Tag
                         return EndTag(tagStart, tags, tagBlockWrapper);
                     case HtmlSymbolType.Bang:
+                        var nextSymbol = Lookahead();
+
+                        // If the next symbol is a text symbol and is not DOCTYPE then it's an ordinary HTML tag.
+                        if (nextSymbol.Type == HtmlSymbolType.Text &&
+                            !string.Equals(nextSymbol.Content, "DOCTYPE", StringComparison.OrdinalIgnoreCase))
```
dougbu: what about the comment case this branch also handles?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -284,15 +301,31 @@ private bool CData()
...
+                    bangSymbol = CurrentSymbol;
+
+                    var nextSymbol = Lookahead();
+
+                    if (nextSymbol.Type == HtmlSymbolType.Text)
+                    {
+                        tagName = nextSymbol.Content;
+                    }
+                }
+                else if (At(HtmlSymbolType.Text))
```
dougbu: looks like this code throws away the bang completely whereas the start tag handling does not.  why not throw away the bang in both cases and update the current block to indicate it's a non-tag-helper `TagBlock`?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -284,15 +301,31 @@ private bool CData()
...
                 {
                     tagName = CurrentSymbol.Content;
                 }
+
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    String.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
+                    // text tags cannot be targeted by TagHelpers and therefore can't be escaped with '!'.
```
dougbu: unnecessary error.  treat `<!text>` the same as `<text>`.  i.e. use `!` to mean "never a tag helper" even if the result is something other than a tag.

src/Microsoft.AspNet.Razor/Generator/TagHelpers/EscapedTagHelperCodeGenerator.cs
NTaylorMullen: Originally I left it out of the original Span but by doing that I'd be leaving out a character that truly existed in the document. Also, leaving out the character affected all source locations for every element following it and then got compoundingly bad the more `!`'s that were used.

src/Microsoft.AspNet.Razor/Generator/TagHelpers/EscapedTagHelperCodeGenerator.cs
NTaylorMullen: Nope, they're one `Span` always. Also, in order for this code generator to be in the document the resulting span must have AT LEAST `<!` in it. In the case that a third party decides to modify that expectation and enable `literalText` to be nothing then the `target.Content` method to retrieve `literalText` will return an empty string.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -662,9 +704,20 @@ private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable
                 tagName = new HtmlSymbol(CurrentLocation, String.Empty, HtmlSymbolType.Unknown);
             }
 
+            // Need to check if we need to back-track.
+            if (bangSymbol != null)
+            {
+                // Need to ensure the Tokenizer is pointing to the correct location.
+                PutCurrentBack();
+                PutBack(bangSymbol);
+                NextToken();
```
dougbu: might need more to the comment above since these three lines look identical to a single `PutCurrentBack()` call.  probably has something to do with setting `PreviousSymbol` correctly (and this pattern happens a few times in the code) but comment would still help.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: I had that initially but quickly found out that for this code path (expectation = at a special tag = `!`) the `tagBlockWrapper` that's passed in is `null` (ref https://github.com/aspnet/Razor/blob/disableths.187/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs#L136) and therefore nothing gets encapsulated in a `TagBlock` unless we manually encapsulate it.

I really dislike the parser, I was face palming every character of this snippet :sob:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: For a comment `nextSymbol.Type` is `HtmlSymbolType.DoubleHyphen`:
https://github.com/aspnet/Razor/blob/disableths.187/src/Microsoft.AspNet.Razor/Tokenizer/Symbols/HtmlSymbolType.cs#L19

Therefore it falls through to its normal logic (calling into `BangTag()`).

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: It doesn't throw it away, this entire section is only looking ahead. the section below is what captures the `!` tag (really anything after the `/`) by calling into `AcceptUntil(HtmlSymbolType.CloseAngle)`.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Whaaa, a `<text>` tag doesn't look anything like a `<p>` as far as syntax trees go, and due to that limitation is incapable to be targeted by TagHelpers. Same principle applies to `<![CDATA[..`. If someone happens to write `!` we want to treat it as a normal HTML tag, that's what this check ensures will happen.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Ahhhh looks like this change is still sitting on my local box, I transitioned this bit to use the `Lookahead` helper. Will update.

NTaylorMullen:
Sorry @dougbu I just updated the branch, it looks like I had rebased a few changes prior to your comments and didn't push them.

dougbu:
difficult seeing overall design here -- forest for the trees and all that
1) have added many special cases to the parser.  how many of those can be simplified if we let the parser ignore some error cases?
2) the new code generator shouldn't be needed.  the `TagHelperParseTreeRewriter` rewrites the parse tree (as the name says :smiley_cat:); why does the result contain two separate cases for `<NotTagHelper>` and `<!TagHelper>`?  IOW why not rewrite the parse tree to match the `<NotTagHelper>` case?
3) not clear whether a tag helper targeting `<script>` tags is supported

NTaylorMullen:
1. None can be ignored due to the C# block HTML tag logic and outside of C# block HTML tag logic being different.
1. Rewriting the parse tree to remove a `!` from the output essentially leaves an empty window in the editor which is a big nono. The alternative is to reconstruct the `Span` and break it into 3 pieces:
`<` markup, `!` metacode, `tagname` markup. As I write this I kind of like that idea.
1. `TagHelper`s targeting script tags are supported. I do one in my channel9 talk :smile:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -197,6 +197,23 @@ private void TagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags)
                         // End Tag
                         return EndTag(tagStart, tags, tagBlockWrapper);
                     case HtmlSymbolType.Bang:
+                        var nextSymbol = Lookahead();
+
+                        // If the next symbol is a text symbol and is not DOCTYPE then it's an ordinary HTML tag.
+                        if (nextSymbol != null && nextSymbol.Type == HtmlSymbolType.Text &&
```
NTaylorMullen: Spread condition

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Modify `AtSpecialTag` logic.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -653,18 +686,35 @@ private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable
         {
             // If we're at text, it's the name, otherwise the name is ""
             HtmlSymbol tagName;
-            if (At(HtmlSymbolType.Text))
+            HtmlSymbol bangSymbol = null;
+            HtmlSymbol potentialTagNameSymbol;
+
+            if (At(HtmlSymbolType.Bang))
```
NTaylorMullen: Investigate rewriting Symbols to contain `!p` for the `tagName`. This would enable correct error cases for:

```
@{
<!p>
</p>
}
```

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +42,72 @@ private void ScanTagInDocumentContext()
             {
                 if (NextIs(HtmlSymbolType.Bang))
```
NTaylorMullen: See if we can use `AtSpecialTag` here.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Done.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Done

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: Wouldn't work well.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/Resources.resx
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!-- 
+    Microsoft ResX Schema 
+
```
dougbu: nit: use the XML editor and don't let VS add the whitespace back :smiley_cat:

src/Microsoft.AspNet.Razor.Runtime/Resources.resx
```diff
@@ -140,4 +140,7 @@
   <data name="TagHelperDescriptorResolver_EncounteredUnexpectedError" xml:space="preserve">
     <value>Encountered an unexpected error when attempting to resolve tag helper directive '{0}' with value '{1}'. Error: {2}</value>
   </data>
+  <data name="HtmlElementNameAttribute_InvalidElementName" xml:space="preserve">
+    <value>Tag helpers cannot target element name "{0}".</value>
```
dougbu: - use single quotes to quote values
- you're passing `"!"` rather than the invalid element name
- give the user more information e.g. ... `element name '{0}' because it contains a '{1}' character.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -45,6 +47,29 @@ public HtmlElementNameAttribute([NotNull] string tag, [NotNull] params string[]
...
+                    ValidateTagName(tagName);
+                }
+
+                _tags = value;
+            }
+        }
+
+        private static void ValidateTagName(string tagName)
+        {
+            if (string.Equals(tagName, "!", StringComparison.OrdinalIgnoreCase))
```
dougbu: - `Contains()` not `Equals()`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -45,6 +47,29 @@ public HtmlElementNameAttribute([NotNull] string tag, [NotNull] params string[]
...
+            {
+                throw new ArgumentException(Resources.FormatHtmlElementNameAttribute_InvalidElementName("!"));
```
dougbu: pass invalid name as well as the problematic `char`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -45,6 +47,29 @@ public HtmlElementNameAttribute([NotNull] string tag, [NotNull] params string[]
         /// <summary>
         /// An <see cref="IEnumerable{string}"/> of tag names for the <see cref="TagHelper"/> to target.
         /// </summary>
-        public IEnumerable<string> Tags { get; private set; }
+        public IEnumerable<string> Tags
```
dougbu: - go back to auto-property and do validation in the constructors, before they create lists and so on
- nit: remove `private set;`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -45,6 +47,29 @@ public HtmlElementNameAttribute([NotNull] string tag, [NotNull] params string[]
...
+                foreach (var tagName in value)
+                {
+                    ValidateTagName(tagName);
+                }
+
+                _tags = value;
+            }
+        }
+
+        private static void ValidateTagName(string tagName)
```
dougbu: - handle `null` names here too; allow `params` constructor to run through its parameters once
- not that big a deal that this means you'll validate `tag` parameters are non-`null`; `[NotNull]` will eventually prevent that case

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
     public sealed class HtmlElementNameAttribute : Attribute
     {
+        private IEnumerable<string> _tags;
```
dougbu: please no

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -197,9 +200,16 @@ private void TagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags)
                         // End Tag
                         return EndTag(tagStart, tags, tagBlockWrapper);
                     case HtmlSymbolType.Bang:
-                        // Comment
-                        Accept(_bufferedOpenAngle);
-                        return BangTag();
+                        // Comment, CDATA, DOCTYPE, or a parser escaped HTML tag.
```
dougbu: teensy nit: hyphenate "parser-escaped"

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -197,9 +200,16 @@ private void TagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags)
...
+                        if (atSpecialTag)
+                        {
+                            Accept(_bufferedOpenAngle);
+                            return BangTag();
+                        }
+                        else
+                        {
+                            goto default;
```
dougbu: :dancers:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -300,6 +326,8 @@ private bool CData()
                 Accept(_bufferedOpenAngle);
                 Accept(solidus);
```
dougbu: nit: while here please change this variable to something more widely known e.g. `slash`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -284,15 +294,31 @@ private bool CData()
...
                     tagName = CurrentSymbol.Content;
                 }
+
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    String.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
+                    // Text tags cannot be targeted by TagHelpers and therefore can't be escaped with '!'.
+                    bangSymbol == null &&
+                    string.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
```
dougbu: why is `bangSymbol == null` needed?  `tagName` includes the `'!'`, causing the `Equals` to fail when `bangSymbol != null`

NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -652,19 +688,43 @@ private void ParseQuoted(HtmlSymbolType type)
         private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable tagBlockWrapper)
         {
             // If we're at text, it's the name, otherwise the name is ""
```
dougbu: clarify this comment. what is "it"?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -652,19 +688,43 @@ private void ParseQuoted(HtmlSymbolType type)
...
+            {
+                tagName = new HtmlSymbol(bangSymbol.Start, "!" + potentialTagNameSymbol.Content, HtmlSymbolType.Text);
+            }
+            else if (potentialTagNameSymbol.Type == HtmlSymbolType.Text)
+            {
+                tagName = potentialTagNameSymbol;
+            }
+
+            // tagName should never be null here, the above if/else if should take care of all cases.
+            Debug.Assert(tagName != null);
```
dougbu: don't repeat check of `potentialTagNameSymbol.Type`.  remove this `Assert()` and change last `else if` to an `else`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -652,19 +688,43 @@ private void ParseQuoted(HtmlSymbolType type)
...
+
             Tuple<HtmlSymbol, SourceLocation> tag = Tuple.Create(tagName, _lastTagStart);
 
-            if (tags.Count == 0 && String.Equals(tag.Item1.Content, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase))
+            if (tags.Count == 0 &&
+                // Text tags cannot be targeted by TagHelpers and therefore can't be escaped with '!'.
+                bangSymbol == null &&
```
dougbu: again, don't need to check `bangSymbol == null`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +43,102 @@ private void ScanTagInDocumentContext()
...
+                {
+                    OptionalBangEscape();
+
+                    // Parsing a start tag
+                    var scriptTag = At(HtmlSymbolType.Text) &&
+                                    string.Equals(CurrentSymbol.Content, "script", StringComparison.OrdinalIgnoreCase);
+                    Optional(HtmlSymbolType.Text);
+                    TagContent(); // Parse the tag, don't care about the content
+                    Optional(HtmlSymbolType.ForwardSlash);
+                    Optional(HtmlSymbolType.CloseAngle);
```
dougbu: just curious: why is the close angle bracket optional?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +43,102 @@ private void ScanTagInDocumentContext()
...
+                    // This section can accept things like: '</p  >' or '</p>' etc.
+                    Optional(HtmlSymbolType.ForwardSlash);
 
-                    // End tag block
-                    tagBlock.Dispose();
+                    // Whitespace here is invalid (according to the spec)
+                    OptionalBangEscape();
+                    Optional(HtmlSymbolType.Text);
+                    AcceptAll(HtmlSymbolType.WhiteSpace);
+                    Optional(HtmlSymbolType.CloseAngle);
```
dougbu: same curiosity question here

dougbu: though I think I'm forgetting an answer you've provided before, sorry.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +43,102 @@ private void ScanTagInDocumentContext()
...
                 }
+
+                Output(SpanKind.Markup);
+
+                // End tag block
+                tagBlock.Dispose();
+            }
+        }
+
+        private bool AtBangEscape(int lookahead = 0)
```
dougbu: nit: suggest these two shared helpers should be in HtmlMarkupParser.cs

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +43,102 @@ private void ScanTagInDocumentContext()
...
+        {
+            var potentialBang = lookahead == 0 ? CurrentSymbol : Lookahead(lookahead);
```
dougbu: nit: this logic for the look-at-current-pseudo-look-ahead should be in `Lookahead()`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -42,58 +43,102 @@ private void ScanTagInDocumentContext()
...
+            {
+                Output(SpanKind.Markup);
+
+                // Accept the parser escape character '!'.
+                Assert(HtmlSymbolType.Bang);
+                AcceptAndMoveNext();
+
+                // Setup the metacode span that we will be outputing.
+                Span.CodeGenerator = SpanCodeGenerator.Null;
+                Output(SpanKind.MetaCode, AcceptedCharacters.None);
```
dougbu: :dancers:

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
             }
         }
 
+        protected TSymbol Lookahead(int count = 1)
+        {
+            if (count == 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
```
dougbu: `throw` if `count < 0` (a case that's missing).  return `CurrentSymbol` when `count == 0`

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
...
+            }
+
+            // We add 1 in order to store the current symbol.
+            var symbols = new TSymbol[count + 1];
+            var currentSymbol = CurrentSymbol;
+
+            symbols[0] = currentSymbol;
+
+            // We need to lookforward "count" many times.
+            for(var i = 1; i <= count; i++)
```
dougbu: space after `for`

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
...
+            {
+                NextToken();
+                symbols[i] = CurrentSymbol;
+            }
+
+            // We don't want to modify where the Tokenizer is pointing. Restore it to what it was before the lookahead.
```
dougbu: - remove first sentence -- desires are meaningless to a computer and to most readers.
- clarify remaining sentence: `Restore Tokenizer's location to where it was pointing before the look-ahead."

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
             }
         }
 
+        protected TSymbol Lookahead(int count = 1)
```
dougbu: shouldn't this method handle hitting EOF during the look-ahead?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,832 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                    "The code block is missing a closing \"}\" character.  Make sure you have a matching \"}\" " +
+                    "character for all the \"{\" characters within this block, and that none of the \"}\" " +
+                    "characters are being interpreted as markup.";
+                var errorEOFMatchingBrace =
+                    "End of file or an unexpected character was reached before the \"{0}\" tag could be parsed.  " +
+                    "Elements inside markup blocks must be complete. They must either be self-closing " +
+                    "(\"<br />\") or have matching end tags (\"<p>Hello</p>\").  If you intended " +
+                    "to display a \"<\" character, use the \"&lt;\" HTML entity.";
+
+                Func<Func<MarkupBlock>, MarkupBlock> buildPartialStatementBlock = (insideBuilder) =>
```
dougbu: side note: ick to your nested `Func`s

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,832 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData OptOutPartialTagDataCSharp
```
dougbu: - are warnings for `"@{<!p class=\"btn\" >}"` (missing end tag) tested elsewhere?
- either way, should probably test `"@{<!p class=\"btn\" /}" as well

dougbu: oh, and `"@{<!p /}"` as well as the analogous in HTML (no missing end tag cases there of course)

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,832 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                            new RazorError(
+                                string.Format(errorFormatNormalNotStarted, "!p", CultureInfo.InvariantCulture),
+                                absoluteIndex: 36, lineIndex: 0, columnIndex: 36),
+                        }
+                    },
+                };
+            }
+        }
+
+        public static TheoryData OptOutAttributeDataCSharp
```
dougbu: nit: reads like an attribute is opting out, suggest `OptOut_WithAttributeData_CSharp`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,832 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                        {
+                            new RazorError(string.Format(errorFormatUnclosed, "p", CultureInfo.InvariantCulture),
+                            SourceLocation.Zero)
+                        }
+                    },
+                };
+            }
+        }
+
+        public static TheoryData OptOutAttributeDataHTML
```
dougbu: nit: `OptOut_WithAttributeData_HTML`

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EscapedTagHelpers.DesignTime.cs
```diff
@@ -0,0 +1,56 @@
...
+              "something"
+
+#line default
+#line hidden
+            ;
+            #pragma warning restore 219
+        }
+        #line hidden
+        private InputTagHelper __InputTagHelper = null;
+        private InputTagHelper2 __InputTagHelper2 = null;
```
dougbu: huh?  .cshtml file contains just one un-escaped `<input/>` element.  runtime generation has same issue.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/Resources.resx
NTaylorMullen: As ugly as it seems I'd rather have the whitespace. Most people touching the resx files will be doing it via VS and then will end up with the same comment/same changeset (more work to do in the future).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
NTaylorMullen: Will do a null or empty check inside to kill two birds with one stone :smile:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Removing it, doesn't make any sense lmao.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: We need to be able to handle malformed tags. Therefore the only thing not optional in a tag is the beginning `<` and tag name.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: Same answer as above.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: `Lookahead` throws when provided with a `0` value. Only reason why it's ok here is because we'll conditionally lookahead.

The name lookahead doesn't really make much sense when you're not really looking ahead lol.

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
NTaylorMullen: I'd rather keep Lookahead throwing when its 0, i'll change the condition to be `<= 0`

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
NTaylorMullen: It does :smile:. When EOF is hit it starts pushing `null`s into the `symbols` array, essentially it returns null when it hits EOF which is what every other method does as well.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: lol PFffF gota embrace the ick here.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Doneski

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
dougbu: why?  method is less useful then it could be.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EscapedTagHelpers.DesignTime.cs
NTaylorMullen: 2 input TagHelpers for 1 input element :smile: They both run over the element.

NTaylorMullen:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EscapedTagHelpers.DesignTime.cs
dougbu: :ok: but then why is testing focused on the 20% case?  do you have tests where each helper targets a unique tag name (no overlap -- by far the most common case for the next while)

NTaylorMullen: It also includes the <p> tag which has only a single `TagHelper` and attribute.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -29,22 +31,34 @@ public HtmlElementNameAttribute([NotNull] string tag)
...
+        {
+            if (string.IsNullOrWhiteSpace(tagName))
+            {
+                throw new ArgumentException(
+                    Resources.HtmlElementNameAttribute_ElementNameCannotBeNullOrWhitespace);
+            }
+
+            if (tagName.Contains('!'))
+            {
+                throw new ArgumentException(Resources.FormatHtmlElementNameAttribute_InvalidElementName(tagName, '!'));
```
dougbu: sorry for commenting on a commit. surfacing:
> pass parameter name in and include it when you create the `ArgumentException`.  an `ArgumentException` or subclass with `ParamName == null` is very sad (:crying_cat_face:).

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
...
+        {
+            if (count <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
```
dougbu: sorry for commenting on a commit. surfacing:
> still think `Lookahead()`'s handling of `count==0` is confusing and unhelpful.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
```diff
@@ -0,0 +1,57 @@
...
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidTagNameData))]
+        public void SingleArgument_ThrowsOnInvalidTagNames(string tagName, string expectedExceptionMessage)
+        {
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(() => new HtmlElementNameAttribute(tagName));
```
dougbu: sorry for commenting on a commit. surfacing:
> use `Assert.Throws<ArgumentException>()` overload that takes the parameter name

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -651,20 +686,39 @@ private void ParseQuoted(HtmlSymbolType type)
...
             HtmlSymbol tagName;
-            if (At(HtmlSymbolType.Text))
+            HtmlSymbol bangSymbol = null;
+            HtmlSymbol potentialTagNameSymbol;
+
+            if (At(HtmlSymbolType.Bang))
+            {
+                bangSymbol = CurrentSymbol;
+
+                potentialTagNameSymbol = Lookahead();
```
dougbu: nit: suggest making the parameter to `Lookahead()` required.  otherwise reader needs to hit `F12` to see whether it will return the current, next, or some random symbol.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -275,30 +285,47 @@ private bool CData()
...
+                    // Text tags cannot be targeted by TagHelpers and therefore can't be escaped with '!'.
+                    string.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
                     matched)
                 {
-                    return EndTextTag(solidus, tagBlockWrapper);
+                    return EndTextTag(forwardSlash, tagBlockWrapper);
                 }
                 Accept(_bufferedOpenAngle);
-                Accept(solidus);
+                Accept(forwardSlash);
```
dougbu: so `</!text>` will reach here and then go where?  i.e. does this eventually end up writing `</text>` out?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -710,6 +764,7 @@ private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable
                 return true;
             }
             Accept(_bufferedOpenAngle);
```
dougbu: - so `<!text>` will reach here and then go where?  i.e. does this eventually end up writing `<text>` out?  hoping this is at least consistent with end tag handling
- nit: blank line after closing brace (yeah, unless next line is another closing brace or part of a larger construct)

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -184,5 +184,38 @@ private void OtherParserBlock()
             Initialize(Span);
             NextToken();
         }
+
+        private bool AtBangEscape(int lookahead = 0)
```
dougbu: why is the parameter optional?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -275,30 +285,47 @@ private bool CData()
...
+                else if (At(HtmlSymbolType.Text))
                 {
                     tagName = CurrentSymbol.Content;
                 }
+
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    String.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
+                    // Text tags cannot be targeted by TagHelpers and therefore can't be escaped with '!'.
```
dougbu: nit: on 3rd reading I'm not sure this comment adds much.  may be more useful to say `// Note tagName may contain a '!' escape character.  This ensures `</!text>` doesn't match here.`  might also be the place for code comments on what happens to `</!text>` after it isn't special-cased here.
- similar comment for the similar start tag code later in this file

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.cs
```diff
@@ -75,6 +76,39 @@ protected void Initialize(SpanBuilder span)
...
+            }
+
+            // We add 1 in order to store the current symbol.
+            var symbols = new TSymbol[count + 1];
+            var currentSymbol = CurrentSymbol;
+
+            symbols[0] = currentSymbol;
+
+            // We need to lookforward "count" many times.
```
dougbu: nit: "look forward" (two words)

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
```diff
@@ -0,0 +1,57 @@
...
+                    { " ", nullOrWhitespaceTagNameError },
+                    { Environment.NewLine + " ", nullOrWhitespaceTagNameError },
+                    { null, nullOrWhitespaceTagNameError },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(InvalidTagNameData))]
+        public void SingleArgument_ThrowsOnInvalidTagNames(string tagName, string expectedExceptionMessage)
```
dougbu: nit: `SingleArgumentConstructor_...` and `MultipleArgumentConstructor_...` below

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [Theory]
+        [MemberData(nameof(InvalidTagNameData))]
+        public void MultipleArgument_ThrowsOnInvalidTagNames(string tagName, string expectedExceptionMessage)
+        {
+            // Act & Assert
+            var exception = Assert.Throws<ArgumentException>(() => new HtmlElementNameAttribute("p", tagName));
+            Assert.Equal(exception.Message, expectedExceptionMessage);
+        }
+    }
```
dougbu: nit: covered one and two arguments but may be worth testing `new HtmlElementNameAttribute("div", "span", tagName, "hr")` or similar case

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,877 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var blockFactory = new BlockFactory(factory);
+
+                // documentContent, expectedOutput
+                return new TheoryData<string, MarkupBlock>
+                {
+                    {
```
dougbu: how about `"<!p /"` and `"<!p class=\"btn\" /" cases?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,877 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                    return new MarkupBlock(
+                        factory.EmptyHtml(),
+                        new StatementBlock(
+                            factory.CodeTransition(),
+                            factory.MetaCode("{").Accepts(AcceptedCharacters.None),
+                            insideBuilder()));
+                };
+
+                // documentContent, expectedOutput, expectedErrors
+                return new TheoryData<string, MarkupBlock, RazorError[]>
```
dougbu: notice you're not testing `<!text>` or `</!text>`.  probably need that.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,877 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                            factory.EmptyCSharp().AsStatement(),
+                            factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                        factory.EmptyHtml());
+                };
+
+                // documentContent, expectedOutput, expectedErrors
+                return new TheoryData<string, MarkupBlock, RazorError[]>
+                {
+                    {
+                        "@{<!p></!p>}",
```
dougbu: - add cases like `@{<text></!text>}` too
- should also cover one or two cases w/ a regular span between mismatched tags e.g. `@{<!p>words and spaces</!p>}`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,877 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                            factory.EmptyCSharp().AsStatement(),
+                            factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                        factory.EmptyHtml());
+                };
+
+                // documentContent, expectedOutput, expectedErrors
+                return new TheoryData<string, MarkupBlock, RazorError[]>
+                {
+                    {
+                        "@{<!p class=\"btn\"></!p>}",
```
dougbu: is there now a success case like `@{<!text>some words</!text>}` which leads to output similar to a `@{<!span>some words</!span>}` input?

dougbu:
:watch: liking it and think you're close

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
NTaylorMullen: Not gonna use `nameOf` here to avoid potential misconceptions :smile: .

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Yup, that would generate an actual text tag. Essentially a `<!text>` would be treated like any other HTML tag at this point.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Same answer as above but with an end tag.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
NTaylorMullen: Will change it to not be optional and will change name to `IsBangEscape`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/HtmlElementNameAttributeTest.cs
NTaylorMullen: Will modify the two arg ctor since it's equivalent to X many arg ctor.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Not sure what you mean.

NTaylorMullen: Actually I think I know what you're saying, adding tons of tests for next iteration :smile:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Done.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Done

NTaylorMullen:
Updated... Holy tests :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlElementNameAttribute.cs
```diff
@@ -29,22 +31,37 @@ public HtmlElementNameAttribute([NotNull] string tag)
...
-        public IEnumerable<string> Tags { get; private set; }
+        public IEnumerable<string> Tags { get; }
+
+        private static void ValidateTagName(string tagName)
+        {
+            if (string.IsNullOrWhiteSpace(tagName))
+            {
+                throw new ArgumentException(
+                    Resources.HtmlElementNameAttribute_ElementNameCannotBeNullOrWhitespace,
+                    nameof(tagName));
```
dougbu: do not `throw` an `ArgumentException` using a parameter name of a private method. the parameter that's invalid is either `tag` or `additionalTags`.  pass the correct name into this method.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -275,30 +285,48 @@ private bool CData()
...
                 {
                     tagName = CurrentSymbol.Content;
                 }
+
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    String.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
+                    // Note tagName may contain a '!' escape character. This ensures </!text> doesn't match here. 
+                    // </!text> tags are treated like any other end HTML tag.
```
dougbu: nit: "end HTML tag" -> "escaped HTML end tag"

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -651,20 +687,40 @@ private void ParseQuoted(HtmlSymbolType type)
 
         private bool StartTag(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, IDisposable tagBlockWrapper)
         {
-            // If we're at text, it's the name, otherwise the name is ""
             HtmlSymbol tagName;
```
dougbu: nit: move this down closer to first use, above line 705

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -651,20 +687,40 @@ private void ParseQuoted(HtmlSymbolType type)
...
-                tagName = new HtmlSymbol(CurrentLocation, String.Empty, HtmlSymbolType.Unknown);
+                tagName = potentialTagNameSymbol;
             }
 
             Tuple<HtmlSymbol, SourceLocation> tag = Tuple.Create(tagName, _lastTagStart);
 
-            if (tags.Count == 0 && String.Equals(tag.Item1.Content, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase))
+            if (tags.Count == 0 &&
+                // Note tagName may contain a '!' escape character. This ensures <!text> doesn't match here. 
+                // <!text> tags are treated like any other begin HTML tag.
```
dougbu: nit: "begin HTML tag" -> "escaped HTML start tag" (no "begin" tags)

dougbu:
:watch: still an `ArgumentException` issue

NTaylorMullen:
Updated.

dougbu:
:shipit:

Issue #94, opened 2014-08-22
Implement Order


NTaylorMullen:
Right now if you have multiple TagHelpers that target the same tag element they run in pretty much a random order.  We should implement an attribute that controls what order Tag Helpers are run in.

NTaylorMullen:
53d60159f993c46acc6bf70a3375ef22dbce5338
a0574fb9be8202a260a5a93719e92d7c6d6214ee

PR #277, opened 2015-01-26
Add order attribute and corresponding TagHelperDescriptor property.

NTaylorMullen:
- Added codegen code to pass appropriate order into the TagHelperExecutionContext. Ultimately decided to make order control a runtime feature to allow flexibility in the future.
- Added functional code generation tests and unit tests.
- Tested runtime and parse time features of the OrderAttribute.
- Modified existing codegen test to further validate the Order attribute value.

#94

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITagHelper.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITagHelper.cs
@@ -11,6 +11,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     public interface ITagHelper
     {
         /// <summary>
+        /// Gets the execution order of this <see cref= "ITagHelper" /> relative to others targeting the same element.
+        /// <see cref="ITagHelper"/> instances with lower values are executed first.
+        /// </summary>
+        int Order { get; }
+
+        /// <summary>
         /// Asynchronously executes the <see cref="ITagHelper"/> with the given <paramref name="context"/> and
         /// <paramref name="output"/>.
         /// </summary>
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.IO;
-using System.Text;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -10,6 +10,77 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
     public class TagHelperRunnerTest
     {
+        public static TheoryData TagHelperOrderData
+        {
+            get
+            {
+                // tagHelperOrders, expectedTagHelperOrders
+                return new TheoryData<int[], int[]>
+                {
+                    {
+                        new[] { 1000, int.MaxValue, 0 },
+                        new[] { 0, 1000, int.MaxValue }
+                    },
+                    {
+                        new[] { int.MaxValue, int.MaxValue, int.MinValue },
+                        new[] { int.MinValue, int.MaxValue, int.MaxValue }
+                    },
+                    {
+                        new[] { 0, 0, int.MinValue },
+                        new[] { int.MinValue, 0, 0 }
+                    },
+                    {
+                        new[] { int.MinValue, -1000, 0 },
+                        new[] { int.MinValue, -1000, 0 }
+                    },
+                    {
+                        new[] { 0, 1000, int.MaxValue },
+                        new[] { 0, 1000, int.MaxValue }
+                    },
+                    {
+                        new[] { int.MaxValue, int.MinValue, int.MaxValue, -1000, int.MaxValue, 0 },
+                        new[] { int.MinValue, -1000, 0, int.MaxValue, int.MaxValue, int.MaxValue }
+                    },
+                    {
+                        new[] { 0, 0, 0, 0 },
+                        new[] { 0, 0, 0, 0 }
+                    },
+
+                    {
+                        new[] { 1000, int.MaxValue, 0, -1000, int.MinValue },
+                        new[] { int.MinValue, -1000, 0, 1000, int.MaxValue }
+                    },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(TagHelperOrderData))]
+        public async Task RunAsync_OrdersTagHelpers(
+            int[] tagHelperOrders,
+            int[] expectedTagHelperOrders)
+        {
+            // Arrange
+            var runner = new TagHelperRunner();
+            var executionContext = new TagHelperExecutionContext("p");
+            var processOrder = new List<int>();
+
+            foreach (var order in tagHelperOrders)
+            {
+                var orderedTagHelper = new OrderedTagHelper(order)
+                {
+                    ProcessOrderTracker = processOrder
+                };
+                executionContext.Add(orderedTagHelper);
+            }
+
+            // Act
+            await runner.RunAsync(executionContext);
+
+            // Assert
+            Assert.Equal(expectedTagHelperOrders, processOrder);
+        }
+
         [Fact]
         public async Task RunAsync_ProcessesAllTagHelpers()
         {
@@ -86,6 +157,26 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             public override void Process(TagHelperContext context, TagHelperOutput output)
             {
                 output.Attributes["foo"] = context.AllAttributes["foo"].ToString();
+            }
+        }
+
+        private class OrderedTagHelper : TagHelper
+        {
+            public OrderedTagHelper(int order)
+            {
+                Order = order;
+            }
+
+            public override int Order { get; }
+            public IList<int> ProcessOrderTracker { get; set; }
+
+            public override void Process(TagHelperContext context, TagHelperOutput output)
+            {
+                // If using this class for testing, ensure that ProcessOrderTracker is always set prior to Process
+                // execution.
+                Debug.Assert(ProcessOrderTracker != null);
+
+                ProcessOrderTracker.Add(Order);
             }
         }
     }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelper.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelper.cs
@@ -10,6 +10,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public abstract class TagHelper : ITagHelper
     {
+        /// <inheritdoc />
+        /// <remarks>Default order is <c>0</c>.</remarks>
+        public virtual int Order { get; } = 0;
+
         /// <summary>
         /// Synchronously executes the <see cref="TagHelper"/> with the given <paramref name="context"/> and
         /// <paramref name="output"/>.
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Linq;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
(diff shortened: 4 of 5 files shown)


dougbu:
> Ultimately decided to make order control a runtime feature to allow flexibility in the future.

why add this complexity without a requirement?  the only requirement I'm aware of is preserving the order from the Razor source in the generated HTML.

dougbu:
ignore second sentence about "preserving order".

the requirement in #94 would seem to be more aligned with an `ITagHelper` property of type `int` or a generalization of the `[HtmlElementName]` attribute to cover ordering as well as matching.

NTaylorMullen:
This was designed wayyyy back in the day with @DamianEdwards to be a separate attribute. There was some discussion a while back about combining all `TagHelper` related settings into a single attribute and it was turned down. 

The thought of making the `OrderAttribute` a runtime feature was decided after talking through the implications with @DamianEdwards and waying benefits/cons. Ultimately, since filters and such's Order is implemented at runtime we decided to follow the same pattern; not to mention the added flexibility that's available for the future..

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Used to override <see cref="ITagHelper"/>'s default execution order.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    public class OrderAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="OrderAttribute"/> class.
+        /// </summary>
+        /// <param name="value">The value used to decided <see cref="ITagHelper"/> execution order.</param>
+        public OrderAttribute(long value)
```
pranavkm: `long order`?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -77,6 +80,15 @@ private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors
             return attributeDescriptors;
         }
 
+        private static long GetOrder(Type tagHelperType)
+        {
+            var typeInfo = tagHelperType.GetTypeInfo();
+            var orderAttributes = typeInfo.GetCustomAttributes<OrderAttribute>(inherit: false);
```
pranavkm: I think there's a `GetCustomAttribute` extension method which returns a single value.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -77,6 +80,15 @@ private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors
...
+
+            // If there isn't an attribute specifying the order then use the default order value.
+            return orderAttributes.FirstOrDefault()?.Value ?? DefaultOrder;
```
pranavkm: Just do a null check here rather than have it converted into a `int?`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Used to override <see cref="ITagHelper"/>'s default execution order.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    public class OrderAttribute : Attribute
```
pranavkm: Might want to state higher values get executed first?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
pranavkm: Also, just use an `int`.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -49,6 +50,36 @@ public static TheoryData HtmlCaseData
...
+
+        [Theory]
+        [MemberData(nameof(DescriptorOrderData))]
+        public void CreateDescriptor_GeneratesCorrectOrder(Type tagHelperType, long[] expectedOrders)
+        {
+            // Arrange & Act
+            var descriptors = TagHelperDescriptorFactory.CreateDescriptors(tagHelperType).ToArray();
+
+            // Assert
+            for (var i = 0; i < expectedOrders.Length; i++)
```
pranavkm: `Assert.Equal(expectedOrders, descriptors.Select(d => d.Order));`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
pranavkm: Should this be named differently? OrderAttribute seems too generic a name - `TagHelperExecutionOrderAttribute` maybe?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -395,6 +426,31 @@ private class InheritedNotOverriddenAttributeTagHelper : OverriddenAttributeTagH
         {
         }
 
+        [Order(123)]
```
pranavkm: Could you throw in a tag helper with a negative order?

dougbu:
> making the `OrderAttribute` a runtime feature

1) attributes are design-time features.  what are the requirements leading us to design-time ordering or tag helper execution?
2) we have at least 3 approaches in the stack to control ordering: `Order` properties e.g. `IActionConstraint` and `IOrderedFilter`, explicit calls aka `INestedProvider<T>`, and the `[Display]` attribute's `Order` property (soon; note this one is explicitly about design-time ordering). why create something new for a scenario that seems very similar to existing orderings?

we need a more-inclusive design discussion here.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
NTaylorMullen: /cc @DamianEdwards

NTaylorMullen:
1: In MVC the order is all controlled runtime. Also, having the feature runtime not only falls inline with what we do in other repos but gives us flexibility on exposing it to `TagHelper` authors in the future (we've discussed such things).

2:
@dougbu We couldn't use those attributes in Razor unles we broke them out into their own lightweight assembly. That'd mean the execution order would be an MVC feature which doesn't really make sense in this case. 

There wouldn't be a whole lot of value to creating a standalone binary that contained a unified `Order` attribute that everyone used; BUT maybe a more generic binary that had other commonly used attributes? Not sure if there's a good solution to your concern, but I do see your point.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OrderAttribute.cs
DamianEdwards: RE Name: yeah, we should do something more specific, e.g. `TagHelperOrder`

RE the actual ordering: if it's an order, the lowest number should be first, highest last. If we want the opposite, it should be called `Weight` or `Priority` or similar

NTaylorMullen:
After talking in person, will change `Order` to be a property on `ITagHelper`

NTaylorMullen:
Updated. Just ammended both of the previous commits, the PR has drastically changed since the redesign. Most prior comments don't apply anymore.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITagHelper.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     public interface ITagHelper
     {
         /// <summary>
+        /// The value used to decided <see cref="ITagHelper"/> execution order. Lower values are executed first.
```
dougbu: - decided -> decide (no 'd')
- because the current text is a sentence fragment, better to change wording to `Gets the execution order of this <see cref="ITagHelper"/> relative to others targeting the same element.`
- expand second sentence -- `<see cref="ITagHelper"/> instances with lower values are executed first.`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -10,6 +10,79 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
+            var executionContext = new TagHelperExecutionContext("p");
+            var processOrder = new List<OrderedTagHelper>();
+
+            // Act
+            foreach (var tagHelper in tagHelperAddOrder)
+            {
+                tagHelper.ProcessOrderTracker = processOrder;
+                executionContext.Add(tagHelper);
+            }
+            await runner.RunAsync(executionContext);
```
dougbu: nit: this is the only statement that belongs in `// Act`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -10,6 +10,79 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
+            get
+            {
+                var tagHelper1000 = new OrderedTagHelper(1000);
+                var tagHelperMAX = new OrderedTagHelper(int.MaxValue);
+                var tagHelper0 = new OrderedTagHelper(0);
+                var tagHelperN1000 = new OrderedTagHelper(-1000);
+                var tagHelperMIN = new OrderedTagHelper(int.MinValue);
+
+                // tagHelperAddOrder, expectedTagHelperOrder
+                return new TheoryData<OrderedTagHelper[], OrderedTagHelper[]>
```
dougbu: why isn't this `TheoryData<int[], int[]>`?  this data cannot currently be used in tests running in parallel.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -88,5 +161,25 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+
+        public class OrderedTagHelper : TagHelper
+        {
+            public OrderedTagHelper(int order)
+            {
+                Order = order;
+            }
+
+            public override int Order { get; }
+            public IList<OrderedTagHelper> ProcessOrderTracker { get; set; }
```
dougbu: simplify to `IList<int>` or an `Action<int>`.  nothing needs the tag helpers themselves so either is an improvement.  delegate leaves list management to the test, but that's probably minor.

dougbu:
:watch:

NTaylorMullen:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -10,6 +10,76 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
+            // Arrange
+            var runner = new TagHelperRunner();
+            var executionContext = new TagHelperExecutionContext("p");
+            var processOrder = new List<int>();
+            OrderedTagHelper orderedTagHelper;
+
+            foreach (var order in tagHelperOrders)
+            {
+                orderedTagHelper = new OrderedTagHelper(order);
+                orderedTagHelper.ProcessOrderTracker = processOrder;
```
dougbu: tiny nit: use initialization syntax

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -10,6 +10,76 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
+        [MemberData(nameof(TagHelperOrderData))]
+        public async Task RunAsync_OrdersTagHelpers(
+            int[] tagHelperOrders,
+            int[] expectedTagHelperOrders)
+        {
+            // Arrange
+            var runner = new TagHelperRunner();
+            var executionContext = new TagHelperExecutionContext("p");
+            var processOrder = new List<int>();
+            OrderedTagHelper orderedTagHelper;
```
dougbu: nit: don't bother declaring this outside the `foreach`.  use a `var` variable inside.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -88,5 +158,25 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
...
+            {
+                Order = order;
+            }
+
+            public override int Order { get; }
+            public IList<int> ProcessOrderTracker { get; set; }
+
+            public override void Process(TagHelperContext context, TagHelperOutput output)
+            {
+                // If using this class for testing, ensure that ProcessOrderTracker is always set PRIOR to Process
```
dougbu: nit: don't shout, lowercase "prior"

dougbu:
:shipit:

Issue #271, opened 2015-01-19
Attributes with no value don't invoke IntelliSense.


NTaylorMullen:
If you type: `<custom asp-int-attribute="" />` the document location of the generated `LineMapping` points to the beginning of `asp-int-attribute` and not inside the quotes. In order to get invoked IntelliSense we need to be mapping it correctly.

NTaylorMullen:
7afd78b36a0a1f7bcda276c115bb88b29982a15a

NTaylorMullen:
*Note:* IntelliSense currently does not invoke automatically, you must do ctrl => j once you've written an empty attribute. A good example of this is in the `InputTagHelper` in Mvc where you may expect IntelliSense to be invoked after typing `<input asp-for=""` for your input tag to see potential properties on the pages Model.

Another issue this solved is upon typing in an empty attribute IntelliSense will automatically invoke once it detects a potential match (just like in the C# editor).

Issue #123, opened 2014-09-08
Razor does not correctly parse HTML attributes that have spaces surrounding the equals


NTaylorMullen:
After playing around with the parse tree a bit I noticed that if you were to create HTML elements such as:
```
<div class   =     "foo"></div>
<div class   ="foo"></div>
<div class=    "foo"></div>
```

The Razor parser does not generate an accurate syntax tree (which is used to generate correct "dynamic" attribute code) that can pull the full attribute value out of the HTML output if the inner condition results to false.

Once this has been added we should also support this attribute syntax in Tag Helpers.

yishaigalatzer:
Consider porting back to fix https://aspnetwebstack.codeplex.com/workitem/2149 when this gets fixed

ChrisKlug:
According to the W3.org spec, you are not supposed to be allowed to have a space between the attribute name and = though. Should it still allow that? If it is just a matter of getting the HtmlMarkupParser to accept spaces before and after the = character, I might be able to help out with that if you want me to?

NTaylorMullen:
@ChrisKlug the HTML5 spec allows it. See: http://www.w3.org/TR/html-markup/syntax.html#syntax-attr-double-quoted

>
A double-quoted attribute value is specified by providing the following parts in exactly the following order:
>
1. an attribute name
1. zero or more space characters
1. a single "=" character
1. zero or more space characters
1. a single """ character
1. an attribute value
1. a """ character

ChrisKlug:
You are absolutely correct. I read the http://www.w3.org/html/wg/drafts/html/master/syntax.html#attribute-name-state state-handling wrong. I guess I should stop questioning things now... :)

NTaylorMullen:
Note place for reaction: https://github.com/aspnet/Razor/commit/4c5f9c5360830b52f28aade2447a7e73e44ea39b#diff-58782e50bed88f4f2f70af4835346999R476

ajaybhargavb:
08c8f9f7ba6f9b48d93a4c6ea407a698fb94e4d8

PR #278, opened 2015-01-27
Fix empty attribute projections for TagHelpers.

NTaylorMullen:
- Added TagHelperParseTreeRewriter tests, attempted to cover all empty attribute edge cases.
- Added Codegen tests to validate output and DesignTimeLineMappings.

#271

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.AspNet.Razor.Generator;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
@@ -90,10 +91,15 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 EditHandler = span.EditHandler,
                 Kind = span.Kind
             };
+
+            // Will contain symbols that represent a single attribute value: <input| class="btn"| />
             var htmlSymbols = span.Symbols.OfType<HtmlSymbol>().ToArray();
             var capturedAttributeValueStart = false;
             var attributeValueStartLocation = span.Start;
-            var symbolOffset = 1;
+
+            // The symbolOffset is initialized to 0 to expect worst case: "class=". If a quote is found later on for
+            // the attribute value the symbolOffset is adjusted accordingly.
+            var symbolOffset = 0;
             string name = null;
 
             // Iterate down through the symbols to find the name and the start of the value.
@@ -104,6 +110,11 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
 
                 if (afterEquals)
                 {
+                    // We've captured all leading whitespace, the attribute name, and an equals with an optional 
+                    // quote/double quote. We're now at: " asp-for='|...'" or " asp-for=|..."
+                    // The goal here is to capture all symbols until the end of the attribute. Note this will not 
+                    // consume an ending quote due to the symbolOffset.
+
                     // When symbols are accepted into SpanBuilders, their locations get altered to be offset by the 
                     // parent which is why we need to mark our start location prior to adding the symbol. 
                     // This is needed to know the location of the attribute value start within the document.
@@ -118,13 +129,24 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 }
                 else if (name == null && symbol.Type == HtmlSymbolType.Text)
                 {
+                    // We've captured all leading whitespace prior to the attribute name.
+                    // We're now at: " |asp-for='...'" or " |asp-for=..."
+                    // The goal here is to capture the attribute name.
+
                     name = symbol.Content;
-                    attributeValueStartLocation = SourceLocation.Advance(span.Start, name);
+                    attributeValueStartLocation = SourceLocation.Advance(attributeValueStartLocation, name);
                 }
                 else if (symbol.Type == HtmlSymbolType.Equals)
                 {
-                    // We've found an '=' symbol, this means that the coming symbols will either be a quote
-                    // or value (in the case that the value is unquoted).
+                    Debug.Assert(
+                        name != null,
+                        "Name should never be null here. The parser should guaruntee an attribute has a name.");
+
+                    // We've captured all leading whitespace and the attribute name.
+                    // We're now at: " asp-for|='...'" or " asp-for|=..."
+                    // The goal here is to consume the equal sign and the optional single/double-quote.
+
+                    // The coming symbols will either be a quote or value (in the case that the value is unquoted).
                     // Spaces after/before the equal symbol are not yet supported:
                     // https://github.com/aspnet/Razor/issues/123
 
@@ -134,26 +156,37 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                     SourceLocation symbolStartLocation;
 
                     // Check for attribute start values, aka single or double quote
-                    if (IsQuote(htmlSymbols[i + 1]))
+                    if ((i + 1) < htmlSymbols.Length && IsQuote(htmlSymbols[i + 1]))
                     {
                         // Move past the attribute start so we can accept the true value.
                         i++;
-                        symbolStartLocation = htmlSymbols[i + 1].Start;
+                        symbolStartLocation = htmlSymbols[i].Start;
+
+                        // If there's a start quote then there must be an end quote to be valid, skip it.
+                        symbolOffset = 1;
                     }
                     else
                     {
                         symbolStartLocation = symbol.Start;
-
-                        // Set the symbol offset to 0 so we don't attempt to skip an end quote that doesn't exist.
-                        symbolOffset = 0;
                     }
 
-                    attributeValueStartLocation = symbolStartLocation +
-                                                  span.Start +
-                                                  new SourceLocation(absoluteIndex: 1,
-                                                                     lineIndex: 0,
-                                                                     characterIndex: 1);
+                    attributeValueStartLocation =
+                        span.Start +
+                        symbolStartLocation +
+                        new SourceLocation(absoluteIndex: 1, lineIndex: 0, characterIndex: 1);
+
                     afterEquals = true;
+                }
+                else if (symbol.Type == HtmlSymbolType.WhiteSpace)
+                {
+                    // We're at the start of the attribute, this branch may be hit on the first iterations of 
+                    // the loop since the parser separates attributes with their spaces included as symbols.
+                    // We're at: "| asp-for='...'" or "| asp-for=..."
+                    // Note: This will not be hit even for situations like asp-for  ="..." because the core Razor 
+                    // parser currently does not know how to handle attributes in that format. This will be addressed
+                    // by https://github.com/aspnet/Razor/issues/123.
+
+                    attributeValueStartLocation = SourceLocation.Advance(attributeValueStartLocation, symbol.Content);
                 }
             }
 
--- a/test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
@@ -213,7 +213,41 @@ namespace Microsoft.AspNet.Razor.Test.Generator
                             BuildLineMapping(1094, 30, 18, 5179, 187, 19, 29),
                             BuildLineMapping(1231, 34, 5279, 192, 0, 1),
                         }
-                    }
+                    },
+                    {
+                        "EmptyAttributeTagHelpers",
+                        "EmptyAttributeTagHelpers.DesignTime",
+                        PAndInputTagHelperDescriptors,
+                        new List<LineMapping>
+                        {
+                            BuildLineMapping(documentAbsoluteIndex: 14,
+                                             documentLineIndex: 0,
+                                             generatedAbsoluteIndex: 493,
+                                             generatedLineIndex: 15,
+                                             characterOffsetIndex: 14,
+                                             contentLength: 11),
+                            BuildLineMapping(documentAbsoluteIndex: 62,
+                                             documentLineIndex: 3,
+                                             documentCharacterOffsetIndex: 26,
+                                             generatedAbsoluteIndex: 1289,
+                                             generatedLineIndex: 39,
+                                             generatedCharacterOffsetIndex: 28,
+                                             contentLength: 0),
+                            BuildLineMapping(documentAbsoluteIndex: 122,
+                                             documentLineIndex: 5,
+                                             generatedAbsoluteIndex: 1634,
+                                             generatedLineIndex: 48,
+                                             characterOffsetIndex: 30,
+                                             contentLength: 0),
+                            BuildLineMapping(documentAbsoluteIndex: 88,
+                                             documentLineIndex: 4,
+                                             documentCharacterOffsetIndex: 12,
+                                             generatedAbsoluteIndex: 1789,
+                                             generatedLineIndex: 54,
+                                             generatedCharacterOffsetIndex: 19,
+                                             contentLength: 0)
+                        }
+                    },
                 };
             }
         }
@@ -245,6 +279,7 @@ namespace Microsoft.AspNet.Razor.Test.Generator
                     { "BasicTagHelpers", PAndInputTagHelperDescriptors },
                     { "BasicTagHelpers.RemoveTagHelper", PAndInputTagHelperDescriptors },
                     { "ComplexTagHelpers", PAndInputTagHelperDescriptors },
+                    { "EmptyAttributeTagHelpers", PAndInputTagHelperDescriptors },
                 };
             }
         }
--- a/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
@@ -19,6 +19,77 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData EmptyAttributeTagHelperData
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+
+                // documentContent, expectedOutput
+                return new TheoryData<string, MarkupBlock>
+                {
+                    {
+                        "<p class=\"\"></p>",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                { "class",  new MarkupBlock() }
+                            }))
+                    },
+                    {
+                        "<p class=''></p>",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                { "class",  new MarkupBlock() }
+                            }))
+                    },
+                    {
+                        "<p class=></p>",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                // We expected a markup node here because attribute values without quotes can only ever
+                                // be a single item, hence don't need to be enclosed by a block.
+                                { "class",  factory.Markup("").With(SpanCodeGenerator.Null) },
+                            }))
+                    },
+                    {
+                        "<p class1='' class2= class3=\"\" />",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                { "class1",  new MarkupBlock() },
+                                { "class2",  factory.Markup("").With(SpanCodeGenerator.Null) },
+                                { "class3",  new MarkupBlock() },
+                            }))
+                    },
+                    {
+                        "<p class1=''class2=\"\"class3= />",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                { "class1",  new MarkupBlock() },
+                                { "class2",  new MarkupBlock() },
+                                { "class3",  factory.Markup("").With(SpanCodeGenerator.Null) },
+                            }))
+                    },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(EmptyAttributeTagHelperData))]
+        public void Rewrite_UnderstandsEmptyAttributeTagHelpers(string documentContent, MarkupBlock expectedOutput)
+        {
+            RunParseTreeRewriterTest(documentContent, expectedOutput, new RazorError[0], "p");
+        }
+
         public static TheoryData<string, MarkupBlock, RazorError[]> MalformedTagHelperAttributeBlockData
         {
             get
--- a/test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
+++ b/test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
@@ -1,0 +1,62 @@
+namespace TestOutput
+{
+    using Microsoft.AspNet.Razor.Runtime.TagHelpers;
+    using System;
+    using System.Threading.Tasks;
+
+    public class EmptyAttributeTagHelpers
+    {
+        private static object @__o;
+        private void @__RazorDesignTimeHelpers__()
+        {
+            #pragma warning disable 219
+            string __tagHelperDirectiveSyntaxHelper = null;
+            __tagHelperDirectiveSyntaxHelper = 
+#line 1 "EmptyAttributeTagHelpers.cshtml"
+              "something"
+
+#line default
+#line hidden
+            ;
+            #pragma warning restore 219
+        }
+        #line hidden
+        private InputTagHelper __InputTagHelper = null;
+        private InputTagHelper2 __InputTagHelper2 = null;
+        private PTagHelper __PTagHelper = null;
+        #line hidden
+        public EmptyAttributeTagHelpers()
+        {
+        }
+
+        #pragma warning disable 1998
+        public override async Task ExecuteAsync()
+        {
+            __InputTagHelper = CreateTagHelper<InputTagHelper>();
+            __InputTagHelper.Type = "";
+            __InputTagHelper2 = CreateTagHelper<InputTagHelper2>();
+            __InputTagHelper2.Type = __InputTagHelper.Type;
+#line 4 "EmptyAttributeTagHelpers.cshtml"
+__InputTagHelper2.Checked = ;
+
+#line default
+#line hidden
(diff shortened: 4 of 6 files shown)


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -134,11 +137,14 @@ public static class TagHelperBlockRewriter
                     SourceLocation symbolStartLocation;
 
                     // Check for attribute start values, aka single or double quote
-                    if (IsQuote(htmlSymbols[i + 1]))
+                    if (i + 1 < htmlSymbols.Length && IsQuote(htmlSymbols[i + 1]))
```
dougbu: nit: parentheses around first `i + 1` to avoid readers having to think

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -148,13 +154,17 @@ public static class TagHelperBlockRewriter
                         symbolOffset = 0;
```
dougbu: why is this necessary?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -148,13 +154,17 @@ public static class TagHelperBlockRewriter
...
+                        span.Start +
+                        symbolStartLocation +
+                        new SourceLocation(absoluteIndex: 1, lineIndex: 0, characterIndex: 1);
+
                     afterEquals = true;
                 }
+                else if (symbol.Type == HtmlSymbolType.WhiteSpace)
+                {
+                    attributeValueStartLocation = SourceLocation.Advance(attributeValueStartLocation, symbol.Content);
+                }
```
dougbu: what is this block for?  expect whitespace before and after the equals sign to be silently skipped.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -119,7 +122,7 @@ public static class TagHelperBlockRewriter
                 else if (name == null && symbol.Type == HtmlSymbolType.Text)
                 {
                     name = symbol.Content;
```
dougbu: this method is complex enough to need comments at the top of almost every block explaining the current (or at least expected) situation.  e.g. `// Found the start of the name,` here

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -119,7 +122,7 @@ public static class TagHelperBlockRewriter
...
-                    attributeValueStartLocation = SourceLocation.Advance(span.Start, name);
+                    attributeValueStartLocation = SourceLocation.Advance(attributeValueStartLocation, name);
                 }
                 else if (symbol.Type == HtmlSymbolType.Equals)
                 {
```
dougbu: probably a long-standing bug but this will get very confused if the user typed `<input type==`.  should add `!afterEquals` to the condition.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: this condition isn't in the right spot because whitespace is allowed between the equals sign and the opening quote or value.  IOW this method needs more states in its machine or to make existing states easier to see:
1. before name (skip whitespace if preceding code hasn't already done that)
2. name found (now skip whitespace)
3. equals found (now skip whitespace)
4. bare or quoted value found (perhaps need to skip whitespace after end of value or the matching quote

side question which branches in this method will `checked`, `@multiple`, `type = @type`, and `type = "@type"` hit?  realize `@multiple` isn't currently supported but I'm curious nonetheless.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,77 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                            new MarkupTagHelperBlock("p",
+                            new Dictionary<string, SyntaxTreeNode>
+                            {
+                                { "class1",  new MarkupBlock() },
+                                { "class2",  factory.Markup("").With(SpanCodeGenerator.Null) },
+                                { "class3",  new MarkupBlock() },
+                            }))
+                    },
+                    {
+                        "<p class1=''class2=\"\"class3= />",
```
dougbu: parser doesn't catch the error here?  space between attribute name/value pairs is required, not optional.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
```diff
@@ -0,0 +1,8 @@
+﻿@addtaghelper "something"
+
+<div>
+    <input type= checked=""class="" />
```
dougbu: not positive parser should accept mistake (missing whitespace) here or on line 6

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -213,7 +213,41 @@ public static TheoryData DesignTimeTagHelperTestData
                             BuildLineMapping(1094, 30, 18, 5179, 187, 19, 29),
                             BuildLineMapping(1231, 34, 5279, 192, 0, 1),
                         }
-                    }
+                    },
```
dougbu: don't see tests for anything but bound `string` attributes.  what about unbound attributes and non-`string` bound attributes?  or am I missing something?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Right now Razor doesn't understand attributes with space around the equals as mentioned on lines 131/132 in this method.

1. `checked` will hit the second branch where we find the name. We will also do work to understand that we didn't find a value.
1. `@multiple` wont be in this method since that will be wrapped in a Block and will be handled by the `TryParseBlock` method.
1. `type = @type` will hit the other method `TryParseBlock` because it's understood as having a "dynamic" value.
1. type = "@type" same as above

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: We don't want to accept the end `"` or `'` into the newly created attribute value.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: This block handles the leading whitespace of attributes. For example the block we're parsing is contained within the | characters: `<input| asp-for="..."| />`. Therefore this branch handles the very first characters of whitespace prior to the attribute name.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: true but `symbolOffset` is initialized to `0`. so why set it to `0` again?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: `afterEquals` is the first condition in the tree. It'll never get here :smile:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: lmao, nope, the Razor parser for some odd reason decided to understand this format but not the mainstream format of having optional spaces around the equal sign.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
NTaylorMullen: Outstanding issue of the Razor parser not fully understanding attributes :sob:

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
NTaylorMullen: `<p>` tags age attribute is an `int`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: :+1: will add comments at the top of each branch stating where the cursor currently points and why.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: well, it'll also handle whitespace between the attribute name and the equals sign.

more generally, what does eating the whitespace prior to the name have to do with empty (quoted in valid cases) attribute values?  (and, why did anything work without this block?)

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Ahh good catch, I reversed the setting this time around to be set to 0 instead of 1 initially. :+1:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: You definitely have a point but if there is space between the attribute name and the equals currently it wont even be understood as an attribute by the Razor parser and will then never make it here :disappointed: . That issue is supposed to be addressed by #123

NTaylorMullen:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -104,6 +107,11 @@ public static class TagHelperBlockRewriter
 
                 if (afterEquals)
                 {
+                    // We've captured all leading whitespace, the attribute name, and an equals with an optional 
+                    // quote/double quote. We're now at: " asp-for='|...'" or " asp-for=|..."
+                    // The goal here is to capture all symbols until the end of the attribute. Note this will not 
+                    // consume an ending quote due to the symbolOffset.
```
dougbu: teensy nit: a second note that whitespace between the attribute value or the closing quote and the next attribute or the end of the tag is not included in `htmlSymbols` may be worthwhile.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -118,11 +126,19 @@ public static class TagHelperBlockRewriter
...
+                    // We're now at: " |asp-for='...'" or " |asp-for=..."
+                    // The goal here is to capture the attribute name.
+
                     name = symbol.Content;
-                    attributeValueStartLocation = SourceLocation.Advance(span.Start, name);
+                    attributeValueStartLocation = SourceLocation.Advance(attributeValueStartLocation, name);
                 }
                 else if (symbol.Type == HtmlSymbolType.Equals)
                 {
+                    // We've captured all leading whitespace and the attribute name.
```
dougbu: if you're sure the parser guarantees a name before the equals, add a `Debug.Assert(name != null)` to the top of this block

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -118,11 +126,19 @@ public static class TagHelperBlockRewriter
...
+                    // We're now at: " asp-for|='...'" or " asp-for|=..."
+                    // The goal here is to consume the equal sign and the optional single/double-quote.
+
                     // We've found an '=' symbol, this means that the coming symbols will either be a quote
```
dougbu: previous paragraph now describes the situation better.  Start with "The coming symbols ..."

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -134,27 +150,37 @@ public static class TagHelperBlockRewriter
                     SourceLocation symbolStartLocation;
 
                     // Check for attribute start values, aka single or double quote
-                    if (IsQuote(htmlSymbols[i + 1]))
+                    if ((i + 1) < htmlSymbols.Length && IsQuote(htmlSymbols[i + 1]))
                     {
                         // Move past the attribute start so we can accept the true value.
                         i++;
-                        symbolStartLocation = htmlSymbols[i + 1].Start;
+                        symbolStartLocation = htmlSymbols[i].Start;
```
dougbu: nit: `htmlSymbols[i]` -> `symbol`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -134,27 +150,37 @@ public static class TagHelperBlockRewriter
...
+                        symbolStartLocation +
+                        new SourceLocation(absoluteIndex: 1, lineIndex: 0, characterIndex: 1);
+
                     afterEquals = true;
                 }
+                else if (symbol.Type == HtmlSymbolType.WhiteSpace)
+                {
+                    // We're at the start of the attribute, this branch will always be hit on the first iterations of 
+                    // the loop. We're at: "| asp-for='...'" or "| asp-for=..."
+                    // Note: This will not be hit even for situations like asp-for  ="..." because the core Razor
```
dougbu: nit: ~~even~~

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -134,27 +150,37 @@ public static class TagHelperBlockRewriter
...
+                    attributeValueStartLocation = 
+                        span.Start +
+                        symbolStartLocation +
+                        new SourceLocation(absoluteIndex: 1, lineIndex: 0, characterIndex: 1);
+
                     afterEquals = true;
                 }
+                else if (symbol.Type == HtmlSymbolType.WhiteSpace)
+                {
+                    // We're at the start of the attribute, this branch will always be hit on the first iterations of
```
dougbu: nit: ~~always~~ since the parser (strangely) separates adjacent attribute name/value pairs

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
```diff
@@ -0,0 +1,62 @@
...
+            __InputTagHelper2 = CreateTagHelper<InputTagHelper2>();
+            __InputTagHelper2.Type = __InputTagHelper.Type;
+#line 6 "EmptyAttributeTagHelpers.cshtml"
+  __InputTagHelper2.Checked = ;
+
+#line default
+#line hidden
+            __PTagHelper = CreateTagHelper<PTagHelper>();
+#line 5 "EmptyAttributeTagHelpers.cshtml"
+__PTagHelper.Age = ;
```
dougbu: side note: do we have a bug about catching this error prior to compilation?  pretty easy for the c# writers to detect this case.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        #pragma warning disable 1998
+        public override async Task ExecuteAsync()
+        {
+            __InputTagHelper = CreateTagHelper<InputTagHelper>();
+            __InputTagHelper.Type = "";
+            __InputTagHelper2 = CreateTagHelper<InputTagHelper2>();
+            __InputTagHelper2.Type = __InputTagHelper.Type;
+#line 4 "EmptyAttributeTagHelpers.cshtml"
+__InputTagHelper2.Checked = ;
```
dougbu: this case is valid HTML.  should handle convert empty `bool` values to `true`.  (`checked`, `checked=''`, and `checked='checked'` all mean `true`; everything else is `false`) .

otherwise, we should warn tag helper authors that Razor isn't smart enough to handle HTML setting `bool` tag helper properties correctly.  suggest they bind them as `string`s and handle the meaning themselves.  that's what the MVC tag helpers currently do (and at least one get's it wrong).

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
```diff
@@ -0,0 +1,8 @@
...
+    <p age=''>
+        <input type=""checked= class="" />
```
dougbu: this line contains an error (no value for `checked`).  why does code get generated?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
dougbu: could you open a bug about this lameness please?  or reference existing bug (if any) near the test case.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: right, sorry; it's just whitespace after equals, missing name, and missing value cases that could trip this code up.  and at least some of those (whitespace) won't reach here due to parser issues.  know missing value _does_ lead to code gen.  what about missing name?

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/287

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: The method itself handles a single attribute and therefore could never contain the whitespace. I don't think adding extra comment details here would add value. It might actually be confusing given the context of the method.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
NTaylorMullen: No we don't I figured the compilation error would be good enough.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
NTaylorMullen: The attribute is bound so I'm not sure if the same mechanics apply, even if it is valid in HTML. In the case that you want your `TagHelper` to have dual functionality (augmenting an existing HTML tag and choosing to or not to use it as a `TagHelper`) then what you're stating definitely makes sense. 

As for the MVC `TagHelper`s we already prefix all `TagHelper` bound attributes to avoid ambiguity so thankfully we don't really have the problem (in my opinion). I don't think we should enable a case such as what you're suggesting because we'd then lose IntelliSense/accurate error handling. Definitely discussion worthy though.

**Note: ** Nothing actionable here on this comment (unless I'm missing something).

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
NTaylorMullen: It's not an error according to the Razor parser. it's essentially parsing the attribute as if it was going to be represented like: `<input checked=true />` minus the value.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
dougbu: should not be letting errors we _hope_ the compiler will catch slip through.  have more information on the actual issue in Razor code.  for example, could let user know about specific invalid empty attribute where compiler can just say the generated code is bogus.  particularly important because compiler errors generally highlight the original Razor source and don't show the generated code.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Done on line underneath this.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
dougbu: it's another invalid case that leads to bogus C#.  might work when `checked` is not bound or bound to a `string` property.  but broken in the context of tag helpers.
not immediately actionable but we need to have a design discussion about errors in tag helper bindings.  probably worth an issue to track that discussion.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/289

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
NTaylorMullen: The issue I just created would apply: https://github.com/aspnet/Razor/issues/289

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/EmptyAttributeTagHelpers.DesignTime.cs
dougbu: agreed, not immediately actionable.  but we should have an issue and design discussion about `bool` tag helper properties and how we should handle them.  I suspect the main question is what would be most useful to tag helper authors?  (FYI when writing `<select/>` tag helper, I went out of my way to define `Multiple` as `string` since Razor wouldn't help if it was a `bool`.)

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/EmptyAttributeTagHelpers.cshtml
dougbu: :+1:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: huh?  my point was "always" isn't true

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: agree this method handles a single attribute.  however the content of `htmlSymbols` isn't obvious.  might just need a comment near where it's defined that mentions it may contain preceding whitespace, will always contain the attribute name, will likely contain equals and a value, and will not contain appended whitespace.  or a shorter comment subsetting that info :smile:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: no, my point was "always" is not true

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -134,27 +154,38 @@ public static class TagHelperBlockRewriter
                     SourceLocation symbolStartLocation;
 
                     // Check for attribute start values, aka single or double quote
-                    if (IsQuote(htmlSymbols[i + 1]))
+                    if ((i + 1) < htmlSymbols.Length && IsQuote(htmlSymbols[i + 1]))
                     {
                         // Move past the attribute start so we can accept the true value.
                         i++;
-                        symbolStartLocation = htmlSymbols[i + 1].Start;
+                        symbolStartLocation = symbol.Start;
```
dougbu: sorry, I was wrong here.  `symbol != htmlSymbols[i]` anymore.  you were right to spell `htmlSymbols[i].Start` out.  (if I'd been correct, would have worked to assign `symbolStartLocation` above the `if`.)

dougbu:
:shipit: with one incorrect suggestion I made reversed and (maybe) a code comment or two improved

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: :+1:

Issue #285, opened 2015-01-29
@addtaghelper and @removetaghelper should allow wild cards.


NTaylorMullen:
Currently to add all `TagHelper`s in an assembly you need to do `@addtaghelper "MyAssembly"`. This should change to be less ambiguous to something like: `@addtaghelper "*, MyAssembly"`. Same goes for `@removetaghelper`. This way the user is always specifying some sort of type name.

This also means adding support for `*` in the type name specification for `@addtaghelper`.

NTaylorMullen:
a78ed663d0a46365f4aa779b233385e9d7154c90
32f0858e8f45ca4d9e59e6b053c752b0ad8210f6

PR #286, opened 2015-01-30
Add ability for wildcards in @addtaghelper and @removetaghelper.

NTaylorMullen:
- @addtaghelper and @removetaghelper can now utilize the '*' wild card to represent 0 or more characters.
- Restricted the @addtaghelper to need the TypeName. @addtaghelper "MyAssemblyName" => @addtaghelper "*, MyAssemblyName".
- Also added tests to ensure cases of @addtaghelper "MyAssemblyName" create an error.
- Regenerated all CS files to folow new format.

#285

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.AspNet.Razor.Parser;
 using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.AspNet.Razor.Text;
@@ -113,8 +114,15 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 return false;
             }
 
-            return string.IsNullOrEmpty(lookupInfo.TypeName) ||
-              string.Equals(descriptor.TypeName, lookupInfo.TypeName, StringComparison.Ordinal);
+            // We need to escape the TypePattern so we can choose to only allow specific regex.
+            var escaped = Regex.Escape(lookupInfo.TypePattern);
+
+            // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire 
+            // string. We also replace any '*' or '?' characters with regex to match appropriate content.
+            var pattern = "^" + escaped.Replace(@"\?", ".?").Replace(@"\*", ".*?") + "$";
+            var regex = new Regex(pattern, RegexOptions.Singleline);
+
+            return regex.IsMatch(descriptor.TypeName);
         }
 
         private static LookupInfo GetLookupInfo(TagHelperDirectiveDescriptor directiveDescriptor,
@@ -128,7 +136,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             // "typeName, assemblyName"
             if (lookupStrings == null ||
                 lookupStrings.Any(string.IsNullOrWhiteSpace) ||
-                (lookupStrings.Length != 1 && lookupStrings.Length != 2))
+                lookupStrings.Length != 2)
             {
                 errorSink.OnError(
                     directiveDescriptor.Location,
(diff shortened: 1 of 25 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -113,8 +114,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
...
             }
 
-            return string.IsNullOrEmpty(lookupInfo.TypeName) ||
-              string.Equals(descriptor.TypeName, lookupInfo.TypeName, StringComparison.Ordinal);
+            // We need to escape the typeMatcher so we can choose to only allow specific regex.
+            var escapedTypeMatcher = Regex.Escape(lookupInfo.TypeMatcher);
+
+            // We surround the escapedTypeMatcher with ^ and $ in order ot ensure a regex match matches the entire 
+            // string. We also replace any '*' characters with regex to match any content.
+            var strRegexMatcher = "^" + escapedTypeMatcher.Replace(@"\*", ".*?") + "$";
```
NTaylorMullen: `regexMatcherPattern`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -113,8 +114,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
                 return false;
             }
 
-            return string.IsNullOrEmpty(lookupInfo.TypeName) ||
-              string.Equals(descriptor.TypeName, lookupInfo.TypeName, StringComparison.Ordinal);
+            // We need to escape the typeMatcher so we can choose to only allow specific regex.
+            var escapedTypeMatcher = Regex.Escape(lookupInfo.TypeMatcher);
```
NTaylorMullen: `escaped`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -160,7 +158,7 @@ private class LookupInfo
         {
             public string AssemblyName { get; set; }
 
-            public string TypeName { get; set; }
+            public string TypeMatcher { get; set; }
```
NTaylorMullen: `TypePattern`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -440,7 +598,13 @@ public void DescriptorResolver_DoesNotResolveTypesForNoTypeResolvingLookupText()
         [Theory]
         [InlineData("")]
         [InlineData(null)]
-        public void DescriptorResolver_CreatesErrorIfNullOrEmptyLookupText_DoesNotThrow(string lookupText)
+        [InlineData("*,")]
+        [InlineData(",")]
+        [InlineData(",,,")]
+        [InlineData("SomeType,")]
+        [InlineData("SomeAssembly")]
+        [InlineData("First,Second,Third")]
```
NTaylorMullen: `" , Foo"`
`"Foo, "`
`"Foo , "`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -113,8 +114,15 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
...
             }
 
-            return string.IsNullOrEmpty(lookupInfo.TypeName) ||
-              string.Equals(descriptor.TypeName, lookupInfo.TypeName, StringComparison.Ordinal);
+            // We need to escape the TypePattern so we can choose to only allow specific regex.
+            var escaped = Regex.Escape(lookupInfo.TypePattern);
+
+            // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire 
+            // string. We also replace any '*' or '?' characters with regex to match appropriate content.
+            var pattern = "^" + escaped.Replace(@"\?", ".?").Replace(@"\*", ".*?") + "$";
```
dougbu: @NTaylorMullen late comment: why `".*?"` and not `".*"` for second replacement?

dougbu: - similar comment on first replacement: shouldn't it be `"."` to match any single character?  when the only wildcard characters are `'?'` and `'*'` e.g. in Windows paths, they usually mean "any single character" and "any zero or more characters".
- separately how would Razor authors discover these wildcards?

NTaylorMullen: `".*?` so we don't over-match what users are looking for (lazy evaluation). As for your second comment as to `'?'` during the code review I thought it was optional, which after now doing some more research realize it is not (too used to regex meaning): https://github.com/aspnet/Razor/issues/297

As for Razor author discovery, that's a problem with all Razor directives currently and is why this issue exists: https://github.com/aspnet/Razor/issues/291

yishaigalatzer: docs for now until we get intellisense

PR #298, opened 2015-02-11
Fix1571 - Supporting self-closing tags for taghelpers

ianhong:
@yishaigalatzer, @NTaylorMullen and @dougbu 

This is the design PR for supporting the self closing tags for TagHelpers. It doesn't contain any unit tests, but I already ran all the MVC functional tests, which contains many samples of self-closing tags. 
Once this design PR is signed-off, I'll add unit tests.

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -22,13 +22,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Internal for testing purposes only.
         /// </summary>
-        internal TagHelperExecutionContext(string tagName)
+        internal TagHelperExecutionContext(string tagName, bool selfClosing)
             : this(tagName,
                    uniqueId: string.Empty,
                    executeChildContentAsync: async () => await Task.FromResult(result: true),
                    startWritingScope: () => { },
                    endWritingScope: () => new StringWriter(),
-                   selfClosing: false)
+                   selfClosing: selfClosing)
         {
         }
 
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -44,7 +44,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
         public TagHelperExecutionContext([NotNull] string tagName,
-                                         bool selfClosing, 
+                                         bool selfClosing,
                                          [NotNull] string uniqueId,
                                          [NotNull] Func<Task> executeChildContentAsync,
                                          [NotNull] Action startWritingScope,
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -24,11 +24,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// </summary>
         internal TagHelperExecutionContext(string tagName, bool selfClosing)
             : this(tagName,
+                   selfClosing,
                    uniqueId: string.Empty,
                    executeChildContentAsync: async () => await Task.FromResult(result: true),
                    startWritingScope: () => { },
-                   endWritingScope: () => new StringWriter(),
-                   selfClosing: selfClosing)
+                   endWritingScope: () => new StringWriter())
         {
         }
 
@@ -36,19 +36,19 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Instantiates a new <see cref="TagHelperExecutionContext"/>.
         /// </summary>
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
+        /// <param name="selfClosing">
+        /// <see cref="bool"/> indicating whether or not the tag in the Razor source was self-closing.
+        /// </param>
         /// <param name="uniqueId">An identifier unique to the HTML element this context is for.</param>
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
-        /// <param name="selfClosing">
-        /// The <c>bool</c> indicating whether or not the tag in the Razor soruce is self-closing.
-        /// </param>
         public TagHelperExecutionContext([NotNull] string tagName,
+                                         bool selfClosing, 
                                          [NotNull] string uniqueId,
                                          [NotNull] Func<Task> executeChildContentAsync,
                                          [NotNull] Action startWritingScope,
-                                         [NotNull] Func<TextWriter> endWritingScope, 
-                                         bool selfClosing)
+                                         [NotNull] Func<TextWriter> endWritingScope)
         {
             _tagHelpers = new List<ITagHelper>();
             _executeChildContentAsync = executeChildContentAsync;
@@ -63,7 +63,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// Indicates whether or not the tag of the HTML element this context is for self-closing.
+        /// Gets a value indicating whether or not the tag in the Razor source was self-closing.
         /// </summary>
         public bool SelfClosing { get; }
 
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -27,7 +27,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                    uniqueId: string.Empty,
                    executeChildContentAsync: async () => await Task.FromResult(result: true),
                    startWritingScope: () => { },
-                   endWritingScope: () => new StringWriter())
+                   endWritingScope: () => new StringWriter(),
+                   selfClosing: false)
         {
         }
 
@@ -39,22 +40,32 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag in the Razor soruce is self-closing.
+        /// </param>
         public TagHelperExecutionContext([NotNull] string tagName,
                                          [NotNull] string uniqueId,
                                          [NotNull] Func<Task> executeChildContentAsync,
                                          [NotNull] Action startWritingScope,
-                                         [NotNull] Func<TextWriter> endWritingScope)
+                                         [NotNull] Func<TextWriter> endWritingScope, 
+                                         bool selfClosing)
         {
             _tagHelpers = new List<ITagHelper>();
             _executeChildContentAsync = executeChildContentAsync;
             _startWritingScope = startWritingScope;
             _endWritingScope = endWritingScope;
 
+            SelfClosing = selfClosing;
             AllAttributes = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
             HTMLAttributes = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             TagName = tagName;
             UniqueId = uniqueId;
         }
+
+        /// <summary>
+        /// Indicates whether or not the tag of the HTML element this context is for self-closing.
+        /// </summary>
+        public bool SelfClosing { get; }
 
         /// <summary>
         /// Indicates if <see cref="GetChildContentAsync"/> has been called.
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     {
         private string _content;
         private bool _contentSet;
-        private bool _isTagNameNullOrWhitespaces;
+        private bool _isTagNameNullOrWhitespace;
         private string _tagName;
 
         // Internal for testing
@@ -55,7 +55,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             set
             {
                 _tagName = value;
-                _isTagNameNullOrWhitespaces = string.IsNullOrWhiteSpace(_tagName);
+                _isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(_tagName);
             }
         }
 
@@ -113,12 +113,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s start tag.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces. Otherwise, the
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespace. Otherwise, the
         /// <see cref="string"/> representation of the <see cref="TagHelperOutput"/>'s start tag.</returns>
         public string GenerateStartTag()
         {
             // Only render a start tag if the tag name is not whitespace
-            if (_isTagNameNullOrWhitespaces)
+            if (_isTagNameNullOrWhitespace)
             {
                 return string.Empty;
             }
@@ -151,12 +151,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s <see cref="PreContent"/>.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
-        /// Otherwise, <see cref="PreContent"/>.
-        /// </returns>
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is not <c>null</c> or whitespace
+        /// and <see cref="SelfClosing"/> is <c>true</c>. Otherwise, <see cref="PreContent"/>.</returns>
         public string GeneratePreContent()
         {
-            if (!_isTagNameNullOrWhitespaces && SelfClosing)
+            if (!_isTagNameNullOrWhitespace && SelfClosing)
             {
                 return string.Empty;
             }
@@ -167,12 +166,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s body.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
-        /// Otherwise, <see cref="Content"/>.
-        /// </returns>
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is not <c>null</c> or whitespace
+        /// and <see cref="SelfClosing"/> is <c>true</c>. Otherwise, <see cref="Content"/>.</returns>
         public string GenerateContent()
         {
-            if (!_isTagNameNullOrWhitespaces && SelfClosing)
+            if (!_isTagNameNullOrWhitespace && SelfClosing)
             {
                 return string.Empty;
             }
@@ -183,12 +181,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s <see cref="PostContent"/>.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
-        /// Otherwise, <see cref="PostContent"/>.
-        /// </returns>
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is not <c>null</c> or whitespace
+        /// and <see cref="SelfClosing"/> is <c>true</c>. Otherwise, <see cref="PostContent"/>.</returns>
         public string GeneratePostContent()
         {
-            if (!_isTagNameNullOrWhitespaces && SelfClosing)
+            if (!_isTagNameNullOrWhitespace && SelfClosing)
             {
                 return string.Empty;
             }
@@ -199,11 +196,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s end tag.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces. Otherwise, the
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespace. Otherwise, the
         /// <see cref="string"/> representation of the <see cref="TagHelperOutput"/>'s end tag.</returns>
         public string GenerateEndTag()
         {
-            if (SelfClosing || _isTagNameNullOrWhitespaces)
+            if (SelfClosing || _isTagNameNullOrWhitespace)
             {
                 return string.Empty;
             }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -16,6 +16,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     {
         private string _content;
         private bool _contentSet;
+        private bool _isTagNameNullOrWhitespaces;
+        private string _tagName;
 
         // Internal for testing
         internal TagHelperOutput(string tagName)
@@ -44,7 +46,18 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <remarks>
         /// A whitespace or <c>null</c> value results in no start or end tag being rendered.
         /// </remarks>
-        public string TagName { get; set; }
+        public string TagName
+        {
+            get
+            {
+                return _tagName;
+            }
+            set
+            {
+                _tagName = value;
+                _isTagNameNullOrWhitespaces = string.IsNullOrWhiteSpace(_tagName);
+            }
+        }
 
         /// <summary>
         /// The HTML element's pre content.
@@ -88,7 +101,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// Indicates whether or not the tag is self closing.
+        /// Indicates whether or not the tag is self-closing.
         /// </summary>
         public bool SelfClosing { get; set; }
 
@@ -100,12 +113,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s start tag.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>string.Empty</c> or whitespace. Otherwise, the
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces. Otherwise, the
         /// <see cref="string"/> representation of the <see cref="TagHelperOutput"/>'s start tag.</returns>
         public string GenerateStartTag()
         {
             // Only render a start tag if the tag name is not whitespace
-            if (string.IsNullOrWhiteSpace(TagName))
+            if (_isTagNameNullOrWhitespaces)
             {
                 return string.Empty;
             }
@@ -138,12 +151,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s <see cref="PreContent"/>.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="SelfClosing"/> is <c>true</c>. <see cref="PreContent"/> 
-        /// otherwise.
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
+        /// Otherwise, <see cref="PreContent"/>.
         /// </returns>
         public string GeneratePreContent()
         {
-            if (SelfClosing)
+            if (!_isTagNameNullOrWhitespaces && SelfClosing)
             {
                 return string.Empty;
             }
@@ -154,11 +167,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s body.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="SelfClosing"/> is <c>true</c>. <see cref="Content"/> otherwise.
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
+        /// Otherwise, <see cref="Content"/>.
         /// </returns>
         public string GenerateContent()
         {
-            if (SelfClosing)
+            if (!_isTagNameNullOrWhitespaces && SelfClosing)
             {
                 return string.Empty;
             }
@@ -169,12 +183,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s <see cref="PostContent"/>.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="SelfClosing"/> is <c>true</c>. <see cref="PostContent"/> 
-        /// otherwise.
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
+        /// Otherwise, <see cref="PostContent"/>.
         /// </returns>
         public string GeneratePostContent()
         {
-            if (SelfClosing)
+            if (!_isTagNameNullOrWhitespaces && SelfClosing)
             {
                 return string.Empty;
             }
@@ -185,11 +199,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s end tag.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>string.Empty</c> or whitespace. Otherwise, the
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces. Otherwise, the
         /// <see cref="string"/> representation of the <see cref="TagHelperOutput"/>'s end tag.</returns>
         public string GenerateEndTag()
         {
-            if (SelfClosing || string.IsNullOrWhiteSpace(TagName))
+            if (SelfClosing || _isTagNameNullOrWhitespaces)
             {
                 return string.Empty;
             }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
@@ -24,7 +24,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 executionContext.AllAttributes,
                 executionContext.UniqueId,
                 executionContext.GetChildContentAsync);
-            var tagHelperOutput = new TagHelperOutput(executionContext.TagName, executionContext.HTMLAttributes);
+            var tagHelperOutput = new TagHelperOutput(executionContext.TagName, executionContext.HTMLAttributes)
+            {
+                SelfClosing = executionContext.SelfClosing,
+            };
             var orderedTagHelpers = executionContext.TagHelpers.OrderBy(tagHelper => tagHelper.Order);
 
             foreach (var tagHelper in orderedTagHelpers)
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
@@ -31,18 +31,23 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag of this scope is self-closing.
+        /// </param>
         /// <returns>A <see cref="TagHelperExecutionContext"/> to use.</returns>
         public TagHelperExecutionContext Begin([NotNull] string tagName,
                                                [NotNull] string uniqueId,
                                                [NotNull] Func<Task> executeChildContentAsync,
                                                [NotNull] Action startWritingScope,
-                                               [NotNull] Func<TextWriter> endWritingScope)
+                                               [NotNull] Func<TextWriter> endWritingScope,
+                                               bool selfClosing)
         {
             var executionContext = new TagHelperExecutionContext(tagName,
                                                                  uniqueId,
                                                                  executeChildContentAsync,
                                                                  startWritingScope,
-                                                                 endWritingScope);
+                                                                 endWritingScope,
+                                                                 selfClosing);
 
             _executionScopes.Push(executionContext);
 
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
@@ -27,27 +27,27 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Starts a <see cref="TagHelperExecutionContext"/> scope.
         /// </summary>
         /// <param name="tagName">The HTML tag name that the scope is associated with.</param>
+        /// <param name="selfClosing">
+        /// <see cref="bool"/> indicating whether or not the tag of this scope is self-closing.
+        /// </param>
         /// <param name="uniqueId">An identifier unique to the HTML element this scope is for.</param>
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
-        /// <param name="selfClosing">
-        /// The <c>bool</c> indicating whether or not the tag of this scope is self-closing.
-        /// </param>
         /// <returns>A <see cref="TagHelperExecutionContext"/> to use.</returns>
         public TagHelperExecutionContext Begin([NotNull] string tagName,
+                                               bool selfClosing,
                                                [NotNull] string uniqueId,
                                                [NotNull] Func<Task> executeChildContentAsync,
                                                [NotNull] Action startWritingScope,
-                                               [NotNull] Func<TextWriter> endWritingScope,
-                                               bool selfClosing)
+                                               [NotNull] Func<TextWriter> endWritingScope)
         {
             var executionContext = new TagHelperExecutionContext(tagName,
+                                                                 selfClosing,
                                                                  uniqueId,
                                                                  executeChildContentAsync,
                                                                  startWritingScope,
-                                                                 endWritingScope,
-                                                                 selfClosing);
+                                                                 endWritingScope);
 
             _executionScopes.Push(executionContext);
 
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
         {
             var tagHelperDescriptors = chunk.Descriptors;
 
-            RenderBeginTagHelperScope(chunk.TagName, chunk.Children);
+            RenderBeginTagHelperScope(chunk.TagName, chunk.Children, chunk.SelfClosing);
 
             RenderTagHelpersCreation(chunk);
 
@@ -89,7 +89,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
             return "__" + descriptor.TypeName.Replace('.', '_');
         }
 
-        private void RenderBeginTagHelperScope(string tagName, IList<Chunk> children)
+        private void RenderBeginTagHelperScope(string tagName, IList<Chunk> children, bool selfClosing)
         {
             // Scopes/execution contexts are a runtime feature.
             if (_designTimeMode)
@@ -134,6 +134,8 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
                    .Write(_tagHelperContext.StartWritingScopeMethodName)
                    .WriteParameterSeparator()
                    .Write(_tagHelperContext.EndWritingScopeMethodName)
+                   .WriteParameterSeparator()
+                   .WriteBooleanLiteral(selfClosing)
                    .WriteEndMethodInvocation();
         }
 
@@ -466,8 +468,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
 
             _writer.Write(ExecutionContextVariableName)
                    .Write(".")
-                   .Write(_tagHelperContext.ExecutionContextOutputPropertyName)
-                   .Write(" = ")
+                   .WriteStartAssignment(_tagHelperContext.ExecutionContextOutputPropertyName)
                    .WriteStartInstanceMethodInvocation(RunnerVariableName,
                                                        _tagHelperContext.RunnerRunAsyncMethodName);
 
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
         {
             var tagHelperDescriptors = chunk.Descriptors;
 
-            RenderBeginTagHelperScope(chunk.TagName, chunk.Children, chunk.SelfClosing);
+            RenderBeginTagHelperScope(chunk.TagName, chunk.SelfClosing, chunk.Children);
 
             RenderTagHelpersCreation(chunk);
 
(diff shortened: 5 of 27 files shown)


msotclas:
Hi __@ianhong__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Youngjune Hong). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -54,6 +54,16 @@ internal TagHelperExecutionContext(string tagName)
             HTMLAttributes = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             TagName = tagName;
             UniqueId = uniqueId;
+            SelfClosing = false;
```
yishaigalatzer: Don't set the default value of a bool.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -24,7 +24,10 @@ public class TagHelperRunner
                 executionContext.AllAttributes,
                 executionContext.UniqueId,
                 executionContext.GetChildContentAsync);
-            var tagHelperOutput = new TagHelperOutput(executionContext.TagName, executionContext.HTMLAttributes);
+            var tagHelperOutput = new TagHelperOutput(executionContext.TagName, executionContext.HTMLAttributes)
+            {
+                SelfClosing = executionContext.SelfClosing,
```
yishaigalatzer: Is there a scenario where we new up output without setting the `SelfClosing` property?

yishaigalatzer:
I don't see what changes the rendering of the HTML in this PR, what am I missing? at least a basic set of tests will be enlightening here

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
ianhong: I don't understand the question, but in the original codes, selfclosing was always 'false', and each taghelper like input could set it true if necessary. In this fix, selfclosing is set based on the original user input tags from cshtml, and individual taghelpers can change it as needed.

ianhong:
Basically, the information flows from user cshtml files -> TagHelperBlock.SelfClosing->TagHelperChunk.SelfClosing -> CSharpTagHelperCodeRenderer, and CSharpTagHelperCodeRenderer generates C# codes setting TagHelperExecutionContext.SelfClosing. I'll add some unit tests tomorrow to clarify this.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
NTaylorMullen: @yishaigalatzer no there's not with this fix in mind. We will always need to be explicit as to if the output is self closing or not since that info will come from what the user originally typed.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -456,6 +457,22 @@ private void RenderTagOutput(string tagOutputMethodName)
...
+            if (_designTimeMode)
+            {
+                return;
+            }
+            if (selfClosing == false)
+            {
+                return;
+            }
+            _writer.Write(ExecutionContextVariableName)
+                   .Write(".SelfClosing")
```
NTaylorMullen: You need a configuration setup for the name `SelfClosing`: `ExecutionContextSelfClosingPropertyName`.

This can be done here: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -456,6 +457,22 @@ private void RenderTagOutput(string tagOutputMethodName)
                    .WriteEndMethodInvocation();
         }
 
+        private void RenderSelfClosing(bool selfClosing)
+        {
+            if (_designTimeMode)
+            {
+                return;
+            }
+            if (selfClosing == false)
```
NTaylorMullen: Lets not assume `selfClosing = false` for the `TagHelperExecutionContext` since it's replaceable. Always set the value.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -456,6 +457,22 @@ private void RenderTagOutput(string tagOutputMethodName)
...
+            {
+                return;
+            }
+            _writer.Write(ExecutionContextVariableName)
```
NTaylorMullen:
```C#
_writer.Write(ExecutionContextVariableName)
       .Write(".")
       .WriteStartAssignment(ExecutionContextSelfClosingPropertyName)
       .WriteBooleanLiteral(selfClosing)
       .WriteLine(";");
```

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -68,6 +70,11 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         }
 
         /// <summary>
+        /// Indicates whether or not the tag is self closing.
+        /// </summary>
+        public bool SelfClosing { get; }
```
NTaylorMullen: `{ get; set; }` for the builder

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -334,7 +334,7 @@ private static string GetTagName(Block tagBlock)
             return textSymbol.Type == HtmlSymbolType.WhiteSpace ? null : textSymbol.Content;
         }
 
-        private static bool IsSelfClosing(Block beginTagBlock)
+        internal static bool IsSelfClosing(Block beginTagBlock)
```
NTaylorMullen: Move this method into the `TagHelperBlockRewriter` and make it `private` again. The one call to this method in this file can just pull the `SelfClosing` value off of the builder.

NTaylorMullen:
Design looks good, let me know when I'm good to go over it in more detail. :watch:

ianhong:
@NTaylorMullen @dougbu and @yishaigalatzer 

I updated this PR per comments and the design discussion. I'd like to update MVC Tag Helpers first to verify the decision while you are reviewing this design PR. I'll send a PR for MVC first.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -143,7 +143,7 @@ public string GenerateStartTag()
         /// </returns>
         public string GeneratePreContent()
         {
-            if (SelfClosing)
+            if (!string.IsNullOrWhiteSpace(TagName) && SelfClosing)
```
dougbu: `TagName` value does _not_ change content and should not be referenced here.  `TagName==null` means "generate the content but not the wrapping start and end tags".

dougbu: - put another way, revert changes to this file
- it may be worth making a different change: update the doc comments for the `SelfClosing` property to include `<remarks/>` that mention `PreContent`, `Content`, and `PostContent` are ignored when `SelfClosing` is `true`.

dougbu: reminded of requirement here offline.  change is :ok: but I still recommend the `<remarks/>` addition.

ianhong: Thanks Doug, I'll make sure add the remark when I send actual PR for this.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -57,6 +57,15 @@ internal TagHelperExecutionContext(string tagName)
         }
 
         /// <summary>
+        /// Indicates whether or not the tag is self closing.
+        /// </summary>
+        public bool SelfClosing
```
dougbu: 1. thought we agreed to name this `OriginalSelfClosing` or `WasOriginallySelfClosing`?
2. "the tag" is a bit ambiguous.  probably want to use "the start tag of the HTML element this context is for". ("the HTML element this context is for" appears elsewhere in this class.)  bit ungainly but at least clear.
3. need to describe original-ness of this value.  definitely include the word "original" in the `<summary/>`.  `<remarks/>` that mention `Output.SelfClosing` reflects changes `ITagHelper` implementations have made may also help.
4. don't think this should be settable since it describes the original start tag.  pass a value in to the `TagHelperExecutionContext` constructor.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
dougbu: since a `TagHelperOutput` is never constructed without immediately setting `SelfClosing`, the value should be a parameter to the constructor.  that way we know for sure nothing is creating these things without providing that setting.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: @dougbu 

1. The `TagHelperExecutionContext` represents the current state of the HTML tag it represents (and shouldn't be changed) therefore `SelfClosing` makes sense here. As for discussing another name, we spit balled the idea a bit but decided to leave it as is from my recollection.
1. :+1: 
1. I don't know if I'd go as far as mentioning the `TagHelperOutput` here because the `TagHelperExecutionContext` has no direct correlation to the `TagHelperOutput` object. The `TagHelperRunner` is the mechanism responsible for joining the two. As for adding a remark stating that this is the "original" value, it should never change since the `TagHelperExecutionContext` is by definition the metadata object for the HTML element, addressing 4 should solve this confusion.
1. I agree, it shouldn't be settable. Does kind of suck though since this means it's an extra property we need to pass into `TagHelperScopeManager.Begin` (that constructs the `TagHelperExecutionContext`).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
NTaylorMullen: I disagree with making this a a constructor required property for the `TagHelperOutput`, the `TagHelperOutput` is a living object which can be modified by users and therefore `SelfClosing` is malleable. That being said we did decide not to touch the property for our own `TagHelper`s in MVC; doesn't mean other `TagHelper` authors will not though. I can even see people constructing their own `TagHelperOutput`s to build complex structures; I'd hate to require them to provide if it's self closing on construction.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -158,7 +158,7 @@ public string GeneratePreContent()
         /// </returns>
         public string GenerateContent()
         {
-            if (SelfClosing)
+            if (!string.IsNullOrWhiteSpace(TagName) && SelfClosing)
```
NTaylorMullen: After looking at this, I think it'd be best to capture if `TagName` is `null` or whitespace in the `TagName`'s setter. This way this line (and all others in the generate methods) could be: `if (!_renderOuterTags && SelfClosing)`. *Note:* I suck at naming, feel free to pick a better name than `_renderOuterTags`.

I bring this up because each of the generate methods is called per HTML element that happens to have `TagHelper`s assocaited with it on the page; seems wasteful to be iterating over the `TagName` looking for `null`/whitespace 3x per HTML element when in reality it can be done once.

ianhong: @NTaylorMullen thanks. that is a good point. How about _isTagNameNullOrEmpty?? I'll update this code when I do actual PR.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -456,6 +457,21 @@ private void RenderTagOutput(string tagOutputMethodName)
...
+        {
+            if (_designTimeMode)
+            {
+                return;
+            }
+
+            _writer.Write(ExecutionContextVariableName)
+                   .Write(".")
+                   .Write(_tagHelperContext.ExecutionContextSelfClosingPropertyName)
+                   .Write(" = ")
```
NTaylorMullen: Don't need this if you make the `Write` above a `WriteStartAssignment`

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
```diff
@@ -150,5 +151,10 @@ public GeneratedTagHelperContext()
         /// Contains information about in-scope tag helpers, HTML attributes, and the tag helpers' output.
         /// </remarks>
         public string ExecutionContextTypeName { get; set; }
+
+        /// <summary>
+        /// The name of the property indicating whether the tag halper in the scope is self-closing or not.
```
NTaylorMullen: halppper

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
ianhong: I copied that codes from a different method in the same class I'll update the original one too when I do the actual PR. :-)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: sure that's not bad.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -29,5 +29,10 @@ public class TagHelperChunk : ChunkBlock
         /// The HTML tag name.
         /// </summary>
         public string TagName { get; set; }
+
+        /// <summary>
+        /// Indicates whether or not the tag is self closing.
+        /// </summary>
+        public bool SelfClosing { get; set; }
```
dougbu: everywhere along this chain except in `TagHelperOutput`, `SelfClosing` should be a get-only auto-property that's set using a new constructor parameter.  this makes the property's nature much more clear.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -40,12 +40,14 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         public TagHelperBlockBuilder(string tagName,
                                      SourceLocation start,
                                      IDictionary<string, SyntaxTreeNode> attributes,
+                                     bool selfClosing,
```
dougbu: this is the pattern you should apply everywhere.  the only issue in `TagHelperBlockBuilder` is that `SelfClosing` should not have a setter.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
NTaylorMullen: All chunks are `get; set;`, just plain poco objects.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -94,6 +95,13 @@ public static class TagHelperBlockRewriter
             return attributes;
         }
 
+        private static bool IsSelfClosing(Block beginTagBlock)
+        {
+            var childSpan = beginTagBlock.Children.Last() as Span;
```
dougbu: use `beginTagBlock.FindLastDescendentSpan()` to avoid problems when the tree is more than one level deep

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
dougbu: I didn't say the property should lack a setter in `TagHelperOutput`.  for that class, my comment is about setting the initial value -- a requirement we should make explicit

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: @NTaylorMullen
3. `TagHelperExecutionContext` has an `Output` property of type `TagHelperOutput`

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
dougbu: ick; but :ok: drop the requirement for the plain POCO objects but impose it where there's already an explicit constructor

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
NTaylorMullen: I understand that. I really don't feel like we should though. `SelfClosing` is a very unimportant property as far as `TagHelperOutput` is concerned. If anything it's just additive. My reasoning in one way falls in line with what we have in MVC for `TagBuilder`, the default `ToString()` doesn't require any mention of rendering the tag self closing or not (`TagRenderMode.Normal`). We follow the same approach here were it's defaulted to not being self-closing and you can change later if you want.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: @dougbu fair point, We probably shouldn't have put that on the `TagHelperExecutionContext` in hindsight. It's only there to output what happened from the result of running the `TagHelper`s. Ultimately we should have created a helper class to handle that responsibility.

NTaylorMullen: My comment also applies tothings like `ExecuteChildContentAsync` we probably shoudl have made a Run overload on the `TagHelperRunner` that handled that.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
dougbu: @NTaylorMullen there are no parallels to `TagBuilder` here because self-closing has no "original" source in the `TagBuilder` case -- it's completely up to the calling code.  in the `TagHelperOutput` case, self-closing comes from one and only one place in 90% of the cases.  yes `TagHelperOutput.SelfClosing` can be overridden (primarily for unusual wrapping scenarios) but there's a clear original value provided in the Razor source file.

ianhong:
@NTaylorMullen  and @dougbu 

Updated per comments as we concluded this morning.

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
```diff
@@ -30,6 +30,7 @@ public GeneratedTagHelperContext()
             ExecutionContextAddTagHelperAttributeMethodName = "AddTagHelperAttribute";
             ExecutionContextAddHtmlAttributeMethodName = "AddHtmlAttribute";
             ExecutionContextOutputPropertyName = "Output";
+            ExecutionContextSelfClosingPropertyName = "SelfClosing";
```
NTaylorMullen: Don't need any changes in this file anymore.

NTaylorMullen:
Design looks good from my end.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -29,5 +48,10 @@ public class TagHelperChunk : ChunkBlock
         /// The HTML tag name.
         /// </summary>
         public string TagName { get; set; }
+
+        /// <summary>
+        /// Indicates whether or not the tag is self-closing.
+        /// </summary>
+        public bool SelfClosing { get; set; }
```
dougbu: this property should not have a setter.  for that matter, does this class need any setters?

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
```diff
@@ -150,5 +150,10 @@ public GeneratedTagHelperContext()
         /// Contains information about in-scope tag helpers, HTML attributes, and the tag helpers' output.
         /// </remarks>
         public string ExecutionContextTypeName { get; set; }
+
+        /// <summary>
+        /// The name of the property indicating whether the tag helper in the scope is self-closing or not.
+        /// </summary>
+        public string ExecutionContextSelfClosingPropertyName { get; set; }
```
dougbu: unused property.  revert changes to this file.

dougbu:
design / API generally fine, modulo a couple of high-level comments

please let us know when you're ready for lower-level comments on the implementation
- suggestions before you make that call: reread XML doc comments and make sure all the words are useful and consistent.  and consistently use named parameters for `true` and `false` -- makes it easier to read even test code.

ianhong:
@dougbu and @NTaylorMullen 

I updated this PR for actual review, and ccing @yishaigalatzer

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
```diff
@@ -177,27 +177,43 @@ public MarkupBlock(IEnumerable<SyntaxTreeNode> children)
 
     public class MarkupTagHelperBlock : TagHelperBlock
     {
-        public MarkupTagHelperBlock(string tagName)
-            : this(tagName, new Dictionary<string, SyntaxTreeNode>())
+        public MarkupTagHelperBlock(string tagName, bool selfClosing = false)
+            : this(tagName, new Dictionary<string, SyntaxTreeNode>(), selfClosing)
         {
```
ianhong: You might prefer explicitly providing the selfclosing parameter values from each test over this default parameter. I tried that but I found >260 build errors. I don't think it is worth updating that many test data. But, let me know if you have a strong opinion on this.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -22,12 +22,13 @@ public class TagHelperExecutionContext
         /// <summary>
         /// Internal for testing purposes only.
         /// </summary>
-        internal TagHelperExecutionContext(string tagName)
+        internal TagHelperExecutionContext(string tagName, bool selfClosing)
             : this(tagName,
```
ianhong: I'm still not sure if I should keep this signature like TagHelperExecutionContext(string tagName), and initialize selfClosing with its the default value (false) just like any other parameters. For now, I changed its signature to simplify some test codes. But, let me know.

dougbu: it's `internal` so changing the signature wasn't required.  but you've done the work and it's not harmful.  I vote for leaving this change in.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -39,17 +40,22 @@ internal TagHelperExecutionContext(string tagName)
...
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag in the Razor soruce is self-closing.
+        /// </param>
         public TagHelperExecutionContext([NotNull] string tagName,
                                          [NotNull] string uniqueId,
                                          [NotNull] Func<Task> executeChildContentAsync,
                                          [NotNull] Action startWritingScope,
-                                         [NotNull] Func<TextWriter> endWritingScope)
+                                         [NotNull] Func<TextWriter> endWritingScope, 
+                                         bool selfClosing)
```
dougbu: nit: might be better to place this parameter immediately after `tagName` in the signature.  keeps everything from the Razor source together.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -57,6 +63,11 @@ internal TagHelperExecutionContext(string tagName)
         }
 
         /// <summary>
+        /// Indicates whether or not the tag of the HTML element this context is for self-closing.
```
dougbu: nits: `/// Gets a value indicating whether or not the tag in the Razor source was self-closing.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -39,17 +40,22 @@ internal TagHelperExecutionContext(string tagName)
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag in the Razor soruce is self-closing.
```
dougbu: - don't use `<c>` to refer to a type: `<see cref="bool"/>` instead
- leave out "The"
- nit: "is" -> "was"

dougbu: and there's a typo in "source"

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -16,6 +16,8 @@ public class TagHelperOutput
     {
         private string _content;
         private bool _contentSet;
+        private bool _isTagNameNullOrWhitespaces;
```
dougbu: remove final "s" from field name; can have numerous spaces in some whitespace

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -138,12 +151,12 @@ public string GenerateStartTag()
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s <see cref="PreContent"/>.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="SelfClosing"/> is <c>true</c>. <see cref="PreContent"/> 
-        /// otherwise.
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces.
+        /// Otherwise, <see cref="PreContent"/>.
```
dougbu: incorrect.  should say
`<c>string.Empty</c> if <see cref="GenerateStartTag"/> will return a non-empty `string` and <see cref="SelfClosing"/> is <c>true</c>`
or perhaps
`<c>string.Empty</c> if <see cref="TagName"/> is not <c>null</c> or whitespace and <see cref="SelfClosing"/> is <c>true</c>`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -100,12 +113,12 @@ public bool ContentSet
         /// <summary>
         /// Generates the <see cref="TagHelperOutput"/>'s start tag.
         /// </summary>
-        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>string.Empty</c> or whitespace. Otherwise, the
+        /// <returns><c>string.Empty</c> if <see cref="TagName"/> is <c>null</c> or whitespaces. Otherwise, the
```
dougbu: always just "whitespace". check for other plural cases too.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: fix other copies of this doc comment

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +31,23 @@ public TagHelperScopeManager()
...
+        /// The <c>bool</c> indicating whether or not the tag of this scope is self-closing.
+        /// </param>
         /// <returns>A <see cref="TagHelperExecutionContext"/> to use.</returns>
         public TagHelperExecutionContext Begin([NotNull] string tagName,
                                                [NotNull] string uniqueId,
                                                [NotNull] Func<Task> executeChildContentAsync,
                                                [NotNull] Action startWritingScope,
-                                               [NotNull] Func<TextWriter> endWritingScope)
+                                               [NotNull] Func<TextWriter> endWritingScope,
+                                               bool selfClosing)
```
dougbu: constructor comments from `TagHelperExecutionContext` apply here

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -89,7 +89,7 @@ internal static string GetVariableName(TagHelperDescriptor descriptor)
             return "__" + descriptor.TypeName.Replace('.', '_');
         }
 
-        private void RenderBeginTagHelperScope(string tagName, IList<Chunk> children)
+        private void RenderBeginTagHelperScope(string tagName, IList<Chunk> children, bool selfClosing)
```
dougbu: move `selfClosing` parameter before `children` parameter

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -12,6 +12,29 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler
...
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag of the tag helpers HTML element is self-closing.
+        /// </param>
+        /// <param name="descriptors">
+        /// The <see cref="TagHelperDescriptor"/>s associated with this tag helpers HTML element.
+        /// </param>
+        public TagHelperChunk(
+            string tagName,
+            IDictionary<string, Chunk> attributes,
+            bool selfClosing,
```
dougbu: nit: it's less important here but `selfClosing` should be before `attributes`. keep `tagName` and `selfClosing` together (and in that order) everywhere we can

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -29,5 +52,10 @@ public class TagHelperChunk : ChunkBlock
         /// The HTML tag name.
         /// </summary>
         public string TagName { get; set; }
+
+        /// <summary>
+        /// Indicates whether or not the tag of the tag helpers HTML element is self-closing.
```
dougbu: - always "Gets" or "Gets or sets" even though existing Razor code is not-so-great about this
- use the same wording for this property as in `TagHelperExecutionContext`

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -12,6 +12,29 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler
     public class TagHelperChunk : ChunkBlock
     {
         /// <summary>
+        /// Instantiates a new <see cref="TagHelperChunk"/>.
+        /// </summary>
+        /// <param name="tagName">The tag name associated with the tag helpers HTML element.</param>
+        /// <param name="attributes">The attributes associated with the tag helpers HTML element.</param>
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag of the tag helpers HTML element is self-closing.
```
dougbu: - use the same doc for this parameter (probably most parameters) as in `TagHelperExecutionContext`

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
```diff
@@ -76,12 +76,8 @@ public override void GenerateStartBlockCode(Block target, CodeGeneratorContext c
...
 
             context.CodeTreeBuilder.StartChunkBlock(
-                new TagHelperChunk
-                {
-                    TagName = tagHelperBlock.TagName,
-                    Attributes = attributes,
-                    Descriptors = _tagHelperDescriptors
-                },
+                new TagHelperChunk(
+                    tagHelperBlock.TagName, attributes, tagHelperBlock.SelfClosing, _tagHelperDescriptors),
```
dougbu: nit: if you're wrapping at all, place all arguments on their own lines

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -35,17 +35,20 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         /// <param name="tagName">An HTML tag name.</param>
         /// <param name="start">Starting location of the <see cref="TagHelperBlock"/>.</param>
         /// <param name="attributes">Attributes of the <see cref="TagHelperBlock"/>.</param>
+        /// <param name="selfClosing">The <c>bool</c> indicating whether or not the current tag is self-closing.</param>
```
dougbu: - use the same doc for this parameter as in `TagHelperExecutionContext`
- also, match parameter order and property comments as well

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -53,10 +56,12 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         // Internal for testing
         internal TagHelperBlockBuilder(string tagName,
                                        IDictionary<string, SyntaxTreeNode> attributes,
+                                       bool selfClosing,
```
dougbu: nit: `internal` but stick with `tagName`, `selfClosing`, `attributes` ordering

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -35,17 +35,20 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         /// <param name="tagName">An HTML tag name.</param>
         /// <param name="start">Starting location of the <see cref="TagHelperBlock"/>.</param>
         /// <param name="attributes">Attributes of the <see cref="TagHelperBlock"/>.</param>
+        /// <param name="selfClosing">The <c>bool</c> indicating whether or not the current tag is self-closing.</param>
         /// <param name="descriptors">The <see cref="TagHelperDescriptor"/>s associated with the current HTML
         /// tag.</param>
         public TagHelperBlockBuilder(string tagName,
                                      SourceLocation start,
```
dougbu: nit: I lean toward placing `selfClosing` before `start` as well as before `attributes`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -207,6 +207,29 @@ public void GeneratePreContent_ReturnsNothingIfSelfClosing()
             Assert.Empty(output);
         }
 
+        [Theory]
+        [InlineData(true, null)]
+        [InlineData(true, "\t")]
+        [InlineData(false, null)]
+        [InlineData(false, "\t")]
+        public void GeneratePreContent_ReturnsPreContentIfTagNameIsEmpty(bool selfClosing, string tagName)
```
dougbu: - reverse parameter order
- nit: update test name to end with "IsWhitespace" or "IsNullOrWhitespace"

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -225,7 +248,7 @@ public void GenerateContent_ReturnsContent()
 
 
         [Fact]
-        public void GenerateContent_ReturnsNothingIfSelfClosing()
+        public void GenerateContent_ReturnsNothingWhenTagNameIsNotEmptyAndSelfClosing()
```
dougbu: nit: "...IfSelfClosingWhenTagNameIsNotNullOrWhitespace"

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -191,7 +191,7 @@ public void GeneratePreContent_ReturnsPreContent()
         }
 
         [Fact]
-        public void GeneratePreContent_ReturnsNothingIfSelfClosing()
+        public void GeneratePreContent_ReturnsNothingWhenTagNameIsNotEmptyAndSelfClosing()
```
dougbu: nit: "...IfSelfClosingWhenTagNameIsNotNullOrWhitespace"

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -241,6 +264,29 @@ public void GenerateContent_ReturnsNothingIfSelfClosing()
             Assert.Empty(output);
         }
 
+        [Theory]
+        [InlineData(true, null)]
+        [InlineData(true, "\t")]
+        [InlineData(false, null)]
+        [InlineData(false, "\t")]
+        public void GenerateContent_ReturnsContentIfTagNameIsEmpty(bool selfClosing, string tagName)
```
dougbu: rinse, repeat from above through rest of class

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
```
dougbu: why?

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
dougbu: it just doesn't matter

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -398,6 +398,20 @@ private static void EvaluateTagHelperBlock(ErrorCollector collector, TagHelperBl
...
             {
+                if (!string.Equals(actual.TagName, expected.TagName, StringComparison.Ordinal))
+                {
+                    collector.AddError(
+                        "{0} - FAILED :: TagName mismatch for TagHelperBlock :: ACTUAL: {1}",
+                        expected.TagName,
+                        actual.TagName);
+                }
+
+                if (actual.SelfClosing != expected.SelfClosing)
```
dougbu: for both this and the above comparison, put expected on the left (as in `Assert.Equal()`)

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/BasicTagHelpers.CustomAttributeCodeGenerator.cs
```diff
@@ -31,7 +31,7 @@ public BasicTagHelpers()
                 WriteLiteral("\r\n        ");
                 __tagHelperExecutionContext = __tagHelperScopeManager.Begin("p", "test", async() => {
                 }
-                , StartWritingScope, EndWritingScope);
+                , StartWritingScope, EndWritingScope, false);
```
dougbu: help future readers and always use named parameters for `bool` values (or `null` or anytime it may be confusing...)

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: I'm indifferent

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +31,23 @@ public TagHelperScopeManager()
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="selfClosing">
+        /// The <c>bool</c> indicating whether or not the tag of this scope is self-closing.
```
NTaylorMullen: `<see cref="bool" />`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -192,5 +194,17 @@ public void Add_MaintainsMultipleTagHelpers()
         private class PTagHelper : TagHelper
         {
         }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void SelfClosing_ReturnsTrueOrFalseAsExpected(bool selfClosing)
```
NTaylorMullen: before private classes / with other tests

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -179,5 +180,25 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
                 ProcessOrderTracker.Add(Order);
             }
         }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task RunAsync_SetTagHelperOutputSelfClosing(bool selfClosing)
```
NTaylorMullen: move with other tests

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -398,6 +398,20 @@ private static void EvaluateTagHelperBlock(ErrorCollector collector, TagHelperBl
...
+                {
+                    collector.AddError(
+                        "{0} - FAILED :: SelfClosing should be {1}", actual.TagName, expected.SelfClosing);
```
NTaylorMullen: use `:: ACTUAL` like error format like above.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/BasicTagHelpers.CustomAttributeCodeGenerator.cs
NTaylorMullen: This is code gen file, not sure we want to start naming parameters in gen'd code.

NTaylorMullen:
:watch:

ianhong:
@NTaylorMullen and @dougbu 
I updated per your comments. Please review.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperChunk.cs
```diff
@@ -12,6 +12,29 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler
...
+        /// </param>
+        public TagHelperChunk(
+            string tagName,
+            bool selfClosing,
+            IDictionary<string, Chunk> attributes,
+            IEnumerable<TagHelperDescriptor> descriptors)
+        {
+            TagName = tagName;
+            Attributes = attributes;
+            SelfClosing = selfClosing;
```
NTaylorMullen: nit: move before attributes to maintain order as in ctor.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -46,17 +50,20 @@ public TagHelperBlockBuilder(TagHelperBlock original)
             Start = start;
             Descriptors = descriptors;
             Type = BlockType.Tag;
+            SelfClosing = selfClosing;
```
NTaylorMullen: nit: move after tagName to coincide with ctor.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -46,17 +50,20 @@ public TagHelperBlockBuilder(TagHelperBlock original)
...
 
         // Internal for testing
         internal TagHelperBlockBuilder(string tagName,
+                                       bool selfClosing,
                                        IDictionary<string, SyntaxTreeNode> attributes,
                                        IEnumerable<SyntaxTreeNode> children)
         {
             TagName = tagName;
             Attributes = attributes;
+            SelfClosing = selfClosing;
```
NTaylorMullen: ditto

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -33,7 +46,8 @@ public class TagHelperExecutionContextTest
                     return Task.FromResult(result: true);
                 },
                 startWritingScope: () => { },
-                endWritingScope: () => writer);
+                endWritingScope: () => writer,
+                selfClosing: false);
```
NTaylorMullen: should be after `tagName`, same below

NTaylorMullen:
Only found small things, :shipit: once @dougbu is good with it and we've signed off on an MVC reaction PR.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: :+1: don't reorder just because you're using named parameters

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperScopeManagerTest.cs
```diff
@@ -24,10 +24,11 @@ public void Begin_CreatesContextWithAppropriateTagName()
...
             var executionContext = scopeManager.Begin("p",
-                                                      string.Empty,
-                                                      DefaultExecuteChildContentAsync,
-                                                      DefaultStartWritingScope,
-                                                      DefaultEndWritingScope);
+                selfClosing: false,
+                uniqueId: string.Empty,
+                executeChildContentAsync: DefaultExecuteChildContentAsync,
+                startWritingScope: DefaultStartWritingScope,
+                endWritingScope: DefaultEndWritingScope);
```
dougbu: like the new indentation but please wrap the first argument as well (throughout this class)

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
```diff
@@ -177,27 +177,43 @@ public MarkupBlock(IEnumerable<SyntaxTreeNode> children)
...
+                                    params SyntaxTreeNode[] children)
+            : this(tagName, new Dictionary<string, SyntaxTreeNode>(), selfClosing, children)
         {
         }
 
         public MarkupTagHelperBlock(string tagName,
                                     IDictionary<string, SyntaxTreeNode> attributes,
                                     params SyntaxTreeNode[] children)
-            : base(new TagHelperBlockBuilder(tagName, attributes, children))
+            : base(new TagHelperBlockBuilder(tagName, false, attributes, children))
```
dougbu: named parameter for `false`

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
```diff
@@ -177,27 +177,43 @@ public MarkupBlock(IEnumerable<SyntaxTreeNode> children)
 
     public class MarkupTagHelperBlock : TagHelperBlock
     {
-        public MarkupTagHelperBlock(string tagName)
-            : this(tagName, new Dictionary<string, SyntaxTreeNode>())
+        public MarkupTagHelperBlock(string tagName, bool selfClosing = false)
```
dougbu: no need for default parameter; just add another constructor overload that doesn't have a `selfClosing` parameter

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
```diff
@@ -177,27 +177,43 @@ public MarkupBlock(IEnumerable<SyntaxTreeNode> children)
...
+            : this(tagName, new Dictionary<string, SyntaxTreeNode>(), selfClosing)
         {
         }
 
         public MarkupTagHelperBlock(string tagName,
-                                    IDictionary<string, SyntaxTreeNode> attributes)
-            : this(tagName, attributes, new SyntaxTreeNode[0])
+                                    IDictionary<string, SyntaxTreeNode> attributes,
+                                    bool selfClosing = false)
+            : this(tagName, attributes, selfClosing, new SyntaxTreeNode[0])
```
dougbu: - reorder parameters to place `selfClosing` after `tagName`, matching the base class
- add another constructor overload rather than using a default parameter value
- make sure all of these constructor overloads are referenced

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingUnitTest.cs
```diff
@@ -131,12 +131,13 @@ public void UsesCorrectUniqueIdForMultipleTagHelperChunksSomeWithSameSameTagHelp
 
         private static TagHelperChunk CreateTagHelperChunk(string tagName, IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
         {
-            return new TagHelperChunk
+            return new TagHelperChunk(
+                tagName,
+                attributes: new Dictionary<string, Chunk>(),
+                selfClosing: false,
```
dougbu: order to match the constructor you're calling

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -40,7 +40,8 @@ public static TheoryData EmptyTagHelperBoundAttributeData
                                 new Dictionary<string, SyntaxTreeNode>
                                 {
                                     { "bound", new MarkupBlock() }
-                                })),
+                                },
+                                selfClosing: true)),
```
dougbu: reorder these arguments too

dougbu:
:watch:

ianhong:
@dougbu Updated per you comments. Please review.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -60,7 +74,8 @@ public class TagHelperExecutionContextTest
                     return Task.FromResult(result: true);
                 },
                 startWritingScope: () => { },
-                endWritingScope: () => new StringWriter());
+                endWritingScope: () => new StringWriter(),
+                selfClosing: false);
```
dougbu: nit: missed one when reordering

dougbu:
:shipit: (just one nit)

ianhong:
23e6264715b0055218611b3f36ac45098f6b0fb1

PR #299, opened 2015-02-11
Change SourceLocation to be serializable.

NTaylorMullen:
- Needed for the DTH so we can send `SourceLocation`s over the wire.

### DIFF
--- a/src/Microsoft.AspNet.Razor/Text/SourceLocation.cs
+++ b/src/Microsoft.AspNet.Razor/Text/SourceLocation.cs
@@ -15,34 +15,24 @@ namespace Microsoft.AspNet.Razor.Text
         public static readonly SourceLocation Undefined = CreateUndefined();
         public static readonly SourceLocation Zero = new SourceLocation(0, 0, 0);
 
-        private int _absoluteIndex;
-        private int _lineIndex;
-        private int _characterIndex;
-
         public SourceLocation(int absoluteIndex, int lineIndex, int characterIndex)
         {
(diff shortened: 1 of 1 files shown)


NTaylorMullen:
Gonna leave this open until the DTH changes are in.

src/Microsoft.AspNet.Razor/Text/SourceLocation.cs
```diff
@@ -15,34 +15,21 @@ public struct SourceLocation : IEquatable<SourceLocation>, IComparable<SourceLoc
...
+            AbsoluteIndex = absoluteIndex;
+            LineIndex = lineIndex;
+            CharacterIndex = characterIndex;
         }
 
-        public int AbsoluteIndex
-        {
-            get { return _absoluteIndex; }
-        }
+        public int AbsoluteIndex { get; set; }
```
rynowak: mutable struct is a pit of failure :cry:

NTaylorMullen: Ya, it's purely so it can be serialized. I guess the other approach would be to provide a custom Json.net serializer for this class on the DTH side.

NTaylorMullen: Talked in person, just going to comment these things to avoid overkill of a custom serializer for a single type.

NTaylorMullen:
Getting this in so Todd can start integrating tooling.

Issue #289, opened 2015-02-02
Create parser error for empty TagHelper bound attribute values.


NTaylorMullen:
Right now if a user creates a `TagHelper` and then provides it with an empty attribute value (ex: `<myth myattr="" />`) we rely on a C# compilation error to notify the user of their mistake. We could create a custom `RazorError` with more information associated with what went wrong.

Also note that this would only apply to `TagHelper`s attributes who don't happen to be typed as `System.String`.

NTaylorMullen:
Moving this to backlog after decision to wait for user feedback.

NTaylorMullen:
Wrong issue, reopening (mistake).

NTaylorMullen:
e14dbdf9be27f23456e87b124b8c2c7c47326e8b

PR #303, opened 2015-02-12
Add parser error for empty TagHelper bound attributes.

NTaylorMullen:
- Errors are only created for TagHelper bound attributes that are not bound to string.
- Added tests to validate proper errors for expected input.

#289

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
@@ -15,6 +15,8 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
 {
     public static class TagHelperBlockRewriter
     {
+        private static readonly string StringTypeName = typeof(string).FullName;
+
         public static TagHelperBlockBuilder Rewrite(string tagName,
                                                     bool validStructure,
                                                     Block tag,
@@ -68,6 +70,23 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 // Only want to track the attribute if we succeeded in parsing its corresponding Block/Span.
                 if (succeeded)
                 {
+                    // Check if it's a bound attribute that is not of type string and happens to be null or whitespace.
+                    string attributeValueType;
+                    if (attributeValueTypes.TryGetValue(attribute.Key, out attributeValueType) &&
+                       !IsStringAttribute(attributeValueType) &&
+                       IsNullOrWhitespaceAttributeValue(attribute.Value))
+                    {
+                        var errorLocation = GetAttributeNameStartLocation(child);
+
+                        errorSink.OnError(
+                            errorLocation,
+                            RazorResources.FormatRewriterError_EmptyTagHelperBoundAttribute(
+                                attribute.Key,
+                                tagName,
+                                attributeValueType),
+                            attribute.Key.Length);
+                    }
+
                     attributes[attribute.Key] = attribute.Value;
                 }
             }
@@ -363,6 +382,34 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
             return builder.Build();
         }
 
+        private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
+        {
+            Span span;
+
+            if (node.IsBlock)
+            {
+                span = ((Block)node).FindFirstDescendentSpan();
+            }
+            else
+            {
+                span = (Span)node;
+            }
+
+            // Span should never be null here, this should only ever be called if an attribute was successfully parsed.
+            Debug.Assert(span != null);
+
+            var nodeStart = span.Parent.Start;
+
+            // Attributes must have at least one non-whitespace character to represent the tagName (even if its a C#
+            // expression).
+            var firstNonWhitespaceSymbol = span
+                .Symbols
+                .OfType<HtmlSymbol>()
+                .First(sym => sym.Type != HtmlSymbolType.WhiteSpace && sym.Type != HtmlSymbolType.NewLine);
+
+            return nodeStart + firstNonWhitespaceSymbol.Start;
+        }
+
         private static KeyValuePair<string, SyntaxTreeNode> CreateMarkupAttribute(
             string name,
             SpanBuilder builder,
@@ -374,12 +421,37 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
             // its value as code. Any non-string value can be any C# value so we need to ensure the SyntaxTreeNode
             // reflects that.
             if (attributeValueTypes.TryGetValue(name, out attributeTypeName) &&
-                !string.Equals(attributeTypeName, typeof(string).FullName, StringComparison.OrdinalIgnoreCase))
+                !IsStringAttribute(attributeTypeName))
             {
                 builder.Kind = SpanKind.Code;
             }
 
             return new KeyValuePair<string, SyntaxTreeNode>(name, builder.Build());
+        }
+
+        private static bool IsNullOrWhitespaceAttributeValue(SyntaxTreeNode attributeValue)
+        {
+            if (attributeValue.IsBlock)
+            {
+                foreach (var span in ((Block)attributeValue).Flatten())
+                {
+                    if (!string.IsNullOrWhiteSpace(span.Content))
+                    {
+                        return false;
+                    }
+                }
+
+                return true;
+            }
+            else
+            {
+                return string.IsNullOrWhiteSpace(((Span)attributeValue).Content);
+            }
+        }
+
+        private static bool IsStringAttribute(string attributeTypeName)
+        {
+            return string.Equals(attributeTypeName, StringTypeName, StringComparison.OrdinalIgnoreCase);
         }
 
         private static bool IsQuote(HtmlSymbol htmlSymbol)
--- a/src/Microsoft.AspNet.Razor/RazorResources.resx
+++ b/src/Microsoft.AspNet.Razor/RazorResources.resx
@@ -422,4 +422,7 @@ Instead, wrap the contents of the block in "{{}}":
     <value>Inline markup blocks (e.g. @&lt;p&gt;content&lt;/p&gt;) must not appear in non-string tag helper attribute values.
  Expected a '{0}' attribute value, not a string.</value>
   </data>
+  <data name="RewriterError_EmptyTagHelperBoundAttribute" xml:space="preserve">
+    <value>Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty or contain only whitespace.</value>
+  </data>
 </root>
--- a/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
@@ -19,6 +19,299 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData EmptyTagHelperBoundAttributeData
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var emptyAttributeError =
+                    "Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of " +
+                    "type '{2}' cannot be empty or contain only whitespace.";
+                var boolTypeName = typeof(bool).FullName;
+
+                // documentContent, expectedOutput, expectedErrors
+                return new TheoryData<string, MarkupBlock, RazorError[]>
+                {
+                    {
+                        "<myth bound='' />",
+                        new MarkupBlock(
+                            new MarkupTagHelperBlock(
+                                "myth",
+                                new Dictionary<string, SyntaxTreeNode>
+                                {
+                                    { "bound", new MarkupBlock() }
+                                })),
+                        new[]
+                        {
+                            new RazorError(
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -68,6 +70,28 @@ public static class TagHelperBlockRewriter
...
                 {
+                    string attributeValueType;
+
+                    // Check if it's a bound attribute that is not of type string. If there's an empty value we need 
+                    // to create an error.
+                    if (attributeValueTypes.TryGetValue(attribute.Key, out attributeValueType) &&
+                       !IsStringAttribute(attributeValueType) &&
+                       (attribute.Value == null ||
+                       (!attribute.Value.IsBlock && string.IsNullOrEmpty(((Span)attribute.Value).Content)) ||
+                       (attribute.Value.IsBlock && ((Block)attribute.Value).Children.Count() == 0)))
```
dougbu: don't need a comment about creating an error but do need one explaining the complicated condition.  would help for example to refactor the `attribute.Value` checks into a new method, as you've done with `IsStringAttribute()`.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -68,6 +70,28 @@ public static class TagHelperBlockRewriter
...
+                    {
+                        var errorLocation = GetAttributeNameStartLocation(child);
```
dougbu: doesn't the `attribute` already have the correct location?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -363,6 +387,35 @@ private static Block RebuildCodeGenerators(Block block)
...
+        private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
+        {
+            var startOffset = node.Start;
+
+            while (node.IsBlock)
+            {
+                // We want to track the start of the parent node. Each iteration of this while loop we will dive deeper
+                // into the Block. Since the attribute name will always be located at .Children[0] we need to track the
+                // corresponding .Start location of the attribute names parent so we can offset its .Start based on the 
+                // leading whitespace.
```
dougbu: nit: leading periods on properties make me want to read this as four sentences; don't think the dots add value

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -68,6 +70,28 @@ public static class TagHelperBlockRewriter
                 // Only want to track the attribute if we succeeded in parsing its corresponding Block/Span.
                 if (succeeded)
                 {
+                    string attributeValueType;
```
dougbu: nit: move this right above the `if`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -363,6 +387,35 @@ private static Block RebuildCodeGenerators(Block block)
...
+                // into the Block. Since the attribute name will always be located at .Children[0] we need to track the
+                // corresponding .Start location of the attribute names parent so we can offset its .Start based on the 
+                // leading whitespace.
+                startOffset = node.Start;
+
+                // An attribute must have at least one child element underneath it
+                node = ((Block)node).Children.First();
+            }
+
+            // At this point the node is not a block so it must be a Span.
```
dougbu: nit: capitalize Block

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -363,6 +387,35 @@ private static Block RebuildCodeGenerators(Block block)
             return builder.Build();
         }
 
+        private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
+        {
+            var startOffset = node.Start;
```
dougbu: name looks wrong. this appears to be the `blockStart` and `firstNonWhitespaceSymbol.Start` provides the offset from that location.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Sadly it does not. The node contains leading whitespace which can be contained in a `Block`. Hence we need to find the first symbol that is not whitespace leading up to the attribute name. To do that we have to dive through blocks to find it :cry:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -363,6 +387,35 @@ private static Block RebuildCodeGenerators(Block block)
...
+                // corresponding .Start location of the attribute names parent so we can offset its .Start based on the 
+                // leading whitespace.
+                startOffset = node.Start;
+
+                // An attribute must have at least one child element underneath it
+                node = ((Block)node).Children.First();
+            }
+
+            // At this point the node is not a block so it must be a Span.
+            var span = (Span)node;
```
dougbu: more succinct and perhaps more readable as:
``` c#
  var span = (node as Block)?.FindFirstDescendentSpan() ?? (Span)node;
  var blockStart = span.Parent.Start;
```
worst case, use an `if` to set `span`.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -422,4 +422,7 @@ Instead, wrap the contents of the block in "{{}}":
     <value>Inline markup blocks (e.g. @&lt;p&gt;content&lt;/p&gt;) must not appear in non-string tag helper attribute values.
  Expected a '{0}' attribute value, not a string.</value>
   </data>
+  <data name="RewriterError_EmptyTagHelperBoundAttribute" xml:space="preserve">
+    <value>Attribute '{0}' on tag helper element '{1}' is a '{2}' tag helper bound attribute with no value. Only '{3}' typed attributes can be without a value.</value>
```
dougbu: not understandable or helpful. do not mention alternatives (a `string` bound attribute for example) that are out of the Razor author's control. suggest `Invalid empty value for attribute '{0}' on tag helper element '{1}'.  Tag helper bound attributes of type '{2}' must have a value.`

dougbu: also I'm not sure "tag helper bound attribute" is going to be meaningful to Razor authors /cc @DamianEdwards ?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: - current test doesn't cause an error for an attribute value containing only whitespace but should
- when could `attribute.Value` be a `Block` with no children?  that is, doesn't `Block`-ness imply the value contains more than just text?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,204 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                    "Attribute '{0}' on tag helper element '{1}' is a '{2}' tag helper bound attribute with no " +
+                    "value. Only '{3}' typed attributes can be without a value.";
+                var boolTypeName = typeof(bool).FullName;
+                var stringTypeName = typeof(string).FullName;
+
+                // documentContent, expectedOutput, expectedErrors
+                return new TheoryData<string, MarkupBlock, RazorError[]>
+                {
+                    {
+                        "<myth bound='' />",
```
dougbu: also test `<myth bound='  ' />`.  I think this currently results in "success" though the generated code will always be erroneous.

NTaylorMullen: It currently does result in success, do you think we should also create errors for whitespace values?

src/Microsoft.AspNet.Razor/RazorResources.resx
DamianEdwards: Perhaps simplify a bit, .e.g `Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty.`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
dougbu: yes, definitely

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: No, being a block simply means there's the potential to have more than 1 part of the value.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Will rename to `nodeStart` since it may not be a `Block`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Done

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -382,6 +429,21 @@ private static Block RebuildCodeGenerators(Block block)
...
 
+        private static bool IsNullOrWhitespaceAttributeValue(SyntaxTreeNode attributeValue)
+        {
+            if (attributeValue.IsBlock)
+            {
+                attributeValue = ((Block)attributeValue).FindFirstDescendentSpan();
+            }
+
+            return attributeValue == null || string.IsNullOrWhiteSpace(((Span)attributeValue).Content);
+        }
```
dougbu: thanks for spreading this out.

I suspect this will cause errors for legal attribute values such as `" true "` or `" @true "`.  should handle the `Block` case using
``` c#
  foreach (var span in ((Block)attributeValue).Flatten())
  {
    if (!string.IsNullOrWhiteSpace(span.Content))
    {
      return false;
    }
  }
  return true;
```

if it's not necessary to check every `Span` in the `Block`, need a code comment explaining why.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -422,4 +422,7 @@ Instead, wrap the contents of the block in "{{}}":
     <value>Inline markup blocks (e.g. @&lt;p&gt;content&lt;/p&gt;) must not appear in non-string tag helper attribute values.
  Expected a '{0}' attribute value, not a string.</value>
   </data>
+  <data name="RewriterError_EmptyTagHelperBoundAttribute" xml:space="preserve">
+    <value>Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty.</value>
```
dougbu: nit: `... cannot be empty or contain only whitespace.`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,255 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                                })),
+                        new[]
+                        {
+                            new RazorError(
+                                string.Format(emptyAttributeError, "bound", "myth", boolTypeName),
+                                absoluteIndex: 6, lineIndex: 0, columnIndex: 6, length: 5)
+                        }
+                    },
+                    {
+                        "<myth bound='    true' />",
```
dougbu: am I correct `<myth bound='    @true' />` leads to an error?  what about `<myth bound='    @(true)' />`?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: You're right :+1:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: You are absolutely correct! super good catch man :+1: :+1: :+1: :+1: :+1:

NTaylorMullen: Done

NTaylorMullen:
Updated.

dougbu:
:shipit:

Issue #264, opened 2014-12-28
Camel-case tag helper directives


glen-84:
When multiple words are joined and the full string is lower-cased (or upper-cased), it makes it difficult to read.

For example:

```c#
@addtaghelper X
```

Would it be possible to make directives case insensitive, or at least allow camel or pascal case?

```c#
@addTagHelper X
```

Unrelated to this ticket: Would it not make sense to use a different prefix for directives, so that they don't look like variables? Something like:

```c#
@!model WebApplication1.Models.LoginViewModel
```

... just a thought.

danroth27:
We think directives should be case sensitive, but we think we should change addtaghelper to addTagHelper, removetaghelper to removeTagHelper, etc.

glen-84:
Sounds good.

dougbu:
Changed title from "RFC: Make directives case insensitive or allow camel case"

dougbu:
fixed in commits fbe331c1abcf840acab6c7f95e898f3e0ec37f66 and https://github.com/aspnet/Mvc/commit/451db6fb1621c0b67f1948085dd0bafe8092badf

PR #305, opened 2015-02-13
Camel-case tag helper directives

dougbu:
- #264

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text.RegularExpressions;
 using Microsoft.AspNet.Razor.Parser;
@@ -16,6 +17,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperDescriptorResolver : ITagHelperDescriptorResolver
     {
+        private static readonly IReadOnlyDictionary<TagHelperDirectiveType, string> _directiveNames =
+            new Dictionary<TagHelperDirectiveType, string>
+            {
+                { TagHelperDirectiveType.AddTagHelper, SyntaxConstants.CSharp.AddTagHelperKeyword },
+                { TagHelperDirectiveType.RemoveTagHelper, SyntaxConstants.CSharp.RemoveTagHelperKeyword },
+            };
+
         private readonly TagHelperTypeResolver _typeResolver;
 
         // internal for testing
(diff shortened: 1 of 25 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -16,6 +16,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperDescriptorResolver : ITagHelperDescriptorResolver
     {
+        private static readonly Dictionary<TagHelperDirectiveType, string> _directiveNames =
```
NTaylorMullen: `IReadOnlyDictionary<TagHelperDirectiveType, string>`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -67,8 +74,7 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
                 }
                 catch (Exception ex)
                 {
-                    var directiveName = "@" + directiveDescriptor.DirectiveType.ToString().ToLowerInvariant();
-
+                    var directiveName = "@" + _directiveNames[directiveDescriptor.DirectiveType];
```
NTaylorMullen: Since we're pulling from a dictionary here we should try and retrieve the directive name, if we fail polyfill with something like `unknown`.

NTaylorMullen:
Only tiny comments, :shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
dougbu: :-1: this class _must_ change if we ever add another directive that reaches here. but I'll add a debug assertion that the dictionary contains the value.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
dougbu: why?  this is `private`

NTaylorMullen: Showcases what it should be used for, only looking up information.

dougbu: disagree but don't care enough not to make the change

dougbu:
updated FYI.  will push once aspnet/Mvc#2013 "react to" PRs is signed off.

NTaylorMullen:
:shipit:

dougbu:
commit fbe331c1abcf840acab6c7f95e898f3e0ec37f66 here and https://github.com/aspnet/Mvc/commit/451db6fb1621c0b67f1948085dd0bafe8092badf

Issue #238, opened 2014-11-22
Add TagHelperContext.Items bag to enable passing values between related tag helpers


DamianEdwards:
Add a new property `public IDictionary<string, object> Items { get; }` to the `TagHelperContext` class to facilitate the passing of arbitrary items between Tag Helpers running on the same element.

e.g.
``` C# 
[HtmlElementName("my-tag")]
public class MyTagHelper : TagHelper {
    public override void Process(TagHelperContext context, TagHelperOutput output) {
        context.Items["message"] = "This came from the MyTagHelper class!";
    }
}

[HtmlElementName("my-tag")]
public class MyOtherTagHelper : TagHelper {
    public override void Process(TagHelperContext context, TagHelperOutput output) {
        var message = context.Items["message"];
    }
}
```

NTaylorMullen:
I don't feel like a single bag to communicate between multiple `TagHelper`s on the same elment has a whole lot of value and doesn't help solve the major problem of communicating between child => parent and vice versa.

Ultimately it will be less common for `TagHelper` authors to write multiple `TagHelper`s that communicate among each other on the same element which is why this item should probably be re-thought to provide greater value.

dougbu:
in an offline discussion, thought we ended with agreement some of the existing property bags (`FormContext.FormData`, `ViewContext.ViewData`, likely a few others) had reasonable semantics but not quite the correct scope. this led us to a proposed design where `public IDictionary<string, object> Items { get; }` is backed by a `CopyOnWriteDictionary` and each `TagHelperContext` would initialize that dictionary to its parent's dictionary.
this design provides
- communication of arbitrary information between tag helpers targeting the same element (no harm though likely not a primary scenario, agreed)
- communication of arbitrary information between tag helpers and any tag helper nested within
- worry-free cleanup of per-helper content -- parent and sibling scopes never see a tag helper's changes to its `Items` content
- communication of arbitrary information from nested tag helpers back to their parent; all that's required is the parent creating its (well-known) `Items` entry before executing its childen

NTaylorMullen:
The way this item is worded leads me to believe the `Items` bag is only accessible to `TagHelpers` running on the same element. If that Items bag is communicated down scope-wise via `CopyOnWriteDictionary` then I'm 100% on board.

DamianEdwards:
:+1:

-----Original Message-----

NTaylorMullen:
b95e73e2b05b528f6f029e1bd93d037d16f47130
[...]

PR #306, opened 2015-02-13
Add Items bag to TagHelperContext.

NTaylorMullen:
- Added an extra parameter to the TagHelperScopeManager so when it creates the TagHelperExecutionContext it can retrieve the parent execution context in order to create the Items bag that eventually exists on the TagHelperContext.
- Add a an Item properties on TagHelperExecutionContext to propagate to the TagHelperContext.
- Updated code generation to pass the parent TagHelperExecutionContext into the TagHelperScopeManager.

#238


NTaylorMullen:
/cc @yishaigalatzer @dougbu So as you can see in this PR i've copied the basis of MVC's `CopyOnWriteDictionary` into the Razor project to be used as the wrapper to the `TagHelperContext.Items` bag. 

MVC version: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Common/CopyOnWriteDictionary.cs

I'm not sure there's a better way around this since MVC's version is located in *Microsoft.AspNet.Mvc.Common* and is required by *Microsoft.AspNet.Mvc.Core*, both of which have no concept of Razor; therefore we can't move the lib into the Razor repo. If you guys have a better alternative I'm all ears :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -18,15 +18,19 @@ public class TagHelperContext
...
         /// asynchronously.</param>
-        public TagHelperContext([NotNull] IDictionary<string, object> allAttributes,
-                                [NotNull] string uniqueId,
-                                [NotNull] Func<Task<string>> getChildContentAsync)
+        public TagHelperContext(
+            [NotNull] IDictionary<string, object> allAttributes,
+            [NotNull] IDictionary<string, object> items,
+            [NotNull] string uniqueId,
+            [NotNull] Func<Task<string>> getChildContentAsync)
         {
```
dougbu: shouldn't have callers deal with the intricacies of `CopyOnWriteDictionary`.  instead, do something similar to `ViewDataDictionary` and provide two constructors:
- a constructor for a new top-level context.  this one initializes `Items` with a regular old `Dictionary<string, object>(StringComparer.OrdinalIgnoreCase)`.
- a constructor that takes either the "parent" `TagHelperContext` (or "outer" `IDictionary<string, object>`).  this one initializes `Items` with `CopyOnWriteDictionary<string, object>(parent.Items, StringComparer.OrdinalIgnoreCase)`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -161,5 +180,136 @@ public Task ExecuteChildContentAsync()
 
             return _childContent;
         }
+
+        private class ItemsCopyOnWriteDictionary : IDictionary<string, object>
```
dougbu: move this to be `private` to `TagHelperContext`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -68,6 +82,11 @@ public bool ChildContentRetrieved
         }
 
         /// <summary>
+        /// TODO
+        /// </summary>
+        public IDictionary<string, object> Items { get; }
```
dougbu: why does this exist?  should be enough to have `Output.Items`

dougbu:
:watch:

yishaigalatzer:
We can discuss sharing stuff from mvc.common with razor. It is just a build time dependency. We might just move the common library somewhere else

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Need to have a publicly visible property to create the `TagHelperContext` with the appropriate dictionary. This is essentially the metadata object used to construct all the requirements for running `TagHelper`s.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Can definitely provide two constructors. However don't want to have the `TagHelperContext` take in the `TagHelperContext` that was used to run the parent set of `TagHelper`s (that's more the `TagHelperExecutionContext`'s responsibility).

The `TagHelperContext` is currently more like the `TagHelperRunningContext` because it only exists within the `TagHelperRunner`. It was designed this way so users could potentially replace just the `TagHelperRunner` and build custom `TagHelper`s (aka call into their own versions `Process` methods with different parameters).

NTaylorMullen:
@yishaigalatzer Lets get together to talk about this because this would end up being some new shared webstack repo which to me at least is iffy.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Note: Items should be case-sensitive, so the created dictionaries will be `StringComparer.Ordinal`. No reason for them to ignore case (not HTML items).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Can't do this sadly, `TagHelperExecutionContext` is the scope manager for executing `TagHelper`s, therefore if this existed in the `TagHelperContext` then we'd be wrapping the `TagHelperExecutionContext`s `Items` which would then never be updated and never passed down to children.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Attempted to do this then realized that the `TagHelperExecutionContext` is the communication channel to the `RazorPage` and is essentially what enables us to communicate the `Items` dictionary to any child `TagHelper`s. It therefore then has to be the one wrapping the `Items` dictionary.

NTaylorMullen:
Updated. Left the `CopyOnWriteDictionary` as a private class until we decide on a home for common web helpers.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Ignore this, it's here until we find a suitable home for common web framework utilities.

dougbu:
@NTaylorMullen will you update this PR once you've pushed aspnet/Common#1?  (will look at it after if that's the case.)

NTaylorMullen:
@dougbu Updated to utilize Common repo.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -18,15 +19,19 @@ public class TagHelperContext
...
         /// <param name="uniqueId">The unique identifier for the source element this <see cref="TagHelperContext" /> 
         /// applies to.</param>
         /// <param name="getChildContentAsync">A delegate used to execute and retrieve the rendered child content 
         /// asynchronously.</param>
-        public TagHelperContext([NotNull] IDictionary<string, object> allAttributes,
-                                [NotNull] string uniqueId,
-                                [NotNull] Func<Task<string>> getChildContentAsync)
+        public TagHelperContext(
+            [NotNull] IDictionary<string, object> allAttributes,
+            [NotNull] IDictionary<string, object> parentItems,
```
dougbu: odd to pass member collections into a constructor.  why not instead change `TagHelperRunner.RunAsync()` to pass `executionContext.Parent?.Output` in?  then `TagHelperExecutionContext` wouldn't need strange `Items` property and `TagHelperOutput` would be the clear owner of its `Items`.

dougbu: or if `TagHelperExecutionContext` _must_ own the `Items` dictionary, expose a property of type `TagHelperExecutionContext` in this class and let tag helper authors get to the `Items` through that property.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -13,6 +13,146 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
+                parentExecutionContext: parentExecutionContext);
+
+            // Assert
+            Assert.Collection(
+                executionContext.Items,
+                (entry) =>
+                {
+                    Assert.Equal("test-entry", entry.Key, StringComparer.Ordinal);
+                    Assert.Equal(1234, entry.Value);
+                });
```
dougbu: nit: for single-element dictionaries, easier to read as
``` c#
  var keyValuePair = Assert.Single(executionContext.Items);
  Assert.Equal("test-entry", keyValuePair.Key, StringComparer.Ordinal);
  Assert.Equal(1234, keyValuePair.Value);
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
dougbu: sorry, meant `TagHelperContext`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -22,6 +23,7 @@ public class TagHelperRunner
         {
             var tagHelperContext = new TagHelperContext(
                 executionContext.AllAttributes,
+                executionContext.Items,
```
dougbu: should confirm `TagHelperRunner` sets up `TagHelperContext` as expected in `TagHelperRunnerTest`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -37,6 +42,15 @@ public class TagHelperContext
         public IDictionary<string, object> AllAttributes { get; }
 
         /// <summary>
+        /// Gets the collection of items used to communicate with child <see cref="ITagHelper"/>s.
+        /// </summary>
+        /// <remarks>
+        /// This <see cref="IDictionary{string, object}"/> is copy-on-write in order to not affect parent
+        /// <see cref="ITagHelper"/>s.
+        /// </remarks>
+        public IDictionary<string, object> Items { get; }
```
dougbu: need unit tests confirming this property works as expected.  should have a `TagHelperContextTest` test class in general.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: That feels pretty heavy handed and seems to defeat the purpose of the `TagHelperExecutionContext` for the `Items` dictionary. I don't see the value in having the `TagHelperContext` own the `CopyOnWriteDictionary` wrapped items. In the end `TagHelperContext` and `TagHelperOutput` are just utility classes sent into `TagHelper` process methods where `TagHelperExecutionContext` is the real state holder behind the scenes.

One thing I want to change one of these days is to no longer expose `TagHelperOutput` to the `RazorPage`, that way users can easily replace that part of `TagHelper` execution, for now that's pretty much the only thing that comes back out into the `RazorPage`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Not a fan of validating properties are set via ctor, but will add one.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperRunner.cs
NTaylorMullen: Done.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Done.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
dougbu: @NTaylorMullen I'm obviously not making my proposal clear.  Concretely...
``` c#
public class TagHelperExecutionContext
{
  ...
  public TagHelperExecutionContext Parent { get; }

  public TagHelperContext TagHelperContext { get; set; }
}

public class TagHelperRunner
{
  public ... RunAsync(TagHelperExecutionContext executionContext)
  {
    var tagHelperContext = new TagHelperContext(
      executionContext.UniqueId,
      executionContext.Parent?.TagHelperContext,
      executionContext.GetChildContentAsync);
    executionContext.TagHelperContext = tagHelperContext;
    ...
  }
}

public class TagHelperContext
{
  public TagHelperContext(..., TagHelperContext parentContext, ...)
  {
    if (parentContext == null)
    {
      // Top-level TagHelperContext.
      Items = new Dictionary<string, object>(StringComparer.Ordinal);
    }
    else
    {
      Items = new CopyOnWriteDictionary<string, object>(
        parentContext.Items, 
        StringComparer.Ordinal);
    }
  }
  ...
  public IDictionary<string, object> Items { get; }
}
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
dougbu: @yishaigalatzer is this another property bag that should be of type `IDictionary<object, object>`?  Would allow the `SelectTagHelper` to use the key `typeof(SelectTagHelper)` when passing information to contained `OptionTagHelper` instances.

dougbu:
:watch:

NTaylorMullen:
Updated after discussion.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -18,15 +19,19 @@ public class TagHelperContext
         /// Instantiates a new <see cref="TagHelperContext"/>.
         /// </summary>
         /// <param name="allAttributes">Every attribute associated with the current HTML element.</param>
+        /// <param name="parentItems">Items that have been created by parent <see cref="ITagHelper"/>s.</param>
```
dougbu: information is no longer about parent.  rename to `items` and use property description.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -37,6 +42,15 @@ public class TagHelperContext
         public IDictionary<string, object> AllAttributes { get; }
 
         /// <summary>
+        /// Gets the collection of items used to communicate with child <see cref="ITagHelper"/>s.
```
dougbu: "child" isn't quite correct; perhaps "other", "in-scope", or "parent and child"?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -37,6 +42,15 @@ public class TagHelperContext
...
+        /// </summary>
+        /// <remarks>
+        /// This <see cref="IDictionary{object, object}"/> is copy-on-write in order to not affect parent
+        /// <see cref="ITagHelper"/>s.
```
dougbu: misleading since communication back to parent `ITagHelper`s is completely supported. suggest `in order to ensure items added to this collection are visible only to other <see cref="ITagHelper"/>s targeting this or child elements.`

might not need "this or" since, while it'll work (modulo order), it's not a primary use case.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -39,17 +42,20 @@ internal TagHelperExecutionContext(string tagName)
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="parentItems">Gets the collection of items used to communicate with child <see cref="ITagHelper"/>s</param>
```
dougbu: wrap long line

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -27,7 +29,8 @@ internal TagHelperExecutionContext(string tagName)
                    uniqueId: string.Empty,
                    executeChildContentAsync: async () => await Task.FromResult(result: true),
                    startWritingScope: () => { },
-                   endWritingScope: () => new StringWriter())
+                   endWritingScope: () => new StringWriter(),
+                   parentItems: new Dictionary<object, object>())
```
dougbu: same issues with names and descriptions exist in this class as in `TagHelperContext`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +32,36 @@ public TagHelperScopeManager()
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="parentExecutionContext">The parent <see cref="TagHelperExecutionContext"/>.</param>
```
dougbu: nit: suggest placing this parameter above the various `Func` and `Action` parameters, as you've done in the `TagHelperContext` constructor

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -39,17 +42,20 @@ internal TagHelperExecutionContext(string tagName)
...
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
+        /// <param name="parentItems">Gets the collection of items used to communicate with child <see cref="ITagHelper"/>s</param>
         public TagHelperExecutionContext([NotNull] string tagName,
                                          [NotNull] string uniqueId,
                                          [NotNull] Func<Task> executeChildContentAsync,
                                          [NotNull] Action startWritingScope,
-                                         [NotNull] Func<TextWriter> endWritingScope)
+                                         [NotNull] Func<TextWriter> endWritingScope,
+                                         [NotNull] IDictionary<object, object> parentItems)
```
dougbu: nit: suggest placing this parameter above the various `Func` and `Action` parameters, as you've done in the `TagHelperContext` constructor (for complete consistency, between the `tagName` and `uniqueId` parameters)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +32,36 @@ public TagHelperScopeManager()
...
+                                               TagHelperExecutionContext parentExecutionContext)
         {
-            var executionContext = new TagHelperExecutionContext(tagName,
-                                                                 uniqueId,
-                                                                 executeChildContentAsync,
-                                                                 startWritingScope,
-                                                                 endWritingScope);
+            IDictionary<object, object> parentItems;
+
+            // If we're not wrapped by another TagHelper then there will not be a parentExecutionContext.
```
dougbu: teensy nit: comma after TagHelper

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +32,36 @@ public TagHelperScopeManager()
...
-                                               [NotNull] Func<TextWriter> endWritingScope)
+                                               [NotNull] Func<TextWriter> endWritingScope,
+                                               TagHelperExecutionContext parentExecutionContext)
         {
-            var executionContext = new TagHelperExecutionContext(tagName,
-                                                                 uniqueId,
-                                                                 executeChildContentAsync,
-                                                                 startWritingScope,
-                                                                 endWritingScope);
+            IDictionary<object, object> parentItems;
```
dougbu: just `items`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -31,18 +32,36 @@ public TagHelperScopeManager()
...
+
+            // If we're not wrapped by another TagHelper then there will not be a parentExecutionContext.
+            if (parentExecutionContext != null)
+            {
+                parentItems = new CopyOnWriteDictionary<object, object>(
+                    parentExecutionContext.Items,
+                    comparer: null);
```
dougbu: will this fail once `[NotNull]` is enforced?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperContextTest.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperContextTest
+    {
+        [Fact]
+        public void Items_IsConstructedItems()
```
dougbu: nit: `Constructor_SetsProperties_AsExpected` since test doesn't just check `Items`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperScopeManagerTest.cs
```diff
@@ -17,6 +17,127 @@ public class TagHelperScopeManagerTest
             async () => await Task.FromResult(result: true);
 
         [Fact]
+        public void Begin_DoesNotRequireParentExecutionContext()
```
dougbu: good coverage but worth testing to confirm "does not affect" is shallow.  that is, test parent adding a test class instance to the parent's `Items`, make a change to a property of that instance in child execution context, and confirm parent sees modification.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
dougbu: hmm, sources are normally the first parameter in copy constructors and start-sub-scope methods.  probably good for `parentExecutionContext` to be first here.

dougbu:
:watch: getting close (but is this blocked by XRE bug?)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
NTaylorMullen: Yup, talked about this in person.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperScopeManagerTest.cs
NTaylorMullen: Will add this test with a dictionary being the complex object.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -35,15 +38,19 @@ internal TagHelperExecutionContext(string tagName)
         /// Instantiates a new <see cref="TagHelperExecutionContext"/>.
         /// </summary>
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
+        /// <param name="items">The collection of items used to communicate with child
```
dougbu: sorry for commenting on a commit; this is where I started seeing "child"

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperScopeManagerTest.cs
```diff
@@ -17,17 +18,168 @@ public class TagHelperScopeManagerTest
...
+                endWritingScope: DefaultEndWritingScope);
+
+            // Assert
+            var executionContextItem = Assert.Single(executionContext.Items);
+            Assert.Equal("test-entry", (string)executionContextItem.Key, StringComparer.Ordinal);
+            Assert.Equal(1234, executionContextItem.Value);
+        }
+
+        [Fact]
+        public void Begin_ReturnedExecutionContext_ComplexObjectItemEntriesAreVisibleToParent()
```
dougbu: name is too specific; much closer to `Begin_DoesShallowCopyOfParentItems`
- if you want to be almost this specific `Begin_SetsUpItems_SoParentSeesChangesChildrenMakeToExistingItems` -- expect few dictionaries within dictionaries but lots of complex objects in `Items` where child changes a property value

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -26,23 +27,42 @@ public TagHelperScopeManager()
...
         /// <param name="startWritingScope">A delegate used to start a writing scope in a Razor page.</param>
         /// <param name="endWritingScope">A delegate used to end a writing scope in a Razor page.</param>
         /// <returns>A <see cref="TagHelperExecutionContext"/> to use.</returns>
-        public TagHelperExecutionContext Begin([NotNull] string tagName,
-                                               [NotNull] string uniqueId,
-                                               [NotNull] Func<Task> executeChildContentAsync,
-                                               [NotNull] Action startWritingScope,
-                                               [NotNull] Func<TextWriter> endWritingScope)
+        public TagHelperExecutionContext Begin(
+            TagHelperExecutionContext parentExecutionContext,
```
dougbu: shoot, why does this parameter exist?  doesn't `TagHelperScopeManager` already know the `parentExecutionContext` value e.g. it can calculate `_executionScopes.Count == 0 ? null : _executionScopes.Peek()`?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -26,23 +27,42 @@ public TagHelperScopeManager()
...
-                                                                 startWritingScope,
-                                                                 endWritingScope);
+            IDictionary<object, object> items;
+
+            // If we're not wrapped by another TagHelper, then there will not be a parentExecutionContext.
+            if (parentExecutionContext != null)
+            {
+                items = new CopyOnWriteDictionary<object, object>(
+                    parentExecutionContext.Items,
+                    comparer: null);
```
dougbu: pass `EqualityComparer<object>.Default` instead

dougbu: thanks @rynowak !

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
NTaylorMullen: You're absolutely right :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
NTaylorMullen: :+1:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -35,15 +38,19 @@ internal TagHelperExecutionContext(string tagName)
         /// Instantiates a new <see cref="TagHelperExecutionContext"/>.
         /// </summary>
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
+        /// <param name="items">The collection of items used to communicate with other 
+        /// <see cref="ITagHelper"/>s</param>
```
dougbu: add trailing period

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -32,17 +33,35 @@ public TagHelperScopeManager()
...
-            var executionContext = new TagHelperExecutionContext(tagName,
-                                                                 uniqueId,
-                                                                 executeChildContentAsync,
-                                                                 startWritingScope,
-                                                                 endWritingScope);
+            IDictionary<object, object> items;
+            var parentExecutionContext = _executionScopes.Count > 0 ? _executionScopes.Peek() : null;
+
+            // If we're not wrapped by another TagHelper, then there will not be a parentExecutionContext.
+            if (parentExecutionContext != null)
```
dougbu: nit: this could be `if (_executionScopes.Count != 0)` (or `> 0` as you prefer).  then you wouldn't need the ternary above

dougbu:
:shipit:
if possible, please check in as 2 commits with first reacting to the Common repo additions

Issue #297, opened 2015-02-09
Modify '?' in @addtaghelper and @removetaghelper to mean 1 character and not "optionally 1 character"


NTaylorMullen:
Right now providing: `@addtaghelper a?c, MyAssembly"` will match type names `abc` AND `ac`. After understanding windows paths better we should modify the current behavior to mean `?` => 1 character.

NTaylorMullen:
f53b6bb0e3915f9f70fee83a15ff847d03219341

PR #307, opened 2015-02-16
Change '?' in add and remove TagHelper to mean 1 character.

NTaylorMullen:
- Modified tests to account for new behavior of the '?' character in @addtaghelper and @removetaghelper.

#297

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -129,7 +129,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
             // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire 
             // string. We also replace any '*' or '?' characters with regex to match appropriate content.
-            var pattern = "^" + escaped.Replace(@"\?", ".?").Replace(@"\*", ".*?") + "$";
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -129,7 +129,8 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
 
             // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire 
             // string. We also replace any '*' or '?' characters with regex to match appropriate content.
-            var pattern = "^" + escaped.Replace(@"\?", ".?").Replace(@"\*", ".*?") + "$";
+            // '*' matches 0 or more characters and '?' matches 1 character.
```
dougbu: nit: mention '*' isn't greedy

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -297,7 +297,7 @@ public static TheoryData ResolveDirectiveDescriptorsData
                         new []
```
dougbu: nit: update a couple of the higher-level tag helper tests e.g. one of the test Razor input files to use a more complicated pattern than '*'

dougbu:
:shipit: (just nits)

PR #308, opened 2015-02-19
Changes to support use of TagHelpers in applications with precompiled

pranavkm:
views.

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -26,18 +26,23 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
         private readonly TagHelperTypeResolver _typeResolver;
 
-        // internal for testing
-        internal TagHelperDescriptorResolver(TagHelperTypeResolver typeResolver)
-        {
-            _typeResolver = typeResolver;
-        }
-
         /// <summary>
         /// Instantiates a new instance of the <see cref="TagHelperDescriptorResolver"/> class.
         /// </summary>
         public TagHelperDescriptorResolver()
             : this(new TagHelperTypeResolver())
         {
+
+        }
+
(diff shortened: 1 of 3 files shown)


dougbu:
looks like this may have an impact on tooling.  are they ready?

dougbu:
:watch:

pranavkm:
> looks like this may have an impact on tooling. are they ready? 

I'm not sure what this is about?

dougbu:
@pranavkm tooling subclasses `ResolveDescriptorsInAssembly` in order to override `ResolveDescriptorsInAssembly()`.  they'll get a default `TagHelperTypeResolver` instance unless they make a change to call the other (new) `base` constructor.  another way to phrase my question is: is that instance the correct one for tooling scenarios?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -112,7 +117,8 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
             var tagHelperTypes = _typeResolver.Resolve(assemblyName, documentLocation, errorSink);
 
             // Convert types to TagHelperDescriptors
-            var descriptors = tagHelperTypes.SelectMany(TagHelperDescriptorFactory.CreateDescriptors);
+            var descriptors = tagHelperTypes.SelectMany(
+                                        type => TagHelperDescriptorFactory.CreateDescriptors(assemblyName, type));
```
NTaylorMullen: just hit enter for spacing, easier to maintain.

pranavkm:
> is that instance the correct one for tooling scenarios?

Yes, this is specifically meant for precompilation - tooling can continue working with their current approach of using the default `TagHelperTypeResolver`. A change they could do is follow the same pattern that Mvc does - call the new ctor passing in their custom type resolver. This would avoid them having to specify a custom `ITagHelperDescriptorResolver`.

dougbu:
:shipit:

PR #310, opened 2015-02-20
Changes for tag helper precompilation with tests

pranavkm:


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -689,24 +689,38 @@ public void DescriptorResolver_DoesNotReturnInvalidTagHelpersWhenSpecified()
             Assert.Empty(descriptors);
         }
 
+        public static TheoryData<string> DescriptorResolver_IgnoresSpacesData
```
pranavkm: I totally missed updating tests when I first sent the PR. The `TagHelperDescriptorFactoryTests` changes are pretty trivial. This one is a bit more interesting - in the old code, we were flowing `Microsoft.AspNet.Razor.Runtime.TagHelpers.Valid_PlainTagHelper,MyAssembly` as the lookup text but resolving the types through the current assembly.
Consequently, the descriptor had an assembly name different than the lookup text and the test at https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs#L122 would fail resulting in an empty descriptor result.

With the change I made, the descriptor uses the passed in assembly name. Consequently, the check would pass and we'd have a TagHelperDescriptor that matched the passed in lookup text (in terms of type name + assembly). This is essentially how we make precompilation work - we produce a secondary assembly but have it behave as if the assembly name matched the original assembly.

If this was intentionally designed to behave this way, I can change this.

pranavkm:
cc @NTaylorMullen

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -31,14 +31,14 @@ public static class TagHelperDescriptorFactory
...
+        /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
         /// <param name="type">The type to create a <see cref="TagHelperDescriptor"/> from.</param>
         /// <returns>A <see cref="TagHelperDescriptor"/> that describes the given <paramref name="type"/>.</returns>
-        public static IEnumerable<TagHelperDescriptor> CreateDescriptors(Type type)
+        public static IEnumerable<TagHelperDescriptor> CreateDescriptors(string assemblyName, Type type)
         {
             var tagNames = GetTagNames(type);
             var typeName = type.FullName;
             var attributeDescriptors = GetAttributeDescriptors(type);
-            var assemblyName = type.GetTypeInfo().Assembly.GetName().Name;
```
NTaylorMullen: Why did this no longer work?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
NTaylorMullen: Definitely an interesting approach; turns out its definitely needed. Generating a second assembly is more of an escape hatch for pre-compilation but the generated descriptors need to match the ones that are created in the non-split assembly. Soooo :+1:

NTaylorMullen:
:shipit:

Issue #309, opened 2015-02-20
Support a directive to set a prefix that opts elements into Tag Helper processing (parse time)


DamianEdwards:
We should support a new directive that allows the app to set a prefix that is used to opt elements into Tag Helper processing at parse time. It should default to nothing (no prefix). This is the functional opposite of the `!` prefix for opting elements out of Tag Helper processing.

The directive should be hierarchical so that it can be set globally (in _GlobalImport.cshtml) or at any folder level.

It can be set to any string literal that the razor parser considers valid as part of an HTML element name.

.e.g. in _GlobalImport.cshtml
``` HTML
@tagHelperPrefix "th:"
```
in Views/Home/Index.cshtml
``` HTML
<input class="not-processed-for-tag-helpers" />
<th:input class="processed-for-tag-helpers" />
<cache>Not processed for tag helpers</cache>
<th:cache>Processed for tag helpers</th:cache>
```

NTaylorMullen:
@DamianEdwards what do you think should happen in the case that a user has two of these on a single page? Error right?

DamianEdwards:
@NTaylorMullen yep, error.

NTaylorMullen:
4d97a544f13a9bb1cf6d21a1c622a50af22add91
4bd02badda18e7097dbc38bcd67b48b7965f022f

PR #313, opened 2015-02-25
Add tagHelperPrefix directive.

NTaylorMullen:
Took the approach of attaching the prefix to `TagHelperDescriptor`s to enable tooling to separate `Prefix`s from `TagName`s in the document. Ran design by @rynowak.

*Change notes:*

- Updated TagHelperDescriptor to have a Prefix property. This enables new tooling scenarios such as refactoring prefixes or even giving them their own classification.
- Added invalid prefix cases '@', '!' and whitespace.
- Added TagHelperPrefix chunks, codegenerator, parsing logic to flow the directive through the Razor pipeline.
 Added parsing, sub-rewriting, rewriting and code generation tests for runtime/designtime.
- Fixed existing tests to utilize new class structures.
- Expanded block types to enable easy testing of prefixed TagHelpers.

#309

/cc @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -23,7 +23,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             {
                 { TagHelperDirectiveType.AddTagHelper, SyntaxConstants.CSharp.AddTagHelperKeyword },
                 { TagHelperDirectiveType.RemoveTagHelper, SyntaxConstants.CSharp.RemoveTagHelperKeyword },
+                { TagHelperDirectiveType.TagHelperPrefix, SyntaxConstants.CSharp.TagHelperPrefixKeyword },
             };
+        private static readonly HashSet<char> InvalidNonWhitespacePrefixCharacters =
+            new HashSet<char>(new[] { '@', '!', '<', '!', '/', '?', '[', '>', ']', '=', '"', '\'' });
 
         private readonly TagHelperTypeResolver _typeResolver;
 
@@ -51,7 +54,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             var resolvedDescriptors = new HashSet<TagHelperDescriptor>(TagHelperDescriptorComparer.Default);
 
-            foreach (var directiveDescriptor in context.DirectiveDescriptors)
+            // tagHelperPrefix directives do not affect which TagHelperDescriptors are added or removed from the final
+            // list, need to remove them.
+            var actionableDirectiveDescriptors = context.DirectiveDescriptors.Where(
+                directive => directive.DirectiveType != TagHelperDirectiveType.TagHelperPrefix);
+
+            foreach (var directiveDescriptor in actionableDirectiveDescriptors)
             {
                 try
                 {
@@ -69,9 +77,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                     }
                     else if (directiveDescriptor.DirectiveType == TagHelperDirectiveType.AddTagHelper)
                     {
-                        var descriptors = ResolveDescriptorsInAssembly(lookupInfo.AssemblyName,
-                                                                       directiveDescriptor.Location,
-                                                                       context.ErrorSink);
+                        var descriptors = ResolveDescriptorsInAssembly(
+                            lookupInfo.AssemblyName,
+                            directiveDescriptor.Location,
+                            context.ErrorSink);
 
                         // Only use descriptors that match our lookup info
                         descriptors = descriptors.Where(descriptor => MatchesLookupInfo(descriptor, lookupInfo));
@@ -89,12 +98,14 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         directiveDescriptor.Location,
                         Resources.FormatTagHelperDescriptorResolver_EncounteredUnexpectedError(
                             "@" + directiveName,
-                            directiveDescriptor.LookupText,
+                            directiveDescriptor.DirectiveText,
                             ex.Message));
                 }
             }
 
-            return resolvedDescriptors;
+            var prefixedDescriptors = PrefixDescriptors(context, resolvedDescriptors);
+
+            return prefixedDescriptors;
         }
 
         /// <summary>
@@ -110,9 +121,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <returns><see cref="TagHelperDescriptor"/>s for <see cref="ITagHelper"/>s from the given
         /// <paramref name="assemblyName"/>.</returns>
         // This is meant to be overridden by tooling to enable assembly level caching.
-        protected virtual IEnumerable<TagHelperDescriptor> ResolveDescriptorsInAssembly(string assemblyName,
-                                                                                        SourceLocation documentLocation,
-                                                                                        ParserErrorSink errorSink)
+        protected virtual IEnumerable<TagHelperDescriptor> ResolveDescriptorsInAssembly(
+            string assemblyName,
+            SourceLocation documentLocation,
+            ParserErrorSink errorSink)
         {
             // Resolve valid tag helper types from the assembly.
             var tagHelperTypes = _typeResolver.Resolve(assemblyName, documentLocation, errorSink);
@@ -122,6 +134,84 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 type => TagHelperDescriptorFactory.CreateDescriptors(assemblyName, type));
 
             return descriptors;
+        }
+
+        private static IEnumerable<TagHelperDescriptor> PrefixDescriptors(
+            TagHelperDescriptorResolutionContext context,
+            IEnumerable<TagHelperDescriptor> descriptors)
+        {
+            var tagHelperPrefix = ResolveTagHelperPrefix(context);
+
+            if (!string.IsNullOrEmpty(tagHelperPrefix))
+            {
+                return descriptors.Select(descriptor =>
+                    new TagHelperDescriptor(
+                        tagHelperPrefix,
+                        descriptor.TagName,
+                        descriptor.TypeName,
+                        descriptor.AssemblyName,
+                        descriptor.Attributes));
+            }
+
+            return descriptors;
+        }
+
+        private static string ResolveTagHelperPrefix(TagHelperDescriptorResolutionContext context)
+        {
+            var prefixDirectiveDescriptors = context.DirectiveDescriptors.Where(
+                descriptor => descriptor.DirectiveType == TagHelperDirectiveType.TagHelperPrefix);
+
+            TagHelperDirectiveDescriptor prefixDirective = null;
+
+            foreach (var directive in prefixDirectiveDescriptors)
+            {
+                if (prefixDirective == null)
+                {
+                    prefixDirective = directive;
+                }
+                else
+                {
+                    // For each invalid @tagHelperPrefix we need to create an error.
+                    context.ErrorSink.OnError(
+                        directive.Location,
+                        Resources.FormatTagHelperDescriptorResolver_InvalidTagHelperDirective(
+                            SyntaxConstants.CSharp.TagHelperPrefixKeyword));
+                }
+            }
+
+            var prefix = prefixDirective?.DirectiveText;
+
+            if (prefix != null && !EnsureValidPrefix(prefix, prefixDirective.Location, context.ErrorSink))
+            {
+                prefix = null;
+            }
+
+            return prefix;
+        }
+
+        private static bool EnsureValidPrefix(
+            string prefix,
+            SourceLocation directiveLocation,
+            ParserErrorSink errorSink)
+        {
+            foreach (var character in prefix)
+            {
+                // Prefixes are correlated with tag names, tag names cannot have whitespace.
+                if (char.IsWhiteSpace(character) ||
+                    InvalidNonWhitespacePrefixCharacters.Contains(character))
+                {
+                    errorSink.OnError(
+                        directiveLocation,
+                        Resources.FormatTagHelperDescriptorResolver_InvalidTagHelperPrefixValue(
+                            SyntaxConstants.CSharp.TagHelperPrefixKeyword,
+                            character,
+                            prefix));
+
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo lookupInfo)
@@ -146,7 +236,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         private static LookupInfo GetLookupInfo(TagHelperDirectiveDescriptor directiveDescriptor,
                                                 ParserErrorSink errorSink)
         {
-            var lookupText = directiveDescriptor.LookupText;
+            var lookupText = directiveDescriptor.DirectiveText;
             var lookupStrings = lookupText?.Split(new[] { ',' });
 
             // Ensure that we have valid lookupStrings to work with. Valid formats are:
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpDesignTimeHelpersVisitor.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpDesignTimeHelpersVisitor.cs
@@ -58,17 +58,22 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
             }
         }
 
+        protected override void Visit(TagHelperPrefixDirectiveChunk chunk)
+        {
+            VisitTagHelperDirectiveChunk(chunk.Prefix, chunk);
+        }
+
         protected override void Visit(AddTagHelperChunk chunk)
         {
-            VisitAddOrRemoveTagHelperChunk(chunk.LookupText, chunk);
+            VisitTagHelperDirectiveChunk(chunk.LookupText, chunk);
         }
 
         protected override void Visit(RemoveTagHelperChunk chunk)
         {
-            VisitAddOrRemoveTagHelperChunk(chunk.LookupText, chunk);
+            VisitTagHelperDirectiveChunk(chunk.LookupText, chunk);
         }
 
-        private void VisitAddOrRemoveTagHelperChunk(string lookupText, Chunk chunk)
+        private void VisitTagHelperDirectiveChunk(string text, Chunk chunk)
         {
             // We should always be in design time mode because of the calling AcceptTree method verification.
             Debug.Assert(Context.Host.DesignTimeMode);
@@ -81,12 +86,10 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
 
             Writer.WriteStartAssignment(TagHelperDirectiveSyntaxHelper);
 
-            // The parsing mechanism for the add/remove TagHelper chunk (CSharpCodeParser.TagHelperDirective())
-            // removes quotes that surround the lookupText.
+            // The parsing mechanism for a TagHelper directive chunk (CSharpCodeParser.TagHelperDirective())
+            // removes quotes that surround the text.
             _csharpCodeVisitor.CreateExpressionCodeMapping(
-                string.Format(
-                    CultureInfo.InvariantCulture,
-                    "\"{0}\"", lookupText),
+                string.Format(CultureInfo.InvariantCulture, "\"{0}\"", text),
                 chunk);
 
             Writer.WriteLine(";");
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperPrefixDirectiveChunk.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperPrefixDirectiveChunk.cs
@@ -1,0 +1,17 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.Generator.Compiler
+{
+    /// <summary>
+    /// A <see cref="Chunk"/> for the <c>@tagHelperPrefix</c> directive.
+    /// </summary>
+    public class TagHelperPrefixDirectiveChunk : Chunk
+    {
+        /// <summary>
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available 
+        /// tag helpers.
+        /// </summary>
+        public string Prefix { get; set; }
+    }
+}
--- a/src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
@@ -75,9 +75,11 @@ namespace Microsoft.AspNet.Razor.Generator
                 codeGenerator.Context.CodeTreeBuilder = new CodeTreeBuilder();
             }
 
+            var unprefixedTagName = tagHelperBlock.TagName.Substring(_tagHelperDescriptors.First().Prefix.Length);
+
             context.CodeTreeBuilder.StartChunkBlock(
                 new TagHelperChunk(
-                    tagHelperBlock.TagName,
+                    unprefixedTagName,
                     tagHelperBlock.SelfClosing,
                     attributes,
                     _tagHelperDescriptors),
--- a/src/Microsoft.AspNet.Razor/Generator/TagHelpers/TagHelperPrefixDirectiveCodeGenerator.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/TagHelpers/TagHelperPrefixDirectiveCodeGenerator.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Parser.SyntaxTree;
+
+namespace Microsoft.AspNet.Razor.Generator
+{
+    /// <summary>
+    /// A <see cref="SpanCodeGenerator"/> responsible for generating 
+    /// <see cref="Compiler.TagHelperPrefixDirectiveChunk"/>s.
+    /// </summary>
+    public class TagHelperPrefixDirectiveCodeGenerator : SpanCodeGenerator
+    {
+        /// <summary>
+        /// Instantiates a new <see cref="TagHelperPrefixDirectiveCodeGenerator"/>.
+        /// </summary>
+        /// <param name="prefix">
+        /// Text used as a required prefix when matching HTML.
+        /// </param>
+        public TagHelperPrefixDirectiveCodeGenerator(string prefix)
+        {
+            Prefix = prefix;
+        }
+
+        /// <summary>
+        /// Text used as a required prefix when matching HTML.
+        /// </summary>
+        public string Prefix { get; }
+
+        /// <summary>
+        /// Generates <see cref="Compiler.TagHelperPrefixDirectiveChunk"/>s.
+        /// </summary>
+        /// <param name="target">
+        /// The <see cref="Span"/> responsible for this <see cref="TagHelperPrefixDirectiveCodeGenerator"/>.
+        /// </param>
+        /// <param name="context">A <see cref="CodeGeneratorContext"/> instance that contains information about
+        /// the current code generation process.</param>
+        public override void GenerateCode(Span target, CodeGeneratorContext context)
+        {
+            context.CodeTreeBuilder.AddTagHelperPrefixDirectiveChunk(Prefix, target);
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
@@ -18,6 +18,7 @@ namespace Microsoft.AspNet.Razor.Parser
     {
         private void SetupDirectives()
         {
+            MapDirectives(TagHelperPrefixDirective, SyntaxConstants.CSharp.TagHelperPrefixKeyword);
             MapDirectives(AddTagHelperDirective, SyntaxConstants.CSharp.AddTagHelperKeyword);
             MapDirectives(RemoveTagHelperDirective, SyntaxConstants.CSharp.RemoveTagHelperKeyword);
             MapDirectives(InheritsDirective, SyntaxConstants.CSharp.InheritsKeyword);
@@ -27,14 +28,27 @@ namespace Microsoft.AspNet.Razor.Parser
             MapDirectives(LayoutDirective, SyntaxConstants.CSharp.LayoutKeyword);
         }
 
+        protected virtual void TagHelperPrefixDirective()
+        {
+            TagHelperDirective(
+                SyntaxConstants.CSharp.TagHelperPrefixKeyword,
+                prefix => new TagHelperPrefixDirectiveCodeGenerator(prefix));
+        }
+
         protected virtual void AddTagHelperDirective()
         {
-            TagHelperDirective(SyntaxConstants.CSharp.AddTagHelperKeyword, removeTagHelperDescriptors: false);
+            TagHelperDirective(
+                SyntaxConstants.CSharp.AddTagHelperKeyword,
+                lookupText =>
+                    new AddOrRemoveTagHelperCodeGenerator(removeTagHelperDescriptors: false, lookupText: lookupText));
         }
 
         protected virtual void RemoveTagHelperDirective()
         {
-            TagHelperDirective(SyntaxConstants.CSharp.RemoveTagHelperKeyword, removeTagHelperDescriptors: true);
+            TagHelperDirective(
+                SyntaxConstants.CSharp.RemoveTagHelperKeyword, 
+                lookupText =>
+                    new AddOrRemoveTagHelperCodeGenerator(removeTagHelperDescriptors: true, lookupText: lookupText));
         }
 
         protected virtual void LayoutDirective()
@@ -450,7 +464,7 @@ namespace Microsoft.AspNet.Razor.Parser
             Output(SpanKind.Code);
         }
 
-        private void TagHelperDirective(string keyword, bool removeTagHelperDescriptors)
+        private void TagHelperDirective(string keyword, Func<string, ISpanCodeGenerator> buildCodeGenerator)
         {
             AssertDirective(keyword);
 
@@ -492,8 +506,7 @@ namespace Microsoft.AspNet.Razor.Parser
                     // renders the C# to colorize the user provided value. We trim the quotes around the user's value
                     // so when we render the code we can project the users value into double quotes to not invoke C#
                     // IntelliSense.
-                    Span.CodeGenerator =
-                        new AddOrRemoveTagHelperCodeGenerator(removeTagHelperDescriptors, rawValue.Trim('"'));
+                    Span.CodeGenerator = buildCodeGenerator(rawValue.Trim('"'));
                 }
 
                 // We expect the directive to be surrounded in quotes.
--- a/src/Microsoft.AspNet.Razor/ParserResults.cs
+++ b/src/Microsoft.AspNet.Razor/ParserResults.cs
@@ -54,6 +54,7 @@ namespace Microsoft.AspNet.Razor
             TagHelperDescriptors = tagHelperDescriptors;
             ErrorSink = errorSink;
             ParserErrors = errorSink.Errors;
+            Prefix = tagHelperDescriptors.FirstOrDefault()?.Prefix;
         }
 
         /// <summary>
@@ -81,5 +82,10 @@ namespace Microsoft.AspNet.Razor
         /// The <see cref="TagHelperDescriptor"/>s found for the current Razor document.
         /// </summary>
         public IEnumerable<TagHelperDescriptor> TagHelperDescriptors { get; }
+
+        /// <summary>
+        /// Text used as a required prefix when matching HTML.
+        /// </summary>
+        public string Prefix { get; }
     }
 }
--- a/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
+++ b/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
@@ -17,10 +17,11 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         internal TagHelperDescriptor([NotNull] string tagName,
                                      [NotNull] string typeName,
                                      [NotNull] string assemblyName)
-            : this(tagName,
-                   typeName,
-                   assemblyName,
-                   Enumerable.Empty<TagHelperAttributeDescriptor>())
+            : this(
+                tagName,
+                typeName,
+                assemblyName,
+                attributes: Enumerable.Empty<TagHelperAttributeDescriptor>())
         {
         }
 
@@ -35,21 +36,66 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         /// <param name="attributes">
         /// The <see cref="TagHelperAttributeDescriptor"/>s to request from the HTML tag.
         /// </param>
-        public TagHelperDescriptor([NotNull] string tagName,
-                                   [NotNull] string typeName,
-                                   [NotNull] string assemblyName,
-                                   [NotNull] IEnumerable<TagHelperAttributeDescriptor> attributes)
+        public TagHelperDescriptor(
+            [NotNull] string tagName,
+            [NotNull] string typeName,
+            [NotNull] string assemblyName,
+            [NotNull] IEnumerable<TagHelperAttributeDescriptor> attributes)
+            : this(
+                prefix: string.Empty,
+                tagName: tagName, 
+                typeName: typeName, 
+                assemblyName: assemblyName, 
+                attributes: attributes)
         {
+        }
+
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="TagHelperDescriptor"/> class with the given
+        /// <paramref name="attributes"/>.
+        /// </summary>
+        /// <param name="prefix">
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available 
+        /// tag helpers.
+        /// </param>
+        /// <param name="tagName">The tag name that the tag helper targets. '*' indicates a catch-all
+        /// <see cref="TagHelperDescriptor"/> which applies to every HTML tag.</param>
+        /// <param name="typeName">The full name of the tag helper class.</param>
+        /// <param name="assemblyName">The name of the assembly containing the tag helper class.</param>
+        /// <param name="attributes">
+        /// The <see cref="TagHelperAttributeDescriptor"/>s to request from the HTML tag.
+        /// </param>
+        public TagHelperDescriptor(
+            string prefix,
+            [NotNull] string tagName,
+            [NotNull] string typeName,
+            [NotNull] string assemblyName,
+            [NotNull] IEnumerable<TagHelperAttributeDescriptor> attributes)
+        {
+            Prefix = prefix ?? string.Empty;
             TagName = tagName;
+            FullTagName = Prefix + TagName;
             TypeName = typeName;
             AssemblyName = assemblyName;
             Attributes = new List<TagHelperAttributeDescriptor>(attributes);
         }
 
         /// <summary>
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available 
+        /// tag helpers.
+        /// </summary>
+        public string Prefix { get; private set; }
+
+        /// <summary>
         /// The tag name that the tag helper should target.
         /// </summary>
         public string TagName { get; private set; }
+
+        /// <summary>
+        /// The full tag name that is required for the tag helper to target an HTML element.
+        /// </summary>
+        /// <remarks>This is equivalent to <see cref="Prefix"/> and <see cref="TagName"/> concatenated.</remarks>
+        public string FullTagName { get; private set; }
 
         /// <summary>
         /// The full name of the tag helper class.
--- a/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
+++ b/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
@@ -29,12 +29,14 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         /// <c>false</c> otherwise.</returns>
         /// <remarks>
         /// Determines equality based on <see cref="TagHelperDescriptor.TypeName"/>,
-        /// <see cref="TagHelperDescriptor.AssemblyName"/> and <see cref="TagHelperDescriptor.TagName"/>.
+        /// <see cref="TagHelperDescriptor.AssemblyName"/>, <see cref="TagHelperDescriptor.TagName"/> and
+        /// <see cref="TagHelperDescriptor.Prefix"/>.
         /// </remarks>
         public bool Equals(TagHelperDescriptor descriptorX, TagHelperDescriptor descriptorY)
         {
             return string.Equals(descriptorX.TypeName, descriptorY.TypeName, StringComparison.Ordinal) &&
                    string.Equals(descriptorX.TagName, descriptorY.TagName, StringComparison.OrdinalIgnoreCase) &&
+                   string.Equals(descriptorX.Prefix, descriptorY.Prefix, StringComparison.OrdinalIgnoreCase) &&
                    string.Equals(descriptorX.AssemblyName, descriptorY.AssemblyName, StringComparison.Ordinal);
         }
 
--- a/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
+++ b/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         private const string CatchAllDescriptorTarget = "*";
 
         private IDictionary<string, HashSet<TagHelperDescriptor>> _registrations;
+        private string _tagHelperPrefix;
 
         /// <summary>
         /// Instantiates a new instance of the <see cref="TagHelperDescriptorProvider"/>.
@@ -43,10 +44,18 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         {
             HashSet<TagHelperDescriptor> descriptors;
 
-            // Ensure there's an ISet to use.
+            if (!string.IsNullOrEmpty(_tagHelperPrefix) &&
+                (tagName.Length <= _tagHelperPrefix.Length ||
+                !tagName.StartsWith(_tagHelperPrefix, StringComparison.OrdinalIgnoreCase)))
+            {
+                // The tagName doesn't have the tag helper prefix, we can short circuit.
+                return Enumerable.Empty<TagHelperDescriptor>();
+            }
+
+            // Ensure there's a HashSet to use.
             if (!_registrations.TryGetValue(CatchAllDescriptorTarget, out descriptors))
             {
-                descriptors = new HashSet<TagHelperDescriptor>();
+                descriptors = new HashSet<TagHelperDescriptor>(TagHelperDescriptorComparer.Default);
             }
 
             // If the requested tag name is the catch-all target, we should short circuit.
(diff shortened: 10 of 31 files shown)


dougbu:
> Added invalid prefix cases '@', '!' and whitespace

No to a blacklist.  Instead confirm a white list, specifically the prefix must contain only allowed HTML tag name characters, including `'-'` and `':'`.  Or just reuse whatever Razor does to confirm tag name composition for now and file a bug to confirm those characters correctly.

dougbu:
> Updated TagHelperDescriptor to have a Prefix property. This enables new tooling scenarios such as refactoring prefixes or even giving them their own classification.

This is not a feature at this time.  Why build in support for a _future_ enhancement when the current feature requires only that `TagHelperDescriptor` contains the correctly-prefixed tag name?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -32,19 +32,25 @@ public static class TagHelperDescriptorFactory
...
-                new TagHelperDescriptor(tagName,
-                                        typeName,
-                                        assemblyName,
-                                        attributeDescriptors));
+                new TagHelperDescriptor(
+                    prefix,
+                    tagName,
+                    typeName,
+                    assemblyName,
+                    attributeDescriptors));
```
dougbu: so much simpler and less-invasive if you create an unchanged `TagHelperDescriptor` with `prefix + tagName` for the `tagName` argument

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -89,7 +97,7 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
                         directiveDescriptor.Location,
                         Resources.FormatTagHelperDescriptorResolver_EncounteredUnexpectedError(
                             "@" + directiveName,
-                            directiveDescriptor.LookupText,
+                            directiveDescriptor.DirectiveText,
```
dougbu: why did this line change?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -32,19 +32,25 @@ public static class TagHelperDescriptorFactory
         /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="type"/>.
         /// </summary>
         /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
+        /// <param name="prefix">The prefix to the created <see cref="TagHelperDescriptor"/>s.</param>
```
dougbu: nit: "to the" -> "for tag names in" throughout PR

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -104,26 +112,85 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
 
             return descriptors;
         }
 
+        private static string ResolveTagHelperPrefix(TagHelperDescriptorResolutionContext context)
+        {
+            var prefixDirectiveDescriptors = context.DirectiveDescriptors.Where(
+                descriptor => descriptor.DirectiveType == TagHelperDirectiveType.TagHelperPrefix);
+
+            if (prefixDirectiveDescriptors.Count() > 1)
```
dougbu: this extra condition is unnecessary since `Skip(1)` will simply return an empty enumeration if `prefixDirectiveDescriptors` is empty or contains one item.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/Visitors/CSharpDesignTimeHelpersVisitor.cs
```diff
@@ -81,13 +86,9 @@ private void VisitAddOrRemoveTagHelperChunk(string lookupText, Chunk chunk)
...
 
-            // The parsing mechanism for the add/remove TagHelper chunk (CSharpCodeParser.TagHelperDirective())
+            // The parsing mechanism for a TagHelper directive chunk (CSharpCodeParser.TagHelperDirective())
             // removes quotes that surround the lookupText.
-            _csharpCodeVisitor.CreateExpressionCodeMapping(
-                string.Format(
-                    CultureInfo.InvariantCulture,
-                    "\"{0}\"", lookupText),
-                chunk);
+            _csharpCodeVisitor.CreateExpressionCodeMapping($"\"{text}\"", chunk);
```
dougbu: this changes the `IFormatProvider` without reason.  leave as-is.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -27,14 +28,27 @@ private void SetupDirectives()
             MapDirectives(LayoutDirective, SyntaxConstants.CSharp.LayoutKeyword);
         }
 
+        protected virtual void TagHelperPrefixDirective()
+        {
+            TagHelperDirective(
```
dougbu: why reuse this method for an additional, fairly-different purpose?  a new method for the prefix directive would be simpler and you would only need to change the existing method's name

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDirectiveDescriptor.cs
```diff
@@ -11,23 +11,24 @@ namespace Microsoft.AspNet.Razor.TagHelpers
     public class TagHelperDirectiveDescriptor
     {
         // Internal for testing purposes.
-        internal TagHelperDirectiveDescriptor(string lookupText,
+        internal TagHelperDirectiveDescriptor(string directiveText,
```
dougbu: existing class is well-named for existing purpose.  why extend it to support this new, quite different directive?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -28,6 +46,7 @@ public class TagHelperDescriptor
         /// Instantiates a new instance of the <see cref="TagHelperDescriptor"/> class with the given
         /// <paramref name="attributes"/>.
         /// </summary>
+        /// <param name="prefix">The prefix of the HTML element name the tag helper targets.</param>
```
dougbu: prefix is a global setting, nothing near the level of individual tag helpers or their descriptions.  do not change this class.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: thinking about this a bit and talking a couple of things through w/ @rynowak, I realize you've got two uses of these descriptors.  on one hand you have descriptor resolution, where the match between add/removeTagHelper directives and discovered tag helpers should use the original (from the tag helper) tag name.  on the other hand, the rewriter needs to match Razor tags against the full name of a tag helper.  note the prefix serves neither purpose and causes repeated string concatenations.

I suggest:
- add the `Prefix` to the `TagHelperDescriptorResolutionContext` (or another high-level context) so tooling can use it (if necessary)
- add a `string ShortName { get; }` property to `TagHelperDescriptor` and change the tag helper filtering code to use that instead.  seems like the less-used `TagName` requirement.  but flip it around to be `TagName` and `FullName` if that minimizes the churn.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: `directiveDescriptor` is now used for `@addTagHelper`, `@removeTagHelper`, `@tagHelperPrefix` so `LookupText` didn't make sense.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: :-1: The format of all tag helper directives is the same, from a parsing perspective they're nearly identical, making a whole other method where the largest change is the CodeGenerator isn't worth it.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
dougbu: ask I ask later -- why reuse the same `TagHelperDirectiveDescriptor` type?  you're mixing concepts and then the code that consumes it has to split things again.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDirectiveDescriptor.cs
NTaylorMullen: It's a `TagHelper` related directive, all of which have the same basic format.

dougbu: admit I may be incorrect about `TagHelperDirective`.  but at this point you're dealing with separate concepts that are handled on distinct code paths in the consuming code.  no longer dealing with format at all.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: First bullet point: `TagHelperDescriptor`'s building process logic exists in the `TagHelperDescriptorResolver`; therefore I feel that anything that affects them should also exist in the `TagHelperDescriptorResolver`, not to mention they all hold the same basic string format/directive structure. 

For your second bullet point I don't see the value in abstracting the `Prefix` into a combination of the element name + prefix from a `TagHelperDescriptor`s perspective. The `TagHelperDescriptor` should be extremely dumb/purely informational, consumers should deal with understanding what it applies to in the document.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDirectiveDescriptor.cs
NTaylorMullen: Sure the end purpose of the tag helper directive may be different but their core concepts are all the same: affect the creation of `TagHelperDescriptor`s  Their core concept is handled by the `TagHelperDescriptorResolver` who only understands provided `TagHelperDirectiveDescriptor`s, no reason to build a separate code path outside of this core `TagHelperDescriptor` building mechanism

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: the information to put in the extremely dumb `TagHelperDescriptor` is what the code requires.  everything is doing matches against either the short or the long name.  so provide those.  don't force repeated concatenations or surface a "prefix" concept when the code just wants a single string to match.

NTaylorMullen: I'd much rather not add complexity to the `TagHelperDescriptor`, it should represent everything that makes up a valid `TagHelper`, in this case the HTMLs `TagName` (what should eventually hit the page) and its optional `Prefix`. Not to mention the editor uses these `TagHelperDescriptor`s and will also use the `Prefix` property directly.

NTaylorMullen:
@dougbu sadly Razor doesn't use any complex logic to validate a tag name which is why I used the blacklist approach.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: Talked in person.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -51,7 +51,12 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
         {
```
dougbu: also update `_directiveNames` at the top of this class

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -124,6 +133,79 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
+            // For each invalid @tagHelperPrefix we need to create an error.
+            foreach (var erroredContext in prefixDirectiveDescriptors.Skip(1))
+            {
+                context.ErrorSink.OnError(
+                    erroredContext.Location,
+                    Resources.FormatTagHelperDescriptorResolver_InvalidTagHelperDirective(
+                        SyntaxConstants.CSharp.TagHelperPrefixKeyword));
+            }
+
+            var prefixDirective = prefixDirectiveDescriptors.FirstOrDefault();
```
dougbu: now see you could do a single loop through `prefixDirectiveDescriptors`.  might be more clear e.g.
``` c#
TagHelperDirective prefixDirective = null;
foreach (var directive in prefixDirectiveDescriptors)
{
  if (prefixDirective == null)
  {
    prefixDirective = directive;
  }
  else
  {
    context.ErrorSink.OnError(...);
  }
}
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -124,6 +133,79 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
+            return descriptors;
+        }
+
+        private static string ResolveTagHelperPrefix(TagHelperDescriptorResolutionContext context)
+        {
+            var prefixDirectiveDescriptors = context.DirectiveDescriptors.Where(
+                descriptor => descriptor.DirectiveType == TagHelperDirectiveType.TagHelperPrefix);
+
+            // For each invalid @tagHelperPrefix we need to create an error.
+            foreach (var erroredContext in prefixDirectiveDescriptors.Skip(1))
```
dougbu: `erroredContext` -> `directive` ("errored" isn't a word and the variable doesn't hold a context)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -124,6 +133,79 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
+                    Resources.FormatTagHelperDescriptorResolver_InvalidTagHelperDirective(
+                        SyntaxConstants.CSharp.TagHelperPrefixKeyword));
+            }
+
+            var prefixDirective = prefixDirectiveDescriptors.FirstOrDefault();
+            var prefix = prefixDirective?.DirectiveText;
+
+            if (prefix == null || !EnsureValidPrefix(prefix, prefixDirective.Location, context.ErrorSink))
+            {
+                prefix = string.Empty;
```
dougbu: why not return `null`?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -124,6 +133,79 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
+            ParserErrorSink errorSink)
+        {
+            foreach (var character in prefix)
+            {
+                // Prefixes are correlated with tag names, tag names cannot have whitespace.
+                if (char.IsWhiteSpace(character) ||
+                    // Reserved Razor transition symbol
+                    character == '@' ||
+                    // Reserved TagHelper opt-out character.
+                    character == '!')
```
dougbu: from `HtmlSymbolType` I'd expect this blacklist (generally poor idea despite the existing mess w.r.t. character exclusion but :ok: for now) to be much longer.  should at least exclude everything that tokenizer will ensure is not found in a `HtmlSymbolType.Text` token.

src/Microsoft.AspNet.Razor/Generator/TagHelpers/TagHelperPrefixDirectiveCodeGenerator.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Parser.SyntaxTree;
+
+namespace Microsoft.AspNet.Razor.Generator
+{
+    /// <summary>
+    /// A <see cref="SpanCodeGenerator"/> responsible for generating <see cref="Compiler.TagHelperPrefixDirectiveChunk"/>s.
```
dougbu: long line

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperPrefixDirectiveChunk.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.AspNet.Razor.Generator.Compiler
+{
+    /// <summary>
+    /// A <see cref="Chunk"/> for the @tagHelperPrefix directive.
+    /// </summary>
+    public class TagHelperPrefixDirectiveChunk : Chunk
+    {
+        /// <summary>
+        /// Text used as the prefix for <see cref="TagHelpers.TagHelperDescriptor"/>s.
```
dougbu: nit: storing a value in a descriptor isn't a "use". suggest `Text used as a required prefix when matching HTML start and end tags in the Razor source to available tag helpers.`  (might work without "in the Razor source")

similar comments about the descriptor rather than actual use appear elsewhere in this PR.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -28,7 +28,11 @@ public static class TagHelperBlockRewriter
             var attributes = GetTagAttributes(tagName, validStructure, tag, descriptors, errorSink);
             var selfClosing = IsSelfClosing(tag);
 
-            return new TagHelperBlockBuilder(tagName, selfClosing, start, attributes, descriptors);
+            // All descriptors for an element should share a common prefix, grab the first.
+            var prefix = descriptors.First().Prefix;
+            var tagHelperTagName = tagName.Substring(prefix.Length);
```
dougbu: why does this code exist?  the descriptor should contain all `string`s of interest; no `Substring` required

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -42,6 +43,11 @@ public TagHelperBlock(TagHelperBlockBuilder source)
         }
 
         /// <summary>
+        /// The tag helper prefix.
+        /// </summary>
+        public string Prefix { get; }
```
dougbu: you have the `Descriptors`, why surface the `Prefix` again?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -75,6 +83,11 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         }
 
         /// <summary>
+        /// The tag helper prefix.
+        /// </summary>
+        public string Prefix { get; set; }
```
dougbu: have the descriptors; why is this exposed?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -141,11 +141,15 @@ private bool TryRewriteTagHelper(Block tagBlock, RewritingContext context)
...
 
-                var tagNameScope = _tagStack.Count > 0 ? _tagStack.Peek().TagName : string.Empty;
+                var parentTagHelper = _tagStack.Count > 0 ? _tagStack.Peek() : null;
 
                 // Validate that our end tag helper matches the currently scoped tag helper, if not we
                 // need to error.
-                if (tagNameScope.Equals(tagName, StringComparison.OrdinalIgnoreCase))
+                if (parentTagHelper != null &&
+                    string.Equals(
+                        parentTagHelper.Prefix + parentTagHelper.TagName,
```
dougbu: do not do another concatenation.  something is wrong with the information available to this method.  probably best to simply have `parentTagHelper.TagName` contain the full name that was matched for the start tag.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -35,23 +36,64 @@ public class TagHelperDescriptor
         /// <param name="attributes">
         /// The <see cref="TagHelperAttributeDescriptor"/>s to request from the HTML tag.
         /// </param>
-        public TagHelperDescriptor([NotNull] string tagName,
-                                   [NotNull] string typeName,
-                                   [NotNull] string assemblyName,
-                                   [NotNull] IEnumerable<TagHelperAttributeDescriptor> attributes)
+        public TagHelperDescriptor(
```
dougbu: what call sites remain for this legacy constructor?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
```diff
@@ -29,12 +29,14 @@ public class TagHelperDescriptorComparer : IEqualityComparer<TagHelperDescriptor
         /// <c>false</c> otherwise.</returns>
         /// <remarks>
         /// Determines equality based on <see cref="TagHelperDescriptor.TypeName"/>,
-        /// <see cref="TagHelperDescriptor.AssemblyName"/> and <see cref="TagHelperDescriptor.TagName"/>.
+        /// <see cref="TagHelperDescriptor.AssemblyName"/> <see cref="TagHelperDescriptor.TagName"/> and
```
dougbu: nit: comma after `<see cref="TagHelperDescriptor.AssemblyName"/>`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -35,23 +36,64 @@ public class TagHelperDescriptor
...
+            FullTagName = Prefix + TagName;
             TypeName = typeName;
             AssemblyName = assemblyName;
             Attributes = new List<TagHelperAttributeDescriptor>(attributes);
         }
 
         /// <summary>
+        /// The prefix of the HTML element name the tag helper targets.
+        /// </summary>
+        public string Prefix { get; private set; }
```
dougbu: why does this have a setter?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -35,23 +36,64 @@ public class TagHelperDescriptor
...
+
+        /// <summary>
         /// The tag name that the tag helper should target.
         /// </summary>
         public string TagName { get; private set; }
```
dougbu: while here only: why does this have a setter?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -35,23 +36,64 @@ public class TagHelperDescriptor
...
 
         /// <summary>
+        /// The full tag name that is required for the tag helper to target an HTML element.
+        /// </summary>
+        /// <remarks>This is equivalent to <see cref="Prefix"/> and <see cref="TagName"/> concatenated.</remarks>
+        public string FullTagName { get; }
```
dougbu: this property is set but unused otherwise.  should be used (exclusively) when matching tag helper names from the Razor source.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -76,7 +78,9 @@ private void Register(TagHelperDescriptor descriptor)
             if (!_registrations.TryGetValue(descriptor.TagName, out descriptorSet))
             {
                 descriptorSet = new HashSet<TagHelperDescriptor>(TagHelperDescriptorComparer.Default);
-                _registrations[descriptor.TagName] = descriptorSet;
+
+                // We want to register TagHelperDescriptors by their prefix and tag name for easy lookup.
+                _registrations[descriptor.Prefix + descriptor.TagName] = descriptorSet;
```
dougbu: use `descriptor.FullTagName` here too

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -41,6 +42,465 @@ private static TagHelperDescriptor Valid_InheritedTagHelperDescriptor
...
+                    "Invalid tag helper directive '{0}' value. '{1} is not allowed in prefix '{2}'.";
+
+                return new TheoryData<Dictionary<string, IEnumerable<Type>>, // descriptorAssemblyLookups
+                                      IEnumerable<TagHelperDirectiveDescriptor>, // directiveDescriptors
+                                      IEnumerable<TagHelperDescriptor>, // expectedDescriptors
+                                      IEnumerable <RazorError>> // expectedErrors
+                {
+                    {
+                        defaultAssemblyLookups,
+                        new []
```
dougbu: tiny tiny nit: no space between `new` and `[]`

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: We want the `TagName` to be the exact case that it is in the document, a `TagHelperDescriptor` is case insensitive where this block element is here to represent what the user typed.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: This is the ctor used by the `TagHelperDescriptorFactory`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: This class was built so it could eventually be serialized.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: Same answer as above

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
NTaylorMullen: This is the answer to this that I agreed on: https://github.com/aspnet/Razor/pull/313#discussion_r25491125

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: to confirm: when all associated tag helpers no-op, the prefix will _not_ appear in the generated HTML.  if so, fine, that sounds consistent with the opt-out (`!`) handling.

but the prefix isn't used later.  need only the short name when generating the `TagHelperScopeManager.Begin()` call.  so passing it around together with descriptors containing exactly the same `string` is not worthwhile.

what may be useful is calculating the global prefix length once per parse, say at the top of this class.  then to get the short name from the full matched text, only need the `Substring` call.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: I asked due to the visible inconsistency between `Prefix` and `FullTagName`.  Guess it's `FullTagName` that needs to change.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: No reason anymore, :+1:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: `TagHelperBlockRewriter` is a static class, will not pass the prefix onto the `TagHelperBlock`/`TagHelperBlockBuilder` but will need to calculate the prefix length for each rewrite.

NTaylorMullen: Actually scratch that, need this for testing scenarios and also in the `TagHelperParseTreeRewriter` matching scenario, we can't just pull from the `TagHelperDescriptor`s because their tagname may be `*`, just realized the `*` case is a hole I need to fix.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Have to based on our convo here: https://github.com/aspnet/Razor/pull/313#discussion_r25537282

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: FullTagName is a meta property, that's set via the ctor, it doesn't need to have a setter.

dougbu: constructors aren't run during deserialization

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: This actually can't happen due to catch-all tag helpers

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeTree/Chunks/TagHelpers/TagHelperPrefixDirectiveChunk.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.Generator.Compiler
+{
+    /// <summary>
+    /// A <see cref="Chunk"/> for the @tagHelperPrefix directive.
```
dougbu: nit: <c>@tagHelperPrefix</c>

src/Microsoft.AspNet.Razor/Generator/TagHelpers/TagHelperPrefixDirectiveCodeGenerator.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <see cref="Compiler.TagHelperPrefixDirectiveChunk"/>s.
+    /// </summary>
+    public class TagHelperPrefixDirectiveCodeGenerator : SpanCodeGenerator
+    {
+        /// <summary>
+        /// Instantiates a new <see cref="TagHelperPrefixDirectiveCodeGenerator"/>.
+        /// </summary>
+        /// <param name="prefix">
+        /// Text used as <see cref="TagHelpers.TagHelperDescriptor"/>'s 
+        /// <see cref="TagHelpers.TagHelperDescriptor.Prefix"/>.
```
dougbu: use your "Text used as a required prefix when matching HTML ..." wording in all relevant doc comments

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -124,6 +136,85 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
...
+                }
+                else
+                {
+                    // For each invalid @tagHelperPrefix we need to create an error.
+                    context.ErrorSink.OnError(
+                        directive.Location,
+                        Resources.FormatTagHelperDescriptorResolver_InvalidTagHelperDirective(
+                            SyntaxConstants.CSharp.TagHelperPrefixKeyword));
+                }
+
```
dougbu: nit: extra blank line

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -42,6 +43,12 @@ public TagHelperBlock(TagHelperBlockBuilder source)
         }
 
         /// <summary>
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available 
+        /// tag helpers.
+        /// </summary>
+        public string Prefix { get; }
```
dougbu: this property should not exist.  it's currently used only when creating a `TagHelperBlockBuilder` from a `TagHelperBlock original`.  that overload is unused.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -75,6 +83,12 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         }
 
         /// <summary>
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available 
+        /// tag helpers.
+        /// </summary>
+        public string Prefix { get; set; }
```
dougbu: this class should have a single purpose -- gathering the information necessary to create a `TagHelperBlock`.  this property exists because it's also used on a stack in `TagHelperParseTreeRewriter`.  if you must use it that way, save the `FullTagName` you actually need in `TagHelperParseTreeRewriter.TryRewriteTagHelper()` to match an end tag to its start tag.

btw the current code saves the descriptor's prefix and so matches `<SomeTag>` with `</SomeTag>` if the configured prefix is "some" only because you're doing a case-insensitive match.  will be much more efficient if you save the exact text from the Razor source because the user likely used the same case.  (side note: I'm also not sure if HTML case-insensitivity extends to matching start and end tags with different cases.)

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -28,7 +28,11 @@ public static class TagHelperBlockRewriter
...
+
+            return new TagHelperBlockBuilder(prefix, tagHelperTagName, selfClosing, start, attributes, descriptors);
```
dougbu: as mentioned in `TagHelperBlockBuilder` either don't pass `prefix` or pass `tagName` (to become `FullTagName`) instead

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
NTaylorMullen: For the `TagHelperBlock` this property is used for testing, I could mark it internal.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
NTaylorMullen: Will make a `FullTagName` property, remove this and the `FullTagName` property will calculate it on the fly by looking at the descriptors. This way it makes the builder accurate depending on its descriptors and reduces the amount of work we do outside of the builder.

dougbu: sounds right but please clarify "calculate it": what needs to be calculated?  since a `TagHelperBlockBuilder` is created when full name (from the Razor source) is known, I hope you don't mean `FullTagName`.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
dougbu: should not have extra members (other than sometimes extra constructor overloads) in product code just for testing.  why test anything related to the prefix?  the important things are matching tags with descriptors and start tags with end tags.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -28,7 +28,11 @@ public static class TagHelperBlockRewriter
             var attributes = GetTagAttributes(tagName, validStructure, tag, descriptors, errorSink);
             var selfClosing = IsSelfClosing(tag);
 
-            return new TagHelperBlockBuilder(tagName, selfClosing, start, attributes, descriptors);
+            // All descriptors for an element should share a common prefix, grab the first.
+            var prefix = descriptors.First().Prefix;
```
dougbu: neither `prefix` nor `prefix.Length` change for the lifetime of a `TagHelperParseTreeRewriter` instance.  calculate length once in that class's constructor and pass it into this method.

src/Microsoft.AspNet.Razor/ParserResults.cs
```diff
@@ -81,5 +82,10 @@ public class ParserResults
         /// The <see cref="TagHelperDescriptor"/>s found for the current Razor document.
         /// </summary>
         public IEnumerable<TagHelperDescriptor> TagHelperDescriptors { get; }
+
+        /// <summary>
+        /// The prefix required to use tag helpers.
```
dougbu: "Text used as a required prefix when matching HTML ..." wording makes sense here too.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: need `private set;` or value will always be `null` after deserialization

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDirectiveType.cs
```diff
@@ -16,6 +16,11 @@ public enum TagHelperDirectiveType
         /// <summary>
         /// A @removeTagHelper directive.
         /// </summary>
-        RemoveTagHelper
+        RemoveTagHelper,
+
+        /// <summary>
+        /// A @tagHelperPrefix directive.
```
dougbu: nit (because comment is currently consistent with that of other values): `<c>@tagHelperPrefix</c>`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -41,6 +42,465 @@ private static TagHelperDescriptor Valid_InheritedTagHelperDescriptor
...
+                                      IEnumerable<TagHelperDirectiveDescriptor>, // directiveDescriptors
+                                      IEnumerable<TagHelperDescriptor>, // expectedDescriptors
+                                      IEnumerable<RazorError>> // expectedErrors
+                {
+                    {
+                        defaultAssemblyLookups,
+                        new[]
+                        {
+                            new TagHelperDirectiveDescriptor(
+                                "th:",
```
dougbu: test other valid prefixes such as "th-" or just "th".  current coverage makes it seem like the colon is required.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -72,11 +80,21 @@ private void Register(TagHelperDescriptor descriptor)
...
-            if (!_registrations.TryGetValue(descriptor.TagName, out descriptorSet))
+            if (_tagHelperPrefix == null)
+            {
+                _tagHelperPrefix = descriptor.Prefix;
+            }
+
+            var registrationKey = 
+                string.Equals(descriptor.TagName, CatchAllDescriptorTarget, StringComparison.Ordinal) ? 
+                CatchAllDescriptorTarget : 
+                descriptor.FullTagName;
```
dougbu: nit: trailing whitespace.  suggest installing the trailing whitespace visualizer (thx @madskristensen !)

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -75,6 +83,12 @@ public TagHelperBlockBuilder(TagHelperBlock original)
         }
 
         /// <summary>
+        /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available
```
dougbu: nit: trailing whitespace

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -41,6 +42,465 @@ private static TagHelperDescriptor Valid_InheritedTagHelperDescriptor
...
+            Assert.Equal(expectedErrors, errorSink.Errors);
+            Assert.Equal(expectedDescriptors.Count(), descriptors.Count());
+
+            foreach (var expectedDescriptor in expectedDescriptors)
+            {
+                Assert.Contains(expectedDescriptor, descriptors, TagHelperDescriptorComparer.Default);
+            }
+        }
+
+        public static TheoryData ResolveDirectiveDescriptorsTagHelperPrefixData
```
dougbu: test a `TagHelperDirective` of type `TagHelperDirectiveType.TagHelperPrefix` with an empty prefix.  is that allowed or not?  can't tell from the tests in this class.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -59,50 +30,64 @@ public static TheoryData TagHelperDescriptorFlowTestData
...
-                        PAndInputTagHelperDescriptors,
+                        DefaultPAndInputTagHelperDescriptors,
                         Enumerable.Empty<TagHelperDescriptor>(),
                         false
                     },
                     {
+                        "BasicTagHelpers.Prefixed",
+                        "BasicTagHelpers.Prefixed",
+                        BuildPAndInputTagHelperDescriptors(prefix: "th:"),
+                        BuildPAndInputTagHelperDescriptors(prefix: "th:"),
```
dougbu: - vary the valid prefixes in these more end-to-end tests as well
- nit: create a property containing the descriptors with this prefix.  (could do that using a deep copy of `DefaultPAndInputTagHelperDescriptors` instead of churning that code.)

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpDirectivesTest.cs
```diff
@@ -13,6 +13,88 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
     public class CSharpDirectivesTest : CsHtmlCodeParserTestBase
     {
         [Fact]
+        public void TagHelperPrefixDirective_Succeeds()
```
dougbu: also test `"@tagHelperPrefix \"\""`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDescriptorProviderTest.cs
```diff
@@ -10,7 +10,78 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
     public class TagHelperDescriptorProviderTest
     {
         [Fact]
-        public void TagHelperDescriptorProvider_GetTagHelpersReturnsNothingForUnregisteredTags()
+        public void GetTagHelpers_OnlyUnderstandsSinglePrefix()
```
dougbu: test what happens when a catch-all descriptor has prefix "th" and `// Act` calls `provider.GetTagHelpers("th")`.  suspect for example `<th/>` will lead to an odd error or generation of a self-closing tag with no name.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,142 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData PrefixedTagHelperBoundData
```
dougbu: - mix up prefixes
- test `<myth />` when prefix is "myth"

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
NTaylorMullen: Will change prefix here, going to keep the BuildDescriptor bits

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDescriptorProviderTest.cs
NTaylorMullen: :+1:, adding case in the `TagHelperDescriptorProvider`.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: done

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
NTaylorMullen: done

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
NTaylorMullen: done

NTaylorMullen:
Updated. Moved all prefix/Full tag name logic for the `TagHelperBlock`/`TagHelperBlockBuilder` into the `TagHelperCodeGenerator`, this way the `SyntaxTreeNode`s only ever represent what is in the document (the full tag name).

src/Microsoft.AspNet.Razor/Generator/TagHelperCodeGenerator.cs
```diff
@@ -75,9 +75,11 @@ public override void GenerateStartBlockCode(Block target, CodeGeneratorContext c
                 codeGenerator.Context.CodeTreeBuilder = new CodeTreeBuilder();
             }
 
+            var unprefixedTagName = tagHelperBlock.TagName.Substring(_tagHelperDescriptors.First().Prefix.Length);
```
dougbu: nit: though this is done only once per `TagHelperCodeGenerator`, might be slightly cleaner to set a `readonly int _prefixLength` field in the constructor

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -141,11 +141,15 @@ private bool TryRewriteTagHelper(Block tagBlock, RewritingContext context)
             {
                 // We're in an end tag helper block.
 
-                var tagNameScope = _tagStack.Count > 0 ? _tagStack.Peek().TagName : string.Empty;
+                var parentTagHelper = _tagStack.Count > 0 ? _tagStack.Peek() : null;
```
dougbu: don't need to change this file.  back to using just the `TagName` value (or `null`).

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -43,10 +44,18 @@ public IEnumerable<TagHelperDescriptor> GetTagHelpers(string tagName)
         {
             HashSet<TagHelperDescriptor> descriptors;
 
-            // Ensure there's an ISet to use.
+            if (!string.IsNullOrEmpty(_tagHelperPrefix) &&
+                (!tagName.StartsWith(_tagHelperPrefix, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(tagName, _tagHelperPrefix, StringComparison.OrdinalIgnoreCase)))
```
dougbu: nit: easier to read and avoids repeated comparisons of the same characters as
``` c#
 (!string.IsNullOrEmpty(_tagHelperPrefix) &&
  (tagName.Length <= _tagHelperPrefix.Length ||
   !tagName.StartsWith(_tagHelperPrefix, StringComparison.OrdinalIgnoreCase))
```

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -59,50 +30,64 @@ public static TheoryData TagHelperDescriptorFlowTestData
...
-                        PAndInputTagHelperDescriptors,
+                        DefaultPAndInputTagHelperDescriptors,
                         Enumerable.Empty<TagHelperDescriptor>(),
                         false
                     },
                     {
+                        "BasicTagHelpers.Prefixed",
+                        "BasicTagHelpers.Prefixed",
+                        BuildPAndInputTagHelperDescriptors(prefix: "THS"),
+                        BuildPAndInputTagHelperDescriptors(prefix: "THS"),
```
dougbu: nit: create a `static` `PrefixedPAndInputTagHelperDescriptors` property right next to `DefaultPAndInputTagHelperDescriptors`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,282 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var blockFactory = new BlockFactory(factory);
+                var availableDescriptors1 = new TagHelperDescriptor[]
```
dougbu: pls give these more specific names.  bit easy to forget what "1" means by the end of this data set.  suggest `thPrefixedDescriptors`, `prefixPrefixedDescriptors`, `mythPrefixedDescriptors`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,282 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+                                name: "bound",
+                                propertyName: "Bound",
+                                typeName: typeof(bool).FullName),
+                        })
+                };
+                var availableDescriptors3 = new TagHelperDescriptor[]
+                {
+                    new TagHelperDescriptor(
+                        prefix: "myth",
+                        tagName: "myth",
```
dougbu: I wasn't specific enough in my `<myth/>` suggestion.  meant this should be a catch-all descriptor.  covers the case where Razor source uses a tag name that exactly matches the prefix of a catch-all descriptor.

dougbu: (as-is, this descriptor would only match `<mythmyth ... />`)

dougbu:
:shipit: after reverting `TagHelperParseTreeRewriter` and making non-nit test changes; send out another iteration if I'm missing something about `TagHelperParseTreeRewriter`

Issue #279, opened 2015-01-27
TagHelperOutput doesn't allow multiple non-bound TagHelper attributes with identical names.


NTaylorMullen:
This item adds the infrastructure needed to complete https://github.com/aspnet/Razor/issues/220 and https://github.com/aspnet/Razor/issues/96.

Ultimately we should be able to have tags such as:
`<input class="btn" class="btn-success" />` which generate the correct output. Today this type of tag which happens to be a `TagHelper` would generate:
`<input class="btn-success" />` because last attribute in wins with no way for the `TagHelper` author to chose the other value.

To properly complete this item we need to create an attribute abstraction that has many types, one of which is a one key => many values (in addition to the default 1=>1 mapping). Other types that could be added in the future are key only, C# etc.

NTaylorMullen:
Assigned this to RC due to #220 and #96 being assigned to the milestone (this is a pre-requisite).

dougbu:
@NTaylorMullen, could you clarify why this issue is a prerequisite to #220 (single attributes w/ no value e.g. `checked`) or #96 (single C# attributes e.g. `@checked`)?

NTaylorMullen:
The correct way to fix #96/#220 is to enable different types of attribute values in the `TagHelperOutput`. Any other fix would be more or less a hack/half-fix. By having a single attribute abstraction that can handle 1->many, no value, 1->1, C# attributes users then only have a single abstraction they need to work with without risk of ambiguity between things like:

`<input checked="" />` and `<input checked />`

NTaylorMullen:
7c604d2b1141f8dc3dc78e14aaacb8f489c3d1e6
dcee0d2ff718dace8087f07f15fa24c9db579131
https://github.com/aspnet/Mvc/commit/f152ea700419b8244677e10adf02aec2a9dc7f9e

PR #322, opened 2015-03-13
[DESIGN] Add TagHelperOutput.Attributes replacement.

NTaylorMullen:
**NOTE:** This is very designy. Just implemented the bare minimum for #279 to get some feedback on how the attribute data structure will work.

- Added a TagHelperAttributes object that's used to hold 1=>many attributes.
- Added a TagHelperAttribute object which can be used to represent attributes that may be minimized.

#279

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
@@ -1,0 +1,50 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperAttribute<TValue> where TValue : class
+    {
+        private string _key;
+
+        public TagHelperAttribute()
+        {
+        }
+
+        public TagHelperAttribute([NotNull] string key)
+            : this(key, value: null)
+        {
+        }
+
+        public TagHelperAttribute([NotNull] string key, TValue value)
+        {
+            Key = key;
+            Value = value;
+        }
+
+        public string Key
+        {
+            get
+            {
+                return _key;
+            }
+            set
+            {
+                // Keys aren't allowed to be null.
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(Key));
+                }
+
+                _key = value;
+            }
+        }
+
+        public TValue Value { get; set; }
+
+        public bool Minimized { get; set; }
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
@@ -6,7 +6,7 @@ using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
-    public class TagHelperAttribute<TValue> where TValue : class
+    public class TagHelperAttribute<TValue>
     {
         private string _key;
 
@@ -15,8 +15,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         public TagHelperAttribute([NotNull] string key)
-            : this(key, value: null)
         {
+            Key = key;
         }
 
         public TagHelperAttribute([NotNull] string key, TValue value)
@@ -36,7 +36,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 // Keys aren't allowed to be null.
                 if (value == null)
                 {
-                    throw new ArgumentNullException(nameof(Key));
+                    throw new ArgumentNullException(nameof(value));
                 }
 
                 _key = value;
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
@@ -10,7 +10,6 @@ using Microsoft.Framework.Internal;
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
     public class TagHelperAttributes<TAttributeValue> : IList<TagHelperAttribute<TAttributeValue>>
-        where TAttributeValue : class
     {
         private readonly IList<TagHelperAttribute<TAttributeValue>> _attributes;
 
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
@@ -1,0 +1,150 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperAttributes<TAttributeValue> : IList<TagHelperAttribute<TAttributeValue>>
+        where TAttributeValue : class
+    {
+        private readonly IList<TagHelperAttribute<TAttributeValue>> _attributes;
+
+        public TagHelperAttributes()
+        {
+            _attributes = new List<TagHelperAttribute<TAttributeValue>>();
+        }
+
+        public int Count => _attributes.Count;
+
+        public bool IsReadOnly => false;
+
+        public IEnumerable<string> Keys => _attributes.Select(attribute => attribute.Key);
+
+        public IEnumerable<TAttributeValue> Values => _attributes.Select(attribute => attribute.Value);
+
+        public TagHelperAttribute<TAttributeValue> this[[NotNull] string key] =>
+            _attributes.LastOrDefault(attribute => MatchesAttribute(key, attribute));
+
+        public TagHelperAttribute<TAttributeValue> this[int index]
+        {
+            get
+            {
+                return _attributes[index];
+            }
+            set
+            {
+                _attributes[index] = value;
+            }
+        }
+
+        public bool TryGetAttribute([NotNull] string key, out TagHelperAttribute<TAttributeValue> attribute)
+        {
+            attribute = this[key];
+
+            return attribute != null;
+        }
+
+        public bool TryGetAttributes([NotNull] string key, out IEnumerable<TagHelperAttribute<TAttributeValue>> attributes)
+        {
+            attributes = _attributes.Where(attribute => MatchesAttribute(key, attribute));
+
+            return attributes.Any();
+        }
+
+        public void Add([NotNull] string key)
+        {
+            _attributes.Add(new TagHelperAttribute<TAttributeValue>(key)
+            {
+                Minimized = true
+            });
+        }
+
+        public void Add([NotNull] string key, TAttributeValue value)
+        {
+            _attributes.Add(new TagHelperAttribute<TAttributeValue>(key, value));
+        }
+
+        public void Add([NotNull] TagHelperAttribute<TAttributeValue> attribute)
+        {
+            _attributes.Add(attribute);
+        }
+
+        public void Insert(int index, TagHelperAttribute<TAttributeValue> item)
+        {
+            _attributes.Insert(index, item);
+        }
+
+        public void CopyTo([NotNull] TagHelperAttribute<TAttributeValue>[] attributes, int index)
+        {
+            _attributes.CopyTo(attributes, index);
+        }
+
+        public bool Remove([NotNull] string key)
+        {
+            IEnumerable<TagHelperAttribute<TAttributeValue>> attributes;
+            if (TryGetAttributes(key, out attributes))
+            {
+                foreach (var attribute in attributes)
+                {
+                    Remove(attribute);
+                }
+
+                return true;
+            }
+
+            return false;
+        }
+
+        public bool Remove([NotNull] TagHelperAttribute<TAttributeValue> item)
+        {
+            // REVIEW: Do we want this to be a comparer remove? If so it'd remove more than one.
+            return _attributes.Remove(item);
+        }
+
+        public void RemoveAt(int index)
+        {
+            _attributes.RemoveAt(index);
+        }
+
+        public void Clear()
+        {
+            _attributes.Clear();
+        }
+
+        public bool Contains([NotNull] string key)
+        {
+            return _attributes.Any(attribute => MatchesAttribute(key, attribute));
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+            get
+            {
+                return _key;
+            }
+            set
+            {
+                // Keys aren't allowed to be null.
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(Key));
```
rynowak: `value` is the actual parameter name. #DesignPR

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,150 @@
...
+            _attributes = new List<TagHelperAttribute<TAttributeValue>>();
+        }
+
+        public int Count => _attributes.Count;
+
+        public bool IsReadOnly => false;
+
+        public IEnumerable<string> Keys => _attributes.Select(attribute => attribute.Key);
+
+        public IEnumerable<TAttributeValue> Values => _attributes.Select(attribute => attribute.Value);
```
rynowak: Might want to look into the contract for IDictionary.Keys/Values. The fact that these are just linq queries and not realized collections can cause problems related to modification while iterating.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        public TagHelperAttribute<TAttributeValue> this[[NotNull] string key] =>
+            _attributes.LastOrDefault(attribute => MatchesAttribute(key, attribute));
```
rynowak: this seems a little over the top with the new syntax

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperAttribute<TValue> where TValue : class
```
rynowak: ` where TValue : class` why?

rynowak: Will we ever use this for anything other than object and string? IMO the generic is valuable if it's user-extensible, and reduces clarity if it's not.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        private TagHelperAttributeComparer()
+        {
+        }
+
+        public bool Equals(TagHelperAttribute<TAttributeValue> attributeX, TagHelperAttribute<TAttributeValue> attributeY)
+        {
+            return 
+                string.Equals(attributeX.Key, attributeY.Key, StringComparison.OrdinalIgnoreCase) &&
+                Equals(attributeX.Value, attributeY.Value);
```
rynowak: `object.Equals` right?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public int GetHashCode(TagHelperAttribute<TAttributeValue> attribute)
+        {
+            return HashCodeCombiner
+                .Start()
+                .Add(attribute.Key, StringComparer.OrdinalIgnoreCase)
+                .Add(attribute.Value)
+                .CombinedHash;
```
rynowak: Will this ever be called? This looks scary from a perf POV.

rynowak: nvm test code don't keeer.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+                    throw new ArgumentNullException(nameof(Key));
+                }
+
+                _key = value;
+            }
+        }
+
+        public TValue Value { get; set; }
+
+        public bool Minimized { get; set; }
```
rynowak: Are you worried about the impact of sharing a mutable 'holder' object across collections? This makes implementing something like a shallow copy non-trivial.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
NTaylorMullen: I wanted to stay away from dictionaries for `TagHelperAttributes` to not conflate the two, ex: x many keys same name => many values for `TagHelperAttributes` which can't be done in a dictionary.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
NTaylorMullen: Whatchou mean?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: Ahh :+1: was thinking of combining `Minimized` into `Value` by being `null` which is where `: class` came into play but realized that's not possible since `TagHelperAttribute<object>` is used for `TagHelperContext.AllAttributes`.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeComparer.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: We want this to be mutable in the case that a user retrieves an attribute and decides to change its rendering to be minimized. Would be kind of lame having to remove then re-add. /cc @DamianEdwards in-case he feels like it'd be ok from a user perspective

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
rynowak: I'm not suggesting that we nessesarily need `IDictionary<K,V>` here, I'm suggesting that you take a look at how `.Keys` and `.Values` are **implemented**.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
rynowak: this would be much more understandable with the regular old property definition syntax. Yeah I know, using the new thing awards many PranavPoints®

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
rynowak: Yep. and the side-effect of mutability is the need to make defensive copies :laughing:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                return _key;
+            }
+            set
+            {
+                // Keys aren't allowed to be null.
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
```
dougbu: fyi (since this is still [Design]) [guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#argument-null-checking) say to use `[param: NotNull]` above the setter instead.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: also, should tag helper authors even be able to override the user's wishes in this way?  for one thing, it's an odd feature to support (minification is something we've recommended doing as a post-processing step).  for another, these instances can be replaced and therefore don't need to be updated in-place.  all in all, make attribute updates aka replacement possible but not easy and especially not easy to do accidentally.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperAttributes<TAttributeValue> : IList<TagHelperAttribute<TAttributeValue>>
```
dougbu: probably a good idea to instead inherit from `System.Collections.ObjectModel.Collection<TagHelperAttribute<TAttributeValue>>`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        private readonly IList<TagHelperAttribute<TAttributeValue>> _attributes;
```
dougbu: use `List<TagHelperAttribute<TAttributeValue>>` within the class for efficiency

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
dougbu: in general this class looks inconsistent with everything else in Razor, less readable, and like @NTaylorMullen had fun with the new syntax.  was it an obfuscation exercise -- how much can you cram on a line or two?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
dougbu: we should separate responsibilities here.  rather than one class, should have two:
1) implement `IList<TagHelperAttribute<TAttributeValue>>` to maintain all attributes form the Razor source or just use a standard `List<T>`
2) implement `IDictionary<string, IList<TagHelperAttribute<TAttributeValue>>>` to support attribute lookup or just use a standard `Dictionary<TKey, TValue>`. (haven't looked far enough ahead to see these properties and methods used but they likely are)
 - this could be a projection over the `List<T>` above to avoid the two getting out of sync

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+            _attributes.Add(new TagHelperAttribute<TAttributeValue>(key)
+            {
+                Minimized = true
+            });
+        }
+
+        public void Add([NotNull] string key, TAttributeValue value)
+        {
+            _attributes.Add(new TagHelperAttribute<TAttributeValue>(key, value));
+        }
```
dougbu: these `Add` helper methods hide too much.  require explicit `TagHelperAttribute` creation in the caller.  if these are just for testing, make them `internal`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public bool Contains([NotNull] string key)
+        {
+            return _attributes.Any(attribute => MatchesAttribute(key, attribute));
+        }
+
+        public bool Contains([NotNull] TagHelperAttribute<TAttributeValue> item)
+        {
+            // REVIEW: Do we want this to be a comparer check?
```
dougbu: underlying list would use correct default if `TagHelperAttribute<TAttributeValue>` implemented `IEquatable<T>`.  if callers want to use a different comparer, they can use the Linq extension method.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+
+                return true;
+            }
+
+            return false;
+        }
+
+        public bool Remove([NotNull] TagHelperAttribute<TAttributeValue> item)
+        {
+            // REVIEW: Do we want this to be a comparer remove? If so it'd remove more than one.
```
dougbu: see `Contains` [answer](https://github.com/aspnet/Razor/pull/322/files#r26773313)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttributes.cs
```diff
@@ -0,0 +1,149 @@
...
+            return _attributes.Any(attribute => MatchesAttribute(key, attribute));
+        }
+
+        public bool Contains([NotNull] TagHelperAttribute<TAttributeValue> item)
+        {
+            // REVIEW: Do we want this to be a comparer check?
+            return _attributes.Contains(item);
+        }
+
+        public int IndexOf([NotNull] TagHelperAttribute<TAttributeValue> item)
```
dougbu: this method would also pick up the correct comparer if `TagHelperAttribute<TAttributeValue>` implemented `IEquatable<T>`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -153,14 +154,14 @@ public void HtmlAttributes_IgnoresCase(string originalName, string updatedName)
         {
             // Arrange
             var executionContext = new TagHelperExecutionContext("p", selfClosing: false);
-            executionContext.HTMLAttributes[originalName] = "hello";
+            executionContext.HTMLAttributes[originalName].Value = "hello";
```
dougbu: hmm, `public static implicit operator TagHelperAttribute(string value)` may be worthwhile.  would work well without setters on `TagHelperAttribute` properties as well.

NTaylorMullen:
Closing this. Will re-open a new Design PR after I've rewritten a bit of this to account for recent changes (super stale PR here).

Issue #169, opened 2014-09-30
Await TagHelper generated runtime objects.


NTaylorMullen:
For a temporary stop gap the `TagHelperRunner.RunAsync` method is `.Result`ed until #85 and https://github.com/aspnet/Mvc/issues/845 are complete.  Once these two are complete we should make the it be `await`ed. 

This also applies to the new `ExecuteChildContentAsync`/`GetChildContentAsync` methods after the content mode redesign.

PR #325, opened 2015-03-17
await TagHelper generated objects

ajaybhargavb:
Issue #169 
@NTaylorMullen


NTaylorMullen:
:shipit: once you've verified that everything works as intended with each directive during runtime.

dougbu:
:shipit: once you're done @NTaylorMullen's verification _and_ got sign-off on a reaction PR in MVC.  we have at least a few unit tests that confirm the generated source.

Issue #326, opened 2015-03-20
Multiple [TargetElement] attributes on a single TagHelper should de-dup if multiple match.


NTaylorMullen:
If I have:

```C#
[TargetElement(Attributes = "a")]
[TargetElement(Attributes = "b")]
public class FooTagHelper : TagHelper {}
```

And I write `<foo a="" b="" />` the `FooTagHelper` class should not be run twice.

NTaylorMullen:
842549bba9f0deefc2c6000da6c4b5c594aa2c9d

PR #327, opened 2015-03-20
De-dupe TagHelperDescriptors based on Type for rendering.

NTaylorMullen:
- This can occur if you have multiple [TargetElement] attributes that overlap. Ultimately the descriptor is the same because its the same type, just the required attributes differ.
- Added tests to validate.

#326

### DIFF
--- a/test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
@@ -19,6 +19,56 @@ namespace Microsoft.AspNet.Razor.Test.Generator
         private static IEnumerable<TagHelperDescriptor> PrefixedPAndInputTagHelperDescriptors
             => BuildPAndInputTagHelperDescriptors("THS");
 
+        private static IEnumerable<TagHelperDescriptor> DuplicateTargetTagHelperDescriptors
+        {
+            get
+            {
+                var inputTypePropertyInfo = typeof(TestType).GetProperty("Type");
+                var inputCheckedPropertyInfo = typeof(TestType).GetProperty("Checked");
+                return new[]
+                {
+                    new TagHelperDescriptor(
+                        tagName: "*",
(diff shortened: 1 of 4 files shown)


NTaylorMullen:
Getting this in now, feel free to leave comments. I'll do a commit to address any feedback. 

/cc @ajaybhargavb you should be good now :smile:

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -19,6 +19,56 @@ private static IEnumerable<TagHelperDescriptor> DefaultPAndInputTagHelperDescrip
...
+                        },
+                        requiredAttributes: new[] { "type" }),
+                    new TagHelperDescriptor(
+                        tagName: "*",
+                        typeName: "CatchAllTagHelper",
+                        assemblyName: "SomeAssembly",
+                        attributes: new TagHelperAttributeDescriptor[]
+                        {
+                            new TagHelperAttributeDescriptor("type", inputTypePropertyInfo),
+                            new TagHelperAttributeDescriptor("checked", inputCheckedPropertyInfo)
```
dougbu: tiny nit: this looks like the tag helper author just wrote an extra, useless `[TargetElement]`.  a more realistic scenario would involve something like `[TargetElement(Attributes = "type")]` and `[TargetElement(Attributes = "checked")]` on the same helper.  the author didn't do anything wrong in this case but `<whatever type="type" checked="checked" />` would lead to duplication.

so, minor leaning toward removing lines 45 and 64

dougbu:
:shipit:

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
NTaylorMullen: Actualllly there's a flaw in the test here. If the `TagHelper` type is the same type the attribute descriptors will all be the same. Will update and get it into dev.

PR #328, opened 2015-03-21
Fix build break in razor tests

rynowak:
This message changed because we got a fix in CoreCLR for a bug in the
host that was impacting Type.GetType and breaking data contract
serializer.

### DIFF
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperTypeResolverTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperTypeResolverTest.cs
@@ -46,8 +46,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 #if DNX451
                 "abcd' or one of its dependencies. The system cannot find the file specified.";
 #else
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperTypeResolverTest.cs
```diff
@@ -46,8 +46,8 @@ public void TypeResolver_RecordsErrorWhenCannotResolveAssembly()
 #if DNX451
                 "abcd' or one of its dependencies. The system cannot find the file specified.";
 #else
-                "abcd, Culture=neutral, PublicKeyToken=null' or one of its dependencies. Could not find or load a " +
-                "specific file. (Exception from HRESULT: 0x80131621)";
+                "abcd, Culture=neutral, PublicKeyToken=null' or one of its dependencies. " +
+                "The system cannot find the file specified.";
 #endif
```
dougbu: nit: would be nice (now or the next time we change this file) to reduce the platform-specific part to what's actually different e.g.
``` c#
#if !DNX451
    `", Culture=neutral, PublicKeyToken=null" +`
#endif
```

dougbu:
:shipit:

Issue #330, opened 2015-03-23
Make RazorError deserializable


NTaylorMullen:
Right now RazorError is not deserializable, it needs to be in order to work during design time.

NTaylorMullen:
05eb00b25828f6f7c3f04f69081aed2d56fe0c01

PR #331, opened 2015-03-23
Update RazorError to be deserialized.

NTaylorMullen:
- Added tests to ensure RazorError can be serialized/deserialized.

#330


NTaylorMullen:
Signed off by @pranavkm in person.

PR #351, opened 2015-04-08
Update resource names

victorhurdugaci:
React to DNX resource name changes.

@dougbu or @NTaylorMullen or both can you please review?

### DIFF
--- a/test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
@@ -11,13 +11,13 @@ namespace Microsoft.AspNet.Razor.Test
     {
         public const string ResourceNameFormat = "{0}.TestFiles.{1}";
 
-        public string ResourceName { get; set; }
+        public string ResourceName { get; }
         public Assembly Assembly { get; set; }
 
         public TestFile(string resName, Assembly asm)
         {
-            ResourceName = resName;
             Assembly = asm;
+            ResourceName = Assembly.GetName().Name + "." + resName.Replace('/', '.');
         }
 
         public static TestFile Create(string localResourceName)


test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
```diff
@@ -27,7 +27,9 @@ public static TestFile Create(string localResourceName)
 
         public Stream OpenRead()
         {
-            var strm = Assembly.GetManifestResourceStream(ResourceName);
+            string resourceName = Assembly.GetName().Name + "." + ResourceName.Replace('/', '.');
```
NTaylorMullen: var

test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
```diff
@@ -27,7 +27,9 @@ public static TestFile Create(string localResourceName)
...
+
+            var strm = Assembly.GetManifestResourceStream(resourceName);
             if (strm == null)
             {
                 Assert.True(false, string.Format("Manifest resource: {0} not found", ResourceName));
```
NTaylorMullen: `ResourceName` is no longer valid here because it's not pointing to the same location as requested.

test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
NTaylorMullen: Actually looks like we don't set `ResourceName` anywhere. Make it `public string ResourceName { get; }` and do this calculation in the constructor:

```C#
public TestFile(string resName, Assembly asm)
{
    ResourceName = Assembly.GetName().Name + "." + resName.Replace('/', '.');
    Assembly = asm;
}

public string ResourceName { get; }

public Assembly Assembly { get; }
```

test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
NTaylorMullen: This wont be needed once you do https://github.com/aspnet/Razor/pull/351/files#r28018029

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
```diff
@@ -27,6 +27,8 @@ public static TestFile Create(string localResourceName)
 
         public Stream OpenRead()
         {
+
```
NTaylorMullen: Undo these line changes.

test/Microsoft.AspNet.Razor.Test/Utils/TestFile.cs
```diff
@@ -11,13 +11,13 @@ public class TestFile
...
 
-        public string ResourceName { get; set; }
+        public string ResourceName { get; }
         public Assembly Assembly { get; set; }
 
         public TestFile(string resName, Assembly asm)
         {
-            ResourceName = resName;
             Assembly = asm;
+            ResourceName = Assembly.GetName().Name + "." + resName.Replace('/', '.'); ;
```
NTaylorMullen: `; ;`

NTaylorMullen:
:shipit: once you've addressed comments.

Issue #341, opened 2015-04-01
Add support to allow Tag Helpers to generate content before and after source element


DamianEdwards:
Tag Helpers can specify content to appear before and after the contents of the HTML element they're operating on. Sometimes however, a Tag Helper will want to generate content that appears before or after the element itself, e.g. the `<link>` and `<script>` Tag Helpers in MVC. Currently, that requires them to manually rebuild the original element by suppressing the default output and constructing the original HTML using the information from `TagHelperContext` and `TagHelperOutput`. This should be simpler for Tag Helpers to do.

We should add new properties to `TagHelperOutput` that enable this, e.g.:

``` C#
public TagHelperContent PreElement { get; }
public TagHelperContent PostElement { get; }
```

### Example

__Source file__:
```HTML
<my-tag-helper>
    Content in source
</my-tag-helper>
```
__Tag Helper file__:
``` C#
public void Process(TagHelperContext context, TagHelperOutput output)
{
    var nl = Environment.NewLine;
    var br = "<br />" + nl;
    output.PreElement.Append("This will appear before source element" + br);
    output.PreContent.Append(nl + "This will appear before source content" + br);
    output.PostContent.Append(br + "This will appear after source content" + nl);
    output.PostElement.Append(br + "This will appear after source element");
}
```
__Output__:
```HTML
This will appear before source element<br />
<my-tag-helper>
This will appear before source content<br />
    Content in source<br />
This will appear after source content
</my-tag-helper><br />
This will appear after source element
```

NTaylorMullen:
c9c8e802a7ead380fcc60c6c7baa5924e8713db8

https://github.com/aspnet/Mvc/commit/2c4c35e126d9287053cdaa772a5f9ca307650f7c

PR #355, opened 2015-04-13
Add PreElement and PostElement to TagHelperOutput.

NTaylorMullen:
- These two new properties will enable TagHelper authors to render content before and after the TagHelper's HTML element.
- Added tests to correspond with existing test coverage.
- Modified existing tests to double check for TagHelperOutput.Pre/PostElement.

#341

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -12,10 +12,6 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperOutput
     {
-        private readonly DefaultTagHelperContent _preContent;
-        private readonly DefaultTagHelperContent _content;
-        private readonly DefaultTagHelperContent _postContent;
-
         // Internal for testing
         internal TagHelperOutput(string tagName)
             : this(tagName, new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase))
@@ -33,9 +29,6 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             TagName = tagName;
             Attributes = new Dictionary<string, object>(attributes, StringComparer.OrdinalIgnoreCase);
-            _preContent = new DefaultTagHelperContent();
-            _content = new DefaultTagHelperContent();
-            _postContent = new DefaultTagHelperContent();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
/cc @dougbu @pranavkm 

Working on an MVC reaction PR which will be more interesting/more test changes :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -47,6 +51,18 @@ internal TagHelperOutput(string tagName)
         public string TagName { get; set; }
 
         /// <summary>
+        /// The HTML element's pre element content.
```
pranavkm: Consider rephrasing this? For instance:

> Content that precedes the HTML element.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -33,9 +35,11 @@ internal TagHelperOutput(string tagName)
         {
             TagName = tagName;
             Attributes = new Dictionary<string, object>(attributes, StringComparer.OrdinalIgnoreCase);
+            _preElement = new DefaultTagHelperContent();
```
Eilon: Any reason these aren't initialized on the fields directly?

NTaylorMullen: Nope, it's just the pattern that we've used in Mvc/Razor since I've been on the team :smile:

Eilon:
Just minor stuff but otherwise :shipit: .

BTW this would be an interesting type of class to look at from a perf perspective because there are a lot of allocations for what might be uncommon scenarios. E.g. always creating all the empty pre/post/whatever contents even if they're never used. *If* it's a perf problem for memory allocations (causing GC mayhem) we could create all the default content objects on demand.

NTaylorMullen:
@Eilon We actually had similar concerns when we moved over to the "new" `TagHelperOutput` buffer model instead of strings during its CR. Filed https://github.com/aspnet/Razor/issues/358

Eilon:
Thanks!

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
pranavkm: We could use property initializers for these now - `public TagHelperContent PreElement { get; } = new DefaultTagHelperContent();`

pranavkm:
:shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: :+1:

Issue #342, opened 2015-04-01
TagHelper attributes with data- have surprising behavior


rynowak:
Consider the following:
```
@{ var pwnt = "Hello, world!"; }
<p data-th-color="color: red; @pwnt">
            This should be red.
</p>


[TargetElement("p", Attributes = "data-th-color")]
    public class ColorTagHelper : TagHelper
    {
        [HtmlAttributeName("data-th-color")]
        public string Color { get; set; }

        public override void Process(TagHelperContext context, TagHelperOutput output)
        {
            output.Attributes.Add("style", Color);
        }
    }
```

Inside the taghelper, the value of `Color` is `"Hello, world!"` instead of the expected `"color: red; Hello, world!"`. According to @NTaylorMullen this is because razor has special handling of data-* attributes.

Changing the attribute to something that doesn't begin with data- will give the expected behavior.

dougbu:
whatever the special case is, the **Principal of Least Surprise** says Razor is broken here.  either binding an attribute named `data-*` is an error or it should do the Right Thing:tm: i.e. behave like any other bound attribute.

NTaylorMullen:
Agree.

Curious: @anurse why did we special case `data-` in Razor?

anurse:
We special-cased it for conditional attributes. Conditional attributes are designed to make it easier to write HTML that User Agents can properly interpret. We do this by making `checked="@foo"` render as `checked="checked"` if `foo` is `true`, or removing it entirely if foo is `false` or `null`.

However, `data-` attributes are explicitly called out as being reserved solely for the purpose of the end-application-developer:

> These attributes are not intended for use by software that is not known to the administrators of the site that uses the attributes. For generic extensions that are to be used by multiple independent tools, either this specification should be extended to provide the feature explicitly, or a technology like microdata should be used (with a standardised vocabulary). (http://www.w3.org/html/wg/drafts/html/master/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes)

Thus, they are explicitly not implemented by User Agents and we didn't want conditional attributes getting in the way of existing application code (and turning something like `data-enabled="@foo"` suddenly vanishing when the JavaScript that reads it expects a literal string `false`).

For a similar reason, I think we should take @dougbu 's first suggestion here and disallow binding Tag Helpers to `data-` attributes.

NTaylorMullen:
@DamianEdwards do you feel like we should be able to bind `data-` attributes? Or should it result in an error?

[...]

PR #356, opened 2015-04-13
Create error when TagHelper binds to HTML attribute starting w/ data-.

NTaylorMullen:
- Added TagHelperDescriptorFactory tests to validate TagHelperAttributeDescriptor creation.

#342

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
+++ b/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
@@ -153,4 +153,7 @@
   <data name="TagHelperDescriptorFactory_Tag" xml:space="preserve">
     <value>Tag</value>
   </data>
+  <data name="TagHelperDescriptorFactory_InvalidBoundAttributeName" xml:space="preserve">
+    <value>Invalid tag helper bound property '{0}.{1}'. Tag helpers cannot bind to HTML attributes beginning with '{2}'.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -16,6 +16,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public static class TagHelperDescriptorFactory
     {
+        private const string DataDashPrefix = "data-";
         private const string TagHelperNameEnding = "TagHelper";
         private const string HtmlCaseRegexReplacement = "-$1$2";
 
@@ -44,7 +45,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             [NotNull] ErrorSink errorSink)
         {
             var typeInfo = type.GetTypeInfo();
-            var attributeDescriptors = GetAttributeDescriptors(type);
+            var attributeDescriptors = GetAttributeDescriptors(type, errorSink);
             var targetElementAttributes = GetValidTargetElementAttributes(typeInfo, errorSink);
             var tagHelperDescriptors =
                 BuildTagHelperDescriptors(
@@ -201,12 +202,45 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             return validName;
         }
 
-        private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(Type type)
+        private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(
+            Type type, 
+            ErrorSink errorSink)
         {
-            var properties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
-            var attributeDescriptors = properties.Select(ToAttributeDescriptor);
+            var accessibleProperties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
+            var attributeDescriptors = new List<TagHelperAttributeDescriptor>();
+
+            foreach (var property in accessibleProperties)
+            {
+                var descriptor = ToAttributeDescriptor(property);
+                if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
+                {
+                    attributeDescriptors.Add(descriptor);
+                }
+            }
 
             return attributeDescriptors;
+        }
+
+        private static bool ValidateTagHelperAttributeDescriptor(
+            TagHelperAttributeDescriptor attributeDescriptor,
+            Type parentType,
+            ErrorSink errorSink)
+        {
+            // data-* attributes are explicitly not implemented by user agents and are not intended for use on 
+            // the server; therefore it's invalid for TagHelpers to bind to them.
+            if (attributeDescriptor.Name.StartsWith(DataDashPrefix, StringComparison.OrdinalIgnoreCase))
+            {
+                errorSink.OnError(
+                    SourceLocation.Zero, 
+                    Resources.FormatTagHelperDescriptorFactory_InvalidBoundAttributeName(
+                        attributeDescriptor.PropertyName,
+                        parentType.FullName,
+                        DataDashPrefix));
+
+                return false;
+            }
+
+            return true;
         }
 
         private static TagHelperAttributeDescriptor ToAttributeDescriptor(PropertyInfo property)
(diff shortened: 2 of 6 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeDescriptorComparer.cs
```diff
@@ -0,0 +1,39 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Razor.TagHelpers;
+using Microsoft.Internal.Web.Utils;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class CaseSensitiveTagHelperAttributeDescriptorComparer : IEqualityComparer<TagHelperAttributeDescriptor>
```
NTaylorMullen: Moved this as-is with the exception of going from `CaseSensitiveAttributeDescriptorComparer` => `CaseSensitiveTagHelperAttributeDescriptorComparer` from the main `TagHelperDescriptor` comparer class.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -201,14 +202,42 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
             return validName;
         }
 
-        private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(Type type)
+        private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(
+            Type type, 
+            ErrorSink errorSink)
         {
-            var properties = type.GetRuntimeProperties().Where(IsValidProperty);
+            var properties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
```
dougbu: don't bury this name change in this PR.  if you dislike the name, fix it separately,  (suggest a FYI PR and immediate push)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -201,14 +202,42 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
             var attributeDescriptors = properties.Select(ToAttributeDescriptor);
 
+            attributeDescriptors = attributeDescriptors.Where(
+                descriptor => ValidTagHelperAttributeDescriptor(descriptor, type, errorSink));
+
             return attributeDescriptors;
         }
 
+        private static bool ValidTagHelperAttributeDescriptor(
```
dougbu: why isn't this part of the existing `IsValid` method?  nesting the `where` clauses instead of joining the two conditions just to avoid calculating the name twice?  the extra complication (and state machine) seems like the wrong trade-off.

src/Microsoft.AspNet.Razor.Runtime/Resources.resx
```diff
@@ -153,4 +153,7 @@
   <data name="TagHelperDescriptorFactory_Tag" xml:space="preserve">
     <value>Tag</value>
   </data>
+  <data name="TagHelperDescriptorFactory_InvalidBoundAttributeName" xml:space="preserve">
+    <value>Invalid tag helper bound property '{0}' binding to HTML attribute '{1}' for type '{2}'. Tag helpers cannot bind to attributes beginning with '{3}'.</value>
```
dougbu: first sentence doesn't read good.  suggest leaving out `binding to HTML attribute '{1}'` and simplifying to `Invalid tag helper bound property '{0}.{1}'.`  `{0}` is the type and `{1}` is the property.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: The existing `IsValidProperty` method was misleading to begin with which is why i changed it to `IsAccessibleProperty`. Reason why I didn't group this in with it was because I'd have to recalculate the resulting attribute name to find out if it is valid or not. An alternative if you don't like the extra `where` is to do:

```C#
var attributeDescriptors = new List<TagHelperAttributeDescriptor>();
foreach (var property in properties)
{
    var descriptor = ToAttributeDescriptor(property);
    if (ValidTagHelperAttributeDescriptor(descriptor, type, errorSink))
    {
        attributeDescriptors.Add(descriptor)
    }
}
```

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -861,6 +861,111 @@ public void GetCommaSeparatedValues_OutputsEmptyArrayForNullValue()
             Assert.Empty(result);
         }
 
+        public static TheoryData InvalidTagHelperAttributeDescriptorData
+        {
+            get
+            {
+                var errorFormat = "Invalid tag helper bound property '{0}' binding to HTML attribute " +
+                    "'{1}' for type '{2}'. Tag helpers cannot bind to attributes beginning with 'data-'.";
```
dougbu: great case where including the full error strings in the test data would improve understanding

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: still nesting state machines.  how about
``` c#
foreach (var property in type.GetRuntimeProperties())
{
  if (!IsValidProperty())
  {
    continue;
  }

  var descriptor ...
}
```

I'm also not excited about creating the `TagHelperAttributeDescriptor` only to throw it away if the name turns out to start with `data-`.  probably better to calculate the name, validate it, and pass it to `ToAttributeDescriptor()` (along with the `PropertyInfo`).

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: Gonna leave as is.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Nah :-1: passing around the generated name is ugly and harder to understand. Throwing away the `TagHelperAttributeDescriptor` only occurs on pages that can't run anyways so it's not a problem.

NTaylorMullen:
Updated.

dougbu:
:shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -201,14 +202,47 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
+            TagHelperAttributeDescriptor attributeDescriptor,
+            Type parentType,
+            ErrorSink errorSink)
+        {
+            // data-* attributes are explicitly not implemented by user agents and are not intended for use on 
+            // the server; therefore it's invalid for TagHelpers to bind to them.
+            if (attributeDescriptor.Name.StartsWith(DataDashPrefix, StringComparison.OrdinalIgnoreCase))
+            {
+                errorSink.OnError(
+                    SourceLocation.Zero,
```
Eilon: Does this mean the developer won't know where the error was? Or is that flowed some other way?

NTaylorMullen: We notify the developer what class/property caused the error via the error message, it's our primary outlet for informing the dev if they screw up when creating their own `TagHelper`s (incredibly unlikely to ever see this error unless you're creating your own `TagHelper`s). 

We *could* have the error light up the `@addTagHelper` on the page to show which directive is the cause of the error but I worry that it may be semi-misleading if the directive is well formed. This is the format we've been following for all the other `TagHelperDescriptor` resolution mechanics but I don't feel strongly one way or another about tying the errors to the `@addTagHelper` directives.

Eilon: Oh I see, I was thinking this was for the consumer of the tag helper, but you're saying it's for the author. Then I think this is fine.

PR #357, opened 2015-04-14
Source location

pranavkm:
Ties into https://github.com/aspnet/Mvc/issues/2321

### DIFF
--- a/src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
+++ b/src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Internal.Web.Utils
         public HashCodeCombiner Add<TValue>(TValue value, IEqualityComparer<TValue> comparer)
         {
             var hashCode = value != null ? comparer.GetHashCode(value) : 0;
-            return Add(0);
+            return Add(hashCode);
         }
 
         public static HashCodeCombiner Start()
--- a/src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
+++ b/src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
@@ -49,7 +49,8 @@ namespace Microsoft.Internal.Web.Utils
 
         public HashCodeCombiner Add<TValue>(TValue value, IEqualityComparer<TValue> comparer)
         {
-            return Add(comparer.GetHashCode(value));
+            var hashCode = value != null ? comparer.GetHashCode(value) : 0;
+            return Add(0);
         }
 
         public static HashCodeCombiner Start()
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
@@ -190,7 +190,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
         /// <returns>The current instance of <see cref="CSharpCodeWriter"/>.</returns>
         public CSharpCodeWriter WriteLineNumberDirective(SourceLocation location, string file)
         {
-            if (!string.IsNullOrEmpty(location.FilePath))
+            if (location.FilePath != null)
             {
                 file = location.FilePath;
             }
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
@@ -190,18 +190,18 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
         /// <returns>The current instance of <see cref="CSharpCodeWriter"/>.</returns>
         public CSharpCodeWriter WriteLineNumberDirective(SourceLocation location, string file)
         {
-            return WriteLineNumberDirective(location.LineIndex + 1, file);
-        }
+            if (!string.IsNullOrEmpty(location.FilePath))
+            {
+                file = location.FilePath;
+            }
 
-        public CSharpCodeWriter WriteLineNumberDirective(int lineNumber, string file)
-        {
             if (!string.IsNullOrEmpty(LastWrite) &&
                 !LastWrite.EndsWith(NewLine, StringComparison.Ordinal))
             {
                 WriteLine();
             }
 
-            var lineNumberAsString = lineNumber.ToString(CultureInfo.InvariantCulture);
+            var lineNumberAsString = (location.LineIndex + 1).ToString(CultureInfo.InvariantCulture);
             return Write("#line ").Write(lineNumberAsString).Write(" \"").Write(file).WriteLine("\"");
         }
 
--- a/src/Microsoft.AspNet.Razor/RazorResources.resx
+++ b/src/Microsoft.AspNet.Razor/RazorResources.resx
@@ -411,6 +411,6 @@ Instead, wrap the contents of the block in "{{}}":
     <value>Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty or contain only whitespace.</value>
   </data>
   <data name="SourceLocationFilePathDoesNotMatch" xml:space="preserve">
-    <value>Cannot perform operations on '{0}' instances with different file paths.</value>
+    <value>Cannot perform '{1}' operations on '{0}' instances with different file paths.</value>
   </data>
 </root>
--- a/src/Microsoft.AspNet.Razor/RazorResources.resx
+++ b/src/Microsoft.AspNet.Razor/RazorResources.resx
@@ -410,4 +410,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="RewriterError_EmptyTagHelperBoundAttribute" xml:space="preserve">
     <value>Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty or contain only whitespace.</value>
   </data>
+  <data name="SourceLocationFilePathDoesNotMatch" xml:space="preserve">
+    <value>Cannot perform operations on '{0}' instances with different file paths.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Razor/SourceLocation.cs
+++ b/src/Microsoft.AspNet.Razor/SourceLocation.cs
@@ -4,9 +4,13 @@
 using System;
 using System.Globalization;
 using Microsoft.AspNet.Razor.Text;
+using Microsoft.Internal.Web.Utils;
 
 namespace Microsoft.AspNet.Razor
 {
+    /// <summary>
+    /// A location in a Razor file.
+    /// </summary>
 #if NET45
     // No Serializable attribute in CoreCLR (no need for it anymore?)
     [Serializable]
@@ -16,12 +20,37 @@ namespace Microsoft.AspNet.Razor
         public static readonly SourceLocation Undefined = CreateUndefined();
         public static readonly SourceLocation Zero = new SourceLocation(0, 0, 0);
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="SourceLocation"/>.
+        /// </summary>
+        /// <param name="absoluteIndex">The absolute index.</param>
+        /// <param name="lineIndex">The line index.</param>
+        /// <param name="characterIndex">The character index.</param>
         public SourceLocation(int absoluteIndex, int lineIndex, int characterIndex)
+            : this(filePath: null, absoluteIndex: absoluteIndex, lineIndex: lineIndex, characterIndex: characterIndex)
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="SourceLocation"/>.
+        /// </summary>
+        /// <param name="filePath">The file path.</param>
+        /// <param name="absoluteIndex">The absolute index.</param>
+        /// <param name="lineIndex">The line index.</param>
+        /// <param name="characterIndex">The character index.</param>
+        public SourceLocation(string filePath, int absoluteIndex, int lineIndex, int characterIndex)
+        {
+            FilePath = filePath;
             AbsoluteIndex = absoluteIndex;
             LineIndex = lineIndex;
             CharacterIndex = characterIndex;
         }
+
+        /// <summary>
+        /// Path of the file.
+        /// </summary>
+        /// <remarks>When <c>null</c>, the parser assumes it is the file currently being processed.</remarks>
+        public string FilePath { get; set; }
 
         /// <remarks>Set property is only accessible for deserialization purposes.</remarks>
         public int AbsoluteIndex { get; set; }
@@ -35,6 +64,7 @@ namespace Microsoft.AspNet.Razor
         /// <remarks>Set property is only accessible for deserialization purposes.</remarks>
         public int CharacterIndex { get; set; }
 
+        /// <inheritdoc />
         public override string ToString()
         {
             return string.Format(
@@ -45,62 +75,57 @@ namespace Microsoft.AspNet.Razor
                 CharacterIndex);
         }
 
+
+        /// <inheritdoc />
         public override bool Equals(object obj)
         {
             return (obj is SourceLocation) && Equals((SourceLocation)obj);
         }
 
+        /// <inheritdoc />
         public override int GetHashCode()
         {
             // LineIndex and CharacterIndex can be calculated from AbsoluteIndex and the document content.
-            return AbsoluteIndex;
+            return HashCodeCombiner.Start()
+                .Add(FilePath, StringComparer.Ordinal)
+                .Add(AbsoluteIndex)
+                .CombinedHash;
         }
 
+        /// <inheritdoc />
         public bool Equals(SourceLocation other)
         {
-            return 
+            return string.Equals(FilePath, other.FilePath, StringComparison.Ordinal) &&
                 AbsoluteIndex == other.AbsoluteIndex &&
                 LineIndex == other.LineIndex &&
                 CharacterIndex == other.CharacterIndex;
         }
 
+        /// <inheritdoc />
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
         public int CompareTo(SourceLocation other)
         {
+            var filePathOrdinal = string.Compare(FilePath, other.FilePath, StringComparison.Ordinal);
+            if (filePathOrdinal != 0)
+            {
+                return filePathOrdinal;
+            }
+
             return AbsoluteIndex.CompareTo(other.AbsoluteIndex);
         }
 
+        /// <summary>
+        /// Advances the <see cref="SourceLocation"/> by the length of the <paramref name="text" />.
+        /// </summary>
+        /// <param name="left">The <see cref="SourceLocation"/> to advance.</param>
+        /// <param name="text">The <see cref="string"/> to advance <paramref name="left"/> by.</param>
+        /// <returns>The advanced <see cref="SourceLocation"/>.</returns>
         public static SourceLocation Advance(SourceLocation left, string text)
         {
             var tracker = new SourceLocationTracker(left);
             tracker.UpdateLocation(text);
             return tracker.CurrentLocation;
-        }
-
-        public static SourceLocation Add(SourceLocation left, SourceLocation right)
-        {
-            if (right.LineIndex > 0)
-            {
-                // Column index doesn't matter
-                return new SourceLocation(
-                    left.AbsoluteIndex + right.AbsoluteIndex,
-                    left.LineIndex + right.LineIndex,
-                    right.CharacterIndex);
-            }
-            else
-            {
-                return new SourceLocation(
-                    left.AbsoluteIndex + right.AbsoluteIndex,
-                    left.LineIndex + right.LineIndex,
-                    left.CharacterIndex + right.CharacterIndex);
-            }
-        }
-
-        public static SourceLocation Subtract(SourceLocation left, SourceLocation right)
-        {
-            return new SourceLocation(
-                left.AbsoluteIndex - right.AbsoluteIndex,
-                left.LineIndex - right.LineIndex,
-                left.LineIndex != right.LineIndex ? left.CharacterIndex : left.CharacterIndex - right.CharacterIndex);
         }
 
         private static SourceLocation CreateUndefined()
@@ -112,34 +137,108 @@ namespace Microsoft.AspNet.Razor
             return sl;
         }
 
+        /// <summary>
+        /// Adds two <see cref="SourceLocation"/>s.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the sum of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
+        public static SourceLocation operator +(SourceLocation left, SourceLocation right)
+        {
+            if (!string.Equals(left.FilePath, right.FilePath, StringComparison.Ordinal))
+            {
+                throw new ArgumentException(
+                    RazorResources.FormatSourceLocationFilePathDoesNotMatch(nameof(SourceLocation)),
+                    nameof(right));
+            }
+
+            if (right.LineIndex > 0)
+            {
+                // Column index doesn't matter
+                return new SourceLocation(
+                    left.FilePath,
+                    left.AbsoluteIndex + right.AbsoluteIndex,
+                    left.LineIndex + right.LineIndex,
+                    right.CharacterIndex);
+            }
+            else
+            {
+                return new SourceLocation(
+                    left.FilePath,
+                    left.AbsoluteIndex + right.AbsoluteIndex,
+                    left.LineIndex + right.LineIndex,
+                    left.CharacterIndex + right.CharacterIndex);
+            }
+        }
+
+        /// <summary>
+        /// Subtracts two <see cref="SourceLocation"/>s.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the difference of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
+        public static SourceLocation operator -(SourceLocation left, SourceLocation right)
+        {
+            if (!string.Equals(left.FilePath, right.FilePath, StringComparison.Ordinal))
+            {
+                throw new ArgumentException(
+                    RazorResources.FormatSourceLocationFilePathDoesNotMatch(nameof(SourceLocation)),
+                    nameof(right));
+            }
+
+            return new SourceLocation(
+                left.FilePath,
+                left.AbsoluteIndex - right.AbsoluteIndex,
+                left.LineIndex - right.LineIndex,
+                left.LineIndex != right.LineIndex ? left.CharacterIndex : left.CharacterIndex - right.CharacterIndex);
+        }
+
+        /// <summary>
+        /// Determines whether the first operand is lesser than the second operand.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> is lesser than <paramref name="right"/>.</returns>
         public static bool operator <(SourceLocation left, SourceLocation right)
         {
             return left.CompareTo(right) < 0;
         }
 
+        /// <summary>
+        /// Determines whether the first operand is greater than the second operand.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> is greater than <paramref name="right"/>.</returns>
         public static bool operator >(SourceLocation left, SourceLocation right)
         {
             return left.CompareTo(right) > 0;
         }
 
+        /// <summary>
+        /// Determines whether the operands are equal.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> and <paramref name="right"/> are equal.</returns>
         public static bool operator ==(SourceLocation left, SourceLocation right)
         {
             return left.Equals(right);
         }
 
+        /// <summary>
+        /// Determines whether the operands are not equal.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> and <paramref name="right"/> are not equal.</returns>
         public static bool operator !=(SourceLocation left, SourceLocation right)
         {
             return !left.Equals(right);
-        }
-
-        public static SourceLocation operator +(SourceLocation left, SourceLocation right)
-        {
-            return Add(left, right);
-        }
-
-        public static SourceLocation operator -(SourceLocation left, SourceLocation right)
-        {
-            return Subtract(left, right);
         }
     }
 }
--- a/src/Microsoft.AspNet.Razor/SourceLocation.cs
+++ b/src/Microsoft.AspNet.Razor/SourceLocation.cs
@@ -48,7 +48,7 @@ namespace Microsoft.AspNet.Razor
         /// <param name="absoluteIndex">The absolute index.</param>
         /// <param name="lineIndex">The line index.</param>
         /// <param name="characterIndex">The character index.</param>
-        public SourceLocation([NotNull] string filePath, int absoluteIndex, int lineIndex, int characterIndex)
+        public SourceLocation(string filePath, int absoluteIndex, int lineIndex, int characterIndex)
         {
             FilePath = filePath;
             AbsoluteIndex = absoluteIndex;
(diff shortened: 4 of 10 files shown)


src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -112,33 +164,43 @@ private static SourceLocation CreateUndefined()
             return sl;
         }
 
-        public static bool operator <(SourceLocation left, SourceLocation right)
```
pranavkm: I'm not sure how attached we are to these operators - they are scarcely used in our code.

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,49 @@ public override string ToString()
...
         {
-            return 
+            return string.Equals(FilePath, other.FilePath, StringComparison.OrdinalIgnoreCase) &&
                 AbsoluteIndex == other.AbsoluteIndex &&
                 LineIndex == other.LineIndex &&
                 CharacterIndex == other.CharacterIndex;
         }
 
+        /// <inheritdoc />
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
```
pranavkm: Will fix.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
```diff
@@ -190,18 +190,18 @@ public CSharpCodeWriter WriteReturn(string value, bool endLine)
         /// <returns>The current instance of <see cref="CSharpCodeWriter"/>.</returns>
         public CSharpCodeWriter WriteLineNumberDirective(SourceLocation location, string file)
```
NTaylorMullen: We talked about the shared hate of this but in the value of less work :+1:

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -16,13 +20,38 @@ public struct SourceLocation : IEquatable<SourceLocation>, IComparable<SourceLoc
...
+            FilePath = filePath;
             AbsoluteIndex = absoluteIndex;
             LineIndex = lineIndex;
             CharacterIndex = characterIndex;
         }
 
+        /// <summary>
+        /// Path of the file.
+        /// </summary>
+        /// <remarks>When null, the parser assumes it is the file currently being processed.</remarks>
```
NTaylorMullen: `<c>null</c>`

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,49 @@ public override string ToString()
...
             // LineIndex and CharacterIndex can be calculated from AbsoluteIndex and the document content.
-            return AbsoluteIndex;
+            return HashCodeCombiner.Start().Add(FilePath).Add(AbsoluteIndex).CombinedHash;
         }
 
+        /// <inheritdoc />
         public bool Equals(SourceLocation other)
         {
-            return 
+            return string.Equals(FilePath, other.FilePath, StringComparison.OrdinalIgnoreCase) &&
```
NTaylorMullen: Should we assume case insensitive file system?

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -112,33 +164,43 @@ private static SourceLocation CreateUndefined()
...
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the difference of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
         public static SourceLocation operator -(SourceLocation left, SourceLocation right)
         {
+            if (!string.Equals(left.FilePath, right.FilePath, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new ArgumentException(
```
NTaylorMullen: You do this file path check => argument exception in 3 places. Refactor them into an `EnsureFilePaths` method.

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +15,238 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+            // Assert
+            Assert.Equal(hashCodeA, hashCodeB);
+        }
+
+        [Fact]
+        public void GetHashCode_ReturnsHashCode_UsingFilePathAndAbsoluteIndex_WhenFilePathIsNonNull()
+        {
+            // Arrange
+            var sourceLocationA = new SourceLocation("some-path", 3, 53, 94);
+            var sourceLocationB = new SourceLocation("some-path", 3, 37, 46);
```
NTaylorMullen: This doesn't validate that it uses the file path and absolute index. Could be just like `GetHashCode_ReturnsHashCode_UsingAbsoluteIndex` where `FilePath` is being ignored.

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +15,238 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+
+        [Fact]
+        public void CompareTo_Throws_IfSourceLocationsDoNotMatch()
+        {
+            // Arrange
+            var sourceLocationA = new SourceLocation("a-path", 1, 1, 1);
+            var sourceLocationB = new SourceLocation("b-path", 1, 1, 1);
+
+            // Act and Assert
+            var ex = ExceptionAssert.ThrowsArgument(() => sourceLocationA.CompareTo(sourceLocationB),
```
NTaylorMullen: No need to capture return value.

NTaylorMullen:
- Add test for `SourceLocationTracker` verifying that the file paths flow with updated `SourceLocations`, just need to test the `UpdateLocation` method.
- Add test that validates codegen prioritizes `SourceLocation.FilePath` over the contexts source file.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -16,13 +20,38 @@ public struct SourceLocation : IEquatable<SourceLocation>, IComparable<SourceLoc
         public static readonly SourceLocation Undefined = CreateUndefined();
         public static readonly SourceLocation Zero = new SourceLocation(0, 0, 0);
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="SourceLocation"/>.
+        /// </summary>
+        /// <param name="absoluteIndex">The absolute index.</param>
+        /// <param name="lineIndex">The line index.</param>
+        /// <param name="characterIndex">The character index.</param>
         public SourceLocation(int absoluteIndex, int lineIndex, int characterIndex)
```
dougbu: what scenarios will continue to use this constructor?  for example, could it be `internal` and used only in tests?

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: also this `Equals()` doesn't match `GetHashCode()`: source locations for file `Hello.world` may equal those for `HELLO.world` but their hash codes will be different.  must either change `GetHashCode()` to use `Add(filePath, StringComparison.OrdinalIgnoreCase)` or use `StringComparison.Ordinal` here.  not sure which is the better choice given oddities in lower layers w.r.t. case sensitivity.

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,49 @@ public override string ToString()
...
                 LineIndex == other.LineIndex &&
                 CharacterIndex == other.CharacterIndex;
         }
 
+        /// <inheritdoc />
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
         public int CompareTo(SourceLocation other)
         {
+            if (!string.Equals(FilePath, other.FilePath, StringComparison.OrdinalIgnoreCase))
```
dougbu: why `throw` instead of comparing the `FilePath` values?

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -76,12 +125,13 @@ public static SourceLocation Advance(SourceLocation left, string text)
             return tracker.CurrentLocation;
         }
 
-        public static SourceLocation Add(SourceLocation left, SourceLocation right)
+        private static SourceLocation Add(SourceLocation left, SourceLocation right)
```
dougbu: implementation of this and `Subtract()` assumes the two locations are in the same file.  that case cries out for an `ArgumentException`.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: I think we generally avoid throwing the exception from `Ensure` methods to ensure we get the most accurate stack traces.

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: hmm, makes partial sense given error checking in the `+` and `-` operators.  not clear however why these methods are separate.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/SourceLocation.cs
NTaylorMullen: I don't think that warrants duplication of code 3x over. You're right Ensure is probably the wrong term for it maybe `ValidateFilePaths`?

pranavkm: Nah, it's 2 lines of code - it can stay as is.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: I can merge this in - was trying to avoid too much diff.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: I wasn't super sure if it made sense to compare file paths. For instance, if you were to try sorting `SourceLocation { FilePath = "/Views/_GlobalImport", AbsoluteIndex = 10 }` and `SourceLocation { FilePath = null, AbsoluteIndex = 30 }`, how would you expect the results to be? In Mvc's case, an imported `Chunk` belongs to the page's `CodeTree`, so the one from `GlobalImport` would be first.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
pranavkm: :100:

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: We use this one all over the place - I'm being pretty lazy in this PR and specifically using the other ctor when creating inherited Chunks in Mvc. The existing code continues to use this ctor.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: Based on a conversation with @loudej yesterday about case sensitivity in paths - this should be `Ordinal` since it's source is programmatic and not something the user specifies.

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: pls do merge

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: the `Microsoft.AspNet.Mvc.TagHelpers.Internal\GlobbingUrlBuilder+PathComparer` class defines a reasonable comparison you could crib from here.  but the exact semantic is less important than throwing an unexpected exception during a comparison.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: Also confirmed with Todd that these operators aren't used in tooling.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: I guess if we don't necessarily care about the semantics of the Comparison, just the fact that it doesn't throw - I'll use `StringComparison.Ordinal` here.

pranavkm:
Updated

src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
```diff
@@ -49,7 +49,8 @@ public HashCodeCombiner Add(object o)
 
         public HashCodeCombiner Add<TValue>(TValue value, IEqualityComparer<TValue> comparer)
         {
-            return Add(comparer.GetHashCode(value));
+            var hashCode = value != null ? comparer.GetHashCode(value) : 0;
+            return Add(0);
```
dougbu: - why doesn't the comparer handle `null`?
- adding the wrong value to the hash. either return `Add(hashCode)` or do what `Add(object)` does:
``` c#
Add(hashCode);
return this;
```

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -410,4 +410,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="RewriterError_EmptyTagHelperBoundAttribute" xml:space="preserve">
     <value>Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty or contain only whitespace.</value>
   </data>
+  <data name="SourceLocationFilePathDoesNotMatch" xml:space="preserve">
+    <value>Cannot perform operations on '{0}' instances with different file paths.</value>
```
dougbu: might be helpful to add another format parameter e.g. `Cannot perform {1} operations ...`.  pass in `+` or `-` as appropriate.

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -16,13 +20,38 @@ public struct SourceLocation : IEquatable<SourceLocation>, IComparable<SourceLoc
...
+            FilePath = filePath;
             AbsoluteIndex = absoluteIndex;
             LineIndex = lineIndex;
             CharacterIndex = characterIndex;
         }
 
+        /// <summary>
+        /// Path of the file.
+        /// </summary>
+        /// <remarks>When <c>null</c>, the parser assumes it is the file currently being processed.</remarks>
```
dougbu: "it" isn't clear.  suggest "the location is in the file ..."

src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
dougbu: also, if the comparer can't handle `null` suggest making a similar change to `Add(object)`

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,52 @@ public override string ToString()
...
+        /// <inheritdoc />
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
         public int CompareTo(SourceLocation other)
         {
+            var filePathOrdinal = string.Compare(FilePath, other.FilePath, StringComparison.Ordinal);
+            if (filePathOrdinal != 0)
+            {
+                return filePathOrdinal;
+            }
```
dougbu: looks good

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,52 @@ public override string ToString()
...
+            return string.Equals(FilePath, other.FilePath, StringComparison.Ordinal) &&
                 AbsoluteIndex == other.AbsoluteIndex &&
                 LineIndex == other.LineIndex &&
                 CharacterIndex == other.CharacterIndex;
         }
 
+        /// <inheritdoc />
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
         public int CompareTo(SourceLocation other)
```
dougbu: `[NotNull]`

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,52 @@ public override string ToString()
...
             // LineIndex and CharacterIndex can be calculated from AbsoluteIndex and the document content.
-            return AbsoluteIndex;
+            return HashCodeCombiner.Start()
+                .Add(FilePath, StringComparer.Ordinal)
+                .Add(AbsoluteIndex)
+                .CombinedHash;
         }
 
+        /// <inheritdoc />
         public bool Equals(SourceLocation other)
```
dougbu: `[NotNull]`

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,30 +75,52 @@ public override string ToString()
...
             return AbsoluteIndex.CompareTo(other.AbsoluteIndex);
         }
 
+        /// <summary>
+        /// Advances the <see cref="SourceLocation"/> by the length of the <paramref name="text" />.
+        /// </summary>
+        /// <param name="left">The <see cref="SourceLocation"/> to advance.</param>
+        /// <param name="text">The <see cref="string"/> to advance <paramref name="left"/> by.</param>
+        /// <returns>The advanced <see cref="SourceLocation"/>.</returns>
         public static SourceLocation Advance(SourceLocation left, string text)
```
dougbu: is it fine if both parameters are `null`?

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -76,12 +128,37 @@ public static SourceLocation Advance(SourceLocation left, string text)
...
+
+        /// <summary>
+        /// Adds two <see cref="SourceLocation"/>s.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the sum of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
+        public static SourceLocation operator +(SourceLocation left, SourceLocation right)
```
dougbu: `[NotNull]` for both parameters; generally `public` constructors and methods in this PR need a sweep

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -89,57 +166,79 @@ public static SourceLocation Add(SourceLocation left, SourceLocation right)
...
-            sl.CharacterIndex = -1;
-            return sl;
-        }
-
+        /// <summary>
+        /// Determines whether the first operand is lesser than the second operand.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> is lesser than <paramref name="right"/>.</returns>
```
dougbu: lesser -> less

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -76,12 +128,37 @@ public static SourceLocation Advance(SourceLocation left, string text)
             return tracker.CurrentLocation;
         }
 
-        public static SourceLocation Add(SourceLocation left, SourceLocation right)
+        private static SourceLocation CreateUndefined()
         {
+            var sl = new SourceLocation();
```
NTaylorMullen: Since you're touching it just do 

```C#
var sl = new SourceLocation
{
    Absoluteindex = -1,
    LineIndex = -1,
    CharacterIndex = -1
};
```

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -89,57 +166,79 @@ public static SourceLocation Add(SourceLocation left, SourceLocation right)
...
         {
+            if (!string.Equals(left.FilePath, right.FilePath, StringComparison.Ordinal))
+            {
+                throw new ArgumentException(
+                    RazorResources.FormatSourceLocationFilePathDoesNotMatch(nameof(SourceLocation)),
+                    nameof(right));
+            }
+
             return new SourceLocation(
+                left.FilePath,
```
dougbu: including the file in a relative location seems a bit nonsensical.  one possibility would be for `-` to return a `SourceLocation` with `FilePath==null` (maintaining current comparison of input `FilePath` values) and `+` to `throw` if _both_ `SourceLocations` have `FilePath!=null`, then return a new `SourceLocation` with the non-`null` `FilePath` (if any).

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
NTaylorMullen: Still didn't address this guy.

src/Microsoft.AspNet.Razor/Text/SourceLocationTracker.cs
```diff
@@ -85,7 +85,11 @@ private void UpdateInternalState()
 
         private void RecalculateSourceLocation()
         {
-            _currentLocation = new SourceLocation(_absoluteIndex, _lineIndex, _characterIndex);
+            _currentLocation = new SourceLocation(
+                _currentLocation.FilePath,
```
NTaylorMullen: Missing test that validates `SourceLocationTracker` flows the file path.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
```diff
@@ -190,18 +190,18 @@ public CSharpCodeWriter WriteReturn(string value, bool endLine)
...
         {
-            return WriteLineNumberDirective(location.LineIndex + 1, file);
-        }
+            if (!string.IsNullOrEmpty(location.FilePath))
```
NTaylorMullen: Just check for `null` or update the remarks for `SourceLocation`/anywhere it states that if its `null` it defaults to the currently parsed file..

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Razor.Test/Generator/Compiler/CSharp/CSharpCodeWriterTest.cs
```diff
@@ -0,0 +1,48 @@
...
+            // Act
+            writer.WriteLineNumberDirective(sourceLocation, filePath);
+            var code = writer.GenerateCode();
+
+            // Assert
+            Assert.Equal(expected, code);
+        }
+
+        [Fact]
+        public void WriteLineNumberDirectives_UsesSourceLocationFilePath_IfAvailable()
```
dougbu: nit: no 's' in `WriteLineNumberDirective_`

test/Microsoft.AspNet.Razor.Test/Generator/Compiler/CSharp/CSharpCodeWriterTest.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
+{
+    public class CSharpCodeWriterTest
```
dougbu: add a test for the fallback case i.e. using `new SourceLocation(10, 4, 3)`

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
dougbu: @NTaylorMullen's comment hasn't been addressed.  should be something like
``` c#
var sourceLocationA = new SourceLocation(3, 53, 94);
var sourceLocationB = new SourceLocation("some-path", 3, 53, 94);
```
then verify the hash codes are _different_

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +16,262 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+            // Act
+            var hashCodeA = sourceLocationA.GetHashCode();
+            var hashCodeB = sourceLocationB.GetHashCode();
+
+            // Assert
+            Assert.Equal(hashCodeA, hashCodeB);
+        }
+
+        [Fact]
+        public void Equal_ReturnsFalse_IfIndexesDiffer()
```
dougbu: this indicates a longstanding bug in the `SourceLocation` implementation.  hash codes should be different for values that don't compare equal.  put another way, either `Equals()` should only check the file path and absolute index or `GetHashCode()` should include the line and character indices in its calculation.

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +16,262 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+            Assert.Equal(path, result.FilePath);
+            Assert.Equal(2, result.AbsoluteIndex);
+            Assert.Equal(5, result.LineIndex);
+            Assert.Equal(6, result.CharacterIndex);
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("path-to-file")]
+        public void Advance_PreservesSourceLocationFilePath(string path)
```
dougbu: nit: `Advance`?

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +16,262 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+
+            // Assert
+            Assert.Equal(path, result.FilePath);
+            Assert.Equal(6, result.AbsoluteIndex);
+            Assert.Equal(5, result.LineIndex);
+            Assert.Equal(9, result.CharacterIndex);
+        }
+
+        [Fact]
+        public void Subtract_Throws_IfSourceLocationsDoNotMatch()
```
dougbu: `_IfFilePathsDoNotMatch`

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +16,262 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+
+            // Act
+            var result = sourceLocationA.CompareTo(sourceLocationB);
+
+            // Assert
+            Assert.Equal(indexA.CompareTo(indexB), result);
+        }
+
+        [Fact]
+        public void Add_Throws_IfSourceLocationsDoNotMatch()
```
dougbu: `_IfFilePathsDoNotMatch`

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: was this fixed?

dougbu:
:watch:

NTaylorMullen:
Should investigate travis failures. It's pasing fine on dev branch.

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: Not on value types.

src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs
pranavkm: Unfortunately the comparer performs a null check so this ends up throwing. I'll fix the hash code

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: I'm neutral about this - subtraction isn't used anywhere in the codebase. Additions are used where the right operand is a relative location and it's used in calculating the offset from document start. So it sort of makes sense to preserve the document location in that case.

src/Microsoft.AspNet.Razor/Text/SourceLocationTracker.cs
pranavkm: There's one in SourceLocationTest which verifies this.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpCodeWriter.cs
pranavkm: :+1:

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
pranavkm: I'm not comfortable making that change since we might have code that relies on this quirk. Leaving it as is.

pranavkm:
Updated

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -4,25 +4,65 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="SourceLocation"/>.
+        /// </summary>
+        /// <param name="filePath">The file path.</param>
+        /// <param name="absoluteIndex">The absolute index.</param>
+        /// <param name="lineIndex">The line index.</param>
+        /// <param name="characterIndex">The character index.</param>
+        public SourceLocation([NotNull] string filePath, int absoluteIndex, int lineIndex, int characterIndex)
```
dougbu: can't use `[NotNull]` here.  breaks other constructor.

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: nevermind

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -89,57 +165,82 @@ public static SourceLocation Add(SourceLocation left, SourceLocation right)
...
+                lineIndex: left.LineIndex - right.LineIndex,
+                characterIndex: characterIndex);
         }
 
+        /// <summary>
+        /// Determines whether the first operand is less than the second operand.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns><c>true</c> if <paramref name="left"/> is less than <paramref name="right"/>.</returns>
```
dougbu: suggest duplicating the `<exception/>` documentation from `CompareTo()` and `Equals()` in the methods that use them

test/Microsoft.AspNet.Razor.Test/Generator/Compiler/CSharp/CSharpCodeWriterTest.cs
dougbu: thx

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
```diff
@@ -14,9 +16,268 @@ public void ConstructorWithLineAndCharacterIndexSetsAssociatedProperties()
...
+            ExceptionAssert.ThrowsArgument(
+                () => { var result = sourceLocationA + sourceLocationB; },
+                "right",
+                $"Cannot perform '+' operations on 'SourceLocation' instances with different file paths.");
+        }
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("same-path")]
+        public void Add_IgnoresCharacterIndexIfLineIndexIsNonZero(string path)
```
dougbu: add `Right` before `LineIndex`

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: nevermind

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,43 +86,78 @@ public override string ToString()
...
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the sum of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
+        public static SourceLocation operator +(SourceLocation left, SourceLocation right)
         {
+            if (!string.Equals(left.FilePath, right.FilePath, StringComparison.Ordinal))
+            {
+                throw new ArgumentException(
```
dougbu: did you say `+` and `-` are rarely used?  wondering because `(A - B) + C` will always `throw` now.  this is why I thought we should allow one `FilePath` to be `null` even if the other isn't.

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: :+1:

test/Microsoft.AspNet.Razor.Test/SourceLocationTest.cs
dougbu: filed #362

dougbu:
:watch: mainly for `(A - B) + C` discussion

src/Microsoft.AspNet.Razor/SourceLocation.cs
pranavkm: `+` is used in a couple of places, `-` isn't used anywhere in our code (we don't know if tooling \ other code outside of Mvc & Razor rely on this)

pranavkm:
Updated.

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -45,43 +86,82 @@ public override string ToString()
...
 
-        public static SourceLocation Add(SourceLocation left, SourceLocation right)
+        /// <summary>
+        /// Adds two <see cref="SourceLocation"/>s.
+        /// </summary>
+        /// <param name="left">The left operand.</param>
+        /// <param name="right">The right operand.</param>
+        /// <returns>A <see cref="SourceLocation"/> that is the sum of the left and right operands.</returns>
+        /// <exception cref="ArgumentException">if the <see cref="FilePath"/> of the left and right operands
+        /// are different.</exception>
```
dougbu: add `, and neither is null`

dougbu:
:shipit:

Issue #332, opened 2015-03-24
Tooling: Directives don't format correctly when entering newline.


NTaylorMullen:
Typing `@inherits ....` results in improperly indented markup on the following line. Looked this over and I believe this is due to the `AcceptedCharacters` associated with directives.

NTaylorMullen:
Investigated this a bit more and it looks like this has always been an issue with Razor (repro'd it with Mvc5 / VS2013), bumping this down lower on my priority list.

NTaylorMullen:
fedd53aab807f526d835ae3ada75a3e678cd309c

https://github.com/aspnet/Mvc/commit/eaa58bc91393832ca833436b80779088c1630262

PR #360, opened 2015-04-14
Update Razor directives to format correctly when entering newline.

NTaylorMullen:
- Existing Razor directives `@layout`, `@inherits`, `@addTagHelper`, `@tagHelperPrefix` and `@removeTagHelper` should only ever span a single line and need to cause a re-parse when a newline is entered during design time. To do this modified their `AcceptedCharacters` to accept anything other than newline rather than anything.
- Updated existing tests to now expect `AcceptedCharacters.AnyExceptNewline` when directives are present.

**Note:** Manually tested this in my own VS' editor and all worked.

#332

**Mvc react PR:** https://github.com/aspnet/Mvc/pull/2385


dougbu:
:shipit: from me but agree @pranavkm should chime in

pranavkm:
:shipit:

Issue #182, opened 2014-10-07
Add [NotBound] property to indicate that a property should not come from HTML


NTaylorMullen:
Right now all public get; set; properties on a `TagHelper` are automatically bound to HTML with no way to opt-out. Adding a `[NotBound]` attribute (or another attribute name) would enable users to opt-out of `TagHelper` binding.

Example:
```C#
class DivTagHelper : TagHelper
{
    [NotBound]
    public string MyValue { get; set; } // Is not auto-filled and is not expected to be represented in HTML.
}
```

dougbu:
de95f67
aspnet/Mvc@c162209

Issue #364, opened 2015-04-25
Duplicate HTML attribute names should be invalid


dougbu:
Razor intentionally does not support binding a single attribute from the .cshtml file to multiple attributes in a tag helper.  However `TagHelperDescriptorFactory` creates `TagHelperDescriptor`s for such tab helpers and does not report an error to the `ErrorSink`.

e.g. the following tag helper should not be valid since the calculated name of one property matches the provided name of the other.
``` c#
private class DuplicateAttributeNameTagHelper
{
    public string MyNameIsLegion { get; set; }

    [HtmlAttributeName("my-name-is-legion")]
    public string Fred { get; set; }
}
```

dougbu:
also restore test at https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs#L497

danroth27:
We would only fix this based on customer feedback.

PR #365, opened 2015-04-27
Add `HtmlAttributeNotBoundAttribute`

dougbu:
- #182
- ignore otherwise-bound (i.e. `public`) properties in tag helpers

nits:
- add more `TagHelperDescriptorFactory` tests e.g. of `internal` properties
- remove `private` setter from `HtmlAttributeNameAttribute`
- use `null` propagation to shorten `IsAccessibleProperty` expression
- clean up some trailing whitespace

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNotBoundAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNotBoundAttribute.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Indicates the associated <see cref="ITagHelper"/> property should not be bound to HTML attributes.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
+    public sealed class HtmlAttributeNotBoundAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="HtmlAttributeNotBoundAttribute"/> class.
+        /// </summary>
+        public HtmlAttributeNotBoundAttribute()
+        {
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -203,7 +203,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(
-            Type type, 
+            Type type,
             ErrorSink errorSink)
         {
             var accessibleProperties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
@@ -226,12 +226,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             Type parentType,
             ErrorSink errorSink)
         {
-            // data-* attributes are explicitly not implemented by user agents and are not intended for use on 
+            // data-* attributes are explicitly not implemented by user agents and are not intended for use on
             // the server; therefore it's invalid for TagHelpers to bind to them.
             if (attributeDescriptor.Name.StartsWith(DataDashPrefix, StringComparison.OrdinalIgnoreCase))
             {
                 errorSink.OnError(
-                    SourceLocation.Zero, 
+                    SourceLocation.Zero,
                     Resources.FormatTagHelperDescriptorFactory_InvalidBoundAttributeName(
                         attributeDescriptor.PropertyName,
                         parentType.FullName,
@@ -255,10 +255,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
         private static bool IsAccessibleProperty(PropertyInfo property)
         {
-            return property.GetMethod != null &&
-                   property.GetMethod.IsPublic &&
-                   property.SetMethod != null &&
-                   property.SetMethod.IsPublic;
+            // Accessible properties are those with public getters and setters and without [HtmlAttributeNotBound].
+            return property.GetMethod?.IsPublic == true &&
+                property.SetMethod?.IsPublic == true &&
+                property.GetCustomAttribute<HtmlAttributeNotBoundAttribute>(inherit: false) == null;
         }
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -467,6 +467,51 @@ public void CreateDescriptor_OnlyAcceptsPropertiesWithPublicGetAndSet()
...
+                typeof(NotBoundAttributeTagHelper),
+                errorSink);
+
+            // Assert
+            Assert.Empty(errorSink.Errors);
+            var descriptor = Assert.Single(descriptors);
+            Assert.Equal(expectedDescriptor, descriptor, CaseSensitiveTagHelperDescriptorComparer.Default);
+        }
+
+        [Fact(Skip = "#364")]
```
NTaylorMullen: Should make sure to add a link to this line in the #364 issue

NTaylorMullen:
:shipit:

dougbu:
de95f67

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -255,10 +255,10 @@ private static TagHelperAttributeDescriptor ToAttributeDescriptor(PropertyInfo p
...
         private static bool IsAccessibleProperty(PropertyInfo property)
         {
-            return property.GetMethod != null &&
-                   property.GetMethod.IsPublic &&
-                   property.SetMethod != null &&
-                   property.SetMethod.IsPublic;
+            // Accessible properties are those with public getters and setters and without [HtmlAttributeNotBound].
+            return property.GetMethod?.IsPublic == true &&
+                property.SetMethod?.IsPublic == true &&
+                property.GetCustomAttribute<HtmlAttributeNotBoundAttribute>(inherit: false) == null;
```
Eilon: Why `inherit: false`?

Eilon: What happens if there's a base class tag helper with an abstract/virtual property that is annotated with the attribute, and that is overridden in a derived tag helper? Does this ignore that? Add a test?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNotBoundAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Indicates the associated <see cref="ITagHelper"/> property should not be bound to HTML attributes.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
+    public sealed class HtmlAttributeNotBoundAttribute : Attribute
```
Eilon: This name is a bit of a mouthful. Maybe `NonHtmlAttributeAttribute` to be more consistent with https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/NonControllerAttribute.cs#L13? (Yes, I know I'm suggesting `XyzAttributeAttribute` but we wouldn't be the first ones to do that, and it makes sense.)

@DamianEdwards thoughts on name?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: the `inherit` parameter is ignored for properties, see https://msdn.microsoft.com/en-us/library/kff8s254.aspx#remarksToggle  I included the argument for consistency with the `Inherited` setting on the attribute itself.

Eilon: Hmm ok then why are we ignoring inheritance entirely? Shouldn't we allow that? Seems dangerous otherwise.

dougbu: > why are we ignoring inheritance entirely?

I asked the same question ages ago. The decision was to have all attributes `sealed` with `Inherited = false`. IIRC the intent was to require explicit attributes in a subclass of a tag helper.

In this particular case, note properties inherited from a tag helper bring along their attributes.  `Inherited` primarily matters for properties when a `virtual` property has an `override` in the subclass. See `TagHelperDescriptorFactoryTest` for some examples.

Eilon: Oh weird. This looks wrong: https://github.com/aspnet/Razor/blob/de95f67400c8f20cf9cfe063c291720bf7d71cce/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs#L292-L293

What was the reasoning for not inheriting the attributes?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNotBoundAttribute.cs
NTaylorMullen: I liked `HtmlAttributeNotBound` because when typing the other attribute that modifies TagHelper` properties, `HtmlAttributeName`, you also end up with the `HtmlAttributeNotBound` as an option via IntelliSense.

Eilon: The problem with the name `HtmlAttributeNotBound` is that it implies that it's an HTML attribute that isn't bound. But doesn't it rather mean that "this isn't an HTML attribute at all"? I'm all for having nice Intellisense, but only if the type name actually makes sense :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Don't you think it'd be odd if by doing:

https://github.com/aspnet/Razor/blob/de95f67400c8f20cf9cfe063c291720bf7d71cce/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs#L1126-L1129

You end up with a different `HtmlAttributeName` association? I feel like it'd be misleading to see `public override string ValidAttribute1 { get; set; }` (no attribute) but you utilize it like `<inherited-overridden-attribute SomethingElse="..." />` with no attribute visible.

I don't feel very strongly either way sense inheriting from `TagHelper`s will be extremely rare; I just like the approach of clarity. Aka if you override you re-specify `HtmlAttributeName` to make it clear how to use it when re-scanning your own `TagHelper`s.

Eilon: Why would overriding the property *change* the name? That seems far weirder. The most common case is someone would want to change some behavior (e.g. in the getter compute something, or in the setter invalidate something). If they want to change the name, they can re-apply the name attribute to alter it.

Issue #362, opened 2015-04-16
`Equals()` and `GetHashCode()` implementations are inconsistent


dougbu:
`SourceLocation.GetHashCode()` and `SourceLocation.Equals()` use different properties, a fairly basic no-no.  problems are unlikely unless one of the source locations is corrupt.  nonetheless this quirk should be removed.

in particular `GetHashCode()` ignores `LineIndex` and `CharacterIndex` while `Equals()` includes them in its comparison.  `Equals()` is the odd one out because `CompareTo()` also ignores `LineIndex` and `CharacterIndex`.

e.g. the following test succeeds
``` c#
// Arrange
var sourceLocationA = new SourceLocation(10, 3, 4);
var sourceLocationB = new SourceLocation(10, 45, 8754);

// Act
var compare = sourceLocationA.CompareTo(sourceLocationB);
var equals = sourceLocationA.Equals(sourceLocationB);
var hashCodeA = sourceLocationA.GetHashCode();
var hashCodeB = sourceLocationB.GetHashCode();

// Assert
Assert.Equal(0, compare);
Assert.False(equals);    // !!!
Assert.Equal(hashCodeA, hashCodeB);
```

original discussion at https://github.com/aspnet/Razor/pull/357#discussion_r28467249

Eilon:
That doesn't seem like a real bug, does it? It's perfectly fine to implement `GetHashCode()` to return `0` and then to implement `Equals()` to do a proper comparison. Nothing will break. The only contract is that if two instances are compared and have `Equals()` be true that their `GetHashCode()` returns the same value.

dougbu:
I haven't seen anything except `Equals()` consistency with `GetHashCode()` but it's possible nothing will break due to that inconsistency here.  now a bit more concerned about `Equals()` returning `false` in scenarios where `CompareTo()` returns `0`.

danroth27:
@dougbu Do the Right Thing

dougbu:
Updated subject since `SourceLocation.GetHashCode()` was fine and `SourceLocation.Equals()` versus `SourceLocation.CompareTo()` is only one example of `Equals()` or `GetHashCode()` issues.

dougbu:
6384977

PR #366, opened 2015-04-27
Improve `GetHashCode()` implementations and their consistency with `Equals()`

dougbu:
- #362 and more
- make `SourceLocation.Equals()` and `SourceLocation.CompareTo()` consistent
- use only immutable values in `GetHashCode()` calculations
 - avoid `object.GetHashCode()`; that calculates hash of reference
 - make a few more properties obviously immutable; remove `private` setters
 - BUT this leads to some static `GetHashCode()` values
- add `SyntaxTreeNode.GetEquivalenceHash()`
- correct important typo in `TagHelperDescriptorComparer`
- ensure `Equals()` does not `throw` an NRE e.g. in `LineMapping`

Add lots of tests
- not for all updated classes but enough to see impact of odd choices
 - assymetric `Equals()` implementations due to specificity in subclasses
 - some `Equals()` check for an `interface` implementation, not their `class`

nits:
- remove redundant null checks e.g. when also done in `base.Equals()`
- add `StringComparer.Ordinal` if `StringComparison.Ordinal` used in `Equals()`

### DIFF
--- a/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
+++ b/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
@@ -19,12 +19,18 @@ namespace Microsoft.AspNet.Razor.Editor
     public class ImplicitExpressionEditHandler : SpanEditHandler
     {
         private readonly ISet<string> _keywords;
+        private readonly IReadOnlyCollection<string> _readOnlyKeywords;
 
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "Func<T> is the recommended delegate type and requires this level of nesting.")]
         public ImplicitExpressionEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer, ISet<string> keywords, bool acceptTrailingDot)
             : base(tokenizer)
         {
             _keywords = keywords ?? new HashSet<string>();
+
+            // HashSet<T> implements IReadOnlyCollection<T> as of 4.6, but does not for 4.5.1. If the runtime cast
+            // succeeds, avoid creating a new collection.
+            _readOnlyKeywords = (_keywords as IReadOnlyCollection<string>) ?? _keywords.ToArray();
+
             AcceptTrailingDot = acceptTrailingDot;
         }
 
@@ -34,7 +40,7 @@ namespace Microsoft.AspNet.Razor.Editor
         {
             get
             {
-                return (IReadOnlyCollection<string>)_keywords;
+                return _readOnlyKeywords;
             }
         }
 
--- a/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
+++ b/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
@@ -18,18 +18,25 @@ namespace Microsoft.AspNet.Razor.Editor
 {
     public class ImplicitExpressionEditHandler : SpanEditHandler
     {
-        private static readonly int TypeHashCode = typeof(ImplicitExpressionEditHandler).GetHashCode();
+        private readonly ISet<string> _keywords;
 
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "Func<T> is the recommended delegate type and requires this level of nesting.")]
         public ImplicitExpressionEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer, ISet<string> keywords, bool acceptTrailingDot)
             : base(tokenizer)
         {
-            Keywords = keywords ?? new HashSet<string>();
+            _keywords = keywords ?? new HashSet<string>();
             AcceptTrailingDot = acceptTrailingDot;
         }
 
         public bool AcceptTrailingDot { get; }
-        public ISet<string> Keywords { get; }
+
+        public IReadOnlyCollection<string> Keywords
+        {
+            get
+            {
+                return (IReadOnlyCollection<string>)_keywords;
+            }
+        }
 
         public override string ToString()
         {
@@ -40,15 +47,15 @@ namespace Microsoft.AspNet.Razor.Editor
         {
             var other = obj as ImplicitExpressionEditHandler;
             return base.Equals(other) &&
-                Keywords.SetEquals(other.Keywords) &&
+                _keywords.SetEquals(other._keywords) &&
                 AcceptTrailingDot == other.AcceptTrailingDot;
         }
 
         public override int GetHashCode()
         {
-            // Hash code should include only immutable properties but Equals also checks the type.
+            // Hash code should include only immutable properties and base has none.
             return HashCodeCombiner.Start()
-                .Add(TypeHashCode)
+                .Add(Keywords)
                 .Add(AcceptTrailingDot)
                 .CombinedHash;
         }
@@ -306,7 +313,7 @@ namespace Microsoft.AspNet.Razor.Editor
         {
             using (var reader = new StringReader(newContent))
             {
-                return Keywords.Contains(reader.ReadWhile(ParserHelpers.IsIdentifierPart));
+                return _keywords.Contains(reader.ReadWhile(ParserHelpers.IsIdentifierPart));
             }
         }
     }
--- a/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
+++ b/src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
@@ -18,15 +18,18 @@ namespace Microsoft.AspNet.Razor.Editor
 {
     public class ImplicitExpressionEditHandler : SpanEditHandler
     {
+        private static readonly int TypeHashCode = typeof(ImplicitExpressionEditHandler).GetHashCode();
+
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "Func<T> is the recommended delegate type and requires this level of nesting.")]
         public ImplicitExpressionEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer, ISet<string> keywords, bool acceptTrailingDot)
             : base(tokenizer)
         {
-            Initialize(keywords, acceptTrailingDot);
+            Keywords = keywords ?? new HashSet<string>();
+            AcceptTrailingDot = acceptTrailingDot;
         }
 
-        public bool AcceptTrailingDot { get; private set; }
-        public ISet<string> Keywords { get; private set; }
+        public bool AcceptTrailingDot { get; }
+        public ISet<string> Keywords { get; }
 
         public override string ToString()
         {
@@ -36,18 +39,17 @@ namespace Microsoft.AspNet.Razor.Editor
         public override bool Equals(object obj)
         {
             var other = obj as ImplicitExpressionEditHandler;
-            return other != null &&
-                   base.Equals(other) &&
-                   Keywords.SetEquals(other.Keywords) &&
-                   AcceptTrailingDot == other.AcceptTrailingDot;
+            return base.Equals(other) &&
+                Keywords.SetEquals(other.Keywords) &&
+                AcceptTrailingDot == other.AcceptTrailingDot;
         }
 
         public override int GetHashCode()
         {
+            // Hash code should include only immutable properties but Equals also checks the type.
             return HashCodeCombiner.Start()
-                .Add(base.GetHashCode())
+                .Add(TypeHashCode)
                 .Add(AcceptTrailingDot)
-                .Add(Keywords)
                 .CombinedHash;
         }
 
@@ -102,12 +104,6 @@ namespace Microsoft.AspNet.Razor.Editor
             }
 
             return PartialParseResult.Rejected;
-        }
-
-        private void Initialize(ISet<string> keywords, bool acceptTrailingDot)
-        {
-            Keywords = keywords ?? new HashSet<string>();
-            AcceptTrailingDot = acceptTrailingDot;
         }
 
         // A dotless commit is the process of inserting a '.' with an intellisense selection.
--- a/src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
+++ b/src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
@@ -170,8 +170,8 @@ namespace Microsoft.AspNet.Razor.Editor
         public override bool Equals(object obj)
         {
             var other = obj as SpanEditHandler;
-            return obj != null &&
-                GetType() == obj.GetType() &&
+            return other != null &&
+                GetType() == other.GetType() &&
                 AcceptedCharacters == other.AcceptedCharacters &&
                 EditorHints == other.EditorHints;
         }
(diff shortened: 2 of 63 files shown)


dougbu:
@Eilon let's talk about the issues shown in new tests

dougbu:
@Eilon, @NTaylorMullen updated and ready to go.  would prefer to defer additional tests a bit...

src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -18,15 +18,25 @@ namespace Microsoft.AspNet.Razor.Editor
...
-            Initialize(keywords, acceptTrailingDot);
+            _keywords = keywords ?? new HashSet<string>();
+            AcceptTrailingDot = acceptTrailingDot;
         }
 
-        public bool AcceptTrailingDot { get; private set; }
-        public ISet<string> Keywords { get; private set; }
+        public bool AcceptTrailingDot { get; }
+
+        public IReadOnlyCollection<string> Keywords
```
NTaylorMullen: Why not just `Keywords { get; }` and do the cast in the ctor?

src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
```diff
@@ -169,17 +170,16 @@ public override string ToString()
         public override bool Equals(object obj)
         {
             var other = obj as SpanEditHandler;
-            return other != null &&
-                   AcceptedCharacters == other.AcceptedCharacters &&
-                   EditorHints == other.EditorHints;
+            return obj != null &&
```
NTaylorMullen: `obj` => `other` same with new line.

src/Microsoft.AspNet.Razor/Generator/AddImportCodeGenerator.cs
```diff
@@ -39,16 +39,14 @@ public override bool Equals(object obj)
...
         }
 
         public override int GetHashCode()
         {
-            return HashCodeCombiner.Start()
-                .Add(Namespace)
-                .Add(NamespaceKeywordLength)
-                .CombinedHash;
+            // Hash code should include only immutable properties.
+            return Namespace == null ? 0 : StringComparer.Ordinal.GetHashCode(Namespace);
```
NTaylorMullen: Does `StringComparer.Ordinal.GetHashCode` not like `null`?

src/Microsoft.AspNet.Razor/Editor/ImplicitExpressionEditHandler.cs
dougbu: this way we don't have to pay the penalty of querying using only Linq. besides that `IReadOnlyCollection<T>` doesn't support `SetEquals()`.

src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
dougbu: why?  if the `GetType()` comparison succeeds, the conditional cast must also succeed.  this allows the compiler to skip the cast attempt when one of the other comparisons fails.

src/Microsoft.AspNet.Razor/Generator/AddImportCodeGenerator.cs
dougbu: correct; it's a sad :panda_face: when given `null`

src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
NTaylorMullen: It makes it harder to scan but I don't feel strongly about it.

src/Microsoft.AspNet.Razor/Generator/DynamicAttributeBlockCodeGenerator.cs
```diff
@@ -47,14 +46,12 @@ public override bool Equals(object obj)
         {
             var other = obj as DynamicAttributeBlockCodeGenerator;
             return other != null &&
```
NTaylorMullen: Can just call `base.Equals` which will also validate the type.

src/Microsoft.AspNet.Razor/Generator/ExpressionCodeGenerator.cs
```diff
@@ -30,12 +32,13 @@ public override string ToString()
 
         public override bool Equals(object obj)
         {
-            return obj is ExpressionCodeGenerator;
+            return obj != null &&
```
NTaylorMullen: Even the the `Equals` paradigm on these classes is odd, we may want to move this bit to `HybridCodeGenerator` to stay consistent. Don't feel strongly though.

src/Microsoft.AspNet.Razor/Generator/LiteralAttributeCodeGenerator.cs
```diff
@@ -62,9 +63,9 @@ public override bool Equals(object obj)
         {
             var other = obj as LiteralAttributeCodeGenerator;
             return other != null &&
```
NTaylorMullen: Did we not want to ensure symmetric `Equals`?In order to do that we should be calling down into base.equals

src/Microsoft.AspNet.Razor/Generator/ExpressionCodeGenerator.cs
dougbu: `HybridCodeGenerator` is `abstract` (for some reason) and has no other subclass.  so it really doesn't matter today.

my preference would be to delete `HybridCodeGenerator` but that doesn't fit in this PR at all. so I'll leave this as-is.

src/Microsoft.AspNet.Razor/Generator/LiteralAttributeCodeGenerator.cs
dougbu: unnecessary. more concretely, `Equals()` remains commutative until and unless someone creates a `LiteralAttributeCodeGenerator` subclass. not likely.

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
```diff
@@ -111,14 +112,18 @@ public override bool Equals(object obj)
...
+                ChildrenEqual(Children, other.Children);
         }
 
         public override int GetHashCode()
         {
-            return (int)Type;
+            return HashCodeCombiner.Start()
+                .Add(Type)
+                .Add(CodeGenerator)
+                .Add(Children)
```
NTaylorMullen: Thinking we shouldn't add `Children`, the `Parent` property is internal and we set it in a few locations; ultimately just to not get ourselves in trouble.

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Span.cs
```diff
@@ -126,29 +126,32 @@ public override bool EquivalentTo(SyntaxTreeNode node)
         {
             var other = node as Span;
             return other != null &&
```
NTaylorMullen: Why not the normal `.GetType()` comparisons?

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Span.cs
```diff
@@ -126,29 +126,32 @@ public override bool EquivalentTo(SyntaxTreeNode node)
...
-                   Kind.Equals(other.Kind) &&
-                   Start.Equals(other.Start) &&
-                   EditHandler.Equals(other.EditHandler) &&
-                   string.Equals(other.Content, Content, StringComparison.Ordinal);
+                Kind.Equals(other.Kind) &&
+                Start.Equals(other.Start) &&
+                EditHandler.Equals(other.EditHandler) &&
+                string.Equals(other.Content, Content, StringComparison.Ordinal);
```
NTaylorMullen: Hmm curious as to why we never added the `CodeGenerator` comparison here. /cc @anurse

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -116,9 +116,9 @@ public override string ToString()
         public bool Equals(TagHelperBlock other)
         {
             return other != null &&
-                   TagName == other.TagName &&
-                   Attributes.SequenceEqual(other.Attributes) &&
-                   base.Equals(other);
+                TagName == other.TagName &&
```
NTaylorMullen: Since we're touching this lets change to `string.Equals`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -116,9 +116,9 @@ public override string ToString()
...
+                Attributes.SequenceEqual(other.Attributes) &&
+                base.Equals(other);
```
NTaylorMullen: Do this first and remove the `other != null` bit.

src/Microsoft.AspNet.Razor/RazorError.cs
```diff
@@ -46,19 +61,23 @@ public override string ToString()
 
         public override bool Equals(object obj)
         {
-            var err = obj as RazorError;
-            return (err != null) && Equals(err);
+            var error = obj as RazorError;
+            return Equals(error);
```
NTaylorMullen: Symmetric equals question again.

src/Microsoft.AspNet.Razor/RazorError.cs
```diff
@@ -46,19 +61,23 @@ public override string ToString()
...
         }
 
         public override int GetHashCode()
         {
-            return base.GetHashCode();
+            return HashCodeCombiner.Start()
+                .Add(Message, StringComparer.Ordinal)
+                .Add(Location)
```
NTaylorMullen: Include `Length`

src/Microsoft.AspNet.Razor/SourceLocation.cs
```diff
@@ -89,13 +89,13 @@ public override string ToString()
         /// <inheritdoc />
         public override bool Equals(object obj)
         {
-            return (obj is SourceLocation) && Equals((SourceLocation)obj);
+            return obj is SourceLocation &&
+                Equals((SourceLocation)obj);
```
NTaylorMullen: just do `as` instead of this cast and the above `is`

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
dougbu: nah, `Parent` is not an issue. `SyntaxTreeNode` is `abstract` and neither `Block` nor `Span` include it in their `Equals()` or `GetHashCode()` implementations.

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Span.cs
dougbu: only need `GetType()` when a subclass exists

dougbu: and this implementation is still checking the type; see `as`

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Span.cs
dougbu: this is `EquivalentTo()` which is intended as a subset of `Equals()`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlock.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/RazorError.cs
dougbu: have no `RazorError` subclass scenario; same for every case where `Equals()` does not check for `Type` match.

src/Microsoft.AspNet.Razor/RazorError.cs
dougbu: no, `Length` is not included in `Equals()` calculation. so equal values with different `Length`s would have different hash codes. we have an answer for that case: don't do that.

src/Microsoft.AspNet.Razor/SourceLocation.cs
dougbu: absolutely not. `SourceLocation` is a `struct`.

src/Microsoft.AspNet.Razor/Editor/SpanEditHandler.cs
dougbu: I don't feel strongly either but :ok:

src/Microsoft.AspNet.Razor/Generator/DynamicAttributeBlockCodeGenerator.cs
dougbu: no need as mentioned elsewhere when there's no subclass

src/Microsoft.AspNet.Razor/Generator/LiteralAttributeCodeGenerator.cs
NTaylorMullen: Isn't consistency the goal though?

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
NTaylorMullen: Hmm, making me think for consistency we should add the Equals to `SyntaxTreeNode` then, sounds wrong not to.

src/Microsoft.AspNet.Razor/RazorError.cs
NTaylorMullen: We should be comparing it. An error with unequal `Length`'s is not the same error.

src/Microsoft.AspNet.Razor/SourceLocation.cs
NTaylorMullen: Ahhh :+1: dur

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
NTaylorMullen: Talked in person :-1: for adding `Parent`

src/Microsoft.AspNet.Razor/Generator/LiteralAttributeCodeGenerator.cs
dougbu: as discussed offline, consistency between `Equals()` and `GetHashCode()` is the goal. that's not the same as doing the same (somewhat inefficient in these base classes) things in all `Equals()` implementations.

src/Microsoft.AspNet.Razor/RazorError.cs
dougbu: separate issue if it is an issue at all.

BTW two errors with identical messages complaining about the same start point in the code probably _should_ be considered equal. so checking `Length` would just make `Equals()` less efficient.

dougbu:
@NTaylor rebased and updated

NTaylorMullen:
:shipit:

NTaylorMullen:
Removed the shipit, looks like the casts to `IReadOnlyCollection` are breaking PR for travis-ci.

https://travis-ci.org/aspnet/Razor/builds/60466235

dougbu:
Added commit similar to that in aspnet/Mvc#2471

NTaylorMullen:
:shipit:

dougbu:
6384977

Issue #359, opened 2015-04-14
We should remove the @layout directive.


NTaylorMullen:
`@layout "....."` is equivalent to doing:
```
@{
    Layout = "...."
}
```

And is for the most part not used in Razor. We should remove it due to its limited capability.

danroth27:
MVC 5 compat

NTaylorMullen:
I think you should reconsider!

danroth27:
Please add an announcement for this in the https://github.com/aspnet/Announcements repo.

NTaylorMullen:
2d20063ba5a17188b6d9591dff4cbe2e6aac7426

maxtoroq:
Limited capability? What do you mean? It does what is was designed to do. If it does little, why's that a bad thing? I love the layout directive, it makes layout assignment more clear and declarative.

Now, I understand sometimes you have to break stuff to make advancements, but this is such a little cool feature, where's the value in removing it?

Also, your argument against backcompat is simply stupid. All breaking changes, even the small ones, add to the sum total of incompatibilities. The *we've broken a lot already, no harm in breaking some more* attitude is disrespectful to the people who depend on those features.

NTaylorMullen:
@maxtoroq Having to maintain this unique directive. when `@layout "foo"` can be re-written as `@{ Layout = "foo" }` was the reason.

At the core Razor level the `@layout` directive behaved significantly different than every other Razor directive (it was contextually aware) and therefore required us to consider the impact when implementing new features.

Lets looks at its impact on one of the newer features `TagHelper`s:

```HTML
<myTagHelper>
    @layout "Foo"
</myTagHelper>
```

If `myTagHelper` suppresses its output and doesn't execute its innards the `Layout` will not be set. This isn't the case for the other Razor directives (`@model`, `@using` etc.) because they aren't contextually aware, they affect parsing.

maxtoroq:
Why would `@layout` be allowed in a tag helper in the first place? Can't you make it a top-level directive only?

NTaylorMullen:
@maxtoroq that's how directives have always worked, they can be placed anywhere and (should only) affect parsing. We could make it a top-level directive but that wouldn't abide by the current expected definition of a Razor directive.

`@{ Layout = "foo" }` as a replacement isn't that horrible is it?

PR #367, opened 2015-04-27
Remove layout directive.

NTaylorMullen:
- Removed all instances of the layout directive from src and test projects.

#359


dougbu:
:shipit: after you've done the one-line Mvc reaction

PR #368, opened 2015-04-28
Remove extra `[NotNullAttribute]`

dougbu:
- add necessary build-time dependency
- add missing `using Microsoft.Framework.Internal;` statements
- remove incorrect NotNullArgument.cs file entirely

nit: clean up some trailing whitespace


NTaylorMullen:
Don't we need to add the `NotNullAttribute` package to the test projects since there's no transitive dependencies for shared source packages?

:watch:

dougbu:
> need to add the `NotNullAttribute` package to the test projects

no. the test projects don't use `[NotNull]` themselves. even if they did, one version is available to them due to the `[InternalsVisibleTo]` from Razor to the two test projects.

NTaylorMullen:
@dougbu What about `Microsoft.AspNet.Runtime.Test`? It has internals visible for both `Microsoft.AspNet.Razor` and `Microsoft.AspNet.Razor.Runtime`. I'd expect that you'd get a type conflict.

dougbu:
> It has internals visible for both Microsoft.AspNet.Razor and Microsoft.AspNet.Razor.Runtime

wow. Razor.Runtime has two `AssemblyInfo.cs` files!

in any case, everything builds and tests file. would be _worse_ if one of the test projects included a class definition for something imported from the runtime projects.

dougbu:
> everything builds and tests file

you were correct. not sure how my builds succeeded before. saw the issue after `git clean`. will send another iteration once I've merge the odd extra `AssemblyInfo.cs`.

dougbu:
updated

NTaylorMullen:
:shipit:

dougbu:
1111405

Issue #183, opened 2014-10-08
Attribute Parser Error


MohammadYounes:
I'm developing an Angular app and using Razor as my view engine, I had a case where I needed to use `ng-pattern` attribute. 

Example:
```html
<input type="text" ng-pattern="/^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@@[a-z0-9]([a-z0-9-]*[a-z0-9])?\.([a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i" />
``` 
Although I'm escaping `@`, the above gives the following error:

```bash
Parser Error Message: "[" is not valid at the start of a code block.  Only identifiers, keywords, comments, "(" and "{" are valid.
```

I was able to work around this using `data-ng-pattern`, but shouldn't this be allowed by Razor ?

yishaigalatzer:
You can escape this way

```html
    <input type="text" ng-pattern="/^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+@("@")[a-z0-9]([a-z0-9-]*[a-z0-9])?\.([a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i" />
```

But yes we are going to try and fix it.

Eilon:
@ajaybhargavb good luck :smile: Please chat with @NTaylorMullen for context.

PR #371, opened 2015-05-01
Fix error with double transition in attribute value

ajaybhargavb:
Issue - #183 
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -316,7 +316,7 @@ namespace Microsoft.AspNet.Razor.Parser
                 var matched = RemoveTag(tags, tagName, tagStart);
 
                 if (tags.Count == 0 &&
-                    // Note tagName may contain a '!' escape character. This ensures </!text> doesn't match here. 
+                    // Note tagName may contain a '!' escape character. This ensures </!text> doesn't match here.
                     // </!text> tags are treated like any other escaped HTML end tag.
                     string.Equals(tagName, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase) &&
                     matched)
@@ -491,7 +491,7 @@ namespace Microsoft.AspNet.Razor.Parser
         private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<HtmlSymbol> nameSymbols)
         {
             // First, determine if this is a 'data-' attribute (since those can't use conditional attributes)
-            LocationTagged<string> name = nameSymbols.GetContent(Span.Start);
+            var name = nameSymbols.GetContent(Span.Start);
             var attributeCanBeConditional = !name.Value.StartsWith("data-", StringComparison.OrdinalIgnoreCase);
 
             // Accept the whitespace and name
@@ -507,7 +507,7 @@ namespace Microsoft.AspNet.Razor.Parser
             }
 
             // We now have the prefix: (i.e. '      foo="')
-            LocationTagged<string> prefix = Span.GetContent();
+            var prefix = Span.GetContent();
 
             if (attributeCanBeConditional)
             {
@@ -521,7 +521,7 @@ namespace Microsoft.AspNet.Razor.Parser
                 }
 
                 // Capture the suffix
-                LocationTagged<string> suffix = new LocationTagged<string>(string.Empty, CurrentLocation);
+                var suffix = new LocationTagged<string>(string.Empty, CurrentLocation);
                 if (quote != HtmlSymbolType.Unknown && At(quote))
                 {
                     suffix = CurrentSymbol.GetContent();
@@ -554,23 +554,45 @@ namespace Microsoft.AspNet.Razor.Parser
         {
             var prefixStart = CurrentLocation;
             var prefix = ReadWhile(sym => sym.Type == HtmlSymbolType.WhiteSpace || sym.Type == HtmlSymbolType.NewLine);
-            Accept(prefix);
 
             if (At(HtmlSymbolType.Transition))
             {
-                var valueStart = CurrentLocation;
-                PutCurrentBack();
-
-                // Output the prefix but as a null-span. DynamicAttributeBlockCodeGenerator will render it
-                Span.CodeGenerator = SpanCodeGenerator.Null;
-
-                // Dynamic value, start a new block and set the code generator
-                using (Context.StartBlock(BlockType.Markup))
+                if (NextIs(HtmlSymbolType.Transition))
                 {
-                    Context.CurrentBlock.CodeGenerator =
-                        new DynamicAttributeBlockCodeGenerator(prefix.GetContent(prefixStart), valueStart);
+                    // Wrapping this in a block so that the ConditionalAttributeCollapser doesn't rewrite it.
+                    using (Context.StartBlock(BlockType.Markup))
+                    {
+                        Accept(prefix);
 
-                    OtherParserBlock();
+                        // Render a single "@" in place of "@@".
+                        Span.CodeGenerator = new LiteralAttributeCodeGenerator(
+                            prefix.GetContent(prefixStart),
+                            new LocationTagged<string>(CurrentSymbol.GetContent(), CurrentLocation));
+                        AcceptAndMoveNext();
+                        Output(SpanKind.Markup, AcceptedCharacters.None);
+
+                        Span.CodeGenerator = SpanCodeGenerator.Null;
+                        AcceptAndMoveNext();
+                        Output(SpanKind.Markup, AcceptedCharacters.None);
+                    }
+                }
+                else
+                {
+                    Accept(prefix);
+                    var valueStart = CurrentLocation;
+                    PutCurrentBack();
+
+                    // Output the prefix but as a null-span. DynamicAttributeBlockCodeGenerator will render it
+                    Span.CodeGenerator = SpanCodeGenerator.Null;
+
+                    // Dynamic value, start a new block and set the code generator
+                    using (Context.StartBlock(BlockType.Markup))
+                    {
+                        Context.CurrentBlock.CodeGenerator =
+                            new DynamicAttributeBlockCodeGenerator(prefix.GetContent(prefixStart), valueStart);
+
+                        OtherParserBlock();
+                    }
                 }
             }
             else if (At(HtmlSymbolType.Text) &&
@@ -578,6 +600,8 @@ namespace Microsoft.AspNet.Razor.Parser
                      CurrentSymbol.Content[0] == '~' &&
                      NextIs(HtmlSymbolType.ForwardSlash))
             {
+                Accept(prefix);
+
                 // Virtual Path value
                 var valueStart = CurrentLocation;
                 VirtualPath();
@@ -587,6 +611,8 @@ namespace Microsoft.AspNet.Razor.Parser
             }
             else
             {
+                Accept(prefix);
+
                 // Literal value
                 // 'quote' should be "Unknown" if not quoted and symbols coming from the tokenizer should never have "Unknown" type.
                 var value = ReadWhile(sym =>
@@ -719,7 +745,7 @@ namespace Microsoft.AspNet.Razor.Parser
             Tuple<HtmlSymbol, SourceLocation> tag = Tuple.Create(tagName, _lastTagStart);
 
             if (tags.Count == 0 &&
-                // Note tagName may contain a '!' escape character. This ensures <!text> doesn't match here. 
+                // Note tagName may contain a '!' escape character. This ensures <!text> doesn't match here.
                 // <!text> tags are treated like any other escaped HTML start tag.
                 string.Equals(tag.Item1.Content, SyntaxConstants.TextTagName, StringComparison.OrdinalIgnoreCase))
             {
--- a/test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
@@ -277,15 +277,14 @@ namespace Microsoft.AspNet.Razor.Test.Framework
             // Evaluate the result
             var collector = new ErrorCollector();
 
-            // Link all the nodes
-            expectedRoot.LinkNodes();
-
             if (expectedRoot == null)
             {
                 Assert.Null(actualRoot);
             }
             else
             {
+                // Link all the nodes
+                expectedRoot.LinkNodes();
                 Assert.NotNull(actualRoot);
                 EvaluateSyntaxTreeNode(collector, actualRoot, expectedRoot);
                 if (collector.Success)
@@ -412,7 +411,7 @@ namespace Microsoft.AspNet.Razor.Test.Framework
                         actual.TagName,
                         actual.SelfClosing);
                 }
-                
+
                 var expectedAttributes = expected.Attributes.GetEnumerator();
                 var actualAttributes = actual.Attributes.GetEnumerator();
 
--- a/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNet.Razor.Parser;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Test.Framework;
 using Microsoft.AspNet.Razor.Text;
+using Microsoft.AspNet.Razor.Tokenizer;
 using Microsoft.AspNet.Razor.Tokenizer.Symbols;
 using Xunit;
 
@@ -704,8 +705,8 @@ catch(bar) { baz(); }", BlockType.Statement, SpanKind.Code);
                                     Factory.Markup("<a"),
                                     new MarkupBlock(
                                         new AttributeBlockCodeGenerator(
-                                            "href", 
-                                            new LocationTagged<string>(" href=\"", 183 + Environment.NewLine.Length * 5, 5, 30), 
+                                            "href",
+                                            new LocationTagged<string>(" href=\"", 183 + Environment.NewLine.Length * 5, 5, 30),
                                             new LocationTagged<string>("\"", 246 + Environment.NewLine.Length * 5, 5, 93)),
                                         Factory.Markup(" href=\"").With(SpanCodeGenerator.Null),
                                         new MarkupBlock(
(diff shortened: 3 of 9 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -558,19 +558,33 @@ private void AttributeValue(HtmlSymbolType quote)
 
             if (At(HtmlSymbolType.Transition))
             {
-                var valueStart = CurrentLocation;
-                PutCurrentBack();
+                if (NextIs(HtmlSymbolType.Transition))
+                {
+                    // Render a single "@" in place of "@@".
+                    var transition = CurrentSymbol;
```
NTaylorMullen: Change this and the next few bits to be:
```C#
Output(SpanKind.Markup);
AcceptAndMoveNext();
Span.CodeGenerator = SpanCodeGenerator.Null;
Output(SpanKind.Markup);
AcceptAndMoveNext();
```

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
                     Factory.Code("        }").AsStatement().Accepts(AcceptedCharacters.None)));
         }
 
+        public static TheoryData ParseBlockData
```
NTaylorMullen: More specific name. `ParseBlockData` is to generic, you're testing the escaping of `@` symbols in attribute values (inside C# blocks).

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
...
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+
+                return new TheoryData<string, Block>
+                {
+                    {
+                        "{<span foo='@@' />}",
```
NTaylorMullen: Build a helper method member (Func) that builds out the statement block so you don't have to repeat the boiler plate over n over.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
...
+                                new MarkupTagBlock(
+                                    factory.Markup("<span"),
+                                    new MarkupBlock(
+                                        new AttributeBlockCodeGenerator("foo", new LocationTagged<string>(" foo='", 6, 0, 6), new LocationTagged<string>("'", 27, 0, 27)),
+                                        factory.Markup(" foo='").With(SpanCodeGenerator.Null),
+                                        factory.Markup("@").With(SpanCodeGenerator.Null),
+                                        factory.Markup("@"),
+                                        new MarkupBlock(
+                                            new DynamicAttributeBlockCodeGenerator(new LocationTagged<string>(string.Empty, 14, 0, 14), 14, 0, 14),
+                                            new ExpressionBlock(
```
NTaylorMullen: Save this value so you don't need to repeat the boiler plate DateTime.Now over n over

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Hmm I think this is an outstanding issue but I believe if you type `@@` anywhere in the document and you continue typing, such as: `@@` => `@DateT@` I don't belive you'll get IntelliSense because the `AcceptedCharacters` is `Any`.

To do this right from an editor perspective we should change the `AcceptedCharacters` for both `@` symbols to be `AcceptedCharacters.None`.

Aka:
```C#
Output(SpanKind.Markup);
AcceptAndMoveNext();
Span.CodeGenerator = SpanCodeGenerator.Null;
Output(SpanKind.Markup, AcceptedCharacters.None);
AcceptAndMoveNext();
Output(SpanKind.Markup, AcceptedCharacters.None);
```

This way we're outputting the second `@` symbol early with a different `AcceptedCharacters` value.  Feel free to update the other document level escaping `@` bits or file an issue.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
...
+                                                factory.Code("DateTime.Now")
+                                                    .AsImplicitExpression(CSharpCodeParser.DefaultKeywords)
+                                                    .Accepts(AcceptedCharacters.NonWhiteSpace))),
+                                        factory.Markup("'").With(SpanCodeGenerator.Null)),
+                                factory.Markup(" />").Accepts(AcceptedCharacters.None))),
+                            factory.EmptyCSharp().AsStatement(),
+                            factory.MetaCode("}").Accepts(AcceptedCharacters.None))
+                    },
+                    {
+                        "{<span foo='@DateTime.Now @@' />}",
```
NTaylorMullen: Add another variation of this test case but without the space between `.Now` and `@@` (do same for document tests)

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
...
+                                        factory.Markup(@"/^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+").With(new LiteralAttributeCodeGenerator(new LocationTagged<string>(string.Empty, 12, 0, 12), new LocationTagged<string>(@"/^[a-z0-9!#$%&'*+\/=?^_`{|}~.-]+", 12, 0, 12))),
+                                        factory.Markup("@").With(SpanCodeGenerator.Null),
+                                        factory.Markup("@"),
+                                        factory.Markup(@"[a-z0-9]([a-z0-9-]*[a-z0-9])?\.([a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i").With(new LiteralAttributeCodeGenerator(new LocationTagged<string>(string.Empty, 46, 0, 46), new LocationTagged<string>(@"[a-z0-9]([a-z0-9-]*[a-z0-9])?\.([a-z0-9]([a-z0-9-]*[a-z0-9])?)*$/i", 46, 0, 46))),
+                                        factory.Markup("\"").With(SpanCodeGenerator.Null)),
+                                factory.Markup(" />").Accepts(AcceptedCharacters.None))),
+                            factory.EmptyCSharp().AsStatement(),
+                            factory.MetaCode("}").Accepts(AcceptedCharacters.None))
+                    },
+                };
```
NTaylorMullen: Add a few tests like you have with `@DateTime.Now` but with the `@(` syntax. Mix in some `@@@(` cases in that too.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,231 @@ public void ParsersCanNestRecursively()
...
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(ParseBlockData))]
+        public void ParseBlock_WithDoubleTransition_DoesNotThrow(string input, Block expected)
+        {
+            ParseBlockTest(input, expected);
```
NTaylorMullen: `// Arrange` above

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpSectionTest.cs
```diff
@@ -443,5 +445,83 @@ public void ParseSectionBlockParsesXmlProcessingInstruction()
                         Factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
                     Factory.EmptyHtml()));
         }
+
+        public static TheoryData ParseBlockData
```
NTaylorMullen: More specific name.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpTemplateTest.cs
```diff
@@ -272,6 +274,50 @@ public void ParseBlockProducesErrorButCorrectlyParsesNestedTemplateInStatementWi
                            GetNestedTemplateError(69));
         }
 
+        private const string TestTemplateWithDoubleTransitionCode = " @<p foo='@@'>Foo #@item</p>";
```
NTaylorMullen: Just inline this in the test

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -281,5 +282,190 @@ public void ParseBlockCanParse1000NestedElements()
             var content = Nested1000.ReadAllText();
             ParseDocument(content);
         }
+
+        public static TheoryData ParseBlockData
```
NTaylorMullen: Apply comments to the CSharpBlock test bits to this file.

NTaylorMullen:
Just thought of another case we should verify: `TagHelper`s whos attributes have escaped `@` symbols.

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
ajaybhargavb: Added.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
ajaybhargavb: Added.

ajaybhargavb:
@NTaylorMullen, Is `TagHelperParseTreeRewriterTest` the right place to add the tag helper tests?

NTaylorMullen:
@ajaybhargavb yup

NTaylorMullen:
@ajaybhargavb Looks like you have some travis-ci failures. Dev isn't failing so something in this PR introduced it.

NTaylorMullen:
Latest iteration looks pretty solid. :watch: for `TagHelper` tests and travis-ci fix.

ajaybhargavb:
Updated. The Travis failure is caused by some Mono bug mentioned here https://github.com/aspnet/dnx/issues/1761.
cc' @pranavkm

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -558,19 +558,32 @@ private void AttributeValue(HtmlSymbolType quote)
...
-                PutCurrentBack();
+                if (NextIs(HtmlSymbolType.Transition))
+                {
+                    // Render a single "@" in place of "@@".
+                    Output(SpanKind.Markup);
+                    AcceptAndMoveNext();
+                    Span.CodeGenerator = SpanCodeGenerator.Null;
+                    Output(SpanKind.Markup, AcceptedCharacters.None);
+                    AcceptAndMoveNext();
+                    Output(SpanKind.Markup, AcceptedCharacters.None);
```
dougbu: - Looks like this is the only significant code change.  I understand #183 was about a parsing error in the context of an attribute value.  Have you confirmed no similar issues exist elsewhere e.g. in strings that are part of directives?  My question isn't a suggestion for additional changes.  Instead I'm hoping we know whether or not there's a cluster of related bugs.
- Unclear why this block calls `Output()` three (four counting the final call in the method) times for two characters.  See the final `Output()` at the bottom of the method is redundant (two successive calls do nothing).  Suspect you need a comment above the first `Output()` mentioning that it outputs the `prefix` (whitespace).

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -277,15 +277,14 @@ public static void EvaluateParseTree(Block actualRoot, Block expectedRoot)
...
-            expectedRoot.LinkNodes();
-
             if (expectedRoot == null)
             {
                 Assert.Null(actualRoot);
             }
             else
             {
+                // Link all the nodes
+                expectedRoot.LinkNodes();
```
dougbu: - curious.  why did this line need to move?
- add a blank line before the `Assert.NotNull()` because that isn't part of linking the nodes.

ajaybhargavb: Its totally unrelated to this change. But the code didn't make sense since the null check for `expectedRoot` comes after `expectedRoot.LinkNodes` method call.
Will add the blank line.

dougbu:
> should verify: `TagHelpers` whos attributes have escaped `@` symbols.

Looks like there are a few such tests now. But the count is low compared to HTML (unbound) attributes elsewhere. Need to cover at least:
- bound `string` attributes (attributes that will be assigned to `string` properties)
- bound non-`string` attributes
- unbound HTML attributes (attributes in elements a tag helper targets that are not themselves bound)
- some of the variations tested in non-tag helper scenarios; just a few

BTW the test coverage in non-tag helper scenarios is, if anything, too much -- hard to see what the variations are.  Suggest a couple of code comments might help to make it clear what's changing.  Is it for example beginning / middle / end / all of the attribute value, whether value is quoted, ...?

dougbu:
:watch:

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -3697,6 +3773,73 @@ public static TheoryData CodeTagHelperAttributesData
...
+                                        "birthday",
+                                        factory.CodeMarkup("DateTime.Now")),
+                                    new KeyValuePair<string, SyntaxTreeNode>(
+                                        "name",
+                                        new MarkupBlock(
+                                            factory.Markup("Time:"),
+                                            factory.Markup(" "),
+                                            factory.Markup("@").With(SpanCodeGenerator.Null).Accepts(AcceptedCharacters.None),
+                                            factory.Markup("@").Accepts(AcceptedCharacters.None),
+                                            dateTimeNow))
```
NTaylorMullen: Might be a bug, where's the space inbetween `@@` and `@DateTime.Now`

NTaylorMullen: Nm, ignore, it's in the `dateTimeNow`

NTaylorMullen:
Looks good to me. One last q: Have you put the binaries in your VS to validate the editing experience? Once you do that and all checks out :shipit:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: let's close on these points first thing Monday

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
dougbu: :+1:

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -281,5 +282,251 @@ public void ParseBlockCanParse1000NestedElements()
...
+                var datetimeBlock = new ExpressionBlock(
+                    factory.CodeTransition(),
+                    factory.Code("DateTime.Now")
+                        .AsImplicitExpression(CSharpCodeParser.DefaultKeywords)
+                        .Accepts(AcceptedCharacters.NonWhiteSpace));
+
+                return new TheoryData<string, Block>
+                {
+                    {
+                        "<span foo='@@' />",
```
dougbu: please add comments here, similar to what you've done in `CSharpBlockTest`

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -736,6 +737,275 @@ public void ParsersCanNestRecursively()
...
+                var datetimeBlock = new ExpressionBlock(
+                    factory.CodeTransition(),
+                    factory.Code("DateTime.Now")
+                        .AsImplicitExpression(CSharpCodeParser.DefaultKeywords)
+                        .Accepts(AcceptedCharacters.NonWhiteSpace));
+
+                return new TheoryData<string, Block>
+                {
+                    {
+                        // Double transition in attribute value
```
dougbu: Good coverage here and in `HtmlDocumentTest`. About the only thing we _might_ want in addition would be `<span foo='@ @' />`.  Closest I see is two comments with a space between them but I may be missing an almost-double-transition case elsewhere.

ajaybhargavb: Added that case.

ajaybhargavb:
Updated.

dougbu:
@ajaybhargavb test coverage looks great and you're very close to :shipit:  Just need to close on the open questions I asked in https://github.com/aspnet/Razor/pull/371#discussion_r29725843

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: @dougbu, I tried looking at the places we special case transitions in [`CSharpCodeParser.Statements`](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs#L493), [`CSharpCodeParser`](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs#L162) and [`HtmlMarkupParser`](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs#L86) and I could not find any similar issue.

dougbu: Based on offline conversation, you've addressed both of my bullets with this comment and your most recent changes.

dougbu:
:shipit:

ajaybhargavb:
Updated. @dougbu @NTaylorMullen

NTaylorMullen:
:watch:  for the `ConditionalAttributeCollapser` test.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -558,19 +558,37 @@ private void AttributeValue(HtmlSymbolType quote)
...
-                var valueStart = CurrentLocation;
-                PutCurrentBack();
+                if (NextIs(HtmlSymbolType.Transition))
+                {
+                    // Render a single "@" in place of "@@".
+                    using (Context.StartBlock(BlockType.Markup))
+                    {
+                        Span.CodeGenerator = new LiteralAttributeCodeGenerator(
+                            prefix.GetContent(prefixStart),
+                            new LocationTagged<string>(CurrentSymbol.GetContent(), CurrentLocation));
```
dougbu: Does it matter what code generator is used for the `prefix`?  More to the point, any downside of using `LiteralAttributeCodeGenerator` for that (in addition to the literal `@`?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -558,19 +558,37 @@ private void AttributeValue(HtmlSymbolType quote)
...
+                        AcceptAndMoveNext();
+                        Output(SpanKind.Markup, AcceptedCharacters.None);
+                        Span.CodeGenerator = SpanCodeGenerator.Null;
```
dougbu: Add a blank line above this.  Accepting and ignoring the second `@` should be more clearly separate.

dougbu:
> put the binaries in your VS to validate the editing experience?

Does this version work much better when tested as @NTaylorMullen describes?

Do we need another test case for one of the scenarios where you saw issues in VS last time?  Guess it might be enough that the expected blocks and code generators appear in the tree...

dougbu:
:watch:

ajaybhargavb:
@dougbu, the issue last time was not when editing but in runtime where the rewriter messed with the generated code. This change fixed it.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: Moved `Accept(prefix)` inside the Markup block.

ajaybhargavb:
Updated.

NTaylorMullen:
:shipit:

dougbu:
:shipit:

Issue #220, opened 2014-11-04
TagHelper attributes with no value cannot be used in TagHelpers


NTaylorMullen:
Razor does not correctly parse HTML attributes that have no value.

After playing around with the parse tree a bit I noticed that if you were to create HTML elements such as:
```
<input checked>
<input checked value>
```

The Razor parser does not generate an accurate syntax tree. This in turn results in the generated code not calling `WriteAttribute` and furthermore with `TagHelper`s does not allow us to pull the attributes from an HTML element easily.

Once this has been added we should also support this attribute syntax in Tag Helpers.

dougbu:
when this bug is fixed, will need to modify the expectations in https://github.com/aspnet/Mvc/tree/dev/test/Microsoft.AspNet.Mvc.FunctionalTests/Compiler/Resources files as well as one HTML helper test file (see https://github.com/aspnet/Mvc/blob/dev/test/WebSites/MvcTagHelpersWebSite/Views/MvcTagHelper_Home/OrderUsingHtmlHelpers.cshtml#L72) that mimics current tag helper results

NTaylorMullen:
With issue #289 all `TagHelper` bound attributes that are not typed `System.String` without a value will be invalid and cause a parse error. Do we want to do something special with say `bool` typed attributes, or any other types?

/cc @DamianEdwards

DamianEdwards:
I think we just shouldn't support bound properties that don't have a value specified, i.e. parser error.

ChrisKlug:
Just ran into this bug/feature/"possibility" and it is quite annoying. Not being able to add a TagHelper that handles the required attribute without having to add a value to the attribute is quite annoying and confusing. I can see this causing a heap of problems...

http://www.w3.org/html/wg/drafts/html/master/syntax.html#attribute-name-state, which is referenced in the code states that a space after the attribute name should move the parsing to the "after attribute name state", which says that if a character is found, it should go to the "attribute name state". 

Doesn't this mean that <input type="text" required /> should see the "required" as its own attribute without a value? Currently, the "required" attribute is combined with "/>" into "runtime />"...

HtmlMarkupParser.cs line 476 has a comment that says "Saw a space or newline after the name, so just skip this attribute and continue around the loop", which doesn't respect the w3.org spec in my opinion. It might not be XHTML or something, but should that really be a concern if HTML5 is what everyone is using, or should be using?

Sorry for long rant...

NTaylorMullen:
[...]

PR #372, opened 2015-05-01
Add support for minimized attributes in TagHelpers.

NTaylorMullen:
- Updated the Razor parser to understand minimized attributes instead of just treating them like plain text. This just involved encompassing minimized attributes in their own blocks just like the other attributes found on the HTML tag.
- Updated TagHelperParseTreeRewriter to only accept minimized attributes for unbound attributes.
- Updated IReadOnlyTagHelperAttribute/TagHelperAttribute to have a Minimized property to indicate that an attribute was minimized.
- Updated parser level block structures to represent minimized attributes as null syntax tree nodes.
- Updated chunk level structures to represent minimized attributes as null chunks.
- Added parse level rewriting tests to validate new TagHelper rewritten structures for minimized attributes.
- Updated existing parser tests to understand minimized attributes.
- Added codegen test to validate understanding of minimized attributes.
- Added TagHelperExecutionContext tests to validate maintaining of runtime TagHelperOutput tests.
- Refactored part of the TagHelperParseTreeRewriterTest file into a base class file so we can make better rewriting tests.

#220

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IReadOnlyTagHelperAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IReadOnlyTagHelperAttribute.cs
@@ -19,5 +19,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Gets the value of the attribute.
         /// </summary>
         object Value { get; }
+
+        /// <summary>
+        /// Gets an indication whether the attribute is minimized or not.
+        /// </summary>
+        /// <remarks>If <c>true</c>, <see cref="Value"/> will be ignored.</remarks>
+        bool Minimized { get; }
     }
 }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
@@ -42,6 +42,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         public object Value { get; set; }
 
         /// <summary>
+        /// Gets or sets an indication whether the attribute is minimized or not.
+        /// </summary>
+        /// <remarks>If <c>true</c>, <see cref="Value"/> will be ignored.</remarks>
+        public bool Minimized { get; set; }
+
+        /// <summary>
         /// Converts the specified <paramref name="value"/> into a <see cref="TagHelperAttribute"/>.
         /// </summary>
         /// <param name="value">The <see cref="Value"/> of the created <see cref="TagHelperAttribute"/>.</param>
@@ -61,7 +67,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             return
                 other != null &&
                 string.Equals(Name, other.Name, StringComparison.OrdinalIgnoreCase) &&
-                Equals(Value, other.Value);
+                Minimized == other.Minimized &&
+                (Minimized || Equals(Value, other.Value));
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
         /// <param name="selfClosing">
         /// <see cref="bool"/> indicating whether or not the tag in the Razor source was self-closing.</param>
-        /// <param name="items">The collection of items used to communicate with other 
+        /// <param name="items">The collection of items used to communicate with other
         /// <see cref="ITagHelper"/>s</param>
         /// <param name="uniqueId">An identifier unique to the HTML element this context is for.</param>
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
@@ -134,6 +134,26 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
+        /// Tracks the minimized HTML attribute in <see cref="AllAttributes"/> and <see cref="HTMLAttributes"/>.
+        /// </summary>
+        /// <param name="name">The minimized HTML attribute name.</param>
+        public void AddMinimizedHtmlAttribute([NotNull] string name)
+        {
+            HTMLAttributes.Add(
+                new TagHelperAttribute
+                {
+                    Name = name,
+                    Minimized = true
+                });
+            AllAttributes.Add(
+                new TagHelperAttribute
+                {
+                    Name = name,
+                    Minimized = true
+                });
+        }
+
+        /// <summary>
         /// Tracks the HTML attribute in <see cref="AllAttributes"/> and <see cref="HTMLAttributes"/>.
         /// </summary>
         /// <param name="name">The HTML attribute name.</param>
@@ -168,7 +188,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// </summary>
         /// <returns>A <see cref="Task"/> that on completion returns the rendered child content.</returns>
         /// <remarks>
-        /// Child content is only executed once. Successive calls to this method or successive executions of the 
+        /// Child content is only executed once. Successive calls to this method or successive executions of the
         /// returned <see cref="Task{TagHelperContent}"/> return a cached result.
         /// </remarks>
         public async Task<TagHelperContent> GetChildContentAsync()
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
@@ -203,6 +203,13 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
                     var firstAttribute = matchingAttributes.First();
                     var attributeValueChunk = firstAttribute.Value;
 
+                    // Minimized attributes are not valid for bound attributes. There will be an error for the bound
+                    // attribute logged by TagHelperBlockRewriter already so we can skip.
+                    if (attributeValueChunk == null)
+                    {
+                        continue;
+                    }
+
                     var attributeValueRecorded = htmlAttributeValues.ContainsKey(attributeDescriptor.Name);
 
                     // Bufferable attributes are attributes that can have Razor code inside of them.
@@ -323,15 +330,21 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
             // Build out the unbound HTML attributes for the tag builder
             foreach (var htmlAttribute in unboundHTMLAttributes)
             {
-                string textValue;
+                string textValue = null;
+                var isPlainTextValue = false;
                 var attributeValue = htmlAttribute.Value;
-                var isPlainTextValue = TryGetPlainTextValue(attributeValue, out textValue);
 
-                // HTML attributes are always strings. So if this value is not plain text i.e. if the value contains
-                // C# code, then we need to buffer it.
-                if (!isPlainTextValue)
+                // A null attribute value means the HTML attribute is minimized.
+                if (attributeValue != null)
                 {
-                    BuildBufferedWritingScope(attributeValue, htmlEncodeValues: true);
+                    isPlainTextValue = TryGetPlainTextValue(attributeValue, out textValue);
+
+                    // HTML attributes are always strings. So if this value is not plain text i.e. if the value contains
+                    // C# code, then we need to buffer it.
+                    if (!isPlainTextValue)
+                    {
+                        BuildBufferedWritingScope(attributeValue, htmlEncodeValues: true);
+                    }
                 }
 
                 // Execution contexts are a runtime feature, therefore no need to add anything to them.
@@ -340,27 +353,39 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
                     continue;
                 }
 
-                _writer
-                    .WriteStartInstanceMethodInvocation(
-                        ExecutionContextVariableName,
-                        _tagHelperContext.ExecutionContextAddHtmlAttributeMethodName)
-                    .WriteStringLiteral(htmlAttribute.Key)
-                    .WriteParameterSeparator()
-                    .WriteStartMethodInvocation(_tagHelperContext.MarkAsHtmlEncodedMethodName);
-
-                // If it's a plain text value then we need to surround the value with quotes.
-                if (isPlainTextValue)
+                // If we have a minimized attribute there is no value
+                if (attributeValue == null)
                 {
-                    _writer.WriteStringLiteral(textValue);
+                    _writer
+                        .WriteStartInstanceMethodInvocation(
+                            ExecutionContextVariableName,
+                            _tagHelperContext.ExecutionContextAddMinimizedHtmlAttributeMethodName)
+                        .WriteStringLiteral(htmlAttribute.Key)
+                        .WriteEndMethodInvocation();
                 }
                 else
                 {
-                    RenderBufferedAttributeValueAccessor(_writer);
-                }
+                    _writer
+                        .WriteStartInstanceMethodInvocation(
+                            ExecutionContextVariableName,
+                            _tagHelperContext.ExecutionContextAddHtmlAttributeMethodName)
+                        .WriteStringLiteral(htmlAttribute.Key)
+                        .WriteParameterSeparator()
+                        .WriteStartMethodInvocation(_tagHelperContext.MarkAsHtmlEncodedMethodName);
 
-                _writer
-                    .WriteEndMethodInvocation(endLine: false)
-                    .WriteEndMethodInvocation();
+                    // If it's a plain text value then we need to surround the value with quotes.
+                    if (isPlainTextValue)
+                    {
+                        _writer.WriteStringLiteral(textValue);
+                    }
+                    else
+                    {
+                        RenderBufferedAttributeValueAccessor(_writer);
+                    }
+
+                    _writer.WriteEndMethodInvocation(endLine: false)
+                        .WriteEndMethodInvocation();
+                }
             }
         }
 
--- a/src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
@@ -19,6 +19,7 @@ namespace Microsoft.AspNet.Razor.Generator
             ScopeManagerEndMethodName = "End";
             ExecutionContextAddMethodName = "Add";
             ExecutionContextAddTagHelperAttributeMethodName = "AddTagHelperAttribute";
+            ExecutionContextAddMinimizedHtmlAttributeMethodName = "AddMinimizedHtmlAttribute";
             ExecutionContextAddHtmlAttributeMethodName = "AddHtmlAttribute";
             ExecutionContextOutputPropertyName = "Output";
             MarkAsHtmlEncodedMethodName = "Html.Raw";
@@ -56,6 +57,11 @@ namespace Microsoft.AspNet.Razor.Generator
         /// The name of the <see cref="ExecutionContextTypeName"/> method used to add tag helper attributes.
         /// </summary>
         public string ExecutionContextAddTagHelperAttributeMethodName { get; set; }
+
+        /// <summary>
+        /// The name of the <see cref="ExecutionContextTypeName"/> method used to add minimized HTML attributes.
+        /// </summary>
+        public string ExecutionContextAddMinimizedHtmlAttributeMethodName { get; set; }
 
         /// <summary>
         /// The name of the <see cref="ExecutionContextTypeName"/> method used to add HTML attributes.
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -473,12 +473,25 @@ namespace Microsoft.AspNet.Razor.Parser
 
             if (!At(HtmlSymbolType.Equals))
             {
-                // Saw a space or newline after the name, so just skip this attribute and continue around the loop
-                Accept(whitespace);
-                Accept(name);
+                // Minimized attribute
+
+                // Output anything prior to the attribute, in most cases this will be the tag name:
+                // |<input| checked />. If in-between other attributes this will noop or output malformed attribute
+                // content (if the previous attribute was malformed).
+                Output(SpanKind.Markup);
+
+                using (Context.StartBlock(BlockType.Markup))
+                {
+                    Accept(whitespace);
+                    Accept(name);
+                    Output(SpanKind.Markup);
+                }
+
                 return;
             }
 
+            // Not a minimized attribute, parse as if it were well-formed (if attribute turns out to be malformed we
+            // will go into recovery).
             Output(SpanKind.Markup);
 
             // Start a new markup block for the attribute
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) .NET Foundation. All rights reserved.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -71,11 +71,12 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 // Only want to track the attribute if we succeeded in parsing its corresponding Block/Span.
                 if (succeeded)
                 {
-                    // Check if it's a bound attribute that is not of type string and happens to be null or whitespace.
+                    // Check if it's a bound attribute that is minimized or not of type string and null or whitespace.
                     string attributeValueType;
                     if (attributeValueTypes.TryGetValue(attribute.Key, out attributeValueType) &&
+                        (attribute.Value == null ||
                         !IsStringAttribute(attributeValueType) &&
-                        IsNullOrWhitespaceAttributeValue(attribute.Value))
+                        IsNullOrWhitespaceAttributeValue(attribute.Value)))
                     {
                         var errorLocation = GetAttributeNameStartLocation(child);
 
@@ -167,7 +168,7 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 {
                     Debug.Assert(
                         name != null,
-                        "Name should never be null here. The parser should guaruntee an attribute has a name.");
+                        "Name should never be null here. The parser should guarantee an attribute has a name.");
 
                     // We've captured all leading whitespace and the attribute name.
                     // We're now at: " asp-for|='...'" or " asp-for|=..."
@@ -239,7 +240,9 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                 return false;
             }
 
-            attribute = CreateMarkupAttribute(name, builder, attributeValueTypes);
+            // If we're not after an equal then we should treat the value as if it were a minimized attribute.
+            var attributeValueBuilder = afterEquals ? builder : null;
+            attribute = CreateMarkupAttribute(name, attributeValueBuilder, attributeValueTypes);
 
             return true;
         }
@@ -432,17 +435,24 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
             IReadOnlyDictionary<string, string> attributeValueTypes)
         {
             string attributeTypeName;
+            Span value = null;
 
-            // If the attribute was requested by the tag helper and doesn't happen to be a string then we need to treat
-            // its value as code. Any non-string value can be any C# value so we need to ensure the SyntaxTreeNode
-            // reflects that.
-            if (attributeValueTypes.TryGetValue(name, out attributeTypeName) &&
-                !IsStringAttribute(attributeTypeName))
+            // Builder will be null in the case of minimized attributes
+            if (builder != null)
             {
-                builder.Kind = SpanKind.Code;
+                // If the attribute was requested by the tag helper and doesn't happen to be a string then we need to treat
+                // its value as code. Any non-string value can be any C# value so we need to ensure the SyntaxTreeNode
+                // reflects that.
+                if (attributeValueTypes.TryGetValue(name, out attributeTypeName) &&
+                    !IsStringAttribute(attributeTypeName))
+                {
+                    builder.Kind = SpanKind.Code;
+                }
+
+                value = builder.Build();
             }
 
-            return new KeyValuePair<string, SyntaxTreeNode>(name, builder.Build());
+            return new KeyValuePair<string, SyntaxTreeNode>(name, value);
         }
 
         private static bool IsNullOrWhitespaceAttributeValue(SyntaxTreeNode attributeValue)
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using Microsoft.Internal.Web.Utils;
 
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
@@ -26,16 +25,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             // Normal comparer (TagHelperAttribute.Equals()) doesn't care about the Name case, in tests we do.
             return attributeX != null &&
                 string.Equals(attributeX.Name, attributeY.Name, StringComparison.Ordinal) &&
-                Equals(attributeX.Value, attributeY.Value);
+                attributeX.Minimized == attributeY.Minimized &&
+                (attributeX.Minimized || Equals(attributeX.Value, attributeY.Value));
         }
 
         public int GetHashCode(IReadOnlyTagHelperAttribute attribute)
         {
-            return HashCodeCombiner
-                .Start()
-                .Add(attribute.Name, StringComparer.Ordinal)
-                .Add(attribute.Value)
-                .CombinedHash;
+            return attribute.GetHashCode();
         }
     }
 }
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             // Arrange & Act
             var executionContext = new TagHelperExecutionContext("p", selfClosing);
 
-            // Assert 
+            // Assert
             Assert.Equal(selfClosing, executionContext.SelfClosing);
         }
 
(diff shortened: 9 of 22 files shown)


dougbu:
> Updated TagHelperParseTreeRewriter to only accept minimized attributes for unbound attributes.

why aren't minimized attributes handled?  bound and unbound attributes use the same `TagHelperAttribute` class.

bound value of a `string` property named `propertyName` with a corresponding minimized attribute in Razor source is `propertyName`.  bound value of a `bool` or `bool?` property with a corresponding minimized attribute in Razor source is `true`.  any other property type is an error.

so far, this seems like #220 part 1 of at least 2.

NTaylorMullen:
@dougbu https://github.com/aspnet/Razor/issues/220#issuecomment-74000770

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -134,6 +134,26 @@ public void Add([NotNull] ITagHelper tagHelper)
         }
 
         /// <summary>
+        /// Tracks the minimized HTML attribute in <see cref="AllAttributes"/> and <see cref="HTMLAttributes"/>.
+        /// </summary>
+        /// <param name="name">The minimized HTML attribute name.</param>
+        public void AddHtmlAttribute([NotNull] string name)
```
dougbu: this will confuse everyone.  suggest either leaving out this "helper" or giving it a name that hints at minimization.

NTaylorMullen: `TagHelperExecutionContext` isn't seen by users. It's only used by codegen.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -345,23 +351,28 @@ private void RenderUnboundHTMLAttributes(IEnumerable<KeyValuePair<string, Chunk>
...
                         ExecutionContextVariableName,
                         _tagHelperContext.ExecutionContextAddHtmlAttributeMethodName)
-                    .WriteStringLiteral(htmlAttribute.Key)
-                    .WriteParameterSeparator()
-                    .WriteStartMethodInvocation(_tagHelperContext.MarkAsHtmlEncodedMethodName);
+                    .WriteStringLiteral(htmlAttribute.Key);
 
-                // If it's a plain text value then we need to surround the value with quotes.
-                if (isPlainTextValue)
+                // If we have a minimized attribute we don't want to provide a value
```
dougbu: agree with describing why but "don't want" provides no information. there simply _is_ no value when `attributeValue == null`.  could repeat something about a minimized attribute (as on line 331) but that seems redundant.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,21 @@ private void BeforeAttribute()
                 return;
             }
 
+            // Minimized attribute
+            // Need to modify this code to account for space inbetween equals when
+            // https://github.com/aspnet/Razor/issues/123 is completed.
```
dougbu: this is a TODO and we should not have such in the code.  add the information linking to this line in #123

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,21 @@ private void BeforeAttribute()
...
             if (!At(HtmlSymbolType.Equals))
             {
-                // Saw a space or newline after the name, so just skip this attribute and continue around the loop
-                Accept(whitespace);
-                Accept(name);
+                // Output anything prior to the attribute, in most cases this will be the tag name:
+                // |<input| checked />
```
dougbu: confusing update that makes more sense given the previous comment at line 476.  why are you changing the code that parses whitespace before an attribute name?  what do the _two_ vertical bars in your example mean?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -239,7 +240,10 @@ private static bool IsSelfClosing(Block beginTagBlock)
                 return false;
             }
 
-            attribute = CreateMarkupAttribute(name, builder, attributeValueTypes);
+            // If we're not after an equal then we should treat the value as if it were a minimized attribute.
+            // Aka null value.
```
dougbu: remove this line. don't introduce an additional term for the same thing.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -431,17 +435,24 @@ private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
...
-                !IsStringAttribute(attributeTypeName))
+            // Builder will be null in the case of minimized attributes
+            if (builder != null)
             {
-                builder.Kind = SpanKind.Code;
+                // If the attribute was requested by the tag helper and doesn't happen to be a string then we need to treat
+                // its value as code. Any non-string value can be any C# value so we need to ensure the SyntaxTreeNode
+                // reflects that.
+                if (attributeValueTypes.TryGetValue(name, out attributeTypeName) &&
+                    !IsStringAttribute(attributeTypeName))
```
dougbu: nit: why is this line wrapped?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -70,7 +70,7 @@ private void RewriteTags(Block input, RewritingContext context)
                 // At this point the child is a Span or Block with Type BlockType.Tag that doesn't happen to be a
                 // tag helper.
 
-                // Add the child to current block. 
+                // Add the child to current block.
```
dougbu: I know I can ignore these changes. but you're making enough of them that they should be in a separate "cleanup trailing whitespace" PR that you commit immediately.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -42,6 +42,13 @@ public TagHelperAttribute(string name, object value)
         public object Value { get; set; }
 
         /// <summary>
+        /// Gets or sets the value indicating if the attribute is minimized.
```
dougbu: `Gets or sets an indication whether the attribute is minimized or not.` i.e. consistent with `IReadOnlyTagHelperAttribute` property.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IReadOnlyTagHelperAttribute.cs
```diff
@@ -19,5 +19,11 @@ public interface IReadOnlyTagHelperAttribute : IEquatable<IReadOnlyTagHelperAttr
         /// Gets the value of the attribute.
         /// </summary>
         object Value { get; }
+
+        /// <summary>
+        /// Indicates whether the attribute is minimized or not.
```
dougbu: `Gets an indication whether ...`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -22,16 +21,13 @@ public bool Equals(IReadOnlyTagHelperAttribute attributeX, IReadOnlyTagHelperAtt
...
         }
 
         public int GetHashCode(IReadOnlyTagHelperAttribute attribute)
         {
-            return HashCodeCombiner
-                .Start()
-                .Add(attribute.Name, StringComparer.Ordinal)
-                .Add(attribute.Value)
-                .CombinedHash;
+            return attribute.GetHashCode();
```
dougbu: :+1: though I don't see how this relates to the rest of this PR.  should at least be separately mentioned in the PR and commit descriptions.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -203,6 +203,54 @@ public void AddHtmlAttribute_MaintainsHTMLAttributes()
...
+
+            // Assert
+            Assert.Equal(
+                expectedAttributes,
+                executionContext.HTMLAttributes,
+                CaseSensitiveTagHelperAttributeComparer.Default);
+        }
+
+        [Fact]
+        public void AddHtmlAttribute_MaintainsHTMLAttributes_Mixed()
```
dougbu: "mixed" means nothing. suggest "SomeMinimized"

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -181,7 +181,7 @@ public void AllAttributes_IgnoresCase(string originalName, string updatedName)
         }
 
         [Fact]
-        public void AddHtmlAttribute_MaintainsHTMLAttributes()
+        public void AddHtmlAttribute_MaintainsHTMLAttributes_Unminimized()
```
dougbu: just churn. we know the base / normal case is not minimized.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -203,6 +203,54 @@ public void AddHtmlAttribute_MaintainsHTMLAttributes()
...
+                { "foo", "bar" }
+            };
+            expectedAttributes.Add(new TagHelperAttribute { Name = "checked", Minimized = true });
+            expectedAttributes.Add(new TagHelperAttribute { Name = "visible", Minimized = true });
+
+            // Act
+            executionContext.AddHtmlAttribute("class", "btn");
+            executionContext.AddHtmlAttribute("foo", "bar");
+            executionContext.AddHtmlAttribute("checked");
+            executionContext.AddHtmlAttribute("visible");
```
dougbu: as I mentioned in the product code, ick.  not at all obvious what this line does.

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -304,6 +304,7 @@ private static void EvaluateSyntaxTreeNode(ErrorCollector collector, SyntaxTreeN
             if (actual == null)
             {
                 AddNullActualError(collector, actual, expected);
+                return;
```
dougbu: why?  or, why did it previously work without this `return`?

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -336,7 +337,14 @@ private static void EvaluateSyntaxTreeNode(ErrorCollector collector, SyntaxTreeN
                 collector.AddMessage("{0} - PASSED :: Attribute names match", expected.Key);
             }
 
-            EvaluateSyntaxTreeNode(collector, actual.Value, expected.Value);
+            if (actual.Value == null && expected.Value == null)
+            {
+                collector.AddMessage("{0} - PASSED :: Attribute values match", expected.Key);
```
dougbu: make message specific to minimized case and don't mention "values" e.g. `Minimized attributes match`.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -19,6 +19,44 @@ private static IEnumerable<TagHelperDescriptor> DefaultPAndInputTagHelperDescrip
         private static IEnumerable<TagHelperDescriptor> PrefixedPAndInputTagHelperDescriptors
             => BuildPAndInputTagHelperDescriptors("THS");
 
+        private static IEnumerable<TagHelperDescriptor> MinimizedTagHelperDescriptors
```
dougbu: what does this name mean?  descriptors can't be minimized can they?

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -70,7 +70,11 @@ public void ParseDocumentCorrectlyHandlesOddlySpacedHTMLElements()
             ParseDocumentTest("<div ><p class = 'bar'> Foo </p></div >",
                 new MarkupBlock(
                     BlockFactory.MarkupTagBlock("<div >"),
-                    BlockFactory.MarkupTagBlock("<p class = 'bar'>"),
+                    new MarkupTagBlock(
+                        Factory.Markup("<p"),
+                        new MarkupBlock(
+                            Factory.Markup(" class")),
+                        Factory.Markup(" = 'bar'>")),
```
dougbu: why did change to support minimized attributes have any affect on how `MarkupTagBlock` is generated for non-minimized case?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -8,16 +8,14 @@
...
 using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.AspNet.Razor.Test.Framework;
 using Microsoft.AspNet.Razor.Text;
-using Microsoft.AspNet.Razor.Tokenizer;
 using Xunit;
 
 namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
-    public class TagHelperParseTreeRewriterTest : CsHtmlMarkupParserTestBase
+    public class TagHelperParseTreeRewriterTest : TagHelperRewritingTestBase
```
dougbu: why?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: this method will confuse _us_

dougbu:
> @dougbu https://github.com/aspnet/Razor/issues/220#issuecomment-74000770

ok though I don't see anything like an understandable / helpful error message if a user writes `<form asp-anti-forgery >...`.  what exactly happens in that case from a Razor author perspective?  where is the error case tested?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: I'm not modifying the parse code that touches whitespace. I'm just putting it in a different location. The bars represent what the `Output(SpanKind.Markup)` will output.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Not going to bother undoing. Having the cleanup trailing whitespace addon auto-removes these bits. More work not doing this then just letting it happen.

test/Microsoft.AspNet.Razor.Test/Framework/ParserTestBase.cs
NTaylorMullen: This test code was never hit before shockingly. Now that I'm expecting `null` values for some `SyntaxTreeNode`s it was being hit and therefore throwing at line 310.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
NTaylorMullen: No they cannot, it's just descriptors used for the `MinimizedTagHelperDescriptors` test case.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
NTaylorMullen: Razor today doesn't understand space between `=`. Therefore, now that the parser understands minimized attributes (but still not space between `=`, that's another issue) it thinks `class` is a minimized attribute.

NTaylorMullen:
In response to https://github.com/aspnet/Razor/pull/372#issuecomment-98272605 : tons of validating tests in the `TagHelperBlockRewriterTest.cs` file. Razor author sees:

`Attribute 'asp-anti-forgerery' on tag helper element 'form' requires a value. Tag helper bound attributes of type 'System.Boolean' cannot be empty or contain only whitespace.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -42,6 +42,13 @@ public TagHelperAttribute(string name, object value)
...
+        /// </summary>
+        /// <remarks>If <c>true</c>, <see cref="Value"/> will be ignored.</remarks>
+        public bool Minimized { get; set; }
+
+
```
pranavkm: Empty line

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: I can't leave out the helper because it needs to do specific work to add a minimized attribute. As for renaming: I don't feel like it warrants an entirely new method name for a different form of an HTML attribute, that to me makes less sense. Going to leave as is.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Long line otherwise.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: So we can re-use the base class bits. Why re-write it all? We already have a few examples of us doing this in Razor tests.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -61,7 +67,8 @@ public bool Equals(IReadOnlyTagHelperAttribute other)
             return
                 other != null &&
                 string.Equals(Name, other.Name, StringComparison.OrdinalIgnoreCase) &&
-                Equals(Value, other.Value);
+                Equals(Value, other.Value) &&
+                Minimized == other.Minimized;
```
dougbu: should probably ignore `Value` when both are `Minimized`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: This warrants a different name because it's doing something different.  In particular, the effect is not the same as a call to `AddHtmlAttribute("attribute-name", value: null)` (the basic expectation of overloads).  So please rename it to `AddMinimizedAttribute()` or `AddMinimizedHtmlAttribute()`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: :+1:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -341,27 +347,34 @@ private void RenderUnboundHTMLAttributes(IEnumerable<KeyValuePair<string, Chunk>
                     continue;
                 }
 
+                var addHtmlAttributeMethodName = attributeValue == null ?
+                    _tagHelperContext.ExecutionContextAddMinimizedHtmlAttributeMethodName :
+                    _tagHelperContext.ExecutionContextAddHtmlAttributeMethodName;
```
dougbu: Remove this ternary and move the similar-but-not-similar-enough `WriteStartInstanceMethodInvocation()` calls into the following `if` block.  Much easier to understand with fewer checks of `attributeValue == null` and all of the method invocation together.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -341,27 +347,34 @@ private void RenderUnboundHTMLAttributes(IEnumerable<KeyValuePair<string, Chunk>
...
-                if (isPlainTextValue)
-                {
-                    _writer.WriteStringLiteral(textValue);
-                }
-                else
+                    .WriteStartInstanceMethodInvocation(ExecutionContextVariableName, addHtmlAttributeMethodName)
+                    .WriteStringLiteral(htmlAttribute.Key);
+
+                // If we have a minimized attribute there is no value
+                if (attributeValue != null)
```
dougbu: put another way, start this with `if (attributeValue == null) { ... } else ...`

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -341,27 +347,34 @@ private void RenderUnboundHTMLAttributes(IEnumerable<KeyValuePair<string, Chunk>
...
-                // If it's a plain text value then we need to surround the value with quotes.
-                if (isPlainTextValue)
-                {
-                    _writer.WriteStringLiteral(textValue);
-                }
-                else
+                    .WriteStartInstanceMethodInvocation(ExecutionContextVariableName, addHtmlAttributeMethodName)
+                    .WriteStringLiteral(htmlAttribute.Key);
+
+                // If we have a minimized attribute there is no value
```
dougbu: Don't repeat the same comment with slightly different words.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: - This change remains confusing.  Very little else but whitespace has changed in this class.  So why does the same condition mean `// Minimized attribute` now but `// Saw a space or newline after the name` before?
- If `// Minimized attribute` is correct, the comment is in the wrong place -- belongs inside the block.  Don't place definitive comments before the check that makes them true.
- `in most cases` must be untrue unless minimized attributes are not supported after other (minimized or not) attributes.  Leave comment out or mention other, equally-likely possibilities.

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
```diff
@@ -58,6 +59,11 @@ public GeneratedTagHelperContext()
         public string ExecutionContextAddTagHelperAttributeMethodName { get; set; }
 
         /// <summary>
+        /// The name of the <see cref="ExecutionContextTypeName"/> method used to add minimized HTML attributes.
+        /// </summary>
+        public string ExecutionContextAddMinimizedHtmlAttributeMethodName { get; set; }
```
dougbu: So there's going to be an MVC reaction PR?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -239,7 +240,9 @@ private static bool IsSelfClosing(Block beginTagBlock)
                 return false;
             }
 
-            attribute = CreateMarkupAttribute(name, builder, attributeValueTypes);
+            // If we're not after an equal then we should treat the value as if it were a minimized attribute.
```
dougbu: why?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
dougbu: To confirm: Nothing in this file has changed except trailing whitespace?

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
NTaylorMullen: Si.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: If we don't run into an `=` it's the end of the attribute, therefore no value.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/MinimizedTagHelpers.cshtml
```diff
@@ -0,0 +1,21 @@
...
+           catchall-unbound-required
+           input-unbound-required
+           bound-string="hello" />
+    <input 
+           nottaghelper
+           class="btn"
+           catchall-unbound-required
+           input-unbound-required
+           input-bound-required-string="hello"
+           input-unbound-required="world" />
```
dougbu: What unique aspect is tested in the last two elements in this file?

NTaylorMullen: Mixing in different spacing between the minimized attributes.

dougbu: Mix whitespace into the first four cases rather than duplicating a couple.  Don't duplicate even more tests.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
dougbu: Rename to match the test files -- `MinimizedTagHelpersDescriptors` or `MinimizedTagHelpers_Descriptors`

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -0,0 +1,1095 @@
...
+                        }
+                    },
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(MinimizedAttributeData_Document))]
+        [MemberData(nameof(MinimizedAttributeData_CSharpBlock))]
+        [MemberData(nameof(MinimizedAttributeData_PartialTags))]
```
dougbu: These tests seems to mix minimized attributes into numerous contexts, all of which use identical attribute parsing.  Fine to mix actual test variants (e.g. `string` v.s. non-string / bound vs. unbound) into a mix of contexts but don't repeat everything for every unrelated context.  Just too much to deal with in this PR or when we make changes in the future.

This is a general Razor testing issue but don't make it worse in this PR.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -1905,7 +1903,12 @@ public static TheoryData OptOut_WithPartialTextTagData
                     {
                         "@{<!text /}",
                         buildPartialStatementBlock(
-                            () => new MarkupBlock(blockFactory.EscapedMarkupTagBlock("<", "text /}"))),
+                            () => new MarkupBlock(
+                                new MarkupTagBlock(
+                                    factory.Markup("<"),
+                                    factory.BangEscape(),
+                                    factory.Markup("text /"),
+                                    new MarkupBlock(factory.Markup("}"))))),
```
dougbu: why not fix `EscapedMarkupTagBlock()`?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperRewritingTestBase.cs
```diff
@@ -0,0 +1,89 @@
...
+        public override ParserContext CreateParserContext(
+            ITextDocument input,
+            ParserBase codeParser,
+            ParserBase markupParser,
+            ErrorSink errorSink)
+        {
+            return base.CreateParserContext(input, codeParser, markupParser, errorSink);
+        }
+
+        protected void EvaluateData(
```
dougbu: Be consistent.  Just make everything in this class `public`.

dougbu:
:watch: for less tests to wade through

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Ah, shoot you're right. I thought I had modified this file. WIll undo :+1:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: This is the parsing change that enables the `TagHelper` rewriter to understand minimized attributes. Attributes are always grouped with the prefixed whitespace before them, prior to this change you would get a single block which contained `<input checked` where now you get one for `<input` and another with ` checked`.

Therefore, this comment is saying "we're about to construct a minimized attribute block, other info does not belong inside of it, Output that now before we construct the block dedicated to the minimized attribute". I'll nuke the comment all together, we don't add them in other places.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
NTaylorMullen: Whenever you touch the Razor parser you do need to create all these insane (seemingly duplicate) variants. It's incredibly fragile and everything floats up to the rewriter.

NTaylorMullen:
Updated. Addressed all except https://github.com/aspnet/Razor/pull/372#discussion_r30013889

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,18 @@ private void BeforeAttribute()
                 return;
             }
 
+            // Minimized attribute
```
dougbu: This comment remains misplaced.  Should be two lines down.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,18 @@ private void BeforeAttribute()
...
             if (!At(HtmlSymbolType.Equals))
             {
-                // Saw a space or newline after the name, so just skip this attribute and continue around the loop
```
dougbu: Most of the comment at this spot in previous iterations was useful.  It described what the `Output()` calls would do.  My complaint was about the last ("in most cases") bit.

dougbu: In particular, remind the reader what has already been accepted i.e. what the first `Output()` will do.  The second `Output()` should be obvious to all.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,18 @@ private void BeforeAttribute()
...
+
+                using (Context.StartBlock(BlockType.Markup))
+                {
+                    Accept(whitespace);
+                    Accept(name);
+                    Output(SpanKind.Markup);
+                }
+
                 return;
             }
```
dougbu: Add a comment right below here reminding user that we're in the mainline scenario at this point i.e. recovery and minimized attribute cases were handled above.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -471,11 +471,18 @@ private void BeforeAttribute()
...
 
+            // Minimized attribute
             if (!At(HtmlSymbolType.Equals))
```
dougbu: Since parser gets confused by whitespace, how will `<element attribute = "value" />` or `<element attribute ="value" />` be parsed now?  Has this change made handling of (perfectly valid) whitespace before the equals _worse_?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -22,16 +21,13 @@ public bool Equals(IReadOnlyTagHelperAttribute attributeX, IReadOnlyTagHelperAtt
                 attributeX == attributeY ||
                 // Normal comparer doesn't care about the Name case, in tests we do.
                 string.Equals(attributeX.Name, attributeY.Name, StringComparison.Ordinal) &&
-                Equals(attributeX.Value, attributeY.Value);
+                Equals(attributeX.Value, attributeY.Value) &&
+                attributeX.Minimized == attributeY.Minimized;
```
dougbu: Rebase and `Minimized` will be handled correctly in the base class.

Unless you want to check both `Minimized` and `Value` even when `Minimzed == true`.  In that case, add a bit more to the comment at line 22.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
dougbu: W.r.t. https://github.com/aspnet/Razor/pull/372/files#r30054041 comment, does seem like handling of whitespace between attribute name and the equals sign is now worse.  How will the `Span` containing `" = 'bar'>"` be treated during code generation e.g. is it ignored?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
dougbu: The Razor parser is a recursive-descent parser with a fair number of special cases, I agree.  However it isn't so fragile that changes in the low-level attribute parsing merit so many unreadable new cases.  For example you've added a single product code clause in the condition about minimized-null-or-empty values for bound non-`string` attributes, not enough to justify testing minimized `string` attributes, minimized unbound attributes, and so on.

Bottom line, it isn't possible to see what matters in this test class.

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
NTaylorMullen: From a non-TagHelper perspective: It'll be treated as equal. Technically there'll be more blocks in the mix but the output will remain unchanged.

From a TagHelper perspective: You'll actually see the code in the HTML elements now. Before it was ignored completely. Aka: `<input checked class="btn" />` resulted in an output.Attributes with 1 attribute (the checked wasn't even acknowledged); poorly formed attributes would either disrupt parsing of other correct attributes or be ignored entirely.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/372#discussion_r30069422

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
NTaylorMullen: There's no base class for this comparer. It's comparing the `IReadOnlyTagHelperAttribute`. This wasn't touched in your PR. Will update to do the `Minimized == true` then no value check though.

NTaylorMullen:
Updated.

NTaylorMullen:
This has been open too long, Will get this in tomorrow afternoon and if more comments come in I will address them.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -202,6 +202,13 @@ protected virtual string GenerateUniqueId()
                     // First attribute wins, even if there's duplicates.
                     var attributeValueChunk = matchingAttributes.First().Value;
 
+                    // Minimized attributes are not valid for bound attributes. There will be an error logged for this
+                    // bound attribute already so we can skip.
```
dougbu: Please make this comment a bit more informative, describing what logs the error.  "There will be" is just too vague.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -473,12 +473,26 @@ private void BeforeAttribute()
...
+                    Accept(name);
+                    Output(SpanKind.Markup);
+                }
+
                 return;
             }
 
+            // Not a minimized attribute, parse as if it were well-formed (if attribute turns out to be malformed we
+            // will go into recovery).
+
```
dougbu: extra blank line should die a horrible, horrible death

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -0,0 +1,909 @@
...
+                            new RazorError(
+                                string.Format(errorFormat, "bound-required-string", "input", stringType),
+                                absoluteIndex: 7,
+                                lineIndex: 0,
+                                columnIndex: 7,
+                                length: 21)
+                        }
+                    },
+                    {
+                        $"<input bound-required-string class='{expressionString}' />",
```
dougbu: Why is it necessary to test `class='{expressionString}'` in combination with almost every other attribute type?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -0,0 +1,909 @@
...
+                    var errors = data[2] as RazorError[];
+
+                    for (var i = 0; i < errors.Length; i++)
+                    {
+                        var error = errors[i];
+                        error.Location = SourceLocation.Advance(error.Location, "@{");
+                    }
+                }
+
+                return documentData;
```
dougbu: Cheating to hide the large amount of repetition but :ok:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
dougbu: Still some extra noise in this test class but much improved.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: thanx

dougbu:
:shipit: after addressing the few remaining comments

Issue #50, opened 2014-05-14
Razor code generator produces unnecessary writeliteral with whitespaces 


NTaylorMullen:
Original Issue on Codeplex: https://aspnetwebstack.codeplex.com/workitem/1548

 Sample code:
```
 <div>@{var a = 1; <text>@a</text>}</div> It translates to: 
```

```C#
 public class Index_cshtml : System.Web.Mvc.WebViewPage { 
 ... 
 public override void Execute() { 
 var a = 1; 
 WriteLiteral(" "); 
 Write(a); 
 WriteLiteral("</div>"); 
 } 
 } 
```
 After code var a = 1; and before code Write(a);, Razor translator generates unnecessary writes of whitespace: WriteLiteral(" ");. 

 And result of this output:
 <div> 1</div> When replace cshtml code by replacing whitespace between ; and @ (this is CODE CONTEXT) by more than one space, for ex., 3 tabs:
```
 <div>@{var a = 1;<text>@a</text>}</div> 
```
It generates: 

```
 var a = 1; 
 WriteLiteral("\t\t\t"); 
 Write(a); 
 WriteLiteral("</div>"); 
```
 So, Razor translator generates unnecessary writes of whitespaces BEFORE @ from CODE CONTEXT.

Eilon:
@ajaybhargavb another interesting Razor bug.

PR #378, opened 2015-05-06
Removed unnecessary whitespace before text tag

ajaybhargavb:
Issue - #50 
@NTaylorMullen @dougbu 

There are a few tests that needs to be updated. Will update them once the product code is signed off.

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
@@ -388,14 +388,14 @@ namespace Microsoft.AspNet.Razor.Parser
             Span.EditHandler.AcceptedCharacters = AcceptedCharacters.Any;
 
             // Accept whitespace but always keep the last whitespace node so we can put it back if necessary
-            var lastWs = AcceptWhiteSpaceInLines();
-            Debug.Assert(lastWs == null || (lastWs.Start.AbsoluteIndex + lastWs.Content.Length == CurrentLocation.AbsoluteIndex));
+            var lastWhitespace = AcceptWhiteSpaceInLines();
+            Debug.Assert(lastWhitespace == null || (lastWhitespace.Start.AbsoluteIndex + lastWhitespace.Content.Length == CurrentLocation.AbsoluteIndex));
 
             if (EndOfFile)
             {
-                if (lastWs != null)
+                if (lastWhitespace != null)
                 {
-                    Accept(lastWs);
+                    Accept(lastWhitespace);
                 }
                 return;
             }
@@ -411,16 +411,23 @@ namespace Microsoft.AspNet.Razor.Parser
             if (Context.DesignTimeMode || !isMarkup)
             {
                 // CODE owns whitespace, MARKUP owns it ONLY in DesignTimeMode.
-                if (lastWs != null)
+                if (lastWhitespace != null)
                 {
-                    Accept(lastWs);
+                    Accept(lastWhitespace);
                 }
             }
             else
             {
+                var nextSymbol = Lookahead(1);
+
                 // MARKUP owns whitespace EXCEPT in DesignTimeMode.
                 PutCurrentBack();
-                PutBack(lastWs);
+
+                // Don't putback the whitespace if it precedes a '<text>' tag.
+                if (nextSymbol != null && !nextSymbol.Content.Equals(SyntaxConstants.TextTagName))
+                {
+                    PutBack(lastWhitespace);
+                }
             }
 
             if (isMarkup)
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -411,16 +411,23 @@ private void Statement(Block block)
...
+
                 // MARKUP owns whitespace EXCEPT in DesignTimeMode.
                 PutCurrentBack();
-                PutBack(lastWs);
+
+                // Don't putback the whitespace if it precedes a '<text>' tag.
+                if (!nextSymbol.Content.Equals(SyntaxConstants.TextTagName))
+                {
+                    PutBack(lastWhitespace);
+                }
```
dougbu: looks like this should mostly fix the problem.  but when `nextSymbol.Content.Equals(SyntaxConstants.TextTagName) && lastWhitespace != null` here (i.e. a non-existent `else if`), should we call `Accept(lastWhitespace)`?  right now nothing owns the whitespace before `<text>`

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/SimpleCodeBlock.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿@{
```
dougbu: rename this to something more specific e.g. `CodeBlockWithTextElement`.  some name that distinguishes this CodeBlock file in terms of what it contains, not the complexity it lacks.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -388,14 +388,14 @@ private void Statement(Block block)
             Span.EditHandler.AcceptedCharacters = AcceptedCharacters.Any;
 
             // Accept whitespace but always keep the last whitespace node so we can put it back if necessary
-            var lastWs = AcceptWhiteSpaceInLines();
-            Debug.Assert(lastWs == null || (lastWs.Start.AbsoluteIndex + lastWs.Content.Length == CurrentLocation.AbsoluteIndex));
+            var lastWhitespace = AcceptWhiteSpaceInLines();
+            Debug.Assert(lastWhitespace == null || (lastWhitespace.Start.AbsoluteIndex + lastWhitespace.Content.Length == CurrentLocation.AbsoluteIndex));
```
dougbu: thanks for the much-improved variable name

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: Two concerns:

1. This method has some wonkyness to it, for instance the two comments on line 423 and 413 completely contradict each other. Moreover, I don't see why Markup should ever own whitespace even in design time mode. That would end up breaking intellisense for C#. We may want to clarify this with @anurse for what his intentions where when writing this bit.
1. This comment could be completely null and void based on the answer to 1, but this seems like the wrong place to put logic handling the `<text>` tag. There's a few locations where we special case `<text>` in `HtmlMarkupParser.Block` but neither should be impacted by what happens in the C# statement block from a conceptual level. Therefore, IF we can allow C# to own whitespace always in statements that would remove the need for this check entirely.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/SimpleCodeBlock.cshtml
```diff
@@ -0,0 +1,3 @@
...
+    var a = 1; <text>@a</text>
```
NTaylorMullen: Add a version with a tab

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
anurse: I think you're right in 1 that there isn't really a need for whitespace before the `<text>` tag to be owned by markup. It probably had something to do with IntelliSense triggering (most of these weird whitespace things did :)) but I can't remember the details now. Double-check that IntelliSense works well around `<text>` tags and I'm happy with the change.

ajaybhargavb:
Updated. @dougbu @NTaylorMullen

NTaylorMullen:
:shipit:

dougbu:
:shipit:

Issue #89, opened 2014-08-18
Support adding attributes to `IDictionary<string, TValue>` properties


NTaylorMullen:
It will look something like this:

```C#
public class MyTagHelper : TagHelper
{
    public Dictionary<string, string> route { get; set; } // Can be <string, T>
    
    [AttributeName("something-")]
    public IEnumerable<KeyValuePair<string, int>> something { get; set; }
    
    public override Process(.....
}
```

This will then enable us to do things like:

```HTML
<mytaghelper routeFoo="hello" routeBar="world" something-foo="Hi" something-bar="earth" />
```

Which will then populate route with:
```
{
    "Foo": "hello",
    "Bar: "world"
}
```

And would populate 'something' with:
```
{
    "foo": "Hi",
    "bar: "earth"
}
```

NTaylorMullen:
Talked in person, can be re-triaged doesn't need to happen now.

dougbu:
@Eilon @danroth27 suggest we should try and get this done early in Beta5.  right now it's languishing in the Backlog.  but it feels more urgent than for example #182.

@DamianEdwards the system could enable authors to differentiate properties that should be bound to a single dictionary expression from ones that should be bound to zero or more attributes. it could also special-case attributes with the exact name (to be dictionary-valued) or ignore such attributes.  i.e. `<a route-aaa="bbb" route="@myDictionary"/>` might be up to the tag helper author, legal, or illegal.  which do you prefer?

Eilon:
Moving out of backlog so we'll revisit in triage.

dougbu:
renamed to "Support adding attributes to `IDictionary<string, TValue>` properties"

dougbu:
- aspnet/Common@aae8e6e
- aspnet/Mvc@61b76fd
- 7dc0508
- 94f2f90
- 60c47c8
- aspnet/Mvc@517c013

Issue #225, opened 2014-11-17
Attribute order from source is not preserved upon rendering


DamianEdwards:
This is annoying. The attributes for elements with Tag Helpers defined do not have their order preserved from the source CSHTML. They appear to be rendered in alphabetical order instead.

`<input type="submit" value="Log in" class="btn btn-default" />`
becomes
`<input class="btn btn-default" type="submit" Value="Log in" />`

dougbu:
current behaviour remains as I described it in previous comment.  the Razor-level behaviour is Just Fine:tm:

however some MVC tag helpers still bind HTML attributes then put them back, moving those attributes to the end.  issues are limited to:
- `<input/>` tag helper binds the `type` and `value` attributes
- `<option/>` tag helper binds the `selected` and `value` attributes

think I'll leave this bug here but fix the above issues in the MVC repo...

dougbu:
requires a few things:
- @NTaylorMullen's fix for #279
- consistently add attributes to `TagHelperExecutionContext` in source order
- use `PostElement` in `<link/>` and `<script/>` tag helpers and merge added attributes for original element
- use `PostElement` for `type="checkbox"` case in `<input/>` tag helper and merge added attributes for original element
- restore `TagHelperAttribute` close to original `AllAttributes` order in `TagHelperContentExtentions.CopyHtmlAttribute()` and (perhaps) rename that method to `CopyHtmlAttributes()`

NTaylorMullen:
Talked about the importance of this fix with @Eilon and @dougbu, not important for beta5. Pushing it out.

NTaylorMullen:
b0b4ac4f3f12299ec1c9301f0c7f67c7646e1c63

PR #380, opened 2015-05-08
[Design] Add `[HtmlAttributeName(..., DictionaryAttributePrefix="prefix")]`

dougbu:
- #89
- support adding attributes (that aren't otherwise bound) to a tag helper dictionary
- add `ObjectCreationExpression`, `Prefix` and more to `TagHelperAttributeDescriptor`
 - determine `string`-ness when creating `TagHelperAttributeDescriptor`s
- handle `Prefix` matches in `TagHelperBlockRewriter`
- refactor `CSharpTagHelperCodeRenderer` to handle new indexer property assignments
 - calculate `string`-ness for a bound attribute using all `TagHelperAttributeDescriptor`s
- extend handling of invalid attribute names to include `[HtmlAttributeName]`

known issue: `ObjectCreationExpression` is a C# expression because otherwise
generation has insufficient information to determine correct constructor.
- no VB or F# (for example) workarounds today

no new tests yet

nits:
- add `TagHelperHelper`
- separate `TagHelperAttributeDescriptorComparer` and `TypeBasedTagHelperDescriptorComparer`
 - encourages reuse and most are already used in multiple classes
- use `<inheritdoc/>` in `TagHelperDescriptorComparer`
 - also give it an explicit constructor

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
@@ -11,6 +11,8 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
     public sealed class HtmlAttributeNameAttribute : Attribute
     {
+        private string _dictionaryAttributePrefix;
+
         /// <summary>
         /// Instantiates a new instance of the <see cref="HtmlAttributeNameAttribute"/> class.
         /// </summary>
@@ -29,5 +31,39 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// HTML attribute name of the associated property.
         /// </summary>
         public string Name { get; }
+
+        /// <summary>
+        /// Gets or sets the prefix used to match HTML attribute names. Matching attributes are added to the
+        /// associated property (an <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/>).
+        /// </summary>
+        /// <remarks>
+        /// If non-<c>null</c> associated property must be compatible with
+        /// <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/> where <c>TKey</c> is
+        /// <see cref="string"/>.
+        /// </remarks>
+        /// <value>
+        /// If associated property is compatible with
+        /// <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/>, default value is <c>Name + "-"</c>.
+        /// Otherwise default value is <c>null</c>.
+        /// </value>
+        public string DictionaryAttributePrefix
+        {
+            get
+            {
+                return _dictionaryAttributePrefix;
+            }
+            set
+            {
+                _dictionaryAttributePrefix = value;
+                DictionaryAttributePrefixSet = true;
+            }
+        }
+
+        /// <summary>
+        /// Gets an indication whether <see cref="DictionaryAttributePrefix"/> has been set. Used to distinguish an
+        /// uninitialized <see cref="DictionaryAttributePrefix"/> value from an explicit <c>null</c> setting.
+        /// </summary>
+        /// <value><c>true</c> if <see cref="DictionaryAttributePrefix"/> was set. <c>false</c> otherwise.</value>
+        public bool DictionaryAttributePrefixSet { get; private set; }
     }
 }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -211,7 +211,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
             foreach (var property in accessibleProperties)
             {
-                var descriptor = ToAttributeDescriptor(property);
+                var descriptor = ToAttributeDescriptor(property, type, errorSink);
                 if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
                 {
                     attributeDescriptors.Add(descriptor);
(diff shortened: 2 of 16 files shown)


src/Microsoft.AspNet.Razor/TagHelpers/TagHelperHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    /// <summary>
+    /// Helper methods related to <see cref="TagHelperDescriptor"/> and related classes.
+    /// </summary>
+    public static class TagHelperHelper
```
dougbu: @rynowak this name's for you

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -20,35 +43,117 @@ internal TagHelperAttributeDescriptor(string name, PropertyInfo propertyInfo)
...
         /// <param name="name">The HTML attribute name.</param>
-        /// <param name="propertyName">The name of the CLR property name that corresponds to the HTML
-        /// attribute.</param>
+        /// <param name="propertyName">The name of the CLR property that corresponds to the HTML attribute.</param>
         /// <param name="typeName">
-        /// The full name of the named (see <paramref name="propertyName"/>) property's
-        /// <see cref="System.Type"/>.
+        /// The full name of the named (see <paramref name="propertyName"/>) property's <see cref="System.Type"/>.
+        /// </param>
+        /// <param name="isStringProperty">
```
dougbu: Previous code was a bit odd, using string comparisons on type names.  This is more explicit and allows us, if we wish, to handle implicit conversions from strings.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -277,5 +378,45 @@ private static string ToHtmlCase(string name)
         {
             return HtmlCaseRegex.Replace(name, HtmlCaseRegexReplacement).ToLowerInvariant();
         }
+
+        private static Type ExtractGenericInterface(Type queryType, Type interfaceType)
```
dougbu: Cribbed this method from MVC repo's Common project.  Will move to Common repo as I make this PR non-design-y, unless I hear objections.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -320,15 +336,16 @@ private static bool IsSelfClosing(Block beginTagBlock)
...
 
-                    attribute = CreateMarkupAttribute(name, spanBuilder, attributeValueTypes);
+                    attribute = CreateMarkupAttribute(name, spanBuilder, isBoundNonStringAttribute);
 
                     return true;
                 }
             }
 
+            // May need to update block Kind of all descendant Spans in this more complex case.
+            // Examples may include name="@aaa + @bbb" and name="@* commented *@ code".
```
dougbu: Plan to remove this note-to-self before removing "[Design]" label from the PR.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperHelper.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// Collection of <see cref="TagHelperDescriptor"/>s targeting the containing element in the Razor source.
+        /// </param>
+        /// <returns>
+        /// <c>true</c> if the attribute named <paramref name="name"/> is bound and no associated property or
+        /// dictionary value (in the case of a <see cref="TagHelperAttributeDescriptor.Prefix"/> match) has
+        /// <see cref="Type"/> <see cref="string"/>. <c>false</c> otherwise e.g. if the property is not bound.
+        /// </returns>
+        public static bool IsBoundNonStringAttribute(
+            [NotNull] string name,
+            [NotNull] IEnumerable<TagHelperDescriptor> descriptors)
```
dougbu: Previous code that determined if an attribute was bound to a `string` property could get very confused when multiple tag helpers bound the same name.  Of course the older code also did not handle prefix matches.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -922,8 +922,9 @@ public static TheoryData InvalidTagHelperAttributeDescriptorData
                         new[] {
                             string.Format(
                                 errorFormat,
+                                typeof(InvalidBoundAttribute).FullName,
                                 nameof(InvalidBoundAttribute.DataSomething),
-                                typeof(InvalidBoundAttribute).FullName)
```
dougbu: Test fix reflects product code fix: Message previously included 'Property.Namespace.Contains.Class`.  Full type name and property name now included in correct order: `'Namespace.Contains.Class.Property`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
```diff
@@ -29,5 +31,39 @@ public HtmlAttributeNameAttribute(string name)
...
+            set
+            {
+                _dictionaryAttributePrefix = value;
+                DictionaryAttributePrefixSet = true;
+            }
+        }
+
+        /// <summary>
+        /// Gets an indication whether <see cref="DictionaryAttributePrefix"/> has been set. Used to distinguish an
+        /// uninitialized <see cref="DictionaryAttributePrefix"/> value from an explicit <c>null</c> setting.
```
NTaylorMullen: q: Why do we care if they set it to `null`? Sounds to protective of the user.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -226,31 +226,132 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
-                                attributeNameAttribute.Name :
-                                ToHtmlCase(property.Name);
+            var attributeName = attributeNameAttribute?.Name ?? ToHtmlCase(property.Name);
+            var isStringProperty = property.PropertyType == typeof(string);
+            string prefix = null;
+            string objectCreationExpression = null;
+            string prefixedValueTypeName = null;
+            var areStringPrefixedValues = false;
+
+            var dictionaryTypeArguments = ExtractGenericInterface(property.PropertyType, typeof(IDictionary<,>))
```
NTaylorMullen: `ExtractGenericInterface` is only used for dictionary, no need for this method to be flexible.

NTaylorMullen: Nm, see the comment about wanting to move the method to common.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -156,8 +166,12 @@ protected virtual string GenerateUniqueId()
 
         private void RenderTagHelpersCreation(
             TagHelperChunk chunk,
-            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors,
+            IEnumerable<KeyValuePair<string, Chunk>> boundStringAttributes)
         {
+            var boundStringAttributeNames =
+                new HashSet<string>(boundStringAttributes.Select(kvp => kvp.Key), StringComparer.OrdinalIgnoreCase);
```
NTaylorMullen: With this approach last wins.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -183,112 +197,164 @@ protected virtual string GenerateUniqueId()
...
+            }
 
-                    // We capture the tag helpers property value accessor so we can retrieve it later (if we need to).
-                    var valueAccessor = string.Format(CultureInfo.InvariantCulture,
-                                                      "{0}.{1}",
-                                                      tagHelperVariableName,
-                                                      attributeDescriptor.PropertyName);
+            // Pass 3: Add all remaining attributes that match a prefix to corresponding dictionary. Use same order as
+            // above loops but skip everything already bound to a specific property of this tag helper.
+            foreach (var attributeDescriptor in attributeDescriptors)
```
NTaylorMullen: I like the multi-phase approach you have but it will make doing https://github.com/aspnet/Razor/issues/225 incredibly difficult :cry:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
dougbu: not about protecting anyone.  if user sets this property to `null`, it means "don't map a prefix for this dictionary" i.e. just let Razor authors assign to the property, not add to the dictionary.  if the property is never set, it means "set up a prefix if this happens to be a dictionary".

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: When doing #225, will still do Pass 1.  Passes 2 and 3 would instead be a per-tag-helper precedence applied while going through the attributes.

Alternatively could add a Pass 4 for #225: Reorder the attribute lists created earlier to match the source order.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
NTaylorMullen: Hmm, that's an interesting functionality. Seems a bit odd that setting the `DictionaryAttributePrefix` to `null` results in the property being directly settable to `Dictionary<string, TValue>` though (seems unrelated). I don't have a great solution but off the top of my head another property or other attribute sound better.

dougbu: The property is _always_ directly settable.  That happens because `HtmlAttributeName.Name` can't be `null`.  This is just about overriding the on-by-default-for-dictionaries `DictionaryAttributePrefix`.

NTaylorMullen: What happens if the user wants to override the prefix though?

You'd have to do:

```C#
public class FooTagHelper
{
    // Need to provide a name given the ctor setup
    [HtmlAttributeName("SomeUnvaluableName", DictionaryAttributePrefix = "ANewName")]
    public IDictionary<string, string> MyDictionary { get; set; }
}
```
Which doesn't make all that much sense and is somewhat ugly.

What if we were to take an approach where we add the `-` to specify dictionary key/values, otherwise if you don't provide the `-` you can set the property directly:

```C#
[HtmlElementName("a")]
public class AnchorTagHelper
{
    public IDictionary<string, string> Params { get; set; }
}
```

Results in

```HTML
@{
    var paramDictionary = new Dictionary<string, string>
    {
        { "a", "b" }
    }
}
<!-- Would create the href appended with ?temp=SomeValue&other=OtherValue -->
<a params-temp="SomeValue" params-other="OtherValue" ...>Hello World</a>
<!-- Would create the href appended with ?a=b -->
<a params="paramDictionary" ...>Hello World</a>
<!-- Would create the href appended with ?a=b&temp=SomeValue&other=OtherValue -->
<a params-temp="SomeValue" params-other="OtherValue" params="paramDictionary" ...>Hello World</a>
```


I feel like this approach would be more flexible/intuitive.

/cc @DamianEdwards

dougbu: We have discussed this one to death and your approach is just weird.  To remind you:
- Dictionary properties are _always_ identical to any other property and may be set in the Razor source if the author chooses.  (System does not support read-only dictionary properties.)
- Prefix matching is optional but on by default for dictionary properties.
- To disable prefix matching, use `[HtmlAttributeName("someName", DictionaryAttributePrefix = null)]`
- To override the default prefix, use `[HtmlAttributeName("someName", DictionaryAttributePrefix = "prefix")]`
- There is never an ambiguity between attribute names and prefixes even in the case of `[HtmlAttributeName("someName", DictionaryAttributePrefix = "someName")]` because non-prefix matches have precedence.

NTaylorMullen: @dougbu Maybe I'm misunderstanding the impl in this PR, there is a lot :smile: . So for dictionary properties by default they act the way I'm suggesting? And if you want to disable the prefixability you can set the prefix to `null`?

dougbu: I have no idea what you're suggesting.  But yes, if you want to disable prefix matching you set `DictionaryAttributePrefix` to `null`.

NTaylorMullen: If a `FooTagHelper` has `public IDictionary<string, string> Query { get; set; }` within it. Is the following valid:

```HTML
@{
    var queriesDict = new Dictionary<string, string>
    {
        { "a", "b" }
    }
}

<foo query-a="...." query-b="..." />

<foo query="queriesDict" />

<foo query="queriesDict" query-a="...." query-b="..." />
```

?

dougbu: Yes.  In addition `<foo query-a="...." query-b="..." query="queriesDict" />` is also valid and the `query` assignment is done first to avoid very strange cases of lost dictionary entries.

NTaylorMullen: Beautiful, alright I completely misunderstood. :+1:, In process of poking through this more in-depth.

dougbu:
Closing [Design].  Let's do the real bits in pieces.

NTaylorMullen:
:+1: thanks @dougbu that'll be fantastic. Hard to consume it all as it is

PR #383, opened 2015-05-11
Add `[HtmlAttributeName(..., DictionaryAttributePrefix="prefix")]` part 1

dougbu:
- related to #89 because we need more descriptor comparers in more places
- separate `TagHelperAttributeDescriptorComparer` and `TypeBasedTagHelperDescriptorComparer`
 - encourages reuse and most will soon be used in multiple classes

nits:
- use `<inheritdoc/>` in `TagHelperDescriptorComparer`
 - also give it an explicit constructor
- make product comparers easier to subclass
 - base test `CaseSensitiveTagHelperAttributeDescriptorComparer` on product code

### DIFF
--- a/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
@@ -20,9 +20,6 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
         internal static readonly string ScopeManagerVariableName = "__tagHelperScopeManager";
         internal static readonly string RunnerVariableName = "__tagHelperRunner";
 
-        private static readonly TagHelperAttributeDescriptorComparer AttributeDescriptorComparer =
-            new TagHelperAttributeDescriptorComparer();
-
         private readonly CSharpCodeWriter _writer;
         private readonly CodeBuilderContext _context;
         private readonly IChunkVisitor _bodyVisitor;
@@ -64,7 +61,7 @@ namespace Microsoft.AspNet.Razor.Generator.Compiler.CSharp
             // multiple TargetElement attributes are on a TagHelper type and matches overlap for an HTML element.
             // Having more than one descriptor with the same TagHelper type results in generated code that runs
             // the same TagHelper X many times (instead of once) over a single HTML element.
-            var tagHelperDescriptors = chunk.Descriptors.Distinct(TypeNameTagHelperDescriptorComparer.Default);
+            var tagHelperDescriptors = chunk.Descriptors.Distinct(TypeBasedTagHelperDescriptorComparer.Default);
 
             RenderBeginTagHelperScope(chunk.TagName, chunk.SelfClosing, chunk.Children);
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
```diff
@@ -20,40 +21,36 @@ public class TagHelperDescriptorComparer : IEqualityComparer<TagHelperDescriptor
...
+        /// <inheritdoc />
         /// <remarks>
         /// Determines equality based on <see cref="TagHelperDescriptor.TypeName"/>,
         /// <see cref="TagHelperDescriptor.AssemblyName"/>, <see cref="TagHelperDescriptor.TagName"/>,
         /// and <see cref="TagHelperDescriptor.RequiredAttributes"/>.
         /// </remarks>
-        public bool Equals(TagHelperDescriptor descriptorX, TagHelperDescriptor descriptorY)
+        public virtual bool Equals(TagHelperDescriptor descriptorX, TagHelperDescriptor descriptorY)
         {
+            if (descriptorX == null || descriptorY == null)
```
NTaylorMullen: Why not just do a `descriptorX == descriptorY ||` prior to the other checks?

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -20,9 +20,6 @@ public class CSharpTagHelperCodeRenderer
         internal static readonly string ScopeManagerVariableName = "__tagHelperScopeManager";
         internal static readonly string RunnerVariableName = "__tagHelperRunner";
 
-        private static readonly TagHelperAttributeDescriptorComparer AttributeDescriptorComparer =
```
NTaylorMullen: Oh wow, we didn't use this at all? failsauce

src/Microsoft.AspNet.Razor/TagHelpers/TypeBasedTagHelperDescriptorComparer.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// Determines equality based on <see cref="TagHelperDescriptor.AssemblyName"/> and
+        /// <see cref="TagHelperDescriptor.TypeName"/>.
+        /// </remarks>
+        public bool Equals(TagHelperDescriptor descriptorX, TagHelperDescriptor descriptorY)
+        {
+            if (descriptorX == null || descriptorY == null)
```
NTaylorMullen: ditto

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
dougbu: Well that does have a slight advantage for identical non-`null` objects.  :ok:

NTaylorMullen:
:shipit: once the `descriptorX == descriptorY` checks are added.

dougbu:
Updated; note the "also" bits.

NTaylorMullen:
:shipit:

dougbu:
7dc0508

PR #385, opened 2015-05-12
Quick fix: Missed `null` checks in 7dc0508

dougbu:


dougbu:
FYI only. Will merge shortly.

PR #390, opened 2015-05-14
Add `[HtmlAttributeName(..., DictionaryAttributePrefix="prefix")]` part II

dougbu:
- relates to #89 because that changes `string` property checks and needs this refactor
- determine `string`-ness when creating `TagHelperAttributeDescriptor`s
 - add `TagHelperAttributeDescriptor.IsStringProperty`
 - avoid repeated `string` comparisons and be more explicit
- ~~calculate `string`-ness for a bound attribute using all `TagHelperAttributeDescriptor`s~~
 - ~~use most specific type when multiple tag helpers bind same attribute to e.g. `object` and `string`~~
 - ~~produce same error for empty attribute values regardless of `TagHelperDescriptor` order~~
 - ~~generate same code regardless of `TagHelperDescriptor` order~~
- refactor `CSharpTagHelperCodeRenderer` to allow reuse of core attribute value rendering
- test all of it

minor:
- fix `TagHelperBlockRewriter.TryParseBlock()` end quote removal when tag is malformed

nits:
- ~~add `TagHelperHelper` though it's fairly simple at this point~~

*new*:
- refactor `TagHelperBlockRewriter` to use a `private` DTO instead of `out` parameters


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -246,11 +246,14 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
-
-            return new TagHelperAttributeDescriptor(attributeName, property.Name, property.PropertyType.FullName);
+            var attributeName = attributeNameAttribute?.Name ?? ToHtmlCase(property.Name);
+            var isStringProperty = property.PropertyType == typeof(string);
+
+            return new TagHelperAttributeDescriptor(
+                attributeName,
+                property.Name,
+                property.PropertyType.FullName,
+                isStringProperty);
```
NTaylorMullen: Why put the `isStringProperty` on the descriptor when it can be mostly derived from the type name? This seems like an unrelated change and is adding unnecessary data to the should be dumb `TagHelperAttributeDescriptor`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            // Find all TagHelperAttributeDescriptor's matching this name.
+            var boundAttributes = descriptors
+                .SelectMany(descriptor => descriptor.Attributes)
+                .Where(attribute => string.Equals(attribute.Name, name, StringComparison.OrdinalIgnoreCase));
+
+            // Check if any matching TagHelperAttributeDescriptor requires a string value.
+            var isStringValue = boundAttributes.Any(attribute => attribute.IsStringProperty);
+
+            return boundAttributes.Any() && !isStringValue;
```
NTaylorMullen: If there's no `boundAttributes` `isStringValue` will be `false`, no need to check `boundAttributes.Any()`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -429,15 +446,12 @@ private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
...
+            bool isBoundNonStringAttribute)
         {
-            string attributeTypeName;
-
-            // If the attribute was requested by the tag helper and doesn't happen to be a string then we need to treat
-            // its value as code. Any non-string value can be any C# value so we need to ensure the SyntaxTreeNode
-            // reflects that.
-            if (attributeValueTypes.TryGetValue(name, out attributeTypeName) &&
-                !IsStringAttribute(attributeTypeName))
+            // If the attribute was requested by a tag helper and but none required it to be a string then we need
```
NTaylorMullen: `tag helper and but` => `tag helper but`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Talked in person :ok:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -298,10 +310,14 @@ private static bool IsSelfClosing(Block beginTagBlock)
             if (!endNode.IsBlock)
             {
                 var endSpan = (Span)endNode;
-                var endSymbol = (HtmlSymbol)endSpan.Symbols.Last();
+
+                // In some malformed cases e.g. <p bar="false', the last Span (false' in the ex.) may contain more
+                // than a single HTML symbol. Do not ignore those other symbols.
+                var symbolCount = endSpan.Symbols.Count();
+                var endSymbol = symbolCount == 1 ? (HtmlSymbol)endSpan.Symbols.First() : null;
```
NTaylorMullen: `.Single()`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -20,35 +25,48 @@ internal TagHelperAttributeDescriptor(string name, PropertyInfo propertyInfo)
...
-                                            string propertyName,
-                                            string typeName)
+        /// <param name="isStringProperty">
+        /// An indication whether this property is of type <see cref="string"/>.
+        /// </param>
+        public TagHelperAttributeDescriptor(
+            [NotNull] string name,
+            [NotNull] string propertyName,
+            [NotNull] string typeName,
+            bool isStringProperty)
```
NTaylorMullen: Optional: Remove this from the ctor params and have it do a check in the ctor body.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Talked in person :ok:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Helper methods related to <see cref="TagHelperDescriptor"/> and related classes.
+    /// </summary>
+    public static class TagHelperDescriptorMatcher
+    {
+        /// <summary>
+        /// Determine the full name of the <see cref="Type"/> of the property corresponding to an attribute named
+        /// <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">Name of the attribute.</param>
```
NTaylorMullen: property name or HTML name? Should be specific

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+            // Find first TagHelperAttributeDescriptor matching this name.
+            var firstBoundAttribute = descriptors
+                .SelectMany(descriptor => descriptor.Attributes)
+                .FirstOrDefault(attribute => string.Equals(attribute.Name, name, StringComparison.OrdinalIgnoreCase));
+            if (firstBoundAttribute == null)
+            {
+                return null;
+            }
+
+            if (!IsBoundNonStringAttribute(name, descriptors))
```
NTaylorMullen: Should leave a comment as to why `firstBoundAttribute.IsStringProperty` isn't enough.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        /// <param name="name">Name of the attribute.</param>
+        /// <param name="descriptors">
+        /// Collection of <see cref="TagHelperDescriptor"/>s targeting the containing element in the Razor source.
+        /// </param>
+        /// <returns>
+        /// <c>true</c> if the attribute named <paramref name="name"/> is bound and no associated property has
+        /// <see cref="Type"/> <see cref="string"/>. <c>false</c> otherwise e.g. if the property is not bound.
+        /// </returns>
+        public static bool IsBoundNonStringAttribute(
```
NTaylorMullen: Might be nice to have an overload that takes `IEnumerable<TagHelperAttributeDescriptor>` that this one calls into so above we don't have to redo the `SelectMany`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeDescriptorComparer.cs
```diff
@@ -24,13 +24,17 @@ public override bool Equals(TagHelperAttributeDescriptor descriptorX, TagHelperA
                 return true;
             }
 
+            // Normal comparer doesn't care about case, in tests we do. Also double-check IsStringProperty though
+            // it is inferred from TypeName.
             return base.Equals(descriptorX, descriptorY) &&
-                // Normal comparer doesn't care about case, in tests we do.
+                descriptorX.IsStringProperty == descriptorY.IsStringProperty &&
```
NTaylorMullen: If the ctor change for `TagHelperAttributeDescriptor` is done, you wont need to do this.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -363,14 +362,18 @@ public void CreateDescriptor_BuildsDescriptorsWithInheritedProperties()
         {
             // Arrange
             var errorSink = new ErrorSink();
-            var intProperty = typeof(InheritedSingleAttributeTagHelper).GetProperty(
-                nameof(InheritedSingleAttributeTagHelper.IntAttribute));
+
+            // Also confirm isStringProperty is calculated correctly.
```
NTaylorMullen: Maybe I'm missing something, aren't these comments only valuable in the case that two `TagHelperDescriptor`s target the same attribute with different types?

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -552,11 +561,22 @@ public void TagHelpers_WithinHelpersAndSections_GeneratesExpectedOutput()
             RunTagHelperTest("TagHelpersInSection", tagHelperDescriptors: tagHelperDescriptors);
         }
 
-        private static IEnumerable<TagHelperDescriptor> BuildPAndInputTagHelperDescriptors(string prefix)
+        private static IEnumerable<TagHelperDescriptor> BuildPAndInputTagHelperDescriptors(
+            string prefix,
+            bool switchAttributes)
```
NTaylorMullen: Test method, just default this to false and don't specify it for the existing `IEnumerable<TagHelperDescriptor>` properties.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
dougbu: Actually will unwind some of this (class will be fairly short but I'm not going to remove it) to do the equivalent of `firstBoundAttribute?.IsStringProperty`.  Not adding intentional support for any cases where multiple tag helpers bind the same attribute to properties of different types.  Some scenarios may work but not because we've complicated the code (as I did here) to do it.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
dougbu: No longer relevant but FYI that would not work.  `isStringValue` was `false` in that case but the method was supposed to return `true` then.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
dougbu: Yes. I'm removing these tests since we aren't worrying about the scenario at all.

dougbu: Oops, no.  This is about the combination of `TagHelperDescriptorFactory` and `TagHelperAttributeDescriptor` (where I've changed the line between them, moving the calculation to `TagHelperAttributeDescriptor`) getting the property right for _each_ attribute descriptor.

dougbu:
@NTaylorMullen updated

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    /// <summary>
+    /// Helper methods related to <see cref="TagHelperDescriptor"/> and related classes.
+    /// </summary>
+    public static class TagHelperDescriptorMatcher
```
NTaylorMullen: I'm horrible with names so I'm not one to offer a better one but `TagHelperDescriptorMatcher` doesn't really make sense for what this class does. Makes me think of an equality comparer. Maybe `TagHelperDescriptorUtility` or `TagHelperDescriptorInspector`?

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -14,10 +14,10 @@ namespace Microsoft.AspNet.Razor.Test.Generator
 {
     public class CSharpTagHelperRenderingTest : TagHelperTestBase
     {
-        private static IEnumerable<TagHelperDescriptor> DefaultPAndInputTagHelperDescriptors
-            => BuildPAndInputTagHelperDescriptors(prefix: string.Empty);
-        private static IEnumerable<TagHelperDescriptor> PrefixedPAndInputTagHelperDescriptors
-            => BuildPAndInputTagHelperDescriptors("THS");
+        private static IEnumerable<TagHelperDescriptor> DefaultPAndInputTagHelperDescriptors { get; }
```
NTaylorMullen: lol, you really don't like the C# 6.0 feature do you?

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDescriptorTest.cs
```diff
@@ -0,0 +1,193 @@
...
+using System.Linq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    public class TagHelperDescriptorTest
+    {
+        [Fact]
+        public void TagHelperDescriptor_CanBeSerialized()
```
NTaylorMullen: These are awesome tests, thanks for adding them due to the setter changes :+1:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -923,7 +923,8 @@ public static TheoryData PrefixedTagHelperBoundData
                             new TagHelperAttributeDescriptor(
                                 name: "bound",
                                 propertyName: "Bound",
-                                typeName: typeof(bool).FullName),
+                                typeName: typeof(bool).FullName,
+                                isStringProperty: false),
```
NTaylorMullen: No need to specify `isStringProperty` for these tests, it'll pull from the `typeName` now

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -3861,9 +3954,17 @@ public static TheoryData CodeTagHelperAttributesData
...
                     {
-                        new TagHelperAttributeDescriptor("age", "Age", typeof(int).FullName),
-                        new TagHelperAttributeDescriptor("birthday", "BirthDay", typeof(DateTime).FullName),
-                        new TagHelperAttributeDescriptor("name", "Name", typeof(string).FullName),
+                        new TagHelperAttributeDescriptor("age", "Age", typeof(int).FullName, isStringProperty: false),
+                        new TagHelperAttributeDescriptor(
+                            "birthday",
+                            "BirthDay",
+                            typeof(DateTime).FullName,
+                            isStringProperty: false),
```
NTaylorMullen: dittos

NTaylorMullen:
:watch: just for the `TagHelperDescriptorMatcher` rename.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
dougbu: prefer the C# 6.0 feature which clearly creates just one `IEnumerable<TagHelperDescriptor>`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
dougbu: Nah. This class matches names against `TagHelperDescriptor`s.  It would be called `TagHelperDescriptorComparer` if it were a comparer.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
dougbu: :ok_hand:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorMatcher.cs
NTaylorMullen: `Matcher` is misleading, it doesn't match `TagHelperDescriptor`s in any way; it takes a group of `TagHelperDescriptor`s and derives information from all of them as a whole. It needs a better name.

dougbu:
@NTaylorMullen updated (first commit corrects tests you just added; they no longer work otherwise)

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -56,40 +52,45 @@ public static class TagHelperBlockRewriter
...
             foreach (var child in attributeChildren)
             {
-                KeyValuePair<string, SyntaxTreeNode> attribute;
-                bool succeeded = true;
-
+                TryParseResult result;
                 if (child.IsBlock)
                 {
-                    succeeded = TryParseBlock(tagName, (Block)child, attributeValueTypes, errorSink, out attribute);
+                    result = TryParseBlock(
```
NTaylorMullen: Should be able to put this all on one line (same with the `TryParseSpan` below)

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -235,24 +235,29 @@ private static bool IsSelfClosing(Block beginTagBlock)
...
+                return null;
             }
 
+            bool isBoundNonStringAttribute;
+            var result = new TryParseResult
+            {
+                IsBoundAttribute = IsBoundAttribute(name, descriptors, out isBoundNonStringAttribute),
+                AttributeName = name,
+            };
+            result.IsBoundNonStringAttribute = isBoundNonStringAttribute;
```
NTaylorMullen: Can't do this in the above property setters?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -475,9 +484,37 @@ private static bool IsNullOrWhitespaceAttributeValue(SyntaxTreeNode attributeVal
...
+        // <paramref name="name"/>.
+        private static string GetPropertyType(string name, IEnumerable<TagHelperDescriptor> descriptors)
         {
-            return string.Equals(attributeTypeName, StringTypeName, StringComparison.OrdinalIgnoreCase);
+            var firstBoundAttribute = FindFirstBoundAttribute(name, descriptors);
+
+            return firstBoundAttribute?.TypeName;
+        }
+
+        // Determines whether an attribute named <paramref name="name"/> is bound to a non-<see cref="string"/> tag
```
NTaylorMullen: Still have doc comments in regular comment

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeDescriptorComparer.cs
NTaylorMullen: Since this is controlled via `TypeName` now I'm not sure it's valuable to check.

NTaylorMullen:
All nits, :shipit:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: done

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: Nah.  Though setters are done in order, it's more clear to keep this "after `IsBoundAttribute`" bit separate.  I was tempted not to use constructor initialization at all but decided the mid ground was most understandable.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: fixed

dougbu:
94f2f90

PR #391, opened 2015-05-14
Quick fix: Add `System.Runtime.Serialization.Primitives` dependency

dougbu:
- tests failing
- root cause is likely similar to reason MVC Core has same dependency
 - something must have changed in the framework packages without manifest fixes

### DIFF
--- a/test/Microsoft.AspNet.Razor.Test/project.json
+++ b/test/Microsoft.AspNet.Razor.Test/project.json
@@ -15,7 +15,8 @@
         "dnxcore50": {
             "dependencies": {
                 "System.Diagnostics.TraceSource": "4.0.0-beta-*",
-                "System.Reflection.TypeExtensions": "4.0.0-beta-*"
+                "System.Reflection.TypeExtensions": "4.0.0-beta-*",
+                "System.Runtime.Serialization.Primitives": "4.0.10-beta-*"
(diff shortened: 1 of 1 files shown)


dougbu:
FYI only; will push shortly

Eilon:
Now that https://github.com/aspnet/dnx/pull/1870 is in, can we undo this? Anything else we need to undo?

davidfowl:
We shouldn't undo it. This is good.

davidfowl:
well only if it makes sense I guess

Eilon:
I'm pretty sure this project isn't explicitly using anything from that package. I can't see how this change is right, but I'm happy to be proven wrong.

davidfowl:
Ah ok then nuke it. It shouldn't be in the runtime either.

Eilon:
@dougbu ?

dougbu:
Added it because testing this project failed miserably without it.  Correct that nothing depends on the package but some other thing does an `Assembly.Load()` and that was blowing up with latest bits.

Does the DNX change improve things so an explicit dependency shouldn't be needed here anymore?  Or have the BCL folks fixed a gap in their manifests?

Eilon:
What's the failing callstack on the `Assembly.Load()`?

Issue #217, opened 2014-10-31
Directive errors don't apply error squiggly to full directive.


NTaylorMullen:
For instance if you do:
```
@model User
@inherits SomePage
```

You get a single character error squigly on the `s` of `inherits`. This should span the entire directive.

PR #392, opened 2015-05-14
Applying error squiggly to full directive

ajaybhargavb:
Issue - #217 
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
@@ -2,10 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
-using Microsoft.AspNet.Razor.Editor;
 using Microsoft.AspNet.Razor.Generator;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Tokenizer.Symbols;
@@ -57,15 +55,19 @@ namespace Microsoft.AspNet.Razor.Parser
 
             // Verify we're on "section" and accept
             AssertDirective(SyntaxConstants.CSharp.SectionKeyword);
+            var startLocation = CurrentLocation;
             AcceptAndMoveNext();
 
             if (nested)
             {
-                Context.OnError(CurrentLocation, RazorResources.FormatParseError_Sections_Cannot_Be_Nested(RazorResources.SectionExample_CS));
+                Context.OnError(
+                    startLocation,
+                    RazorResources.FormatParseError_Sections_Cannot_Be_Nested(RazorResources.SectionExample_CS),
+                    Span.GetContent().Value.Length);
                 errorReported = true;
             }
 
-            IEnumerable<CSharpSymbol> ws = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
+            var whitespace = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
 
             // Get the section name
             var sectionName = string.Empty;
@@ -79,19 +81,19 @@ namespace Microsoft.AspNet.Razor.Parser
                 }
 
                 PutCurrentBack();
-                PutBack(ws);
+                PutBack(whitespace);
                 AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: false));
             }
             else
             {
-                Accept(ws);
+                Accept(whitespace);
                 sectionName = CurrentSymbol.Content;
                 AcceptAndMoveNext();
             }
             Context.CurrentBlock.CodeGenerator = new SectionCodeGenerator(sectionName);
 
             var errorLocation = CurrentLocation;
-            ws = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
+            whitespace = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
 
             // Get the starting brace
             var sawStartingBrace = At(CSharpSymbolType.LeftBrace);
@@ -104,7 +106,7 @@ namespace Microsoft.AspNet.Razor.Parser
                 }
 
                 PutCurrentBack();
-                PutBack(ws);
+                PutBack(whitespace);
                 AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: false));
                 Optional(CSharpSymbolType.NewLine);
                 Output(SpanKind.MetaCode);
@@ -113,7 +115,7 @@ namespace Microsoft.AspNet.Razor.Parser
             }
             else
             {
-                Accept(ws);
+                Accept(whitespace);
             }
 
             // Set up edit handler
@@ -184,7 +186,7 @@ namespace Microsoft.AspNet.Razor.Parser
             if (!At(CSharpSymbolType.RightBrace))
             {
                 editHandler.AutoCompleteString = "}";
-                Context.OnError(block.Start, RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(block.Name, "}", "{"));
+                Context.OnError(blockStart, RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(block.Name, "}", "{"));
                 CompleteBlock();
                 Output(SpanKind.Code);
             }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -57,15 +57,19 @@ protected virtual void SectionDirective()
...
-                Context.OnError(CurrentLocation, RazorResources.FormatParseError_Sections_Cannot_Be_Nested(RazorResources.SectionExample_CS));
+                Context.OnError(
+                    startLocation,
+                    RazorResources.FormatParseError_Sections_Cannot_Be_Nested(RazorResources.SectionExample_CS),
+                    Span.GetContent().Value.Length);
                 errorReported = true;
             }
 
-            IEnumerable<CSharpSymbol> ws = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
+            var whitespace = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: false));
```
dougbu: much appreciate this rename :dancers:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -242,11 +247,16 @@ protected void BaseTypeDirective(string noTypeNameError, Func<string, SpanCodeGe
...
             }
+            var remainingWhitespaceLength = Span.GetContent().Value.Length;
+
             AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: true));
 
             if (EndOfFile || At(CSharpSymbolType.WhiteSpace) || At(CSharpSymbolType.NewLine))
             {
-                Context.OnError(CurrentLocation, noTypeNameError);
+                Context.OnError(
+                    CurrentLocation,
```
dougbu: won't this complain about the end of file and not the directive?

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -287,7 +298,9 @@ private void TagHelperDirective(string keyword, Func<string, ISpanCodeGenerator>
 
             if (EndOfFile || At(CSharpSymbolType.NewLine))
             {
-                Context.OnError(CurrentLocation, RazorResources.FormatParseError_DirectiveMustHaveValue(keyword));
+                Context.OnError(CurrentLocation,
```
dougbu: put first argument on its own line

dougbu: also not sure here: does `CurrentLocation` still reference the directive's spot?

src/Microsoft.AspNet.Razor/RazorError.cs
```diff
@@ -77,7 +77,8 @@ public bool Equals(RazorError other)
         {
             return other != null &&
                 string.Equals(other.Message, Message, StringComparison.Ordinal) &&
-                Location.Equals(other.Location);
+                Location.Equals(other.Location) &&
+                Length.Equals(other.Length);
```
dougbu: do not make this change.  `Length` can be changed and therefore cannot be part of the `RazorError`'s identity.

dougbu: if you need this for test purposes, use a test-specific `IEqualityComparer<RazorError>` instead

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: Yup, you'll need to grab the starting location of the keyword. You can get at that value at the top of this block with `Span.Start` (I think).

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: Nope, it'll be at: `@addTagHelper |`

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -318,8 +331,10 @@ private void TagHelperDirective(string keyword, Func<string, ISpanCodeGenerator>
                 if (!startsWithQuote ||
                     !rawValue.EndsWith("\"", StringComparison.OrdinalIgnoreCase))
                 {
-                    Context.OnError(startLocation,
-                                    RazorResources.FormatParseError_DirectiveMustBeSurroundedByQuotes(keyword));
+                    Context.OnError(
+                        startLocation,
+                        RazorResources.FormatParseError_DirectiveMustBeSurroundedByQuotes(keyword),
+                        keywordWithWhitespace.Length + rawValue.Length);
```
NTaylorMullen: The length here will extend past the end of the directive value due to the `startLocation` occurring here: `@addTagHelper |`.

For this error I wouldn't use the length of the `keyword + rawValue.Length` anyways though. It's referencing the value of the directive being an invalid format. So for: `@addTagHelper "*` I'd highlight the `"*`

src/Microsoft.AspNet.Razor/Parser/ParserContext.cs
```diff
@@ -208,6 +208,14 @@ public void OnError(SourceLocation location, string message)
             _errorSink.OnError(location, message);
```
NTaylorMullen: Make the above overload just call into your newly created one:

```C#
public void OnError(SourceLocation location, string message)
{
    // Default error length is 1 unless specified otherwise.
    OnError(location, message, length: 1);
}
```

src/Microsoft.AspNet.Razor/RazorError.cs
NTaylorMullen: This is `Equals` not `GetHashCode`

dougbu: right; carry on :ok_hand:

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/ParserContext.cs
ajaybhargavb: Changed.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -229,7 +233,9 @@ protected void BaseTypeDirective(string noTypeNameError, Func<string, SpanCodeGe
             Context.CurrentBlock.Type = BlockType.Directive;
 
             // Accept whitespace
-            var remainingWs = AcceptSingleWhiteSpaceCharacter();
+            var remainingWhitespace = AcceptSingleWhiteSpaceCharacter();
+            var keywordStartLocation = Span.Start;
```
NTaylorMullen: To avoid confusion, throw this at the start of this method.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -280,6 +291,8 @@ private void TagHelperDirective(string keyword, Func<string, ISpanCodeGenerator>
 
             var foundWhitespace = At(CSharpSymbolType.WhiteSpace);
             AcceptWhile(CSharpSymbolType.WhiteSpace);
+            var keywordStartLocation = Span.Start;
```
NTaylorMullen: Don't need to do this here. Just grab `CurrentLocation` prior to accepting the directive name. Doing `Span.Start` is a worse case :smile:

NTaylorMullen:
Looks good, :shipit: (Note this will probably break the `@model` tests in MVC because it uses the same parsing bits as `@inherits`)

Will need to do this for the directives in MVC as well.

dougbu:
:shipit: once @NTaylorMullen is pleased

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -287,7 +299,10 @@ private void TagHelperDirective(string keyword, Func<string, ISpanCodeGenerator>
 
             if (EndOfFile || At(CSharpSymbolType.NewLine))
             {
-                Context.OnError(CurrentLocation, RazorResources.FormatParseError_DirectiveMustHaveValue(keyword));
+                Context.OnError(
+                    keywordStartLocation,
+                    RazorResources.FormatParseError_DirectiveMustHaveValue(keyword),
+                    keywordWithWhitespace.Length);
```
dougbu: Thought you were removing the whitespace from the error's length?

dougbu:
:shipit: after addressing my one question i.e. making one last bit consistent w.r.t. whitespace

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
ajaybhargavb: Yup. Missed this. Will change.

PR #394, opened 2015-05-18
Add TagHelper support for unbound data- attributes.

NTaylorMullen:
- Involved updating the HtmlMarkupParser to properly separate data- attributes. Prior to this change `data-foo="abc @DateTime.Now def"` would involve 1 Span for `data-foo="abc` 1 Span for `@DateTime.Now` and 1 Span for `def"`. This was very unique behavior from an attribute standpoint (as far as Razor is concerned) and made it difficult for the TagHelper rewriting system to rewrite attributes.
- Added unit tests to validate the various ways you can write unbound data- attributes.
- Updated the BasicTagHelpers codegeneration test to intermix some unbound data- attributes.

#342


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -553,8 +553,15 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
             }
             else
             {
+                // Output the attribute name, the equals and optional quote. Ex: foo="
+                Output(SpanKind.Markup);
+
                 // Not a "conditional" attribute, so just read the value
                 SkipToAndParseCode(sym => IsEndOfAttributeValue(quote, sym));
+
+                // Output the attribute value (will include everything inbetween the attribugte's quotes).
```
dougbu: some typos here: `in-between` and `attribute's`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -553,8 +553,15 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
             }
             else
             {
+                // Output the attribute name, the equals and optional quote. Ex: foo="
+                Output(SpanKind.Markup);
```
dougbu: Exactly why is this addition required?  `SkipToAndParseCode()` often calls `Output()` itself.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -553,8 +553,15 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
...
+
                 // Not a "conditional" attribute, so just read the value
                 SkipToAndParseCode(sym => IsEndOfAttributeValue(quote, sym));
+
+                // Output the attribute value (will include everything inbetween the attribugte's quotes).
+                Output(SpanKind.Markup);
```
dougbu: And why is this necessary?  Is the important thing splitting the parts up versus ensuring `Output()` is called (which already happens a couple of lines below in this block).

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -14,6 +14,167 @@ namespace Microsoft.AspNet.Razor.TagHelpers
...
+                                            factory.Markup(" middle "),
+                                            dateTimeNow)),
+                                }))
+                    },
+                };
+            }
+        }
+
+
+
```
dougbu: Just one blank line here

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Correct, this and the `Output(SpanKind.Markup)` for the name are important for splitting up the attribute. Typically attributes are always split up like so: `class="hello"` => `|class="|hello|"|` or `class="prefix @DateTime.Now suffix"` => `|class="|prefix| @DateTime.Now| suffix|"|`but this wasn't happening for `data-` attributes, it was grouping everything together.

dougbu: Got it. Worth adding some of this to the commit description for posterity.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: Thanks for the [answer](https://github.com/aspnet/Razor/pull/394#discussion_r30569997) below.

dougbu:
:shipit: after addressing minor comments

Issue #199, opened 2014-10-20
Multiple attributes same name (case insensitive) should bind the first, render the rest


NTaylorMullen:
Right now if there's a tag helper with: `<form action="something" ActIon="else">` last attribute always wins.  After talks with @DamianEdwards we should create a parse level error IF the attribute is bound by the `TagHelper`, otherwise since HTML5 attributes are case insensitive: last attribute wins.

yishaigalatzer:
Please verify and close

NTaylorMullen:
Updated title since this bug was ultimately about two issues (one of which is resolved). First issue was insensitive attribute names causes the parsing system to throw, this was fixed a while back. Second issue was we should create a parse level error if there's a duplicate TagHelper bound attribute, right now last wins.

NTaylorMullen:
Decided to wait for user feedback on this, moving to backlog milestone for now.

yishaigalatzer:
This is not ok to move to backlog, it will be a braking change at that point. We should either fix or wont fix. If we want to delay we can do it for RTM. Lets decide on Monday.

NTaylorMullen:
This is the issue we decided to not do/wait for user feedback on in our meeting (I just never moved it).

yishaigalatzer:
Sure, so we have two options in my mind. We either throw right (or before RTM) now at runtime, or we call it supported.

NTaylorMullen:
Right now this issue has to wait for user feedback so if we are to take that approach we should probably call it supported until users feel differently :smile:

yishaigalatzer:
Lets talk in person. This just doesn't work

dougbu:
Reopening per discussions with @NTaylorMullen and @Eilon. In general Razor should not simply ignore anything in the Razor source. An error in the `<element asp-hello="goodbye" asp-hello="dude" />` case is the Right Thing:tm:.

W.r.t. the above discussion, note browsers do _not_ implement "last one wins". Razor currently uses the actual "first attribute wins" semantic when determining values for attributes duplicated in an element.

Separately Razor now handles the `<input type="a" type="b" />` case when either no tag helper targets the `<input/>` element or `type` is not bound. Both attribute settings are passed through for the browser to interpret.

dougbu:
Also cleared the "Won't fix" label.

dougbu:
[...]

PR #395, opened 2015-05-18
Add `[HtmlAttributeName(..., DictionaryAttributePrefix="prefix")]` part III

dougbu:
- #89 remainder
- support adding attributes (that aren't otherwise bound) to a tag helper dictionary
 - add `ObjectCreationExpression`, `Prefix` and more to `TagHelperAttributeDescriptor`
 - handle `Prefix` matches in `TagHelperBlockRewriter`
 - generate code for dictionary property initialization and indexer property assignments
- extend handling of invalid attribute names to include `[HtmlAttributeName]`

known issue:
- `ObjectCreationExpression` is a C# expression because otherwise generation has
  insufficient information to determine correct constructor.
- no VB or F# (for example) workarounds today

nits:
- improve `TagHelperDescriptorFactory_InvalidBoundAttributeName` wording
 - rename resource to `TagHelperDescriptorFactory_InvalidAttributeNameOrPrefixStart`
- correct order of arguments to `FormatTagHelperDescriptorFactory_InvalidBoundAttributeName`
- correct `TagHelperDescriptorFactoryTest` test names
 - remove a few unnecessary `ToArray()` calls

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -220,7 +220,6 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 var descriptor = ToAttributeDescriptor(property, attributeNameAttribute);
                 if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
                 {
-                    // Does this property also support indexer assignments?
                     bool isInvalid;
                     var indexerDescriptor = ToIndexerAttributeDescriptor(
                         property,
@@ -261,11 +260,6 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             Type parentType,
             ErrorSink errorSink)
         {
-            if (attributeDescriptor == null)
-            {
-                return false;
-            }
-
             var nameOrPrefix = attributeDescriptor.IsIndexer ?
                 Resources.TagHelperDescriptorFactory_Prefix :
                 Resources.TagHelperDescriptorFactory_Name;
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -211,15 +211,46 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             var accessibleProperties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
             var attributeDescriptors = new List<TagHelperAttributeDescriptor>();
 
+            // Keep indexer descriptors separate to avoid sorting the combined list later.
+            var indexerDescriptors = new List<TagHelperAttributeDescriptor>();
+
             foreach (var property in accessibleProperties)
             {
-                bool isExplicitPrefix;
-                var descriptor = ToAttributeDescriptor(property, type, errorSink, out isExplicitPrefix);
-                if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink, isExplicitPrefix))
+                var attributeNameAttribute = property.GetCustomAttribute<HtmlAttributeNameAttribute>(inherit: false);
+                var descriptor = ToAttributeDescriptor(property, attributeNameAttribute);
+                if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
                 {
+                    // Does this property also support indexer assignments?
+                    bool isInvalid;
+                    var indexerDescriptor = ToIndexerAttributeDescriptor(
+                        property,
+                        attributeNameAttribute,
+                        parentType: type,
+                        errorSink: errorSink,
+                        defaultPrefix: descriptor.Name + "-",
+                        isInvalid: out isInvalid);
+
+                    if (indexerDescriptor != null &&
+                        !ValidateTagHelperAttributeDescriptor(indexerDescriptor, type, errorSink))
+                    {
+                        isInvalid = true;
+                    }
+
+                    if (isInvalid)
+                    {
+                        // HtmlAttributeNameAttribute was not valid. Ignore this property completely.
+                        continue;
+                    }
+
                     attributeDescriptors.Add(descriptor);
+                    if (indexerDescriptor != null)
+                    {
+                        indexerDescriptors.Add(indexerDescriptor);
+                    }
                 }
             }
+
+            attributeDescriptors.AddRange(indexerDescriptors);
 
             return attributeDescriptors;
         }
@@ -228,31 +259,22 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         internal static bool ValidateTagHelperAttributeDescriptor(
             TagHelperAttributeDescriptor attributeDescriptor,
             Type parentType,
-            ErrorSink errorSink,
-            bool isExplicitPrefix)
+            ErrorSink errorSink)
         {
             if (attributeDescriptor == null)
             {
                 return false;
             }
 
-            var validName = ValidateTagHelperAttributeNameOrPrefix(
+            var nameOrPrefix = attributeDescriptor.IsIndexer ?
+                Resources.TagHelperDescriptorFactory_Prefix :
+                Resources.TagHelperDescriptorFactory_Name;
+            return ValidateTagHelperAttributeNameOrPrefix(
                 attributeDescriptor.Name,
                 parentType,
                 attributeDescriptor.PropertyName,
                 errorSink,
-                Resources.TagHelperDescriptorFactory_Name);
-
-            // Validate an explicit prefix even if name is invalid. Never validate default prefix because that would
-            // only duplicate the name's errors.
-            var validPrefix = !isExplicitPrefix || ValidateTagHelperAttributeNameOrPrefix(
-                attributeDescriptor.Prefix,
-                parentType,
-                attributeDescriptor.PropertyName,
-                errorSink,
-                Resources.TagHelperDescriptorFactory_Prefix);
-
-            return validName && validPrefix;
+                nameOrPrefix);
         }
 
         private static bool ValidateTagHelperAttributeNameOrPrefix(
@@ -264,7 +286,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             if (string.IsNullOrEmpty(attributeNameOrPrefix))
             {
-                // HtmlAttributeNameAttribute validates Name is non-null and non-empty. Both are valid for Prefix.
+                // HtmlAttributeNameAttribute validates Name is non-null and non-empty. Both are valid for
+                // DictionaryAttributePrefix. (Empty DictionaryAttributePrefix is a corner case which would bind every
+                // attribute of a target element. Likely not particularly useful but unclear what minimum length
+                // should be required and what scenarios a minimum length would break.)
                 return true;
             }
 
@@ -320,20 +345,28 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
         private static TagHelperAttributeDescriptor ToAttributeDescriptor(
             PropertyInfo property,
-            Type parentType,
-            ErrorSink errorSink,
-            out bool isExplicitPrefix)
+            HtmlAttributeNameAttribute attributeNameAttribute)
         {
-            isExplicitPrefix = false;
-
-            var attributeNameAttribute = property.GetCustomAttribute<HtmlAttributeNameAttribute>(inherit: false);
             var attributeName = attributeNameAttribute != null ?
                                 attributeNameAttribute.Name :
                                 ToHtmlCase(property.Name);
-            string prefix = null;
-            string objectCreationExpression = null;
-            string prefixedValueTypeName = null;
 
+            return new TagHelperAttributeDescriptor(
+                attributeName,
+                property.Name,
+                property.PropertyType.FullName,
+                isIndexer: false);
+        }
+
+        private static TagHelperAttributeDescriptor ToIndexerAttributeDescriptor(
+            PropertyInfo property,
+            HtmlAttributeNameAttribute attributeNameAttribute,
+            Type parentType,
+            ErrorSink errorSink,
+            string defaultPrefix,
+            out bool isInvalid)
+        {
+            isInvalid = false;
             var dictionaryTypeArguments = ClosedGenericMatcher.ExtractGenericInterface(
                     property.PropertyType,
                     typeof(IDictionary<,>))
@@ -344,6 +377,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 {
                     // DictionaryAttributePrefix is not supported unless associated with an
                     // IDictionary<string, TValue> property.
+                    isInvalid = true;
                     errorSink.OnError(
                         SourceLocation.Zero,
                         Resources.FormatTagHelperDescriptorFactory_InvalidAttributePrefix(
@@ -352,36 +386,26 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                             nameof(HtmlAttributeNameAttribute),
                             nameof(HtmlAttributeNameAttribute.DictionaryAttributePrefix),
                             "IDictionary<string, TValue>"));
-                    return null;
                 }
-            }
-            else
-            {
-                // Potential prefix case.
-                isExplicitPrefix = attributeNameAttribute?.DictionaryAttributePrefixSet == true;
 
-                prefix =  isExplicitPrefix ? attributeNameAttribute.DictionaryAttributePrefix : (attributeName + "-");
-                if (prefix == null)
-                {
-                    // Explicitly set to null. Ignore.
-                    isExplicitPrefix = false;
-                }
-                else
-                {
-                    objectCreationExpression = GetObjectCreationExpression(
-                        property.PropertyType,
-                        dictionaryTypeArguments);
-                    prefixedValueTypeName = dictionaryTypeArguments[1].FullName;
-                }
+                return null;
+            }
+
+            // Potential prefix case. Use default prefix (based on name)?
+            var useDefault = attributeNameAttribute == null || !attributeNameAttribute.DictionaryAttributePrefixSet;
+
+            var prefix = useDefault ? defaultPrefix : attributeNameAttribute.DictionaryAttributePrefix;
+            if (prefix == null)
+            {
+                // DictionaryAttributePrefix explicitly set to null. Ignore.
+                return null;
             }
 
             return new TagHelperAttributeDescriptor(
-                attributeName,
-                property.Name,
-                property.PropertyType.FullName,
-                prefix,
-                objectCreationExpression,
-                prefixedValueTypeName);
+                name: prefix,
+                propertyName: property.Name,
+                typeName: dictionaryTypeArguments[1].FullName,
+                isIndexer: true);
         }
 
         private static bool IsAccessibleProperty(PropertyInfo property)
@@ -407,32 +431,6 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         private static string ToHtmlCase(string name)
         {
             return HtmlCaseRegex.Replace(name, HtmlCaseRegexReplacement).ToLowerInvariant();
-        }
-
-        private static string GetObjectCreationExpression(Type propertyType, Type[] dictionaryTypeArguments)
-        {
-            var propertyTypeInfo = propertyType.GetTypeInfo();
-
-            // First see if default constructor is available and public.
-            if (!propertyTypeInfo.IsAbstract && !propertyTypeInfo.IsInterface)
-            {
-                var defaultConstructor = propertyTypeInfo.DeclaredConstructors
-                    .SingleOrDefault(constructor => constructor.IsPublic && constructor.GetParameters().Length == 0);
-                if (defaultConstructor != null)
-                {
-                    return $"new { TypeNameHelper.GetTypeDisplayName(propertyType, fullName: true) }()";
-                }
-            }
-
-            // Second try Dictionary<TKey, TValue>.
-            var dictionaryType = typeof(Dictionary<,>).MakeGenericType(dictionaryTypeArguments);
-            if (propertyTypeInfo.IsAssignableFrom(dictionaryType.GetTypeInfo()))
-            {
-                return $"new { TypeNameHelper.GetTypeDisplayName(dictionaryType, fullName: true) }()";
-            }
-
-            // Otherwise rely on the tag helper developer or Razor author to initialize the property.
-            return null;
         }
     }
 }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -38,7 +38,9 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// </summary>
         /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
         /// <param name="type">The type to create a <see cref="TagHelperDescriptor"/> from.</param>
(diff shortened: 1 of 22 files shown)


dougbu:
/cc @NTaylorMullen 
Ready to go except for question on whether Razor should initialize `null` dictionary properties.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/PrefixedAttributeTagHelpers.Reversed.cs
```diff
@@ -0,0 +1,203 @@
...
+            }
+            , StartTagHelperWritingScope, EndTagHelperWritingScope);
+            __InputTagHelper2 = CreateTagHelper<InputTagHelper2>();
+            __tagHelperExecutionContext.Add(__InputTagHelper2);
+            if (__InputTagHelper2.IntDictionaryProperty == null)
+            {
+                __InputTagHelper2.IntDictionaryProperty = new System.Collections.Generic.Dictionary<string, int>();
+            }
+#line 15 "PrefixedAttributeTagHelpers.cshtml"
+ __InputTagHelper2.IntDictionaryProperty = intDictionary;
```
dougbu: Slight quirk alert: Each tag helper gets their own dictionary when Razor initializes it, which is quite nice when different helpers have different (higher-priority) individual bound properties matching the prefix.  However setting the dictionary explicitly in the Razor source gives every tag helper the same dictionary instance even if the expression is something like `new Dictionary<string, int>()` due to the way we cache expression results.

dougbu:
Offline decisions for this PR:
1. Split `TagHelperAttributeDescriptor`s for this scenario in two, using a `public bool IsIndexer` (or other name) to distinguish the indexer match from regular / old assignments to the property. This simplifies eventually supporting indexer assignments when the property does not support a dictionary assignment (though we are _not_ doing that get-only feature today).
2. Add parse error when resolved dictionary key is `string.Empty`. This is almost-always an error case and we don't want to help the Razor author add a few characters after the editor auto-fills the prefix. Example error `The tag helper attribute '{attribute name}' is missing a key. The syntax is <{element} {attribute name}key="value" />`.
3. Remove initialization of `null` dictionary properties. This was Just Weird:tm: and encourages the anti-pattern of uninitialized collection properties. (Again, will eventually support get-only dictionary properties to better support normal patterns.)
4. To replace (3), add a runtime check for a `null` dictionary property and `throw` an `InvalidOperationException` with a `RazorPage`-provided `Message`. Check should be done once per tag helper and only when indexer assignment is used.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -209,48 +211,201 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
                 if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
                 {
+                    // Does this property also support indexer assignments?
+                    bool isInvalid;
+                    var indexerDescriptor = ToIndexerAttributeDescriptor(
+                        property,
+                        attributeNameAttribute,
+                        parentType: type,
+                        errorSink: errorSink,
+                        defaultPrefix: descriptor.Name + "-",
```
NTaylorMullen: Instead of passing in the `attributeNameAttribute` and the `defaultPrefix` just pass in the `descriptor` and leave the `defaultPrefix` resolution to the `ToIndexerAttributeDescriptor`. This also enables you to move `attributeNameAttribute` resolution to `ToAttributeDescriptor` like it was before.

dougbu: :-1: doesn't work because `ToIndexerAttributeDescriptor` needs `attributeNameAttribute`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -209,48 +211,201 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
+            var indexerDescriptors = new List<TagHelperAttributeDescriptor>();
+
             foreach (var property in accessibleProperties)
             {
-                var descriptor = ToAttributeDescriptor(property);
+                var attributeNameAttribute = property.GetCustomAttribute<HtmlAttributeNameAttribute>(inherit: false);
+                var descriptor = ToAttributeDescriptor(property, attributeNameAttribute);
                 if (ValidateTagHelperAttributeDescriptor(descriptor, type, errorSink))
                 {
+                    // Does this property also support indexer assignments?
```
NTaylorMullen: Comment confused me a bit. If because if indexerDescriptor != null then the attribute supports indexer assignments.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -209,48 +211,201 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
             var accessibleProperties = type.GetRuntimeProperties().Where(IsAccessibleProperty);
             var attributeDescriptors = new List<TagHelperAttributeDescriptor>();
 
+            // Keep indexer descriptors separate to avoid sorting the combined list later.
```
NTaylorMullen: Why do we care about order here? Aren't we now checking if the dictionary is `null` and then exploding?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: What change are you asking for?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -209,48 +211,201 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
         }
 
-        private static bool ValidateTagHelperAttributeDescriptor(
+        // Internal for testing.
+        internal static bool ValidateTagHelperAttributeDescriptor(
             TagHelperAttributeDescriptor attributeDescriptor,
             Type parentType,
             ErrorSink errorSink)
         {
+            if (attributeDescriptor == null)
```
NTaylorMullen: Can't ever be `null` given the current call sites can it?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: Order matters because all non-`IsIndexer` attribute descriptors in a particular tag helper have higher precedence and _must_ be matched instead of the `IsIndexer` one (if any).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -209,48 +211,201 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
+            string attributeNameOrPrefix,
+            Type parentType,
+            string propertyName,
+            ErrorSink errorSink,
+            string nameOrPrefix)
+        {
+            if (string.IsNullOrEmpty(attributeNameOrPrefix))
+            {
+                // HtmlAttributeNameAttribute validates Name is non-null and non-empty. Both are valid for
+                // DictionaryAttributePrefix. (Empty DictionaryAttributePrefix is a corner case which would bind every
```
NTaylorMullen: Didn't we decide with @Eilon / @DamianEdwards that the `Empty DictionaryAttributePrefix` scenario was invalid?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: Not anymore but it doesn't hurt much. Will remove anyhow...

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Ah bah, I see below that you handled this error case, should remove that part of this comment.

dougbu: Nope. We decided the empty dictionary key -- an attribute name that would result in `property[""] = value` was invalid.

dougbu: > see below that you handled this error case

No, I handle the empty dictionary key error case. Empty `DictionaryAttributePrefix` is supported.

NTaylorMullen: Why would a user ever bind to empty string? Not a blocking question for the PR, just here for discussion

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: I thought we have it planned to explode in the case that you utilize both mechanisms of setting dictionary attributes.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: :ok:  my head was still stuck in the case where everything was 1 attribute. Thought you could derive it from `descriptor`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Comment looks like it's applying to `isInvalid` which isn't the mechanism you use for indicating if a property supports indexer assignments.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: Not sure what you mean. This is about having `asp-route-name` matching its own property even if a dictionary property with prefix `asp-route-` exists. And the semantic I've implemented here is exactly what we decided on.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -69,10 +69,9 @@ public void RenderTagHelper(TagHelperChunk chunk)
 
             // Determine what attributes exist in the element and divide them up.
             var htmlAttributes = chunk.Attributes;
-            var attributeDescriptors = tagHelperDescriptors.SelectMany(descriptor => descriptor.Attributes);
-            var unboundHtmlAttributes = htmlAttributes.Where(
-                attribute => !attributeDescriptors.Any(
-                    descriptor => string.Equals(attribute.Key, descriptor.Name, StringComparison.OrdinalIgnoreCase)));
+            var attributeDescriptors = tagHelperDescriptors.SelectMany(descriptor => descriptor.Attributes).ToArray();
```
NTaylorMullen: Why `.ToArray()`? We only ever consume it as an enumerable.

dougbu: Because the enumeration is executed as many times as there are HTML attributes in the line immediately below.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -190,81 +190,126 @@ protected virtual string GenerateUniqueId()
...
+                            .Write("if (")
+                            .Write(valueAccessor)
+                            .WriteLine(" == null)");
+                        using (_writer.BuildScope())
+                        {
+                            // System is in Host.NamespaceImports for all MVC scenarios. No need to generate FullName
+                            // of InvalidOperationException type.
+                            _writer
+                                .Write("throw ")
+                                .WriteStartNewObject(typeof(InvalidOperationException).Name)
```
NTaylorMullen: Could also do `nameof(InvalidOperationException)`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: removed comment

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: That was a nit, right? I'll do it but could just as well pass `"InvalidOperationException"` since the name can't be refactored to anything else.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
```diff
@@ -190,81 +190,126 @@ protected virtual string GenerateUniqueId()
...
+                if (attributeValueChunk == null)
+                {
+                    // Minimized attributes are not valid for bound attributes. TagHelperBlockRewriter has already
+                    // logged an error if it was a bound attribute; so we can skip.
+                    continue;
+                }
 
-                if (matchingAttributes.Any())
+                // Find the matching TagHelperAttributeDescriptor.
+                var attributeDescriptor = attributeDescriptors.FirstOrDefault(
```
NTaylorMullen: Kinda hack since this is relying on the Indexers always being at the end of the attribute descriptors. Is the plan to eventually create an error if multiple attribute descriptors for a single `TagHelperDescriptor` target the same attribute?

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Yup nit.

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
```diff
@@ -79,6 +80,21 @@ public GeneratedTagHelperContext()
         public string ExecutionContextOutputPropertyName { get; set; }
 
         /// <summary>
+        /// The name of the method used to format an error message about using an indexer when the tag helper property
+        /// is <c>null</c>.
+        /// </summary>
+        /// <remarks>
+        /// Method signature should be
+        /// <code>
+        /// public string FormatInvalidIndexerAssignment(
```
NTaylorMullen: Love the idea for this, we should file an issue to do this for all of the config properties

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -25,48 +26,77 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
...
+        /// if <paramref name="isIndexer"/> is <c>true</c>, the full name of the indexer's value <see cref="Type"/>.
+        /// </param>
+        /// <param name="isIndexer">
+        /// If <c>true</c> this <see cref="TagHelperAttributeDescriptor"/> is used for dictionary indexer assignments.
+        /// Otherwise this <see cref="TagHelperAttributeDescriptor"/> is used for property assignment.
         /// </param>
+        /// <remarks>
+        /// HTML attribute names are matched case-insensitively, regardless of <paramref name="isIndexer"/>.
+        /// </remarks>
         public TagHelperAttributeDescriptor(
```
NTaylorMullen: If it makes it easier, i'd be ok with another internal ctor where it defaults `isIndexer` to `false`. If it's easier for you as is, leave it be :smile:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1045,6 +863,584 @@ public static TheoryData InvalidTagHelperAttributeDescriptorData
...
+                return new TheoryData<string>
+                {
+                    null,
+                    string.Empty,
+                    "data",
+                    "dataa-",
+                    "ValidName",
+                    "valid-name",
+                    "--valid--name--",
+                    ",,--__..oddly.valid::;;",
```
NTaylorMullen: lmao

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1232,5 +1628,81 @@ private class OverriddenValidBoundAttributeWithInvalidUpperCase : TagHelper
             [HtmlAttributeName("DATA-SOMETHING")]
             public string ValidSomething { get; set; }
         }
+
```
NTaylorMullen: Should add cases where `[HtmlAttributeName("valid-name", DictionaryAttributePrefix = null)]`

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -523,28 +600,42 @@ public static TheoryData RuntimeTimeTagHelperTestData
...
         [MemberData(nameof(RuntimeTimeTagHelperTestData))]
-        public void TagHelpers_GenerateExpectedRuntimeOutput(string testName,
-                                                             IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+        public void TagHelpers_GenerateExpectedRuntimeOutput(
+            string testName,
+            string baseLineName,
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
         {
+            // Arrange
+            if (baseLineName == null)
```
NTaylorMullen: Don't have to do this, `RunTagHelperTest` handles it.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/PrefixedAttributeTagHelpers.cshtml
```diff
@@ -0,0 +1,21 @@
...
+        { "name", "value" },
+    };
+}
+
+<div class="randomNonTagHelperAttribute">
+    <input type="checkbox" int-dictionary="intDictionary" string-dictionary="stringDictionary"/>
+    <input type="password" int-dictionary="intDictionary" int-prefix-garlic="37" int-prefix-grabber="42" />
+    <input type="radio"
+           int-prefix-grabber="42" int-prefix-salt="37" int-prefix-pepper="98" int-prefix-salt="8"
+           string-prefix-grabber="string" string-prefix-paprika="another string"/>
```
NTaylorMullen: Add a complex case for `string-prefix` so we can validate in codegen that it's being buffered correctly.

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
dougbu: Recently reopened #199 but I'm not fixing that as part of this PR. Error should be caught in `TagHelperBlockRewriter` anyhow. And `FirstOrDefault()` would still be necessary because there isn't necessarily a property corresponding to every attribute.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
```diff
@@ -523,28 +600,42 @@ public static TheoryData RuntimeTimeTagHelperTestData
...
+                    },
                 };
             }
         }
 
         [Theory]
         [MemberData(nameof(RuntimeTimeTagHelperTestData))]
-        public void TagHelpers_GenerateExpectedRuntimeOutput(string testName,
-                                                             IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+        public void TagHelpers_GenerateExpectedRuntimeOutput(
```
NTaylorMullen: Run it through DesignTime too to validate that the prefix doesn't explode line mappings.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/Generator/Compiler/CodeBuilder/CSharp/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: I understand now based on our later convo in this thread.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Generator/GeneratedTagHelperContext.cs
dougbu: #406

NTaylorMullen:
Note: I'm completely cool with shipping this and delaying the test changes mentioned in my PR comments for the sake of time.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
dougbu: Leaving as-is.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/PrefixedAttributeTagHelpers.cshtml
dougbu: :ok:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
dougbu: Will do this part in a follow-up PR just for the additional tests.

test/Microsoft.AspNet.Razor.Test/Generator/CSharpTagHelperRenderingTest.cs
dougbu: :ok:

dougbu:
Updated

NTaylorMullen:
:shipit:

dougbu:
60c47c8

Issue #44, opened 2014-05-12
Need to evaluate whether any C# 6 features will require changes to the Razor parser


Eilon:
Please don't close this bug until the C#6 features are finalized (which won't be any time soon). We can use this bug to track potential changes we might need to make to Razor. E.g. imagine a new C# 6 feature has some additional `@` signs in it somewhere - we'll likely have to adapt to that.

Eilon:
Also need to support exception filters:

Today with this code:

```c#
try
{
    throw new InvalidOperationException();
}
    catch (Exception e) if (e.Message.Contains("Operation is not valid"))
{
}
```

You get this error
```
CompilationFailedException: Compilation for 'c:\users\mattgal\documents\visual studio 2015\Projects\WebApplication1\src\WebApplication1\Views\Home\Index.cshtml' failed:
Expected a "{" but found a "if". Block statements must be enclosed in "{" and "}". You cannot use single-statement control-flow statements in CSHTML pages. For example, the following is not allowed:

@if(isLoggedIn)
<p>Hello, @user</p>

Instead, wrap the contents of the block in "{}":

@if(isLoggedIn) {
<p>Hello, @user</p>
} 
```

From TFS bug 1099131.

NTaylorMullen:
I investigated and it looks like the only pieces that aren't working are:

1. Exception Filters
1. Null-Conditional for expressions (works fine in statement blocks)

/cc @DamianEdwards Do you feel strongly one way or the other about exception filters making it?

Eilon:
Just making sure these are tested:

***Using Static Members***
* Tricky because it could change how we parse `@using ...` (or maybe not)
```c#
using static System.Console;
```

***Expression-bodied members***
* Tricky because there's an extra `>` where we might not expect it (or maybe not)
```c#
public double Dist => Sqrt(X * X + Y * Y);  
```

***String interpolation***
* Tricky because all kinds of weird stuff can happen inside a string
```c#
$"{p.Name} is {p.Age} years old."  

var @class = "History";
var happiness = 6;
$"I like taking {@class} class? {(happiness > 5).ToString()}."  
```

NTaylorMullen:
## Status

***Null Propagation***
Breaks by using them in implicit expressions such as `@Foo?.Bar`, `@Foo?()`, `@Foo?[0]`.

***Using Static Members***
Breaks in the scenario @Eilon mentioned: `@using static System.Console`

***String Interpolation***
Breaks in the following scenario.
```
@{
[...]

PR #396, opened 2015-05-18
Add support for null conditionals in implicit expressions.

NTaylorMullen:
- Added case in ImplicitExpression handling to understand question marks.
- text?. is special compared to text. because with text. we currently validate content after text. to determine if it's an expression or if it's a period. Now with ?. we always treat it as an expression because ?. is not a useful sentance of any kind.
- Added unit tests to validate new implicit expression handling
- Added runtime and design time code generation tests to validate null propagation.

#44

*Note: Will need to evaluate the editing experience in the editor once VS bits become available due to the complex nature of implicit expressions.*

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
@@ -388,7 +388,32 @@ namespace Microsoft.AspNet.Razor.Parser
                     }
                     return MethodCallOrArrayIndex(acceptedCharacters);
                 }
-                if (CurrentSymbol.Type == CSharpSymbolType.Dot)
+                if (At(CSharpSymbolType.QuestionMark))
+                {
+                    var next = Lookahead(count: 1);
+
+                    if (next != null)
+                    {
+                        if (next.Type == CSharpSymbolType.Dot)
+                        {
+                            // Accept null conditional dot operator (?.).
+                            AcceptAndMoveNext();
+                            AcceptAndMoveNext();
+
+                            // If the next piece after the ?. is a keyword or identifier then we want to continue.
+                            return At(CSharpSymbolType.Identifier) || At(CSharpSymbolType.Keyword);
+                        }
+                        else if (next.Type == CSharpSymbolType.LeftBracket)
+                        {
+                            // We're at the ? for a null conditional bracket operator (?[).
+                            AcceptAndMoveNext();
+
+                            // Accept the [ and any content inside (it will attempt to balance).
+                            return MethodCallOrArrayIndex(acceptedCharacters);
+                        }
+                    }
+                }
+                else if (At(CSharpSymbolType.Dot))
                 {
                     var dot = CurrentSymbol;
                     if (NextToken())
--- a/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNet.Razor.Parser;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Test.Framework;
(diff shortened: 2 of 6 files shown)


test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
```diff
@@ -17,6 +18,105 @@ public override ParserBase CreateCodeParser()
...
+                    { "foo?!", "foo" },
+                    { "foo?.", "foo?." },
+                    { "foo??.", "foo" },
+                    { "foo?.(abc)", "foo?." },
+                    { "foo?.<p>", "foo?." },
+                    { "foo?.bar", "foo?.bar" },
+                    { "foo?.bar<p>", "foo?.bar" },
+                    { "foo??.bar<p>", "foo" },
+                    { "foo?.bar(false)?.<p>", "foo?.bar(false)?." },
+                    { "foo?.bar(false)?.baz", "foo?.bar(false)?.baz" },
```
dougbu: - mix some tests like `foo??bar` in here
- avoid "foo", "bar", et cetera when adding new tests

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/NullPropagationExpressions.cshtml
```diff
@@ -0,0 +1,13 @@
+﻿@{
+    @ViewBag?.Data
+    @ViewBag.Method?(Hello World)
```
dougbu: Has support for `Method?(...)` been added to the language recently?  Null-conditional syntax previously included only properties (`?.Property`) and indexers (`?[index]`).

dougbu: self-nit: `Property?.Method()` also works :smiley:

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
```diff
@@ -17,6 +18,105 @@ public override ParserBase CreateCodeParser()
...
+                    { "foo?(", "foo?(", AcceptedCharacters.Any, missingEndParenError(5) },
+                    { "foo?(bar", "foo?(bar", AcceptedCharacters.Any, missingEndParenError(5) },
+                    { "foo?(abc)", "foo?(abc)", AcceptedCharacters.NonWhiteSpace, noErrors },
+                    { "foo?(<p>", "foo?(", AcceptedCharacters.Any, missingEndParenError(5) },
+                    { "foo?(bar.<p>", "foo?(bar.", AcceptedCharacters.Any, missingEndParenError(5) },
+                    { "foo??(bar<p>", "foo", AcceptedCharacters.NonWhiteSpace, noErrors },
+                    { "foo?(abc)?", "foo?(abc)", AcceptedCharacters.NonWhiteSpace, noErrors },
+                    { "foo?(abc)?(def", "foo?(abc)?(def", AcceptedCharacters.Any, missingEndParenError(11) },
+                    { "foo?(abc)?(def)", "foo?(abc)?(def)", AcceptedCharacters.NonWhiteSpace, noErrors },
+                    { "foo?(abc)?.bar?(def)", "foo?(abc)?.bar?(def)", AcceptedCharacters.NonWhiteSpace, noErrors },
```
dougbu: Doubt any of the above are valid C#.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -388,7 +388,33 @@ private bool MethodCallOrArrayIndex(AcceptedCharacters acceptedCharacters)
...
+
+                    if (next != null)
+                    {
+                        if (next.Type == CSharpSymbolType.Dot)
+                        {
+                            // Accept null propagation dot operator (?.).
+                            AcceptAndMoveNext();
+                            AcceptAndMoveNext();
+
+                            // If the next piece after the ?. is a keyword or identifier then we want to continue.
```
dougbu: What keyword should work in this context?  E.g. `Value?.nameof(...)` isn't valid.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -388,7 +388,33 @@ private bool MethodCallOrArrayIndex(AcceptedCharacters acceptedCharacters)
...
+                            return At(CSharpSymbolType.Identifier) || At(CSharpSymbolType.Keyword);
+                        }
+                        else if (next.Type == CSharpSymbolType.LeftParenthesis ||
+                            next.Type == CSharpSymbolType.LeftBracket)
+                        {
+                            // It's either ?[ or ?(, accept the question mark because it's a null propagation operator.
```
dougbu: Again, what does `?(` mean?

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -388,7 +388,33 @@ private bool MethodCallOrArrayIndex(AcceptedCharacters acceptedCharacters)
...
-                if (CurrentSymbol.Type == CSharpSymbolType.Dot)
+                if (At(CSharpSymbolType.QuestionMark))
+                {
+                    var next = Lookahead(count: 1);
+
+                    if (next != null)
+                    {
+                        if (next.Type == CSharpSymbolType.Dot)
+                        {
+                            // Accept null propagation dot operator (?.).
```
dougbu: This feature is now consistently called "Null-conditional operators".  See https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
dougbu: https://github.com/dotnet/roslyn/wiki/New-Language-Features-in-C%23-6#null-conditional-operators specifically says
> Note that an invocation (a parenthesized argument list) cannot immediately follow the ? operator – that would lead to too many syntactic ambiguities.

The workaround for delegates is a null-conditional `Invoke` call e.g. `delegate?.Invoke(...)`

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/NullPropagationExpressions.cshtml
NTaylorMullen: Oh man I thought that Method?() was supported for delegates. Just tried it out, I be wrong, will remove.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
NTaylorMullen: How is that not valid? If there's a property which happens to be a delegate named `nameof` it works fine. Same for `class`, `typeof`, `for` etc.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/396#discussion_r30571892

NTaylorMullen:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
```diff
@@ -17,6 +18,96 @@ public override ParserBase CreateCodeParser()
...
+        public static TheoryData NullConditionalOperatorData_Dot
+        {
+            get
+            {
+                // implicitExpression, expectedImplicitExpression
+                return new TheoryData<string, string>
+                {
+                    { "val?", "val" },
+                    { "val??", "val" },
+                    { "val??more", "val" },
```
dougbu: Not something that should change in this PR but this case is interesting. Does the test mean null coalescing expressions aren't supported in implicit expressions, at least outside brackets or parentheses? Might be worth a bug since this case is unlikely to be an typo e.g. would expect whitespace after `??` or `???` if this were English.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Output/NullConditionalExpressions.DesignTime.cs
```diff
@@ -0,0 +1,99 @@
...
+#line hidden
+#line 4 "NullConditionalExpressions.cshtml"
+                               
+    
+
+#line default
+#line hidden
+
+#line 5 "NullConditionalExpressions.cshtml"
+__o = ViewBag?.Method(Value?[23]?.More)(SecondValue)?["key"];
```
dougbu: Not something that should change in this PR but might be worth filing a bug to remove this corner case from implicit expression support. Adjacent parenthesized expressions happen only when you have a `Func<..., Action<...>>`, `Func<..., Func<...>>` or similar nested delegates and you're calling the whole thing in one expression. Seems more likely the expression ends after the first method invocation and the next part begins a new parenthesized sentence.

Fix relative to this PR's version of the files would be at https://github.com/aspnet/Razor/pull/396/files#diff-d2cd6cc81ade624cba92cccb004b6afdR389

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/NullConditionalExpressions.cshtml
```diff
@@ -0,0 +1,11 @@
...
+    @ViewBag?.Data
+    @ViewBag.IntIndexer?[0]
+    @ViewBag.StrIndexer?["key"]
+    @ViewBag?.Method(Value?[23]?.More)(SecondValue)?["key"]
+}
+
+@ViewBag?.Data
+@ViewBag.IntIndexer?[0]
+@ViewBag.StrIndexer?["key"]
+@ViewBag?.Method(Value?[23]?.More)(SecondValue)?["key"]
```
dougbu: Why bother testing nested delegates to this extent? It's a corner case that should be visible in at most one or two low-level (`CSharpImplicitExpressionTest`) cases.

dougbu:
> What keyword should work in this context? E.g. Value?.nameof(...) isn't valid.

> How is that not valid? If there's a property which happens to be a delegate named nameof it works fine. Same for class, typeof, for etc.

`nameof` was a bad example in my comment. You can define a delegate or other property / field named `nameof` or using most other contextual keywords. However the same is not true of `class` or `typeof` unless used with in a C# verbatim (`@`) identifier.

Unfortunately I now see the Razor list of C# keywords (in `CSharpKeywordDetector`) mixes contextual keywords in with `class` and other non-contextual keywords. Carry on :frowning:

dougbu:
:shipit: after minimizing testing of the extreme corner case mentioned above.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpImplicitExpressionTest.cs
NTaylorMullen: Nope, not supported. If you want to do pieces of code that typically have spaces the route to take is `@(...)`. Same reason why `@a+1` doesn't work :smile:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/NullConditionalExpressions.cshtml
NTaylorMullen: This is just combining everything in one, will remove the nested delegate part of this.

Issue #379, opened 2015-05-06
Whitespace at the end of a <text> tag should not be written to output.


ajaybhargavb:
Related issue - #50 

Example:
```C#
 <div>@{<text>@a</text> ↵}</div>
```
produces
```C#
WriteLiteral("  \r\n");
```
instead of
```C#
WriteLiteral("\r\n");
```

The whitespace between `</text>` and `}` should not be written to output.

cc' @dougbu 

(Note:  The ↵ denotes a new line - added that to make the whitespace visible)

ajaybhargavb:
Currently we accept all whitespaces at the end of a tag block. We should special case it for the `<text>` tag.
https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs#L1006

PR #397, opened 2015-05-18
Ignoring whitespace at the end of text tag

ajaybhargavb:
Issue - #379 
@dougbu @NTaylorMullen


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1043,6 +1043,11 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
             if (!Context.DesignTimeMode)
             {
                 AcceptWhile(HtmlSymbolType.WhiteSpace);
+                if (Context.LastSpan.Kind == SpanKind.Transition)
+                {
+                    // Ignore the whitespace at the end of a <text> tag.
+                    Span.ClearSymbols();
```
NTaylorMullen: Can't clear symbols because it'd destroy the tree without accounting for the symbols. The space after `<text>` should be qualified as code.

ajaybhargavb:
Updated. Need to change a few more tests. I'll do that once the product code is accepted.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1043,6 +1043,12 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
             if (!Context.DesignTimeMode)
             {
                 AcceptWhile(HtmlSymbolType.WhiteSpace);
+                if (Context.LastSpan.Kind == SpanKind.Transition)
+                {
+                    // Mark the whitespace at the end of a <text> tag as code.
+                    Span.CodeGenerator = new StatementCodeGenerator();
+                    Output(SpanKind.Code);
```
NTaylorMullen: Dangerous to just `Output(SpanKind.Code)` blindly here. Your goal is to only output the whitespace but another `Output` has not occurred prior to this code execution within this method.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/CS/Source/CodeBlockWithTextElement.cshtml
```diff
@@ -1,4 +1,4 @@
 ﻿@{
-    var a = 1; <text>@a</text>
```
NTaylorMullen: Hmm these tests are all super odd. Since you're touching it can you modify these text tags to have more than just C# expressions? Purpose of them is to output literal markup but we don't actually verify that. Doing this will also make it easier to visualize your changes

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1043,6 +1043,12 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
+                }
                 if (!EndOfFile && CurrentSymbol.Type == HtmlSymbolType.NewLine)
```
dougbu: Add a blank line here to keep things separate

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: @NTaylorMullen is your suggestion to add `Output(SpanKind.Markup)` (matching the `Output()` call at the end of this method) before accepting the non-newline whitespace?  If yes, :+1: from me.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Correct :smile:

NTaylorMullen: Hmm, actually in order to make this truly only affect after transition bits it'd probably be best to only `Output(SpanKind.Markup)` prior to the whitespace IF the last span happens to be a transition. This way the change doesn't affect non after transition based scenarios.

ajaybhargavb:
Updated.

dougbu:
:shipit: when @NTaylorMullen is pleased

NTaylorMullen:
solid, :shipit:

PR #398, opened 2015-05-19
Quick Fix: Remove unneeded NamespaceKeywordLength from AddImportCodeGenerator.

NTaylorMullen:
- This was completely unused outside of the AddImportCodeGenerator.


NTaylorMullen:
FYI only; will push shortly

dougbu:
fyi :shipit:

PR #400, opened 2015-05-19
Add support for static usings.

NTaylorMullen:
- Added special cases to the using statement parser to understand `@using static Foo`.
- Added several unit tests to validate the static using structure.
- Modified existing import code generation tests to validate several static usings.

#44 

*Note: ?w=1 is your friend when reviewing this PR*

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
@@ -101,7 +101,7 @@ namespace Microsoft.AspNet.Razor.Parser
                 // using ( ==> Using Statement
                 UsingStatement(block);
             }
-            else if (At(CSharpSymbolType.Identifier))
+            else if (At(CSharpSymbolType.Identifier) || At(CSharpKeyword.Static))
             {
                 // using Identifier ==> Using Declaration
                 if (!topLevel)
@@ -126,31 +126,40 @@ namespace Microsoft.AspNet.Razor.Parser
             // Set block type to directive
             Context.CurrentBlock.Type = BlockType.Directive;
 
-            // Parse a type name
-            Assert(CSharpSymbolType.Identifier);
-            NamespaceOrTypeName();
-            IEnumerable<CSharpSymbol> ws = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
-            if (At(CSharpSymbolType.Assign))
+            if (At(CSharpSymbolType.Identifier))
             {
-                // Alias
-                Accept(ws);
-                Assert(CSharpSymbolType.Assign);
-                AcceptAndMoveNext();
-
-                AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
-
-                // One more namespace or type name
+                // non-static using
                 NamespaceOrTypeName();
+                var whitespace = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
+                if (At(CSharpSymbolType.Assign))
+                {
+                    // Alias
+                    Accept(whitespace);
+                    Assert(CSharpSymbolType.Assign);
+                    AcceptAndMoveNext();
+
+                    AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
+
+                    // One more namespace or type name
+                    NamespaceOrTypeName();
+                }
+                else
+                {
+                    PutCurrentBack();
+                    PutBack(whitespace);
+                }
             }
-            else
+            else if (At(CSharpKeyword.Static))
             {
-                PutCurrentBack();
-                PutBack(ws);
+                // static using
+                AcceptAndMoveNext();
+                AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: true));
+                NamespaceOrTypeName();
             }
 
             Span.EditHandler.AcceptedCharacters = AcceptedCharacters.AnyExceptNewline;
             Span.CodeGenerator = new AddImportCodeGenerator(
-                Span.GetContent(syms => syms.Skip(1)));
+                Span.GetContent(symbols => symbols.Skip(1)));
 
             // Optional ";"
             if (EnsureCurrent())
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -126,26 +126,35 @@ private void UsingDeclaration()
...
-            if (At(CSharpSymbolType.Assign))
+            if (At(CSharpSymbolType.Identifier))
             {
-                // Alias
-                Accept(ws);
-                Assert(CSharpSymbolType.Assign);
-                AcceptAndMoveNext();
+                // non-static using
+                NamespaceOrTypeName();
+                IEnumerable<CSharpSymbol> ws = ReadWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
```
dougbu: Fix hideous variable name (to `whitespace`) while here.

dougbu: And use `var` more.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -126,26 +126,35 @@ private void UsingDeclaration()
...
+                }
             }
-            else
+            else if (At(CSharpKeyword.Static))
             {
-                PutCurrentBack();
-                PutBack(ws);
+                // static using
+                AcceptAndMoveNext();
+                AcceptWhile(IsSpacingToken(includeNewLines: false, includeComments: true));
```
dougbu: Why must the identifier be on the same line as `@using static`? In a regular .cs file, the following is fine:
``` c#
using
    static
    System.Console;
```

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: That's how our `@using` imports have always been. In general, Razor directives populate only a single line.

Not to mention having an errored using like so:

```HTML
@using 
Favorite Movie: <strong>Abc</strong>
```
Would be terribly confusing to the user because Razor would attempt to do: `using Favorite` at the top of the generated C# file.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: `?w=1` your friend :smile: 

Will fix up the bits I indented.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: Also note: semicolons at the end of an `@using` statement are optional in Razor

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
Eilon: The real reason: Razor isn't C#, so Razor has its own rules.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -126,31 +126,40 @@ private void UsingDeclaration()
...
+                NamespaceOrTypeName();
             }
 
             Span.EditHandler.AcceptedCharacters = AcceptedCharacters.AnyExceptNewline;
             Span.CodeGenerator = new AddImportCodeGenerator(
-                Span.GetContent(syms => syms.Skip(1)));
+                Span.GetContent(symbols => symbols.Skip(1)));
```
dougbu: :dancers:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -126,31 +126,40 @@ private void UsingDeclaration()
             // Set block type to directive
             Context.CurrentBlock.Type = BlockType.Directive;
 
-            // Parse a type name
-            Assert(CSharpSymbolType.Identifier);
```
dougbu: Great to see this `Assert()` go away. Using `Debug.Assert()` to inform the user of an error parsing their Razor file was more than a bit odd.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
```diff
@@ -148,6 +149,53 @@ public void FinallyClause()
...
+                        factory.Code(code)
+                            .AsNamespaceImport(import)
+                            .Accepts(AcceptedCharacters.AnyExceptNewline));
+
+                // document, expectedResult
+                return new TheoryData<string, DirectiveBlock>
+                {
+                    { "@using static", createUsing("using static", " static") },
+                    { "@using static    ", createUsing("using static    ", " static    ") },
+                    { "@using         static    ", createUsing("using         static    ", "         static    ") },
```
dougbu: Do the above 3 cases result in a Razor error? Or does it just cause a compilation error e.g. because the name of the static class (on the next line) is taken as HTML?

dougbu:
:watch: for one open question

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
NTaylorMullen: Causes a compilation error. I considered creating parser errors for these but after investigating what we did for normal `@using` imports (we don't create errors) I decided to follow the same pattern.

NTaylorMullen:
This good to go?

dougbu:
Yes. :shipit:

PR #405, opened 2015-05-20
Marking newline after a text tag as markup

ajaybhargavb:
Followup bug fix to issue - #379
This should also fix Mvc build
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -1049,19 +1049,22 @@ namespace Microsoft.AspNet.Razor.Parser
             tags.Clear();
             if (!Context.DesignTimeMode)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1049,19 +1049,22 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
             tags.Clear();
             if (!Context.DesignTimeMode)
             {
-                if (Context.LastSpan.Kind == SpanKind.Transition)
+                if (CurrentSymbol != null && CurrentSymbol.Type == HtmlSymbolType.WhiteSpace)
```
NTaylorMullen: `At(HtmlSymbolType.WhiteSpace)`

NTaylorMullen:
:shipit:

dougbu:
:shipit:

Issue #361, opened 2015-04-15
Support * wild card in TargetElement Attribute.


NTaylorMullen:
This will enable `TagHelper` authors to target `IEnumerable<KeyValuePair>` attributes (https://github.com/aspnet/Razor/issues/89).

Essentially:

```C#
[TargetElement(Attributes = "q-*")]
public class QueryParamTagHelper : TagHelper { }
```

Use:

```HTML
<!-- A TagHelper because of q-custom -->
<a href="~/home/index" q-custom="AValue">Go to index</a> 
```

danroth27:
@DamianEdwards How important is this? Do we want this for beta5?

DamianEdwards:
Yes, it's needed for the anchor Tag Helper when we finish the work to support binding to dictionary properties.

NTaylorMullen:
0e61b49881f77fb9123792ea12eb6f5e4411078d

PR #408, opened 2015-05-21
Add support for '*' at the end of required attributes.

NTaylorMullen:
- `[TargetElement(Attributes ="prefix-*")]` is now supported.
- Added '*' to the list of invalid non whitespace characters in `TagHelperDescriptorFactory`.
- Modified `TagHelperDescriptorProvider` to respect suffixed wildcards in `TagHelperAttributeDescriptor.Attributes`.
- Added tests to validate wildcard required attributes

#361

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         // https://github.com/aspnet/Razor/issues/165
 
         public static ICollection<char> InvalidNonWhitespaceNameCharacters { get; } = new HashSet<char>(
-            new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'' });
+            new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'', '*' });
 
         /// <summary>
         /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="type"/>.
@@ -169,6 +169,25 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             bool targetingAttributes,
             ErrorSink errorSink)
         {
+            if (!targetingAttributes &&
+                string.Equals(
+                    name,
+                    TagHelperDescriptorProvider.ElementCatchAllTarget,
+                    StringComparison.OrdinalIgnoreCase))
+            {
+                // '*' as the entire name is OK in the TargetElement catch-all case.
+                return true;
+            }
+            else if (targetingAttributes &&
+                name.EndsWith(
+                    TagHelperDescriptorProvider.RequiredAttributeWildcardSuffix,
+                    StringComparison.OrdinalIgnoreCase))
+            {
+                // A single '*' at the end of a required attribute is valid; everywhere else is invalid. Strip it from
+                // the end so we can validate the rest of the name.
+                name = name.Substring(0, name.Length - 1);
+            }
+
             var targetName = targetingAttributes ?
                 Resources.TagHelperDescriptorFactory_Attribute :
                 Resources.TagHelperDescriptorFactory_Tag;
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
@@ -12,16 +12,16 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
     public sealed class TargetElementAttribute : Attribute
     {
-        public const string CatchAllDescriptorTarget = TagHelperDescriptorProvider.CatchAllDescriptorTarget;
+        public const string ElementCatchAllTarget = TagHelperDescriptorProvider.ElementCatchAllTarget;
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


dougbu:
Need to update the `TagHelperDescriptor.RequiredAttributes`

dougbu:
... documentation

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -169,6 +169,26 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
             ErrorSink errorSink)
         {
+            if (!targetingAttributes &&
+                string.Equals(
+                    name,
+                    TagHelperDescriptorProvider.CatchAllDescriptorTarget,
+                    StringComparison.OrdinalIgnoreCase))
+            {
+                // '*' as the entire name is ok if we're not targeting attributes because it's the TargetElement
+                // catch-all case.
```
dougbu: Slightly confusing. Suggest `// '*' as the entire name is OK in the TargetElement catch-all case.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
```diff
@@ -41,14 +41,17 @@ public TargetElementAttribute(string tag)
...
         public string Tag { get; }
 
         /// <summary>
-        /// A comma-separated <see cref="string"/> of attributes the HTML element must contain for the 
+        /// A comma-separated <see cref="string"/> of attributes the HTML element must contain for the
         /// <see cref="ITagHelper"/> to run.
         /// </summary>
+        /// <remarks>
+        /// <c>*</c> at the end of an attribute acts as a wild card.
+        /// </remarks>
```
dougbu: While here please change `attribute` -> `attribute name` throughout this property's documentation.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -14,6 +14,8 @@ public class TagHelperDescriptorProvider
     {
         public const string CatchAllDescriptorTarget = "*";
 
+        public static readonly string RequiredAttributeWildcardSuffix = "*";
```
dougbu: Be consistent with `CatchAllDescriptorTarget` or just remove this property. Suggest both should be `public static readonly` since `CatchAllDescriptorTarget` is not pi.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -14,6 +14,8 @@ public class TagHelperDescriptorProvider
     {
         public const string CatchAllDescriptorTarget = "*";
```
dougbu: If we stick with two properties, should probably rename to `ElementCatchAllTarget` or something more clearly distinct from `RequiredAttributeWildcardSuffix`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -91,7 +93,21 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(string tagName, IEnumerab
...
+                        // '*' at the end of a required attribute indicates: apply to attributes prefixed with the
+                        // required attribute value.
+                        if (requiredAttribute.EndsWith(
+                            RequiredAttributeWildcardSuffix,
+                            StringComparison.OrdinalIgnoreCase))
+                        {
+                            var prefix = requiredAttribute.Substring(0, requiredAttribute.Length - 1);
+
+                            if (!attributeNames.Any(
+                                attributeName => attributeName.StartsWith(prefix, StringComparison.OrdinalIgnoreCase)))
```
dougbu: Should not successfully match the required attribute when `attributeName` is exactly equal to `prefix`. Such a match and a dictionary property with the same prefix will only lead to parse errors.

NTaylorMullen: :+1:

dougbu:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
NTaylorMullen: The `CatchAllDescriptorTarget` is purposefully `public const` so that the `TargetElement` can expose it as a `public const` for users to consume. The `RequiredAttributeWildcardSuffix` isn't supposed to be easily consumable by users so it's just here as `public static readonly` so we can use the same value throughout our code.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
```diff
@@ -41,14 +41,17 @@ public TargetElementAttribute(string tag)
...
         /// that targets all HTML elements with the required <see cref="Attributes"/>.</remarks>
         public string Tag { get; }
 
         /// <summary>
-        /// A comma-separated <see cref="string"/> of attributes the HTML element must contain for the 
+        /// A comma-separated <see cref="string"/> of attribute names the HTML element must contain for the
         /// <see cref="ITagHelper"/> to run.
         /// </summary>
+        /// <remarks>
+        /// <c>*</c> at the end of an attribute name acts as a wild card.
```
dougbu: On second look, I'm not sure the "wild card" mention helps. The asterix changes the `Attributes` entry to a prefix match, nothing as general as a wildcard.

dougbu:
:shipit: with minor doc change

NTaylorMullen:
Will get this in once checkins open back up.

Issue #324, opened 2015-03-17
Don't special case '*' retrieval in TagHelperDescriptorProvider


NTaylorMullen:
Associated convo: https://github.com/aspnet/Razor/pull/316/files#r26606226

danroth27:
@NTaylorMullen The convo is too big. Please paste in the relevant points here. Thanks.

dougbu:
simply about removing code (a few lines) that is never used in any realistic scenario.  exists in case someone uses a low-level internal detail of our tag helper resolution system in a somewhat strange way.  note tag helper authors don't even know `TagHelperDescriptor` exists.

NTaylorMullen:
3ba92523ff90c9d77fc8c1dd5e722927e10e21c1

PR #411, opened 2015-05-29
Remove special casing of '*' in TagHelperDescriptorProvider.

NTaylorMullen:
- Requesting a '*' tagName from the TagHelperDescriptorProvider could only happen if a user was directly calling into it (extremely unlikely). Therefore I've removed the special casing to make the logic more simple.
- Removed tests that expected this behavior.

#324


dougbu:
:shipit:

PR #413, opened 2015-05-31
Avoid errors when generating baselines

dougbu:
- ignore missing expected output resources in this mode
- update `BaselineWriter` to avoid `IOException`s (file access issues)
 - serialize file operations
 - one file handle per file
- repeatedly `rm TestFiles/CodeGenerator/Output/*; `dnx . test`; no errors

also
- update files only if they have changed
 - new `TestFiles.Exists()` method to support this check
- do not check results in `CSharpCodeBuilderTests` if `GENERATE_BASELINES` set

nits:
- update BOM in CodeBlockWithTextElement.cs to avoid future `git diff`s
- use more `var` and improve variable names in `TestFile`
- wrap long lines

### DIFF
--- a/test/Microsoft.AspNet.Razor.Test/CodeGenerators/RazorChunkGeneratorTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/CodeGenerators/RazorChunkGeneratorTest.cs
@@ -195,15 +195,15 @@ namespace Microsoft.AspNet.Razor.Test.Generator
 
             var textOutput = results.GeneratedCode;
 #if GENERATE_BASELINES
+            var outputFile = string.Format(
+                @"test\Microsoft.AspNet.Razor.Test\TestFiles\CodeGenerator\Output\{0}.{1}",
+                baselineName,
+                BaselineExtension);
+
             // Update baseline files if files do not already match.
             if (!string.Equals(expectedOutput, textOutput, StringComparison.Ordinal))
             {
-            BaselineWriter.WriteBaseline(
-                    string.Format(
-                        @"test\Microsoft.AspNet.Razor.Test\TestFiles\ChunkGenerator\Output\{0}.{1}",
-                        baselineName,
-                        BaselineExtension),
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNet.Razor.Test/CodeGenerators/RazorChunkGeneratorTest.cs
```diff
@@ -216,15 +232,68 @@ private Stream NormalizeNewLines(Stream inputStream)
...
+
+            var indent = "            ";
+            lineMappings.AppendLine($"{ indent }var expectedLineMappings = new[]");
+            lineMappings.AppendLine($"{ indent }{{");
+            foreach (var lineMapping in designTimeLineMappings)
+            {
+                var innerIndent = indent + "    ";
+                var documentLocation = lineMapping.DocumentLocation;
+                var generatedLocation = lineMapping.GeneratedLocation;
+                lineMappings.AppendLine($"{ innerIndent }BuildLineMapping(");
```
NTaylorMullen: `nameof(RazorCodeGeneratorTest.BuildLineMapping)`

NTaylorMullen:
:shipit:

test/Microsoft.AspNet.Razor.Test/CodeGenerators/RazorChunkGeneratorTest.cs
dougbu: Not possible since `BuildLineMapping` is defined in a subclass.

Thanks for helping me see how badly out-of-sync the `RazorCodeGeneratorTest<TLanguage>` and `RazorCodeGeneratorTest` classes are with their containing files (`RazorChunkGeneratorTest.cs` and `CSharpRazorChunkGeneratorTest.cs`, respectively). `RazorCodeGeneratorTest` is also a fairly poor name for the subclass.

dougbu:
d545f47

Issue #412, opened 2015-05-30
Error is occurred if there is a link in the render section


pherbel:
Error is occurred if there is a link in the render section.
I don't know exactly why this error is coming. Maybe it is a bug. Please check it.
I'm running on beta4 version.

Error will be this:
error CS0246: The type or namespace name 'TagHelperContent' could not be found (are you missing a using directive or an assembly reference?)

Sample code:
@{
    ViewBag.Title = "MyPage";
}
@section mysection {

    <a href="http://www.test.com">Test</a>
}

This will work:
@{
    ViewBag.Title = "MyPage";
}
@section mysection {

    @*<a href="http://www.test.com">Test</a>*@
}

dougbu:
This and aspnet/Mvc#2595 are duplicates. Closing aspnet/Mvc#2595 since most if not all of the fix will be in this repo.

dougbu:
Self-assigning since aspnet/Mvc#2595 was in this milestone and assigned to me.

dougbu:
113d80c

pherbel:
Thank you.

Azaret:
Thank you, I'll check this out.

PR #414, opened 2015-06-01
Ensure `CodeVisitor`s consistenctly find tag helpers

dougbu:
- #412
- fix `CSharpTagHelperRunnerInitializationVisitor` and `CSharpUsingVisitor`
- fix existing test of the tag helper in a section scenario

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperRunnerInitializationVisitor.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperRunnerInitializationVisitor.cs
@@ -26,6 +26,25 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
             _tagHelperContext = Context.Host.GeneratedClassContext.GeneratedTagHelperContext;
         }
 
+        /// <inheritdoc />
+        public override void Accept(Chunk chunk)
+        {
+            // If at any ParentChunk other than a TagHelperChunk, then dive into its Children to search for more
+            // TagHelperChunk nodes. This method avoids overriding each of the ParentChunk-specific Visit() methods to
+            // dive into Children.
+            var parentChunk = chunk as ParentChunk;
+            if (parentChunk != null && !(parentChunk is TagHelperChunk))
+            {
+                Accept(parentChunk.Children);
+            }
+            else
+            {
+                // If at a TagHelperChunk or any non-ParentChunk, "Accept()" it. This ensures the Visit(TagHelperChunk)
+                // method below is called.
+                base.Accept(chunk);
+            }
+        }
+
         /// <summary>
         /// Writes the TagHelperRunner initialization code to the Writer.
         /// </summary>
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
@@ -23,6 +23,25 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
 
         public IList<string> ImportedUsings { get; set; }
 
+        /// <inheritdoc />
+        public override void Accept(Chunk chunk)
+        {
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperRunnerInitializationVisitor.cs
```diff
@@ -26,6 +26,25 @@ public class CSharpTagHelperRunnerInitializationVisitor : CodeVisitor<CSharpCode
             _tagHelperContext = Context.Host.GeneratedClassContext.GeneratedTagHelperContext;
         }
 
+        /// <inheritdoc />
+        public override void Accept(Chunk chunk)
```
NTaylorMullen: Just do:

```C#
protected override void Visit(ParentChunk chunk)
{
    Accept(chunk.Children);
}
```

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
```diff
@@ -23,6 +23,25 @@ public CSharpUsingVisitor([NotNull] CSharpCodeWriter writer, [NotNull] CodeGener
 
         public IList<string> ImportedUsings { get; set; }
 
+        /// <inheritdoc />
+        public override void Accept(Chunk chunk)
```
NTaylorMullen: ditto

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperRunnerInitializationVisitor.cs
dougbu: That wouldn't work at all. `Visit(ParentChunk)` is called only if the `chunk` isn't one of the many subclasses of `ParentChunk`. This is exactly why the comments below say what they do.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
dougbu: https://github.com/aspnet/Razor/pull/414#discussion_r31458232

dougbu:
Responded to all.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperRunnerInitializationVisitor.cs
NTaylorMullen: Good point :+1: . One thing to note though: With the current approach you'll be diving into every attribute (`DynamicCodeAttributeChunk`, `LiteralCodeAttributeChunk`, `CodeAttributeChunk`) and expression blocks (`ExpressionBlockChunk`) which can't/shouldn't contain directives or `TagHelper`s.

Not horrible but not ideal. It might be good to make a base class that has knowledge about diving into known Razor blocks (template chunks, sections, statements etc.) that can contain directives or complex structures like `TagHelperChunk`s. That way you can just inherit from it in this class and the `CSharpUsingVisitor`.

Since you're updating the `@using` here as well, this problem impacts other directives as well, for instance: any `@model` or `@addTagHelper` inside of sections, statements etc. won't be picked up.

dougbu: > Since you're updating the @using here as well, this problem impacts other directives as well, for instance: any @model or @addTagHelper inside of sections, statements etc. won't be picked up.

Previously `@model`, `@addTagHelper`, and `@using` inside statements or sections were _not_ picked up. Now they are.

Diving into a few extra blocks isn't a big deal -- see existing `CSharpTagHelperFieldDeclarationVisitor` for example.

NTaylorMullen: Alright.

NTaylorMullen:
:shipit:

dougbu:
113d80c

PR #415, opened 2015-06-01
Enable TagHelper attributes to render in the order they were originally.

NTaylorMullen:
- This was accomplished by changing when/how TagHelper attributes are rendered. Previously they were rendered: Bound => Unbound. Now they're rendered inthe order that they exist on the TagHelperChunk.
- Regenerated test files to account for new re-ordering of TagHelper attributes.
- Did not add additional tests since the current tests seem to be heavily impacted by the re-organization of code rendering; effectively showcasing the fix.

#225


dougbu:
> Did not add additional tests since the current tests seem to be heavily impacted by the re-organization of code rendering; effectively showcasing the fix.

Disagree with this choice.  The vast majority of the existing tests only saw `AddTagHelperAttribute()` calls move relative to `CreateTagHelper()` calls, as the `CreateTagHelper()` calls moved into a group.  Need a few more cases (besides `PrefixedAttributeTagHelpers.cshtml` and a couple of others):
- mixing bound and unbound attributes
- involving unbound attributes with duplicate names (this currently isn't tested at all)

Ideally would test whether `TagHelperExecutionContext.AllAttributes` and `...HtmlAttributes` additions end up in the correct order more directly.  Unfortunately, that can be done only in a MVC test and Razor confirmations don't really belong there.  Good news is my spot checks, especially of generation for `PrefixedAttributeTagHelpers.cshtml`, indicate what's tested is working correctly.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -69,13 +69,11 @@ public void RenderTagHelper(TagHelperChunk chunk)
...
             RenderTagHelpersCreation(chunk, tagHelperDescriptors);
 
-            // Determine what attributes exist in the element and divide them up.
-            var htmlAttributes = chunk.Attributes;
-            var attributeDescriptors = tagHelperDescriptors.SelectMany(descriptor => descriptor.Attributes).ToArray();
-            var unboundHtmlAttributes = htmlAttributes.Where(attribute => !attributeDescriptors.Any(
-                attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key)));
+            // This ensures that all created TagHelpers have initialized dictionary bound properties which are used
+            // via TagHelper indexers.
+            RenderTagHelperPropertyVerification(chunk.Attributes, tagHelperDescriptors);
```
dougbu: Why? Can do the same checks in `RenderAttributes()`. This is wasteful and duplicates the loops unnecessarily.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
         {
+            if (_designTimeMode)
+            {
+                // No need to render any dictionary verification in design time mode.
+                return;
+            }
+
             // Track dictionary properties we have confirmed are non-null.
             var confirmedDictionaries = new HashSet<string>(StringComparer.Ordinal);
```
dougbu: This placement is totally wrong and indicates a gap in the current tests. Dictionary properties must be verified per tag helper. So this line must be moved within the loop through the `tagHelperDescriptors`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
-                var attributeName = attributeKeyValuePair.Key;
-                var attributeValueChunk = attributeKeyValuePair.Value;
-                if (attributeValueChunk == null)
+                foreach (var attributeDescriptor in tagHelperDescriptor.Attributes)
                 {
-                    // Minimized attributes are not valid for bound attributes. TagHelperBlockRewriter has already
-                    // logged an error if it was a bound attribute; so we can skip.
-                    continue;
-                }
+                    if (attributeDescriptor.IsIndexer && confirmedDictionaries.Add(attributeDescriptor.PropertyName))
```
dougbu: No. Do not mark a property as confirmed unless the confirmation is actually generated. Again, need more tests.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
 
-                // Find the matching TagHelperAttributeDescriptor.
-                var attributeDescriptor = attributeDescriptors.FirstOrDefault(
-                    descriptor => descriptor.IsNameMatch(attributeName));
-                if (attributeDescriptor == null)
-                {
-                    // Attribute is not bound to a property or indexer in this tag helper.
-                    continue;
-                }
+                        // attributeName will be null in the case that the attribute is not utilized.
```
dougbu: Why change the existing comment? This is more vague and less helpful.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
+            {
+                // No need to render any dictionary verification in design time mode.
+                return;
+            }
+
             // Track dictionary properties we have confirmed are non-null.
             var confirmedDictionaries = new HashSet<string>(StringComparer.Ordinal);
 
-            // First attribute wins, even if there are duplicates.
-            var distinctAttributes = chunkAttributes.Distinct(KeyValuePairKeyComparer.Default);
```
dougbu: This is still useful if this method continues to exist.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
+                foreach (var attributeDescriptor in tagHelperDescriptor.Attributes)
                 {
-                    // Minimized attributes are not valid for bound attributes. TagHelperBlockRewriter has already
-                    // logged an error if it was a bound attribute; so we can skip.
-                    continue;
-                }
+                    if (attributeDescriptor.IsIndexer && confirmedDictionaries.Add(attributeDescriptor.PropertyName))
+                    {
+                        var attributeName = chunkAttributes.FirstOrDefault(
+                            chunkAttribute => attributeDescriptor.IsNameMatch(chunkAttribute.Key)).Key;
```
dougbu: Move `.Key` to a new line. Hard to see now.

dougbu: Better if width fits:
``` c#
var attributeName = chunkAttributes
    .FirstOrDefault(chunkAttribute => attributeDescriptor.IsNameMatch(chunkAttribute.Key))
    .Key;
```

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
-                        dictionaryKey);
                 }
+            }
+        }
+
+        private void RenderAttributes(
+            IList<KeyValuePair<string, Chunk>> chunkAttributes,
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+        {
+            // First attribute wins, even if there are duplicates.
```
dougbu: Comment has nothing to do with the `renderedBoundAttributeNames` variable. Lousy name BTW since they haven't been rendered; use `boundAttributeNames`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
-                // If we haven't recorded this attribute value before then we need to record its value.
-                var attributeValueRecorded = htmlAttributeValues.ContainsKey(attributeName);
-                if (!attributeValueRecorded)
+            foreach (var attribute in chunkAttributes)
+            {
+                var associatedDescriptors = tagHelperDescriptors.Where(descriptor =>
+                    descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key)));
+
+                // Bound attributes will have associated descriptors.
+                if (associatedDescriptors.Any())
```
dougbu: Much more understandable to go through `associatedDescriptors` once and set a `bool` variable if any match is found. This ends up iterating through `associatedDescriptors` an extra time when the attribute is bound.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
+                    // the same attribute we don't execute the attributes value more than once.
+                    string valueLocation = null;
+
+                    foreach (var associatedDescriptor in associatedDescriptors)
+                    {
+                        // There should only be a single TagHelperAttributeDescriptor associated with the attribute.
+                        // TagHelpers cannot have multiple attribute descriptors pointing to the same property unless
+                        // they're indexed.
+                        var associatedAttributeDescriptor = associatedDescriptor.Attributes.First(
+                            attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key));
```
dougbu: Do not "fix" the _separate_ bug about multiple matching attribute descriptors in this way. Leads to unfathomable errors from Linq because `TagHelperBlockRewriter` is not guarding against this situation. Stick with `FirstOrDefault()` and leave #199 alone. (#237 also touches this case but that enhancement is for later.)

dougbu: Ah, the problem here is just the unfathomable comment.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
-                        attributeDescriptor,
-                        bufferableAttribute,
-                        attributeValueChunk,
-                        valueAccessor);
-
-                    // Execution contexts are a runtime feature.
-                    if (_designTimeMode)
+                    if (!renderedBoundAttributeNames.Add(attribute.Key))
                     {
+                        // We already rendered an attribute with the same name that was bound. First wins, skip.
```
dougbu: Don't rewrite existing, understandable comments and don't talk in terms of "we".

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
+            IList<KeyValuePair<string, Chunk>> chunkAttributes,
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+        {
+            // First attribute wins, even if there are duplicates.
+            var renderedBoundAttributeNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-                // If we haven't recorded this attribute value before then we need to record its value.
-                var attributeValueRecorded = htmlAttributeValues.ContainsKey(attributeName);
-                if (!attributeValueRecorded)
+            foreach (var attribute in chunkAttributes)
```
dougbu: Restore `// Go through the HTML attributes in source order, assigning to properties or indexers as we go.` comment, adding the possibility of adding HTML attribute to the execution context.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
+                // logged an error if it was a bound attribute; so we can skip.
+                return null;
+            }
+
+            // We need to capture the tag helper's property value accessor so we can retrieve it later (if we need to).
+            var valueAccessor = string.Format(
+                CultureInfo.InvariantCulture,
+                "{0}.{1}",
+                tagHelperVariableName,
+                attributeDescriptor.PropertyName);
```
dougbu: Why bother when `valueLocation` is non-`null`? This should be within the `if` block below.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
+
+            if (attributeDescriptor.IsIndexer)
+            {
+                var dictionaryKey = attributeName.Substring(attributeDescriptor.Name.Length);
+                valueAccessor += "[\"" + dictionaryKey + "\"]";
```
dougbu: Use string interpolation; easier on the eyes.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
-                        .Write(htmlAttributeValues[attributeName])
-                        .WriteLine(";");
+                    RenderUnboundAttribute(attribute);
+                }
+            }
+        }
+
+        private string RenderBoundAttribute(
+            string tagHelperVariableName,
+            string valueLocation,
```
dougbu: Call this `valueAccessor` and reuse it if `null` is passed.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -251,69 +226,134 @@ protected virtual string GenerateUniqueId()
...
-                        .WriteStartInstanceMethodInvocation(
-                            ExecutionContextVariableName,
-                            _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName)
-                        .WriteStringLiteral(attributeName)
-                        .WriteParameterSeparator()
-                        .Write(valueAccessor)
-                        .WriteEndMethodInvocation();
+                    // We capture the value location of each bound attribute to ensure when multiple TagHelpers request
+                    // the same attribute we don't execute the attributes value more than once.
+                    string valueLocation = null;
```
dougbu: Call this `valueAccessor` and either leave out the comment or duplicate the `// We need to capture the tag helper's property value accessor so we can retrieve it later (if we need to).` comment here.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Got it now. Leave out the comment because it makes sense only where the `HashSet` is checked. Variable name is :ok: though "Names" part doesn't add much.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: It's definitely more execution time but makes the attribute rendering code easier to understand: simplified the `RenderBoundAttribute` by not special casing the `valueAccessor`, not passing `confirmedDictionaries` (would affect the coordinator method), not passing in error information. Ultimately wanted to keep the `RenderBoundAttribute` as simple as possible.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

NTaylorMullen: If it's per `TagHelperDescriptor`, I don't believe we need `confirmedDictionaries` anymore at all with how this method now works. Since it's iterating over the attribute descriptors can always generate the check if there's an associated `chunkAttribute`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Why? We'd still need to do the `FirstOrDefault`. In most cases this `Distinct` will return the original attributes anyways (people rarely use duplicate, same named attributes).

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: The check is different. We used to iterate through the `chunkAttribute` and check to see if there's a descriptor associated. Now we're iterating the `TagHelperAttributeDescriptor`s and seeing if there's a `chunkAttribute` associated.

dougbu: Result is the same; `// Attribute is not bound to a property or indexer in this tag helper.` still applies (within the `if`).

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Not sure I agree with the "as simple as possible" goal but, if so, instead merge `RenderTagHelperPropertyVerification` into `RenderTagHelpersCreation()`. No need of a separate loop through both the descriptors and the attributes. If part of `RenderAttributes()` would avoid iterating through the attributes again; if part of `RenderTagHelpersCreation()`, save a loop through the descriptors.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Perhaps. Let's see the new code.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Going through `associatedDescriptors` first with a `foreach` makes the current `if (!renderedBoundAttributeNames.Add(attribute.Key))` check messy. I feel like it reduces the clarity.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Will do

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -179,68 +177,45 @@ protected virtual string GenerateUniqueId()
...
-            var distinctAttributes = chunkAttributes.Distinct(KeyValuePairKeyComparer.Default);
-
-            // Go through the HTML attributes in source order, assigning to properties or indexers as we go.
-            foreach (var attributeKeyValuePair in distinctAttributes)
+            foreach (var tagHelperDescriptor in tagHelperDescriptors)
             {
-                var attributeName = attributeKeyValuePair.Key;
-                var attributeValueChunk = attributeKeyValuePair.Value;
-                if (attributeValueChunk == null)
+                foreach (var attributeDescriptor in tagHelperDescriptor.Attributes)
```
dougbu: This loop will not handle precedence of non-indexer properties over indexers. For example if `asp-aaa` and `asp-bbb` were used in an element associated with a tag helper that has non-indexers with those names, an unused dictionary would still be initialized.

Need to instead check
``` c#
chunkAttributes.Any(
    keyValuePair => tagHelperDescriptor.Attributes.Any(
        attribute => attribute.IsNameMatch(keyValuePair.Key) && attribute.IsIndexer))
```

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Avoid repeated use of complicated enumerations unless absolutely necessary. Not necessary here; the current code is more wasteful than clear.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/415/files#diff-6f3af99214f3f3e7141c9d7bd2aea6c0R351

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Can't, the `valueAccessor` is used along side of the `valueLocation` https://github.com/aspnet/Razor/pull/415/files#diff-6f3af99214f3f3e7141c9d7bd2aea6c0R351

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: That line could just as easily use the `valueLocation`. This isn't useful unless the parameter is `null`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Yes you can -- and should.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Talked in person, will update names of variables to be better.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Talked in person, will update names of variables to be better.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Talked in person

NTaylorMullen:
@dougbu Multiple tests cover those scenarios. Heck even just `MinimizedTagHelpers` covers both of the scenarios you mentioned and then some.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Added this change then regenerated the code and your dictionary tests actually do cover this scenario but the old output was wrong. In the commit to address code review comments you can see the change for the added throw cases.

NTaylorMullen:
Updated.

dougbu:
> even just `MinimizedTagHelpers` covers both of the scenarios you mentioned and then some by itself.

Minimized tag helper attributes are a special case. More generally we have very little coverage of code generation with multiple bound and unbound attributes in various source orders. And we have just one test (`PrefixedAttributeTagHelpers.cshtml`) involving attributes with duplicate names; that one covers only bound attributes.

NTaylorMullen:
@dougbu The `MinimizedTagHelper`s test has mixes of unbound/bound/required minimized/notminimized all mixed together. In addition to this `TagHelpersInSection`, `BasicTagHelpers`, `ComplexTagHelpers`, `AttributeTargetingTagHelpers`, `SingleTagHelper` all have  mixes of bound/unbound (multiples, not multiples etc.). 

As for duplicate name-unbound attributes. We do only have 1 test covering that which is the `MinimizedTagHelpers` test and as you said it is a minimized attribute. I'll add a duplicate same name, unbound, unminimized to `ComplexTagHelpers`.

NTaylorMullen:
Added update to `ComplexTagHelpers`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -172,75 +166,37 @@ protected virtual string GenerateUniqueId()
...
-            // First attribute wins, even if there are duplicates.
-            var distinctAttributes = chunkAttributes.Distinct(KeyValuePairKeyComparer.Default);
-
-            // Go through the HTML attributes in source order, assigning to properties or indexers as we go.
-            foreach (var attributeKeyValuePair in distinctAttributes)
-            {
-                var attributeName = attributeKeyValuePair.Key;
-                var attributeValueChunk = attributeKeyValuePair.Value;
-                if (attributeValueChunk == null)
+                // Execution contexts and thrown errors are a runtime feature.
```
dougbu: `... and thrown errors ...` confuses rather than helps. Suggest `... and throwing errors for null dictionary properties are runtime features.`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
+            var renderedBoundAttributeNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            // Go through the HTML attributes in source order, assigning to properties or indexers or adding to
+            // TagHelperExecutionContext's HTML attributes as we go.
+            foreach (var attribute in chunkAttributes)
+            {
+                var associatedDescriptors = tagHelperDescriptors.Where(descriptor =>
+                    descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key)));
+
+                // Bound attributes will have associated descriptors.
```
dougbu: Huh? Leave out the odd `will`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
-                        attributeDescriptor,
-                        bufferableAttribute,
-                        attributeValueChunk,
-                        valueAccessor);
-
-                    // Execution contexts are a runtime feature.
-                    if (_designTimeMode)
+                    if (!renderedBoundAttributeNames.Add(attribute.Key))
                     {
+                        // We only need to create attribute values once per HTML element (not once per tag helper).
```
dougbu: Another head-scratcher comment. This `continue` fits with the old `// First attribute wins, even if there are duplicates.` comment, perhaps changed slightly to `// First attribute value wins if there are duplicates.`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
-                    _writer
-                        .WriteStartInstanceMethodInvocation(
-                            ExecutionContextVariableName,
-                            _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName)
-                        .WriteStringLiteral(attributeName)
-                        .WriteParameterSeparator()
-                        .Write(valueAccessor)
-                        .WriteEndMethodInvocation();
+                    // We need to capture the tag helper's property value accessor so we can retrieve it later
+                    // (if we need to).
```
dougbu: `(if we need to)` worked only because more detail was provided earlier in the old code. Change to `... if there are more tag helpers that need the value`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
+            var attributeName = attribute.Key;
+            var attributeValueChunk = attribute.Value;
+            if (attributeValueChunk == null)
+            {
+                // Minimized attributes are not valid for bound attributes. TagHelperBlockRewriter has already
+                // logged an error if it was a bound attribute; so we can skip.
+                return null;
+            }
+
+            // We need to capture the tag helper's property value accessor so we can retrieve it later (if we need to).
```
dougbu: This isn't about capturing anything. Need to know what should be on the left hand side of the assignment. Sorry if I previously thought this comment would fit here.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
+                tagHelperVariableName,
+                attributeDescriptor.PropertyName);
+
+            if (attributeDescriptor.IsIndexer)
+            {
+                var dictionaryKey = attributeName.Substring(attributeDescriptor.Name.Length);
+                currentValueAccessor += $"[\"{dictionaryKey}\"]";
+            }
+
+            // If there's no location to pull the attribute value from then we need to execute the attribute value.
```
dougbu: `execute the attribute value` -> `determine the attribute value`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
+                    .WriteStringLiteral(attributeName)
+                    .WriteParameterSeparator()
+                    .Write(currentValueAccessor)
+                    .WriteEndMethodInvocation();
+
+                return currentValueAccessor;
+            }
+            else
+            {
+                // The attribute value has already been executed and was passed to us as valueLocation, we don't want
```
dougbu: - `has already been executed and was passed to us as valueLocation` -> `has already been determined and accessor was pass to us as `previousValueAccessor`
- additional `execute` -> `determine` (or perhaps `evaluate`) and variable name fix on next line

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -172,75 +166,37 @@ protected virtual string GenerateUniqueId()
...
 
-                if (attributeDescriptor.IsIndexer)
+                // Ensure that all created TagHelpers have initialized dictionary bound properties which are used
+                // via TagHelper indexers.
+                foreach (var chunkAttribute in chunk.Attributes)
                 {
-                    // Need a different valueAccessor in this case. But first need to throw a reasonable Exception at
-                    // runtime if the property is null. The check is not required at design time.
-                    if (!_designTimeMode && confirmedDictionaries.Add(attributeDescriptor.PropertyName))
+                    var associatedAttributeDescriptor = tagHelperDescriptor.Attributes.FirstOrDefault(
```
dougbu: Add a `continue` if `chunkAttribute.Value == null` since no indexer assignment will occur. Also make sure this case is tested. If it's actually impossible to reach here with a bound attribute that has `chunkAttribute.Value == null`, turn the `null` check at the top of `RenderBoundAttribute()` into a `Debug.Assert()` and use same thing here.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -172,75 +166,37 @@ protected virtual string GenerateUniqueId()
...
+                    tagHelperVariableName);
 
-                // We capture the tag helper's property value accessor so we can retrieve it later (if we need to).
-                var valueAccessor = string.Format(
-                    CultureInfo.InvariantCulture,
-                    "{0}.{1}",
-                    tagHelperVariableName,
-                    attributeDescriptor.PropertyName);
+                // Track dictionary properties we have confirmed are non-null.
+                var confirmedDictionaries = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
dougbu: `StringComparer.Ordinal` since C# identifiers are case-sensitive.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -172,75 +166,37 @@ protected virtual string GenerateUniqueId()
...
-                    // runtime if the property is null. The check is not required at design time.
-                    if (!_designTimeMode && confirmedDictionaries.Add(attributeDescriptor.PropertyName))
+                    var associatedAttributeDescriptor = tagHelperDescriptor.Attributes.FirstOrDefault(
+                        attributeDescriptor => attributeDescriptor.IsNameMatch(chunkAttribute.Key));
+
+                    if (associatedAttributeDescriptor != null &&
+                        associatedAttributeDescriptor.IsIndexer &&
+                        confirmedDictionaries.Add(associatedAttributeDescriptor.PropertyName))
                     {
                         _writer
```
dougbu: Add a comment above this line describing the point of this block e.g. `// Throw a reasonable Exception at runtime if the dictionary property is null.` (cribbing from old code)

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -172,75 +166,37 @@ protected virtual string GenerateUniqueId()
...
                             .Write("if (")
-                            .Write(valueAccessor)
-                            .WriteLine(" == null)");
+                            .Write(tagHelperVariableName)
+                            .Write(".")
+                            .Write(associatedAttributeDescriptor.PropertyName)
+                            .Write(" == null)");
```
dougbu: Ick, no Java braces. Stick with `WriteLine()` here.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
-                var attributeValueRecorded = htmlAttributeValues.ContainsKey(attributeName);
-                if (!attributeValueRecorded)
+        private void RenderAttributes(
+            IList<KeyValuePair<string, Chunk>> chunkAttributes,
+            IEnumerable<TagHelperDescriptor> tagHelperDescriptors)
+        {
+            var renderedBoundAttributeNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            // Go through the HTML attributes in source order, assigning to properties or indexers or adding to
+            // TagHelperExecutionContext's HTML attributes as we go.
```
dougbu: Think you mean `HTMLAttributes` but I'm not sure.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Alternatively use `// Haven't seen this attribute value before and need to record its value.`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Alternatively use `// If this attribute value has not been seen before, need to record its value.`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -389,67 +410,64 @@ protected virtual string GenerateUniqueId()
...
         }
 
-        private void RenderUnboundHTMLAttributes(IEnumerable<KeyValuePair<string, Chunk>> unboundHTMLAttributes)
+        private void RenderUnboundAttribute(KeyValuePair<string, Chunk> chunkAttribute)
         {
-            // Build out the unbound HTML attributes for the tag builder
-            foreach (var htmlAttribute in unboundHTMLAttributes)
+            string textValue = null;
+            var isPlainTextValue = false;
+            var attributeValue = chunkAttribute.Value;
```
dougbu: `attributeValueChunk` was a better name. Ideally stop passing around the poorly-named `chunkAttribute` and determine `attributeName` and `attributeValueChunk` in `RenderAttributes()` then pass those around instead.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
+                var bufferableAttribute = attributeDescriptor.IsStringProperty;
+
+                RenderNewAttributeValueAssignment(
+                    attributeDescriptor,
+                    bufferableAttribute,
+                    attributeValueChunk,
+                    currentValueAccessor);
+
+                // Execution contexts are a runtime feature.
+                if (_designTimeMode)
```
dougbu: Slightly more straightforward to place the "inform the context" bit within a `if (!_designTimeMode)` block. The reduced nesting for a single statement isn't that useful here.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +205,135 @@ protected virtual string GenerateUniqueId()
...
-                        bufferableAttribute,
-                        attributeValueChunk,
-                        valueAccessor);
-
-                    // Execution contexts are a runtime feature.
-                    if (_designTimeMode)
+                    if (!renderedBoundAttributeNames.Add(attribute.Key))
                     {
+                        // We only need to create attribute values once per HTML element (not once per tag helper).
                         continue;
```
dougbu: Realized looking at `PrefixedAttributeTagHelpers.cs` that we have an existing bug in our handling of duplicate bound attributes. `TagHelperExecutionContext.AllAttributes` also only contains the first attribute value. If MVC code or a user manually restores such a duplicate attribute to `TagHelperOutput.Attributes`, only first name / value pair can ever be restored. Do _not_ fix this here but do file a bug.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: It's not impossible for a `chunkAttribute.Value == null` to reach here but it's impossible for the `RazorPage` to run at runtime if it's `null` (there'll be an error in the `ErrorSink` for it). Being `null` would indicate there's a bound `TagHelper` attribute with 0 value (error case).

Going to leave as is.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: Discussed offline; :ok: since this particular block doesn't use `chunkAttribute.Value`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Good catch. :+1: 

https://github.com/aspnet/Razor/issues/418

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +206,135 @@ protected virtual string GenerateUniqueId()
...
+                    string valueAccessor = null;
+
+                    foreach (var associatedDescriptor in associatedDescriptors)
+                    {
+                        var associatedAttributeDescriptor = associatedDescriptor.Attributes.First(
+                            attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key));
+                        var tagHelperVariableName = GetVariableName(associatedDescriptor);
+
+                        valueAccessor = RenderBoundAttribute(
+                            attribute.Key,
```
dougbu: Minor: `attribute.Key` is used multiple times in the body of this loop. Should probably define `attributeName` variable just before `associatedDescriptors` is defined.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +206,135 @@ protected virtual string GenerateUniqueId()
...
                 }
                 else
                 {
-                    // The attribute value has already been recorded, lets retrieve it from the stored value
-                    // accessors.
-                    _writer
-                        .WriteStartAssignment(valueAccessor)
-                        .Write(htmlAttributeValues[attributeName])
-                        .WriteLine(";");
+                    RenderUnboundAttribute(attribute.Key, attribute.Value);
```
dougbu: Even less important `attribute.Value` is used twice in the loop body...

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +206,135 @@ protected virtual string GenerateUniqueId()
...
+                }
+            }
+        }
+
+        private string RenderBoundAttribute(
+            string attributeName,
+            Chunk attributeValueChunk,
```
dougbu: :100:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +206,135 @@ protected virtual string GenerateUniqueId()
...
+            string tagHelperVariableName,
+            string previousValueAccessor,
+            TagHelperAttributeDescriptor attributeDescriptor)
+        {
+            if (attributeValueChunk == null)
+            {
+                // Minimized attributes are not valid for bound attributes. TagHelperBlockRewriter has already
+                // logged an error if it was a bound attribute; so we can skip.
+                return null;
```
dougbu: Side note: If an attribute name is duplicated and the first use is minimized, will never render anything for it. Shouldn't matter since this is an error scenario but worth keeping in mind if we ever support minimized bound attributes. About the worst aspect is that errors in the non-minimized attribute value won't show up until the minimized one is removed.

E.g. with
``` html
  <tagHelper bound-attribute bound-attribute="value" />
```
Nothing will be generated for `bound-attribute`. Well, perhaps dictionary initialization (if the binding is to an indexer with a name such as `bound-`) but that's terribly minor.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -249,70 +206,135 @@ protected virtual string GenerateUniqueId()
...
+        }
+
+        private string RenderBoundAttribute(
+            string attributeName,
+            Chunk attributeValueChunk,
+            string tagHelperVariableName,
+            string previousValueAccessor,
+            TagHelperAttributeDescriptor attributeDescriptor)
+        {
+            if (attributeValueChunk == null)
```
dougbu: This check should be outside the `foreach` loop at line 248. No need to do something general for every associated tag helper.

dougbu:
:shipit: after addressing remaining few comments.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

Issue #352, opened 2015-04-08
Consume docs on TagHelper classes to provide IntelliSense at design time.


NTaylorMullen:
Right now when writing `TagHelper`s users consume them in the editor without any added IntelliSense. If we were to discover docs related to `TagHelper` classes and expose those via `TagHelperDescriptor`s we could provide more information to make tooling easier to use.

I believe we should do this at the Razor level to enable cross-platform scenarios.

danroth27:
Need to understand the tooling impact.

NTaylorMullen:
This wont be done for beta5, this is a larger issue that will most likely be post rtm.

NTaylorMullen:
ccf8433f27f0a3e4d20a8a75beefd6a07f0748a8
9055da1894d2df0aaabe642498a94c5001a01981
02a3ca8abb042102980c85e4dd41835aac562602
920358ab0de25b868e8192e62e9af38839588660

PR #419, opened 2015-06-08
Add TagHelper documentation resolution to TagHelperDescriptorResolver.

NTaylorMullen:
- Added `TagHelperUseageDescriptor` and associated factory for the `TagHelperDescriptorFactory` to utilize.
- `TagHelperUseageDescriptor`s are only created during design time.
- CoreCLR is not supported for XML documentation resolution for now. Can revisit this later when we have better tooling integration with CoreCLR.

#352


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,144 @@
...
+            IEnumerable<XElement> members = null;
+            if (!XMLFileMembersCache.TryGetValue(typeAssemblyLocation, out members))
+            {
+                var xmlDocumentationFile = GetXmlDocumentationFile(typeAssemblyLocation);
+
+                // We only want to process the file if it exists. In the case it doesn't, a null value will be added
+                // to the cache to not constantly look for new XML files.
+                if (xmlDocumentationFile != null)
+                {
+                    var xmlDocumentation = XDocument.Load(xmlDocumentationFile.FullName);
```
NTaylorMullen: Suggestion by @rynowak is to break out the parsing code of the XML file into a stand alone class in-case we want to utilize it in other areas down the line.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+                    documentationProvider = new XmlDocumentationProvider(xmlDocumentationFile.FullName);
+                }
+
+                XmlDocumentationProviderCache.Add(typeAssemblyLocation, documentationProvider);
+            }
+
+            // Members will be null if there is no associated XML file for the provided typeAssembly.
+            if (documentationProvider != null)
+            {
+                if (documentationProvider.HasDocumentation(id))
```
NTaylorMullen: Each one of these method calls essentially calls through to linq to find the member and then provide information on it. Options:
1. Leave as is
2. Cache the member lookup in the `documentationProvider`
3. Provide an extra abstraction (`XmlDocumentationMember`) to make retrieval of all these bits easy (kind of like `FileInfo` but for an xml doc member).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+                    assemblyXmlDocumentationFile = fallbackDirectories
+                        .Select(fallbackDiretory =>
+                            new FileInfo(
+                                Path.Combine(assemblyDirectory, fallbackDiretory, assemblyXmlDocumentationName)))
+                        .FirstOrDefault(file => file.Exists);
+                }
+
+                return assemblyXmlDocumentationFile;
+            }
+            catch (PathTooLongException)
```
NTaylorMullen: Instead of doing this could a custom exception thrown by the `XmlDocumentationProvider` that can be caught OR could add a property to the provider like `XmlFileResolved`. Problem with adding a property would then add the requirement to throw whenever one of the `GetSummary` or `GetRemarks` methods is invoked.

Adding this comment for discussion.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+                    var remarks = documentationProvider.GetRemarks(id);
+
+                    return new TagHelperUseageDescriptor(summary, remarks);
+                }
+            }
+
+            return null;
+        }
+
+        private static FileInfo GetXmlDocumentationFile(string typeAssemblyLocation)
```
NTaylorMullen: Debated moving the functionality of this method into another type such as `XmlDocumentationLocator` but thought it might be overkill for now.

NTaylorMullen:
Updated by splitting out the `XmlDocumentationProvider` and added several comments to the PR for discussion.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -44,17 +44,27 @@ public static class TagHelperDescriptorFactory
...
             var typeInfo = type.GetTypeInfo();
-            var attributeDescriptors = GetAttributeDescriptors(type, errorSink);
+            var attributeDescriptors = GetAttributeDescriptors(type, designTime, errorSink);
             var targetElementAttributes = GetValidTargetElementAttributes(typeInfo, errorSink);
+            TagHelperUseageDescriptor typeUseageDescriptor = null;
+
+#if !DNXCORE50
+            // The TagHelperUseageDescriptor is consumed at design time to provide IntelliSense when typing HTML
+            // elements or attributes that are tag helper related.
+            typeUseageDescriptor = designTime ? TagHelperUseageDescriptorFactory.CreateDescriptor(typeInfo) : null;
```
dougbu: typo throughout this PR; should be `Usage`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -246,7 +269,8 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
                         parentType: type,
                         errorSink: errorSink,
                         defaultPrefix: descriptor.Name + "-",
-                        isInvalid: out isInvalid);
+                        isInvalid: out isInvalid,
+                        designTime: designTime);
```
dougbu: leave `out` parameters last. less important but would also be good to place `designTime` consistently just before (or after) `errorSink`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -414,11 +441,35 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
+#if !DNXCORE50
+            // The TagHelperUseageDescriptor is consumed at design time to provide IntelliSense when typing HTML
+            // elements or attributes that are tag helper related.
+            propertyUseageDescriptor = designTime ? TagHelperUseageDescriptorFactory.CreateDescriptor(property) : null;
+#endif
+
+            return new TagHelperAttributeDescriptor(
+                attributeName,
+                property.Name,
+                property.PropertyType.FullName,
```
dougbu: Change would mess up `isIndexer` cases; pass `typeName`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -121,6 +125,7 @@ public IEnumerable<TagHelperDescriptor> Resolve([NotNull] TagHelperDescriptorRes
         protected virtual IEnumerable<TagHelperDescriptor> ResolveDescriptorsInAssembly(
             string assemblyName,
             SourceLocation documentLocation,
+            bool designTime,
```
dougbu: Remove this parameter. Any subclass must provide value when calling a `TagHelperDescriptorResolver` constructor. So that subclass can have its own `private bool _designTime` field.

src/Microsoft.AspNet.Razor.Runtime/project.json
```diff
@@ -6,11 +6,21 @@
         "Microsoft.Framework.BufferEntryCollection.Sources": { "type": "build", "version": "1.0.0-*" },
         "Microsoft.Framework.ClosedGenericMatcher.Sources": { "type": "build", "version": "1.0.0-*" },
         "Microsoft.Framework.CopyOnWriteDictionary.Sources": { "type": "build", "version": "1.0.0-*" },
-        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
+        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" },
```
dougbu: invalid JSON

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -24,6 +24,42 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
...
+            [NotNull] string propertyName,
+            [NotNull] string typeName,
+            bool isIndexer,
+            bool isStringProperty)
+            : this(
+                  name,
+                  propertyName,
+                  typeName,
+                  isIndexer,
+                  isStringProperty: string.Equals(typeName, typeof(string).FullName, StringComparison.Ordinal),
```
dougbu: pass `isStringProperty`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -46,13 +82,15 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
             [NotNull] string name,
             [NotNull] string propertyName,
             [NotNull] string typeName,
-            bool isIndexer)
+            bool isIndexer,
+            TagHelperUseageDescriptor useage)
```
dougbu: Parameter and property need longer names. suggest `usageDescriptor`. Same in `TagHelperDescriptor`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -24,6 +24,42 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
...
+            bool isStringProperty)
+            : this(
+                  name,
+                  propertyName,
+                  typeName,
+                  isIndexer,
+                  isStringProperty: string.Equals(typeName, typeof(string).FullName, StringComparison.Ordinal),
+                  useage: null)
+        {
+        }
```
dougbu: In real PR, update tests to use more-complete constructors and remove these. Same in `TagHelperDescriptor`.

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+#if !DNXCORE50 // Not used in coreclr
```
dougbu: Remove useless comment or describe what doesn't work (why this class isn't used with Core).

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var declaringTypeInfo = propertyInfo.DeclaringType.GetTypeInfo();
+            return $"P:{declaringTypeInfo.FullName}.{propertyInfo.Name}";
+        }
+
+        private XElement GetMember(string id)
+        {
+            var associatedMemeber = _members
+                .FirstOrDefault(element =>
+                    string.Equals(element.Attribute("name").Value, id, StringComparison.Ordinal));
```
dougbu: Is `name` attribute required? If not need `?.`.

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,66 @@
...
+        public string GetRemarks(string id)
+        {
+            var associatedMemeber = GetMember(id);
+
+            return associatedMemeber?.Element("remarks")?.Value.Trim();
+        }
+
+        public bool HasDocumentation(string id)
+        {
+            return GetMember(id) != null;
```
dougbu: `<member>` may not contain `<summary>` or `<remarks>`. Either check that one or both exist or change method name to something like `MayHaveDocumentation()`. But best option is probably to remove this method and change caller to confirm at least one of `summary` or `remarks` are not `null` or empty before creating a usage descriptor.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
dougbu: 4. Do the work _once_. As mentioned in next class, get rid of `HasDocumentation()` and add `string.IsNullOrEmpty()` checks before creating the usage descriptor.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
dougbu: Not creating a `XmlDocumentationProvider` here, just evaluating paths. Leave as-is.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+using System.Reflection;
+using Microsoft.AspNet.Razor.TagHelpers;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public static class TagHelperUseageDescriptorFactory
+    {
+        private static readonly Dictionary<string, XmlDocumentationProvider> XmlDocumentationProviderCache =
+            new Dictionary<string, XmlDocumentationProvider>(StringComparer.Ordinal);
```
dougbu: Have you tested and confirmed this cache is necessary? Given the short flurry of tag helper descriptor resolution compared to the lifetime of the host process, could tie up a lot of memory for little reason. Also much better to add caches once we've measured and determined exactly what needs to be cached.

For now, even though we know the count of tag helper classes per tag helper assembly is likely greater than one, do not add this.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
```diff
@@ -0,0 +1,136 @@
...
+        public static TagHelperUseageDescriptor CreateDescriptor([NotNull] TypeInfo typeInfo)
+        {
+            var id = XmlDocumentationProvider.GetId(typeInfo);
+            return CreateDescriptorCore(typeInfo.Assembly, id);
+        }
+
+        public static TagHelperUseageDescriptor CreateDescriptor([NotNull] PropertyInfo propertyInfo)
+        {
+            var id = XmlDocumentationProvider.GetId(propertyInfo);
+            var declaringTypeAssembly = propertyInfo.DeclaringType.GetTypeInfo().Assembly;
```
dougbu: Throughout this PR, remove `Type` from `TypeAssembly` names. Every assembly contains types. `declaringAssembly` here, `assemblyLocation` (or just `location`) a few lines down, and so on.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
NTaylorMullen: Yes it's required.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
NTaylorMullen: This prevents Razor parsing from going to the file system. We've always pre-emptively added caches for file system like behaviors.

dougbu: There already is a cache of `TagHelperDescriptors` so this is redundant.

NTaylorMullen: That cache isn't in Razor, it's in VS outside of our control entirely.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: :facepalm:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: named parameters failed me. That's how it is in the method signature :/

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUseageDescriptorFactory.cs
NTaylorMullen: Don't feel strongly enough about it. Will remove.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+                {
+                    return null;
+                }
+            }
+
+            XmlDocumentationProvider documentationProvider = null;
+            var xmlDocumentationFile = GetXmlDocumentationFile(assemblyLocation);
+
+            // We only want to process the file if it exists. In the case it doesn't, a null value will be added
+            // to the cache to not constantly look for new XML files.
```
dougbu: Comment is no longer correct

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (xmlDocumentationFile != null)
+            {
+                documentationProvider = new XmlDocumentationProvider(xmlDocumentationFile.FullName);
+            }
+
+            // Members will be null if there is no associated XML file for the provided assembly.
```
dougbu: "Members" ??

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (documentationProvider != null)
```
dougbu: Can be combined with the above `if` and can move `var documentationProvider` inside the block.

dougbu:
Design outline looks good. Ready for the real thing.

NTaylorMullen:
Updated, this is no longer in Design. /cc @dougbu

NTaylorMullen:
Note: Was unable to validate a full end-to-end `TagHelperDescriptorFactory.CreateDescriptor`s creating `TagHelperUsageDescriptor`s due to the lack of XML files from our runtime (it loads projects into memory) and the requirement to have full reflection capability on `Type`s (if we wanted to attempt to fake a `Type`).

There was one alternative I thought of: Create a pre-build script that generates a package and then forces the test project to reference the package instead of the project on test run. Seemed to heavy handed for a minimal scenario so decided not to.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -38,23 +38,37 @@ public static class TagHelperDescriptorFactory
...
         {
             var typeInfo = type.GetTypeInfo();
-            var attributeDescriptors = GetAttributeDescriptors(type, errorSink);
+            var attributeDescriptors = GetAttributeDescriptors(type, designTime, errorSink);
             var targetElementAttributes = GetValidTargetElementAttributes(typeInfo, errorSink);
+            TagHelperUsageDescriptor typeUsageDescriptor = null;
+
+#if !DNXCORE50
+            // The TagHelperUsageDescriptor is consumed at design time to provide IntelliSense when typing HTML
+            // elements or attributes that are tag helper related.
```
dougbu: Don't see much value in this comment or its repetition later in this file. Reads like `<remarks>` in the descriptor classes and says nothing about the line of code. Suggest removing or perhaps just saying `// TagHelperUsageDescriptor information is not needed at runtime because it used only for IntelliSense.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -26,14 +26,16 @@ public class TagHelperDescriptorResolver : ITagHelperDescriptorResolver
...
         private readonly TagHelperTypeResolver _typeResolver;
+        private readonly bool _designTime;
 
         /// <summary>
         /// Instantiates a new instance of the <see cref="TagHelperDescriptorResolver"/> class.
         /// </summary>
-        public TagHelperDescriptorResolver()
-            : this(new TagHelperTypeResolver())
+        /// <param name="designTime">Indicates if the <see cref="TagHelperDescriptorResolver"/> should run in
+        /// design time mode</param>
```
dougbu: Be more specific, similar to `TagHelperDescriptorFactory` doc comments e.g. `Indicates whether resolved <see cref="TagHelperDescriptor"/>s should include design time-specific information`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using Microsoft.AspNet.Razor.TagHelpers;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Factory for <see cref="TagHelperUsageDescriptor"/>s from <see cref="Type"/>s and <see cref="PropertyInfo"/>s.
```
dougbu: Missing a word like "extracting" or "providing" after "for".

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            if (string.IsNullOrWhiteSpace(assemblyLocation) && !string.IsNullOrWhiteSpace(assembly.CodeBase))
+            {
+                var uri = new UriBuilder(assembly.CodeBase);
+
+                // Normalize the path to a UNC path. This will remove things like file:// from start of the uri.Path.
+                assemblyLocation = Uri.UnescapeDataString(uri.Path);
+
+                // Still couldn't resolve a valid assemblyLocation.
+                if (string.IsNullOrWhiteSpace(assemblyLocation))
```
dougbu: Move this check after the containing `if`. Won't currently be reached if `assemblyLocation` is `null` or whitespace and `CodeBase` is also `null` or whitespace.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+            var declaringAssembly = propertyInfo.DeclaringType.Assembly;
+
+            return CreateDescriptorCore(declaringAssembly, id);
+        }
+
+        private static TagHelperUsageDescriptor CreateDescriptorCore(Assembly assembly, string id)
+        {
+            var assemblyLocation = assembly.Location;
+
+            if (string.IsNullOrWhiteSpace(assemblyLocation) && !string.IsNullOrWhiteSpace(assembly.CodeBase))
```
dougbu: Why `IsNullOrWhiteSpace()` and not `IsNullOrEmpty()`?

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,100 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
Eilon: Fix filename casing.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                    assemblyXmlDocumentationFile = fallbackDirectories
+                        .Select(fallbackDiretory =>
+                            new FileInfo(
+                                Path.Combine(assemblyDirectory, fallbackDiretory, assemblyXmlDocumentationName)))
+                        .FirstOrDefault(file => file.Exists);
+                }
+
+                return assemblyXmlDocumentationFile;
+            }
+            catch (PathTooLongException)
```
Eilon: :frowning:

Eilon: But seriously, what's so special about PathTooLong? How was the original assembly loaded to begin with if the path is too long?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                return null;
+            }
+        }
+
+        private static IEnumerable<string> GetCultureFallbackDirectories()
+        {
+            var culture = CultureInfo.CurrentCulture;
+
+            // Following the fall-back process defined by:
+            // https://msdn.microsoft.com/en-us/library/sb6a8618.aspx#cpconpackagingdeployingresourcesanchor1
```
Eilon: Did you determine if this is correct for VS language settings?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                // Still couldn't resolve a valid assemblyLocation.
+                if (string.IsNullOrWhiteSpace(assemblyLocation))
+                {
+                    return null;
+                }
+            }
+
+            var xmlDocumentationFile = GetXmlDocumentationFile(assemblyLocation);
+
+            // We only want to process the file if it exists.
```
dougbu: In general try to avoid "we" and just describe what's going on. Here `// Only process the file if it exists.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                }
+            }
+
+            return null;
+        }
+
+        private static FileInfo GetXmlDocumentationFile(string assemblyLocation)
+        {
+            if (string.IsNullOrWhiteSpace(assemblyLocation) ||
+                assemblyLocation.IndexOfAny(InvalidPathCharacters) != -1)
```
dougbu: Directly using the file system. Why start with checks the file system does?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                var assemblyDirectory = Path.GetDirectoryName(assemblyLocation);
+                var assemblyName = Path.GetFileName(assemblyLocation);
+                var assemblyXmlDocumentationName = Path.ChangeExtension(assemblyName, ".xml");
+                var culture = CultureInfo.CurrentCulture;
+
+                var assemblyXmlDocumentationFile = new FileInfo(
+                    Path.Combine(assemblyDirectory, assemblyXmlDocumentationName));
+
+                // If there's not an XML file side-by-side the .dll it may exist in a culture specific directory.
+                if (!assemblyXmlDocumentationFile.Exists)
```
dougbu: If `assemblyXmlDocumentationFile.Exists`, end up ignoring the culture. Suggest instead falling back to `assemblyDirectory` if the `Select()` below fails to find a culture-specific file.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+            {
+                return null;
+            }
+
+            try
+            {
+                var assemblyDirectory = Path.GetDirectoryName(assemblyLocation);
+                var assemblyName = Path.GetFileName(assemblyLocation);
+                var assemblyXmlDocumentationName = Path.ChangeExtension(assemblyName, ".xml");
+                var culture = CultureInfo.CurrentCulture;
```
dougbu: Unused variable.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+
+                var assemblyXmlDocumentationFile = new FileInfo(
+                    Path.Combine(assemblyDirectory, assemblyXmlDocumentationName));
+
+                // If there's not an XML file side-by-side the .dll it may exist in a culture specific directory.
+                if (!assemblyXmlDocumentationFile.Exists)
+                {
+                    var fallbackDirectories = GetCultureFallbackDirectories();
+                    assemblyXmlDocumentationFile = fallbackDirectories
+                        .Select(fallbackDiretory =>
```
dougbu: typo s/b `fallbackDirectory`

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Xml.Linq;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime
+{
+    /// <summary>
+    /// Enables retrieval of summary and remarks XML documentation from an XML documentation file.
```
dougbu: Huh, what is this "enables" about? Suggest `Retrieves ...` or `Extracts ...`

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
```diff
@@ -0,0 +1,100 @@
...
+        private XElement GetMember(string id)
+        {
+            var associatedMemeber = _members
+                .FirstOrDefault(element =>
+                    string.Equals(element.Attribute("name").Value, id, StringComparison.Ordinal));
+
+            return associatedMemeber;
+        }
+
+        private static string GetUnnestedFullName(Type type)
```
dougbu: Why is this useful? Tag helper classes cannot be nested.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptorComparer.cs
```diff
@@ -31,7 +31,8 @@ protected TagHelperAttributeDescriptorComparer()
         /// <remarks>
         /// Determines equality based on <see cref="TagHelperAttributeDescriptor.IsIndexer"/>,
         /// <see cref="TagHelperAttributeDescriptor.Name"/>, <see cref="TagHelperAttributeDescriptor.PropertyName"/>,
-        /// and <see cref="TagHelperAttributeDescriptor.TypeName"/>. Ignores
+        /// <see cref="TagHelperAttributeDescriptor.TypeName"/>, and
+        /// <see cref="TagHelperAttributeDescriptor.UsageDescriptor"/>. Ignores
```
dougbu: All that needs to change in this file is the comments. Add `UsageDescriptor` to the list of ignored properties because it (like `IsStringProperty`) can be inferred directly from `TypeName`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptor.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    /// <summary>
+    /// A metadata class describing how to use a tag helper at design time.
```
dougbu: The part about `at design time` has nothing to do with this class; that choice is elsewhere. Suggest `A metadata class containing information about tag helper use.`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptor.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// <param name="summary">A summary on how to use a tag helper.</param>
+        /// <param name="remarks">Remarks on how to use a tag helper.</param>
+        public TagHelperUsageDescriptor(string summary, string remarks)
+        {
+            Summary = summary;
+            Remarks = remarks;
+        }
+
+        /// <summary>
+        /// A summary on how to use a tag helper.
```
dougbu: on -> of

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptor.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// </summary>
+        public string Summary { get; }
+
+        /// <summary>
+        /// Remarks on how to use a tag helper.
```
dougbu: on -> about

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptorComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using Microsoft.Internal.Web.Utils;
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    /// <summary>
+    /// An <see cref="IEqualityComparer{TagHelperUsageDescriptor}"/> used to check equality between
+    /// two <see cref="TagHelperUsageDescriptor"/>s.
+    /// </summary>
+    public class TagHelperUsageDescriptorComparer : IEqualityComparer<TagHelperUsageDescriptor>
```
dougbu: Class either doesn't need to exist or should be a test class.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
```diff
@@ -31,7 +31,8 @@ protected TagHelperDescriptorComparer()
         /// <remarks>
         /// Determines equality based on <see cref="TagHelperDescriptor.TypeName"/>,
         /// <see cref="TagHelperDescriptor.AssemblyName"/>, <see cref="TagHelperDescriptor.TagName"/>,
-        /// and <see cref="TagHelperDescriptor.RequiredAttributes"/>.
+        /// <see cref="TagHelperDescriptor.RequiredAttributes"/>, and
+        /// <see cref="TagHelperDescriptor.UsageDescriptor"/>.
```
dougbu: This class should not change. `UsageDescriptor` can be inferred directly from `TypeName`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -39,6 +40,8 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
         /// If <c>true</c> this <see cref="TagHelperAttributeDescriptor"/> is used for dictionary indexer assignments.
         /// Otherwise this <see cref="TagHelperAttributeDescriptor"/> is used for property assignment.
         /// </param>
+        /// <param name="usageDescriptor">The <see cref="TagHelperUsageDescriptor"/> that indicates how to use the
+        /// tag helper attribute during design time.</param>
```
dougbu: No indications here and "during design time" is not worth mentioning. For all documentation of `UsageDescriptor` parameters and properties, say `... that contains information about use of this attribute.` or `... this tag helper`. If you want to add `<remarks>` mentioning the properties are `null` at runtime, I guess that's OK but that ignores all the other cases in which the descriptor will be `null`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptorComparer.cs
```diff
@@ -0,0 +1,61 @@
...
+                string.Equals(descriptorX.Remarks, descriptorY.Remarks, StringComparison.Ordinal);
+        }
+
+        /// <inheritdoc />
+        public int GetHashCode(TagHelperUsageDescriptor descriptor)
+        {
+            if (descriptor == null)
+            {
+                return 0;
+            }
```
dougbu: Remove this and add `[NotNull]` to the parameter. Method is _supposed_ to `throw` if passed a `null` value.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CommonTagHelpers.cs
```diff
@@ -32,4 +35,36 @@ public class NonPublicAccessorTagHelper : TagHelper
...
+    /// <summary>
+    /// The summary for the type.
+    /// </summary>
+    /// <remarks>
+    /// The remarks for the type.
+    /// </remarks>
+    public class DocumentedTagHelper : TagHelper
+    {
+        public static readonly string XmlTestFileLocation =
+            Directory.GetCurrentDirectory() + "/TestFiles/TagHelperDocumentation.xml";
```
dougbu: This doesn't belong here.

test/Microsoft.AspNet.Razor.Runtime.Test/project.json
```diff
@@ -3,6 +3,7 @@
     "dependencies": {
         "Microsoft.AspNet.Razor.Runtime": "4.0.0-*",
         "Microsoft.AspNet.Testing": "1.0.0-*",
+        "Moq": "4.2.1312.1622",
```
dougbu: Bad practise to put a runtime-specific dependency in the general list. Move under frameworks / dnx451 / dependencies.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+
+        public static TheoryData CreateDescriptor_FallbackTypeData
+        {
+            get
+            {
+                // Faking localized dlls only to force the TagHelperUsageDescriptorFactory to search for localized XML.
+                var defaultENGBLocation = DocumentedTestAssemblyLocation.Replace(".dll", "-en-GB.dll");
+                var defaultENLocation = DocumentedTestAssemblyLocation.Replace(".dll", "-en.dll");
```
dougbu: Something is wrong if tests using these locations pass. Product code uses locale-specific sub-directories, not locale-specific XML file names.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
dougbu: This class assumes XML files will always be named identically to the corresponding satellite assembly and found in the same directories. The referenced page doesn't describe how XML files are found, just those satellites. And it's about desktop applications, not (DNX-created) localized packages.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,141 @@
...
+                            new FileInfo(
+                                Path.Combine(assemblyDirectory, fallbackDiretory, assemblyXmlDocumentationName)))
```
dougbu: A partial implementation of the process documented at https://msdn.microsoft.com/en-us/library/sb6a8618.aspx#cpconpackagingdeployingresourcesanchor1  Does not handle fallback assembly names or the precedence of a locale-specific filename in the same directory as the main assembly over one found in a sub-directory. Also does not handle anything about assemblies that are in the GAC, where XML files are likely found only in Reference Assemblies directories.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,316 @@
...
+                var invalidCodeBase = defaultCodeBase + '\0';
+
+                // tagHelperType, expectedUsageDescriptor
+                return new TheoryData<Type, TagHelperUsageDescriptor>
+                {
+                    { CreateDocumentationTagHelperTestType(location: null, codeBase: null), null },
+                    {
+                        CreateDocumentationTagHelperTestType(defaultLocation, codeBase: null),
+                        new TagHelperUsageDescriptor(
+                            summary: "The summary for the type.",
```
dougbu: `<summary>` and `<remarks>` rarely contain just simple text. Need to test more common content such as `<see cref="T:Microsoft.AspNet.Razor.Runtime.TagHelpers.ITagHelper"/> implementation targeting &lt;a&gt; elements.` and `Must be <c>null</c> if <see cref="P:Microsoft.AspNet.Mvc.TagHelpers.AnchorTagHelper.Route"/> is non-<c>null</c>.`.

test/Microsoft.AspNet.Razor.Runtime.Test/XmlDocumentationProviderTest.cs
```diff
@@ -0,0 +1,173 @@
...
+        private static readonly PropertyInfo DocumentedTagHelperRemarksPropertyInfo =
+            DocumentedTagHelperTypeInfo.GetProperty(nameof(DocumentedTagHelper.RemarksProperty));
+        private static readonly PropertyInfo DocumentedTagHelperRemarksSummaryPropertyInfo =
+            DocumentedTagHelperTypeInfo.GetProperty(nameof(DocumentedTagHelper.RemarksAndSummaryProperty));
+
+        [Fact]
+        public void CanReadXml()
+        {
+            // Act. Ensuring that reading the Xml file doesn't throw.
+            var xmlDocumentationProvider = new XmlDocumentationProvider(DocumentedTagHelper.XmlTestFileLocation);
```
dougbu: Remove `var xmlDocumentationProvider = `

test/Microsoft.AspNet.Razor.Runtime.Test/XmlDocumentationProviderTest.cs
```diff
@@ -0,0 +1,173 @@
...
+        public static TheoryData GetSummaryData
+        {
+            get
+            {
+                // id, expectedSummary
+                return new TheoryData<string, string>
+                {
+                    {
+                        XmlDocumentationProvider.GetId(DocumentedTagHelperTypeInfo),
+                        "The summary for the type."
```
dougbu: Rinse and repeat comment in `TagHelperUsageDescriptorFactoryTest` about common `<summary>` and `<remarks>` content.

test/Microsoft.AspNet.Razor.Runtime.Test/XmlDocumentationProviderTest.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            // Act
+            var id = XmlDocumentationProvider.GetId(DocumentedTagHelperRemarksSummaryPropertyInfo);
+
+            // Assert
+            Assert.Equal(expectedId, id, StringComparer.Ordinal);
+        }
+
+        [Fact]
+        public void GetId_TypeInfo_NestedClass()
```
dougbu: Remove all nested class testing. Not a real tag helper scenario.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,316 @@
...
+                    }
+                };
+            }
+        }
+
+        [Theory]
+        [ReplaceCulture]
+        [MemberData(nameof(CreateDescriptor_TypeData))]
+        [MemberData(nameof(CreateDescriptor_FallbackTypeData))]
+        public void CreateDescriptor_Type(Type tagHelperType, TagHelperUsageDescriptor expectedUsageDescriptor)
```
dougbu: Change all tests in the two new classes to have names that describe expectations. This one might be `CreateDescriptor_WithType_ReturnsExpectedDescriptors()`.

test/Microsoft.AspNet.Razor.Runtime.Test/XmlDocumentationProviderTest.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        public static TheoryData GetSummaryData
+        {
+            get
+            {
+                // id, expectedSummary
+                return new TheoryData<string, string>
+                {
+                    {
+                        XmlDocumentationProvider.GetId(DocumentedTagHelperTypeInfo),
```
dougbu: Strange to use tested class in `// Arrange`. Suggest providing the identifiers as literal strings.

test/Microsoft.AspNet.Razor.Runtime.Test/XmlDocumentationProviderTest.cs
```diff
@@ -0,0 +1,173 @@
...
+            // Assert
+            Assert.Equal(expectedId, id, StringComparer.Ordinal);
+        }
+
+        [Fact]
+        public void GetId_PropertyInfo()
+        {
+            // Arrange
+            var expectedId = $"P:{typeof(DocumentedTagHelper).FullName}." +
+                nameof(DocumentedTagHelper.RemarksAndSummaryProperty);
```
dougbu: Don't mix string interpolation and concatenation in the same expression. Pick one.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
Eilon: No need to support GAC reference for tag helpers, because they will never be in the GAC.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: :+1:, I added it because it was the only other exception the `Path` utilities could throw (I already handle the invalid path characters). Will remove.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: It's handled below but it does still attempt to locate the XML file. Will move down :+1:

src/Microsoft.AspNet.Razor.Runtime/XMLDocumentationProvider.cs
NTaylorMullen: Gah, fixed it. Git doesn't like me.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: No reason, being over protective. Will change.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: @Eilon talked with @ToddGrun about how the refs are picked up to provide localized text for other pieces in the editor (HTML editor etc.) and he said that it's just handled via `CultureInfo.CurrentCulture`, they don't do anything special to consume the VS specific setting.

@dougbu This isn't for DNX created packages: today DNX does not support satellite assemblies; @davidfowl said it's currently on the todo. When they do light up support for localized bits we'll revisit this and add support if it's needed (hopefully just works).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptorComparer.cs
NTaylorMullen: That assumes that both `TagHelperAttributeDescriptor`s come from the same assembly which is wrong (we don't have the assembly on the descriptor).

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperUsageDescriptorComparer.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/419#discussion_r32958698

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorComparer.cs
NTaylorMullen: Unlike the `TagHelperAttributeDescriptorComparer` :+1: to this. It can be inferred from both `TypeName` and `AssemblyName`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
NTaylorMullen: Wont add remarks but will change wording to do `contains information about use of this attribute` :+1:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptorComparer.cs
dougbu: My comment assumes we never compare attribute descriptors except within tag helper descriptors. Where is that invalidated? If it's incorrect, a lot more code needs to change.

NTaylorMullen: We don't, actually after doing a find for this attribute descriptor comparer it's only used in tests. Going to move it there in the form of `CaseSensitiveTagHelperAttributeDescriptorComparer`. Will do the same for `TagHelperUsageDescriptorComparer`.

test/Microsoft.AspNet.Razor.Runtime.Test/project.json
NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
NTaylorMullen: We were spoofing the names before to avoid a hit on the XML file that's located in the non-localized folders. Now that we're checking localized folders first prior to an XML file next to the DLL i'll adjust the tests accordingly.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
NTaylorMullen: :+1: found an issue with `.Value` on `XElement`s (it doesn't stringify contents that are also tags). Adding several use-cases of this all over.

NTaylorMullen:
Updated. Pretty Massive, things changed:

- Added support for culture specific assemblies when searching for localized XML.
- Added support for `NeutralResourcesLanguageAttribute` to allow an ultimate fall-back location when searching for XML.
- Moved `TagHelperAttributeDescriptorComparer` from the src projects into test (wasn't being used in src). This also involved changing the `CaseSensitiveTagHelperAttributeDescriptorComparer` to now be the normal `TagHelperAttributeDescriptorComparer`.
- Added several new localization use-cases for XML documentaiton consumption.
- Updated XML files to showcase more complex scenarios.

dougbu:
> Added support for `NeutralResourcesLanguageAttribute` to allow an ultimate fall-back location when searching for XML.

What is this about? `[NeutralResourcesLanguage]` describes resource locations within a set of assemblies, not a location on disc.

I _could_ see a slight enhancement in the future where Razor would use the `[NeutralResourcesLanguage]`'s `CultureName` property to determine the ultimate fallback language. But do that separately -- it only complicates the initial implementation and is clearly separable.

NTaylorMullen:
@dougbu Unless I'm reading the documentation wrong, when `NeutralResourceLanguage.Location` is set to `UltimateResourceFallbackLocation.Satellite` it indicates to utilize the provided culture to look for the culture specific satellite assembly with xml doc.

dougbu:
`UltimateResourceFallbackLocation.Satellite` means the _resources_ for the neutral language (aka the assembly's final fallback locale) are found in an satellite -- and not the main assembly. `UltimateResourceFallbackLocation.CultureName` is the name of that neutral language / fallback locale.

Where does the documentation mention XML docs?

NTaylorMullen:
It doesn't mention XML docs specifically anywhere in the documentation. Taking the same approach as resources ensures that we can load `TagHelper` documentation in the same locale they're seeing their errors/images/etc. in. A consistent experience.

dougbu:
> a consistent experience

Huh? Finding things inside an assembly group is completely different from finding other files that (usually) have the same base name as the assembly or a satellite. For one thing the runtime location of the assembly group is often _not_ where XML doc files are found.

NTaylorMullen:
It's a final fall-back location to search. In the case that there's no XML file for the current culture, in the current directory, of various names, *then* we look for the `NeutralLangaugeAttribute`. 

Are you suggesting we don't provide any `TagHelper` documentation in that case?

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -41,9 +43,12 @@ public TagHelperDescriptorResolver()
         /// specified <paramref name="typeResolver"/>.
         /// </summary>
         /// <param name="typeResolver">The <see cref="TagHelperTypeResolver"/>.</param>
-        public TagHelperDescriptorResolver(TagHelperTypeResolver typeResolver)
+        /// <param name="designTime">Indicates if the <see cref="TagHelperDescriptorResolver"/> should run in
+        /// design time mode</param>
```
dougbu: Use same description as in previous constructor.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,171 @@
...
+            if (string.IsNullOrEmpty(assemblyLocation) && !string.IsNullOrEmpty(assembly.CodeBase))
+            {
+                var uri = new UriBuilder(assembly.CodeBase);
+
+                // Normalize the path to a UNC path. This will remove things like file:// from start of the uri.Path.
+                assemblyLocation = Uri.UnescapeDataString(uri.Path);
+            }
+
+            // Couldn't resolve a valid assemblyLocation.
+            if (string.IsNullOrWhiteSpace(assemblyLocation))
```
dougbu: `IsNullOrEmpty()`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,171 @@
...
+                // Check for a localized XML file for the current culture.
+                var localizedXmlFile = GetLocalizedXmlDocumentationFile(
+                    CultureInfo.CurrentCulture,
+                    assemblyDirectory,
+                    assemblyXmlDocumentationName);
+                FileInfo assemblyXmlDocumentationFile;
+
+                if (localizedXmlFile == null)
+                {
+                    // Check for an uncultured XML file next to the assembly
```
dougbu: uncultured -> "culture-neutral" :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,171 @@
...
+                        if (!xmlFile.Exists)
+                        {
+                            // Look for a culture specific XML file in a culture specific directory.
+                            xmlFile = new FileInfo(
+                                Path.Combine(assemblyDirectory, cultureName, cultureSpecificFileName));
+                        }
+                    }
+
+                    return xmlFile;
+                }).FirstOrDefault(file => file.Exists);
```
dougbu: Looks like an unrolled loop a compiler has spit out in IL. Would be easier to understood if the `ExpandPaths()` method expanded all the paths and the Linq was just `Select(path => new FileInfo(path)).FirstOrDefault(file => file.Exists)`. Also avoids the extra `Exists` check and makes the linked page more relevant within `ExpandPaths()`. The method might be
``` c#
private static IEnumerable<string> ExpandPaths(
    CultureInfo culture,
    string assemblyDirectory,
    string assemblyXmlDocumentationName)
{
    do
    {
        var cultureName = culture.Name;
        var cultureSpecificFileName =
            Path.ChangeExtension(assemblyXmlDocumentationName, cultureName + ".xml");

        yield return Path.Combine(assemblyDirectory, cultureSpecificFileName);
        yield return Path.Combine(assemblyDirectory, cultureName, assemblyXmlDocumentationName));
        yield return Path.Combine(assemblyDirectory, cultureName, cultureSpecificFileName));

        culture = culture.Parent;
    } while (culture != null && culture != CultureInfo.InvariantCulture);
}
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,171 @@
...
+                var assemblyDirectory = Path.GetDirectoryName(assemblyLocation);
+                var assemblyName = Path.GetFileName(assemblyLocation);
+                var assemblyXmlDocumentationName = Path.ChangeExtension(assemblyName, ".xml");
+
+                // Check for a localized XML file for the current culture.
+                var localizedXmlFile = GetLocalizedXmlDocumentationFile(
+                    CultureInfo.CurrentCulture,
+                    assemblyDirectory,
+                    assemblyXmlDocumentationName);
+                FileInfo assemblyXmlDocumentationFile;
```
dougbu: Confusing to have two variables of the same type here; just need one `xmlDocumentationFile` variable.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,171 @@
...
+                    // Check for an uncultured XML file next to the assembly
+                    assemblyXmlDocumentationFile = new FileInfo(
+                        Path.Combine(assemblyDirectory, assemblyXmlDocumentationName));
+                }
+                else
+                {
+                    assemblyXmlDocumentationFile = localizedXmlFile;
+                }
+
+                return assemblyXmlDocumentationFile.Exists == true ? assemblyXmlDocumentationFile : null;
```
dougbu: Minor but this `Exists` check is unnecessary when using a localized file. Realize now you could do the opposite check of what used to be in the `if` block. Then just `return xmlDocumentationFile;` here. I.e.
``` c#
if (xmlDocumentationFile == null)
{
    xmlDocumentationFile = new FileInfo(
        Path.Combine(assemblyDirectory, assemblyXmlDocumentationName));
    if (!xmlDocumentationFile.Exists)
    {
        xmlDocumentationFile = null;
    }
}
```

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -136,6 +141,11 @@ public class TagHelperDescriptor
         /// <remarks>
         /// <c>*</c> at the end of an attribute name acts as a prefix match.
         /// </remarks>
-        public IReadOnlyList<string> RequiredAttributes { get; private set; }
+        public IReadOnlyList<string> RequiredAttributes { get; }
+
+        /// <summary>
+        /// The <see cref="TagHelperUsageDescriptor"/> that indicates how to use the tag helper at design time.
```
dougbu: Stick with the "contains information about" wording for all `TagHelperUsageDescriptor` properties and parameters.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeDescriptorComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public bool Equals(TagHelperAttributeDescriptor descriptorX, TagHelperAttributeDescriptor descriptorY)
+        {
+            if (descriptorX == descriptorY)
+            {
+                return true;
+            }
+
+            return descriptorX != null &&
```
dougbu: `descriptorY` may also be `null`; all we know is they aren't both `null`.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeDescriptorComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+                string.Equals(descriptorX.TypeName, descriptorY.TypeName, StringComparison.Ordinal) &&
+                TagHelperUsageDescriptorComparer.Default.Equals(
+                    descriptorX.UsageDescriptor,
+                    descriptorY.UsageDescriptor);
+        }
+
+        public int GetHashCode(TagHelperAttributeDescriptor descriptor)
+        {
+            // Ignore IsStringProperty because it is directly inferred from TypeName and thus won't vary the hash
+            // bucket. Base comparer does not care about Name case in its hash code but in tests we do.
```
dougbu: Correct out-of-date comment.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperUsageDescriptorComparer.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using Microsoft.Internal.Web.Utils;
+
+namespace Microsoft.AspNet.Razor.TagHelpers
+{
+    public class TagHelperUsageDescriptorComparer : IEqualityComparer<TagHelperUsageDescriptor>
+    {
+        public static readonly TagHelperUsageDescriptorComparer Default = new TagHelperUsageDescriptorComparer();
+
+        protected TagHelperUsageDescriptorComparer()
```
dougbu: `private`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorComparer.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.AspNet.Razor.TagHelpers;
+using Microsoft.Internal.Web.Utils;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperUsageDescriptorComparer : IEqualityComparer<TagHelperUsageDescriptor>
+    {
+        public static readonly TagHelperUsageDescriptorComparer Default = new TagHelperUsageDescriptorComparer();
+
+        protected TagHelperUsageDescriptorComparer()
```
dougbu: `private`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,385 @@
...
+            "TagHelpers.TagHelper\" />.";
+        private const string TestSummaryPropertySuffix = "This <see cref=\"P:Microsoft.AspNet.Razor.Runtime." +
+            "TagHelpers.DocumentedTagHelper.SummaryProperty\" /> is of type <see cref=\"T:System.String\" />.";
+        private const string TestRemarksPropertySuffix = "The <see cref=\"P:Microsoft.AspNet.Razor.Runtime." +
+            "TagHelpers.DocumentedTagHelper.SummaryProperty\" /> may be <c>null</c>.";
+        private const string TestRemarksAndSummaryPropertySummarySuffix = "This is a complex <see cref=\"T:System." +
+            "Collections.Generic.IDictionary`2\" />.";
+        private const string TestRemarksAndSummaryPropertyRemarksSuffix = "<see cref=\"P:Microsoft.AspNet.Razor." +
+            "Runtime.TagHelpers.DocumentedTagHelper.SummaryProperty\" /><see cref=\"P:Microsoft.AspNet.Razor.Runtime" +
+            ".TagHelpers.DocumentedTagHelper.RemarksProperty\" />";
```
dougbu: "Test" is noise, "Suffix" is misleading, and word order is confusing. So `TypeSummary`, `TypeRemarks`, `PropertySummary` (`PropertyWithSummary_Summary` seems a tad redundant), `PropertyRemarks`, `PropertyWithSummaryAndRemarks_Summary`, `PropertyWithSummaryAndRemarks_Remarks`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,385 @@
...
+
+        // These test assemblies don't really exist. They are used to look up corresponding XML for a fake assembly
+        // which is based on the DocumentedTagHelper type.
+        public static readonly string DocumentedTestAssemblyLocation =
+            Directory.GetCurrentDirectory() + "/TestFiles/NotLocalized/TagHelperDocumentation.dll";
+        public static readonly string LocalizedDocumentedTestAssemblyLocation =
+            Directory.GetCurrentDirectory() + "/TestFiles/Localized/TagHelperDocumentation.dll";
+        public static readonly string DocumentedTestAssemblyCodeBaseLocation =
+            "file:///" + DocumentedTestAssemblyLocation;
```
dougbu: Better but remove "Test" here and just `DocumentedAssemblyCodeBase` for the last field.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,385 @@
...
+            // Assert
+            Assert.Equal(expectedUsageDescriptor, usageDescriptor, TagHelperUsageDescriptorComparer.Default);
+        }
+
+        private static Type CreateDocumentationTagHelperTestType(string location, string codeBase)
+        {
+            return CreateTestType<DocumentedTagHelper>(location, codeBase);
+        }
+
+        private static Type CreateTestType<TWrappedType>(string location, string codeBase)
```
dougbu: Suggest `CreateType()`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperUsageDescriptorFactoryTest.cs
```diff
@@ -0,0 +1,385 @@
...
+            using (new CultureReplacer(culture))
+            {
+                usageDescriptor = TagHelperUsageDescriptorFactory.CreateDescriptor(mockPropertyInfo.Object);
+            }
+
+            // Assert
+            Assert.Equal(expectedUsageDescriptor, usageDescriptor, TagHelperUsageDescriptorComparer.Default);
+        }
+
+        private static Type CreateDocumentationTagHelperTestType(string location, string codeBase)
```
dougbu: Suggest `CreateDocumentedTagHelperType()`

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
NTaylorMullen: Hmm, I like it :+1:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperUsageDescriptorFactory.cs
```diff
@@ -0,0 +1,155 @@
...
+        {
+            do
+            {
+                var cultureName = culture.Name;
+                var cultureSpecificFileName =
+                    Path.ChangeExtension(assemblyXmlDocumentationName, cultureName + ".xml");
+
+                yield return Path.Combine(assemblyDirectory, cultureSpecificFileName);
+                yield return Path.Combine(assemblyDirectory, cultureName, assemblyXmlDocumentationName);
+                yield return Path.Combine(assemblyDirectory, cultureName, cultureSpecificFileName);
```
dougbu: I left comments out of my sample code. Please restore the link to `https://msdn.microsoft.com/en-us/library/sb6a8618.aspx#cpconpackagingdeployingresourcesanchor1` (above the `do`) and the short descriptions of each path e.g. `// A culture-specific XML file next to the assembly.` (with blank lines above each comment).

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeDescriptorComparer.cs
dougbu: thx

dougbu:
:shipit:

Issue #370, opened 2015-04-30
Refactor some of TagHelperParseTreeRewriterTest.cs into TagHelperBlockRewriterTest.cs


NTaylorMullen:
Right now pretty much every `TagHelper` rewriting test is in a single file. A while back we refactored the rewriting into two classes, therefore we should split out some of the `TagHelperParseTreeRewriterTest`s.

Right now the class is several thousand lines of tests :smile:

NTaylorMullen:
Taking care of this real quick because it'll speed up my work on another issue.

NTaylorMullen:
619cbc3716bdd294006a81a5da0761b624325d4b

PR #420, opened 2015-06-08
FYI: Refactored several tests from TagHelperParseTreeRewriterTest into Tag…

NTaylorMullen:
…HelperBlockRewriterTest.

- Split the tests into their corresponding impact areas. We were testing a lot of bits that were not specific to the TagHelperParseTreeRewriter in its corresponding tests.

#370


NTaylorMullen:
Just an FYI.

Issue #402, opened 2015-05-19
Exception filters cause Razor errors


NTaylorMullen:
If you write:

```C#
@try
{
    throw new InvalidOperationException();
}
catch (Exception e) when (e.Message.Contains("Operation is not valid"))
{
}
```

You get the following error:

```
CompilationFailedException: Compilation for 'c:\......\Index.cshtml' failed:
Expected a "{" but found a "if". Block statements must be enclosed in "{" and "}". You cannot use single-statement control-flow statements in CSHTML pages. For example, the following is not allowed:

@if(isLoggedIn)
<p>Hello, @user</p>

Instead, wrap the contents of the block in "{}":

@if(isLoggedIn) {
<p>Hello, @user</p>
} 
```

NTaylorMullen:
a1df1702e5376ca3f35164c80f83033ca2b679cc

PR #421, opened 2015-06-09
Add support for C# 6 exception filters.

NTaylorMullen:
- Added new handling of the C# try catch statement to allow exception filters after catch statements.
- Added tests to validate valid and invalid scenarios.

#402

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
@@ -13,6 +13,9 @@ namespace Microsoft.AspNet.Razor.Parser
 {
     public partial class CSharpCodeParser
     {
+        private static readonly Func<CSharpSymbol, bool> IsValidStatementSpacingSymbol =
+            IsSpacingToken(includeNewLines: true, includeComments: true);
+
         private void SetUpKeywords()
         {
             MapKeywords(ConditionalBlock, CSharpKeyword.For, CSharpKeyword.Foreach, CSharpKeyword.While, CSharpKeyword.Switch, CSharpKeyword.Lock);
@@ -247,19 +250,19 @@ namespace Microsoft.AspNet.Razor.Parser
         private void AfterTryClause()
         {
             // Grab whitespace
-            IEnumerable<CSharpSymbol> ws = SkipToNextImportantToken();
+            var whitespace = SkipToNextImportantToken();
 
             // Check for a catch or finally part
             if (At(CSharpKeyword.Catch))
             {
-                Accept(ws);
+                Accept(whitespace);
                 Assert(CSharpKeyword.Catch);
-                ConditionalBlock(topLevel: false);
+                FilterableCatchBlock();
                 AfterTryClause();
             }
             else if (At(CSharpKeyword.Finally))
             {
-                Accept(ws);
+                Accept(whitespace);
                 Assert(CSharpKeyword.Finally);
                 UnconditionalBlock();
             }
@@ -267,7 +270,7 @@ namespace Microsoft.AspNet.Razor.Parser
             {
                 // Return whitespace and end the block
                 PutCurrentBack();
-                PutBack(ws);
+                PutBack(whitespace);
                 Span.EditHandler.AcceptedCharacters = AcceptedCharacters.Any;
             }
         }
@@ -342,6 +345,41 @@ namespace Microsoft.AspNet.Razor.Parser
             AcceptAndMoveNext();
             AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
             ExpectCodeBlock(block);
+        }
+
+        private void FilterableCatchBlock()
+        {
+            Assert(CSharpKeyword.Catch);
+
+            var block = new Block(CurrentSymbol);
+
+            // Accept "catch"
+            AcceptAndMoveNext();
+            AcceptWhile(IsValidStatementSpacingSymbol);
+
+            // Parse the catch condition if present. If not present, let the C# compiler complain.
+            if (AcceptCondition())
+            {
+                AcceptWhile(IsValidStatementSpacingSymbol);
+
+                if (At(CSharpKeyword.When))
+                {
+                    // Accept "when".
+                    AcceptAndMoveNext();
+                    AcceptWhile(IsValidStatementSpacingSymbol);
+
+                    // Parse the filter condition if present. If not present, let the C# compiler complain.
+                    if (!AcceptCondition())
+                    {
+                        // Incomplete condition.
+                        return;
+                    }
+
+                    AcceptWhile(IsValidStatementSpacingSymbol);
+                }
+
+                ExpectCodeBlock(block);
+            }
         }
 
         private void ConditionalBlock(bool topLevel)
--- a/src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
+++ b/src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
@@ -88,7 +88,8 @@ namespace Microsoft.AspNet.Razor.Tokenizer
             { "interface", CSharpKeyword.Interface },
             { "break", CSharpKeyword.Break },
             { "checked", CSharpKeyword.Checked },
-            { "namespace", CSharpKeyword.Namespace }
+            { "namespace", CSharpKeyword.Namespace },
+            { "when", CSharpKeyword.When }
         };
 
         public static CSharpKeyword? SymbolTypeForIdentifier(string id)
--- a/src/Microsoft.AspNet.Razor/Tokenizer/Symbols/CSharpKeyword.cs
+++ b/src/Microsoft.AspNet.Razor/Tokenizer/Symbols/CSharpKeyword.cs
@@ -82,6 +82,7 @@ namespace Microsoft.AspNet.Razor.Tokenizer.Symbols
         Interface,
         Break,
         Checked,
-        Namespace
+        Namespace,
+        When
     }
 }
--- a/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
@@ -137,6 +137,169 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
                                ));
         }
 
+        public static TheoryData ExceptionFilterData
+        {
+            get
+            {
+                var factory = SpanFactory.CreateCsHtml();
+
+                // document, expectedStatement
+                return new TheoryData<string, StatementBlock>
+                {
+                    {
+                        "@try { someMethod(); } catch(Exception) when (true) { handleIO(); }",
+                        new StatementBlock(
+                            factory.CodeTransition(),
+                            factory
+                                .Code("try { someMethod(); } catch(Exception) when (true) { handleIO(); }")
+                                .AsStatement())
+                    },
+                    {
+                        "@try { A(); } catch(Exception) when (true) { B(); } finally { C(); }",
+                        new StatementBlock(
+                            factory.CodeTransition(),
+                            factory
+                                .Code("try { A(); } catch(Exception) when (true) { B(); } finally { C(); }")
+                                .AsStatement()
+                                .Accepts(AcceptedCharacters.None))
+                    },
+                    {
+                        "@try { A(); } catch(Exception) when (true) { B(); } catch(IOException) when (false) { C(); }",
+                        new StatementBlock(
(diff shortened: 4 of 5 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -13,6 +13,9 @@ namespace Microsoft.AspNet.Razor.Parser
 {
     public partial class CSharpCodeParser
     {
+        private static readonly Func<CSharpSymbol, bool> IsValidStatementSpacingSymbol =
+            IsSpacingToken(includeNewLines: true, includeComments: true);
```
dougbu: Why isn't this an `IsSpacingToken()` overload? Does the compiler not find the correct method to call?

src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
```diff
@@ -88,7 +88,8 @@ internal static class CSharpKeywordDetector
             { "interface", CSharpKeyword.Interface },
             { "break", CSharpKeyword.Break },
             { "checked", CSharpKeyword.Checked },
-            { "namespace", CSharpKeyword.Namespace }
+            { "namespace", CSharpKeyword.Namespace },
+            { "when", CSharpKeyword.When }
```
dougbu: This makes `when` a keyword in a few irrelevant contexts. Suggest instead looking for an identifier with the name `"when"` in `FilterableCatchBlock()`. (General problem that many supposed keywords in this collection are actually contextual. Let's not make that worse.)

src/Microsoft.AspNet.Razor/Tokenizer/Symbols/CSharpKeyword.cs
```diff
@@ -82,6 +82,7 @@ public enum CSharpKeyword
         Interface,
         Break,
         Checked,
-        Namespace
+        Namespace,
+        When
```
dougbu: Revert this too

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -344,6 +347,41 @@ private void UnconditionalBlock()
...
+        {
+            Assert(CSharpKeyword.Catch);
+
+            var block = new Block(CurrentSymbol);
+
+            // Accept "catch"
+            AcceptAndMoveNext();
+            AcceptWhile(IsValidStatementSpacingSymbol);
+
+            // Parse the catch condition, if present (if not present, we'll let the C# compiler complain).
```
dougbu: minor: Harder to read than necessary. It's just two sentences.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -344,6 +347,41 @@ private void UnconditionalBlock()
...
+            {
+                AcceptWhile(IsValidStatementSpacingSymbol);
+
+                if (At(CSharpKeyword.When))
+                {
+                    // Accept "when".
+                    AcceptAndMoveNext();
+                    AcceptWhile(IsValidStatementSpacingSymbol);
+
+                    // Parse the filter condition, if present (if not present, we'll let the C# compiler complain).
```
dougbu: ditto on minor comment here

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
dougbu: More generally a `static` method named `IsValidStatementSpacingSymbol` would avoid the extra indirection.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
```diff
@@ -137,6 +137,157 @@ public void CatchClause()
...
+        {
+            get
+            {
+                var factory = SpanFactory.CreateCsHtml();
+
+                // document, expectedStatement
+                return new TheoryData<string, StatementBlock>
+                {
+                    {
+                        "@try { someMethod(); } catch(Exception) when (true) { handleIO(); }",
```
dougbu: Does Razor currently have issues parsing an exception filter nested inside other C# block types? e.g.
``` c#
  @{
    try
    {
    }
    catch (Exception) when (Exception is ArgumentException)
    {
    }
  }
```

I doubt this is a problem. But either way, add one test using an exception filter in a statement block.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
NTaylorMullen: I'd rather stick with consistency here. If we want to break out keywords from this list lets do it in a bulk push instead of piecemealed.

src/Microsoft.AspNet.Razor/Tokenizer/Symbols/CSharpKeyword.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/421#discussion_r32056483

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
NTaylorMullen: Looks like it does :+1:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: Seems that `IsSpacingToken` is called all over with different variations of the parameters. I opted to not go with the `static method` approach so the `Func<CSharpSymbol, bool>` is only ever created once.

dougbu: You're making a trade-off toward performance and away from readability. Why?

NTaylorMullen: Not seeing the trade-off of readability. It's

```C#
AcceptWhile(IsValidStatementSpacingSymbol)
```

Vs.

```C#
AcceptWhile(IsValidStatementSpacingSymbol())
```

NTaylorMullen:
Updated

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
dougbu: That line of code would be identical either way. I'm talking about having to `F12` twice to reach the real code.

src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
dougbu: Consistently wrong is still wrong. Don't make it worse.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
NTaylorMullen: It's the difference of invoking a method to retrieve the value vs. having it stored. Not worth changing.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
```diff
@@ -137,6 +137,168 @@ public void CatchClause()
...
+                        string.Format("{0}{{{0}    B();{0}}}{0}catch(IOException) when (false)", Environment.NewLine) +
+                        string.Format("{0}{{{0}    C();{0}}}", Environment.NewLine),
+                        new StatementBlock(
+                            factory.CodeTransition(),
+                            factory
+                                .Code(
+                                    string.Format("try{0}{{{0}   A();{0}}}{0}catch(Exception) ", Environment.NewLine) +
+                                    string.Format("when (true){0}{{{0}    B();{0}}}{0}", Environment.NewLine) +
+                                    string.Format("catch(IOException) when (false){0}{{{0}    ", Environment.NewLine) +
+                                    $"C();{Environment.NewLine}}}")
```
ajaybhargavb: Minor nit: suggest keeping the string formatting consistent. (more readable)

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
dougbu: Having trouble parsing your responses but agreed offline to let this go.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
```diff
@@ -137,6 +137,168 @@ public void CatchClause()
...
+        public static TheoryData ExceptionFilterErrorData
+        {
+            get
+            {
+                var factory = SpanFactory.CreateCsHtml();
+                var unbalancedParenErrorString = "An opening \"(\" is missing the corresponding closing \")\".";
+                var unbalancedBracketCatchErrorString = "The catch block is missing a closing \"}\" character.  " +
+                    "Make sure you have a matching \"}\" character for all the \"{\" characters within this block, " +
+                    "and that none of the \"}\" characters are being interpreted as markup.";
+                var noErrors = new RazorError[0];
```
ajaybhargavb: Where is this used?

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -254,7 +257,7 @@ private void AfterTryClause()
             {
                 Accept(ws);
```
ajaybhargavb: Unrelated but since you are touching this, change to `var` here https://github.com/aspnet/Razor/pull/421/files#diff-f4f601d3f9c913c93d41ea331034c64fR253

src/Microsoft.AspNet.Razor/Tokenizer/CSharpKeywordDetector.cs
NTaylorMullen: It's not functionally wrong, if you were to label it anything it's extra sensitive to provide an accurate user experience. 

**Explanation:** Today, keywords and identifiers are treated identically with one exception: the editing experience. Keywords have always been seen as pieces of C# that "may" affect the C# parsing mechanics of Razor where identifiers are just filler code. Edit handlers protect this behavior by ensuring the editor re-parses the document when it encounters tokens that are or begin with keywords as the user is typing (the alternative to this behavior is to not re-parse and to just let the C# compiler do the complaining).

That's where the issue comes in; for Razor we must ensure that we know the document structure so we know when to transition into or out of HTML. For the `when` case, it's definitely rare for a user to run into it. But i'd rather the editor ensure it's document state when it does run into the keyword rather than having malformed `try` blocks disrupt the Razor editing experience.

Now to comment on my previous notion of "making a bulk push instead of doing things piecemealed" we could eventually get keywords to behave more contextually to prevent unneeded reparses. However, to do this right would involve new infrastructure and removal or modification of the current bits that make the Razor editing experience what it is today.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
NTaylorMullen: :+1:

ajaybhargavb:
:shipit: after removing unused variable.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpStatementTest.cs
```diff
@@ -137,6 +137,168 @@ public void CatchClause()
...
+                            factory
+                                .Code(
+                                    string.Format("try{0}{{{0}   A();{0}}}{0}catch(Exception) ", Environment.NewLine) +
+                                    string.Format("when (true){0}{{{0}    B();{0}}}{0}", Environment.NewLine) +
+                                    string.Format("catch(IOException) when (false){0}{{{0}    ", Environment.NewLine) +
+                                    $"C();{Environment.NewLine}}}")
+                                .AsStatement())
+                    },
+                    {
+                        "@{try { someMethod(); } catch(Exception) when (true) { handleIO(); }}",
```
dougbu: minor: Add a comment here because it's easy to miss the outermost curly braces (and think this is yet-another `@try` case).

dougbu:
:shipit: with code comment addition and @ajaybhargavb's fix

Issue #172, opened 2014-10-01
Enable support for instrumentation (Page Inspector) for TagHelpers


NTaylorMullen:
In TagHelpers initial state they will not support page inspector (instrumentation).  They will however work around the pre-existing instrumentation.  Meaning pieces of a `TagHelper`s body (if not `ContentBehavior.Modify`) will still communicate with page inspector via instrumentation.  

We need to add support for page inspector in TagHelpers.  However; it's not as simple as just adding it, we need to decide how much page inspector integration we'll have with TagHelpers.

danroth27:
From @jodavis 

Functional issue:
When content is generated by a tag helper (or even left unmodified by a tag helper), Browser Link can't map that element back to source.

Repro steps:
1. In a DNX website using MVC, add an IPageExecutionListenerFeature
2. Set a breakpoint in PageExecutionContext.BeginContext and EndContext
3. Run a Razor page that uses Tag Helpers.

Expected result:
BeginContext and EndContext are called for every range of the rendered page

Actual result:
BeginContext and EndContext are called for most of the rendered page, but not the parts rendered by Tag Helpers.

NTaylorMullen:
ddbd603653b001fcd0985ccc4b166d013c7df854

PR #422, opened 2015-06-10
Enable instrumentation for TagHelpers.

NTaylorMullen:
- Removed old code that disabled instrumentation for TagHelper bodies. Instrumentation will throw out sections for begin/end context that don't make it to the final output.
- Added instrumentation around WriteTagHelperAsync. Any content inside of the TagHelper will get sub-mapped via begin/end context.
- Hand verified BasicTagHelpers.cs begin/end context tests to ensure correctness.

#172


pranavkm:
:shipit:

Issue #382, opened 2015-05-10
Add OutputElementHintAttribute to control inherited TagHelper IntelliSense.


NTaylorMullen:
Right now if you create a `TagHelper` `<foo>` which translates into a `<ul>` you miss out on any `ul` specific HTML IntelliSense when writing `<foo `. It'd be nice to have an attribute that would cause the HTML editor to inherit specified tag IntelliSense. AKA

```C#
[OutputElementHint("ul")]
public class FooTagHelper : TagHelper
{
...
}
```

Which would result in `<foo ` invoking global and `ul` based IntelliSense.

DamianEdwards:
Nice idea, crappy name. I'd go with something like `[OutputElementName("ul")]` or similar. We don't want to include VS-specific names in our APIs. This is really just a "hint" to the IDE in case it cares.

NTaylorMullen:
:smile: :+1:

NTaylorMullen:
@DamianEdwards I'm thinking of adding support for multiple `OutputElementName`s like so:

```C#
[OutputElementName("ul", "div", "p")]
public class FooTagHelper : TagHelper
{
...
}
```

Which would indicate "give me all of the IntelliSense for `ul`, `div`, and `p` tags combined". Thoughts?

danroth27:
@NTaylorMullen Please add a tooling issue to tracking the tooling side of this work.

And the name is OutputElementHintAttribute.

NTaylorMullen:
85be731e2b39ea9a0e2d39cf11775f23c2a2528f
https://github.com/aspnet/Mvc/commit/15cb2ce709e7b118437c9ea39ad66f1ac9c4d5d3
https://github.com/aspnet/RazorTooling/commit/71a5a9f0c4026da3b368f4dbeaf4d210ad509eb8

PR #430, opened 2015-06-23
Add OutputElementHintAttribute to control design time IntelliSense for TagHelpers.

NTaylorMullen:
- Decided to not have the attribute inheritable since `TargetElement` is not inheritable.
- Added tests to validate serialization, deserialization and construction of `TagHelperDescriptor`s with `OutputElementHint`s.

#382

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OutputElementHintAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OutputElementHintAttribute.cs
@@ -1,0 +1,31 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Provides a hint of the <see cref="ITagHelper"/>'s output element.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    public sealed class OutputElementHintAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="OutputElementHintAttribute"/> class.
+        /// </summary>
+        /// <param name="outputElement">
+        /// The HTML element the <see cref="ITagHelper"/> may output.
+        /// </param>
+        public OutputElementHintAttribute([NotNull] string outputElement)
+        {
+            OutputElement = outputElement;
+        }
+
+        /// <summary>
+        /// The HTML element the <see cref="ITagHelper"/> may output.
+        /// </summary>
+        public string OutputElement { get; }
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -82,12 +82,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             IEnumerable<TargetElementAttribute> targetElementAttributes,
             bool designTime)
         {
-            TagHelperUsageDescriptor typeUsageDescriptor = null;
+            TagHelperDesignTimeDescriptor typeDesignTimeDescriptor = null;
 
 #if !DNXCORE50
             if (designTime)
             {
-                typeUsageDescriptor = TagHelperUsageDescriptorFactory.CreateDescriptor(typeInfo.GetType());
+                typeDesignTimeDescriptor = TagHelperDesignTimeDescriptorFactory.CreateDescriptor(typeInfo.GetType());
             }
 #endif
 
@@ -111,7 +111,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         assemblyName,
                         attributeDescriptors,
                         requiredAttributes: Enumerable.Empty<string>(),
-                        usageDescriptor: typeUsageDescriptor)
+                        designTimeDescriptor: typeDesignTimeDescriptor)
                 };
             }
 
@@ -122,7 +122,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         assemblyName,
                         attributeDescriptors,
                         attribute,
-                        typeUsageDescriptor));
+                        typeDesignTimeDescriptor));
         }
 
         private static TagHelperDescriptor BuildTagHelperDescriptor(
@@ -130,7 +130,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             TargetElementAttribute targetElementAttribute,
-            TagHelperUsageDescriptor usageDescriptor)
+            TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
             var requiredAttributes = GetCommaSeparatedValues(targetElementAttribute.Attributes);
 
@@ -140,7 +140,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 assemblyName,
                 attributeDescriptors,
                 requiredAttributes,
-                usageDescriptor);
+                designTimeDescriptor);
         }
 
         private static TagHelperDescriptor BuildTagHelperDescriptor(
@@ -149,7 +149,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             IEnumerable<string> requiredAttributes,
-            TagHelperUsageDescriptor usageDescriptor)
+            TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
             return new TagHelperDescriptor(
                 prefix: string.Empty,
@@ -158,7 +158,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 assemblyName: assemblyName,
                 attributes: attributeDescriptors,
                 requiredAttributes: requiredAttributes,
-                usageDescriptor: usageDescriptor);
+                designTimeDescriptor: designTimeDescriptor);
         }
 
         /// <summary>
@@ -462,12 +462,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             bool isIndexer,
             bool designTime)
         {
-            TagHelperUsageDescriptor propertyUsageDescriptor = null;
+            TagHelperAttributeDesignTimeDescriptor propertyDesignTimeDescriptor = null;
 
 #if !DNXCORE50
             if (designTime)
             {
-                propertyUsageDescriptor = TagHelperUsageDescriptorFactory.CreateDescriptor(property);
+                propertyDesignTimeDescriptor =
+                    TagHelperDesignTimeDescriptorFactory.CreateAttributeDescriptor(property);
             }
 #endif
 
@@ -476,7 +477,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 property.Name,
                 typeName,
                 isIndexer,
-                propertyUsageDescriptor);
+                propertyDesignTimeDescriptor);
         }
 
         private static bool IsAccessibleProperty(PropertyInfo property)
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
@@ -14,40 +14,69 @@ using Microsoft.Framework.Internal;
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
     /// <summary>
-    /// Factory for providing <see cref="TagHelperUsageDescriptor"/>s from <see cref="Type"/>s and
-    /// <see cref="PropertyInfo"/>s.
+    /// Factory for providing <see cref="TagHelperDesignTimeDescriptor"/>s from <see cref="Type"/>s and
+    /// <see cref="TagHelperAttributeDesignTimeDescriptor"/>s from <see cref="PropertyInfo"/>s.
     /// </summary>
-    public static class TagHelperUsageDescriptorFactory
+    public static class TagHelperDesignTimeDescriptorFactory
     {
         /// <summary>
-        /// Creates a <see cref="TagHelperUsageDescriptor"/> from the given <paramref name="type"/>.
+        /// Creates a <see cref="TagHelperDesignTimeDescriptor"/> from the given <paramref name="type"/>.
         /// </summary>
-        /// <param name="type">The <see cref="Type"/> to create a <see cref="TagHelperUsageDescriptor"/> from.</param>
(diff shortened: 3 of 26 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/OutputElementHintAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Provides a hint of the <see cref="ITagHelper"/>'s output element.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```
NTaylorMullen: Decided to not have the attribute inheritable since `TargetElement` is not inheritable.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -83,13 +83,18 @@ public static class TagHelperDescriptorFactory
...
             TagHelperUsageDescriptor typeUsageDescriptor = null;
+            string outputElementHint = null;
 
-#if !DNXCORE50
             if (designTime)
             {
+                var outputElementHintAttribute =
+                    typeInfo.GetCustomAttribute<OutputElementHintAttribute>(inherit: false);
+
+                outputElementHint = outputElementHintAttribute?.OutputElement;
```
dougbu: Doubt the `GetCustomAttribute<T>()` is so expensive that it needs to be inside this condition. Also this factory should only execute once at runtime. So calculate the hint unconditionally.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -147,5 +151,11 @@ public class TagHelperDescriptor
         /// The <see cref="TagHelperUsageDescriptor"/> that indicates how to use the tag helper at design time.
         /// </summary>
         public TagHelperUsageDescriptor UsageDescriptor { get; }
+
+        /// <summary>
+        /// The HTML element the tag helper may output.
+        /// </summary>
+        /// <remarks>Overrides the HTML IntelliSense provided at design time.</remarks>
```
dougbu: This is a Visual Studio and perhaps OmniSharp-specific comment and should not be so definite. Suggest "In IDEs supporting Intellisense, may override the HTML information provided at design time."

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -147,5 +151,11 @@ public class TagHelperDescriptor
...
+        public string OutputElementHint { get; set; }
```
dougbu: What is the expected experience when multiple tag helpers target say `<joe>` and one's hint is "fred" but the other has no hint? How about both target '<joe>' and one's hint is "fred" but the other's is "ginger"?

dougbu: Same question for multiple `UsageDescriptor` values for the same element or attribute.

NTaylorMullen: Hmm good question. We could:

1. Take first hint provided.
1. Concat hints (but if we do this why not allow a single `TagHelper` to provide multiple hints).
1. Error.

/cc @DamianEdwards

NTaylorMullen: `UsageDescriptor`s I feel like should be handled at `DesignTime`. No need to special case them for our purposes. Lot of options there for how to showcase that info to a user.

dougbu: W.r.t. the list, first option would make more sense if it were "Take the last hint provided" (based on execution order) because that helper can override all that ran before it.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: :+1: Updated options.

NTaylorMullen: Wait actually shoot, can't do that. we'd need to construct the object to read the order.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
Eilon: Why waste cycles? The intention of this attribute is that it's design-time only.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
Eilon: I thought when we chatted earlier this week the idea was that you could have lots of hints and they'd all be provided to the tool, which would presumably union together all the Intellisense for all the elements?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: I made this suggestion because it seemed like an unnecessary optimization that mixed the hint with descriptors and required another non-`var` variable.

Eilon: I'd be more tempted to have a whole property called `DesignTimeStuff` that then had properties for each of the design-time only parts of the descriptor, which for now might be just the hint.

dougbu: Like it. `UsageDescriptor` would be in there too.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: @Eilon yes that was the idea. I talked to @DamianEdwards though about allowing multiple hints and he didn't see the value in it. Hence this question :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Could revamp `TagHelperUsageDescriptor` to be the design time stuff property. Would just have to add an extra type maybe like `TagHelperElementUsageDescriptor` which would contain `OutputElementHint` for `TagHelperDescriptor`s. `TagHelperAttributeDescriptor`s would continue to use `TagHelperUsageDescriptor`.

What do you guys think?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
Eilon: We could always have the system *provide* all the tags, and then let the tool decide what to do with it.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
Eilon: I don't know what these things are, so I don't think much :smile:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: Ya that's what I was thinking with 1 & 2 (enforced by the editor) but wanted to present 3 since having conflicting `OutputElementHint`s could potentially indicate that `TagHelper`s will clash. 

3 isn't typically what we've pushed for with other `TagHelper` features since we've normally taken the route of flexibility; but wanted to bring it up since there wasn't a clear path.

Eilon: Tag helpers could always do crazy stuff... in the end one of them will win - that's why they're ordered.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDesignTimeDescriptor.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// </summary>
+    public class TagHelperDesignTimeDescriptor
+    {
+        /// <summary>
+        /// Instantiates a new instance of <see cref="TagHelperDesignTimeDescriptor"/>.
+        /// </summary>
+        /// <param name="summary">A summary on how to use a tag helper.</param>
+        /// <param name="remarks">Remarks on how to use a tag helper.</param>
+        /// <param name="outputElementHint">The HTML element a tag helper may output.</param>
+        public TagHelperDesignTimeDescriptor(string summary, string remarks, string outputElementHint)
```
Eilon: Should this avoid the ctor and just have get/set properties? I'm sure we'll add lots more stuff in the future, and it might be annoying to have to keep making the ctor have more and more params.

Eilon: In fact, same goes for the other descriptors.

Eilon:
Overall looks good, just had a question about the class design.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDesignTimeDescriptor.cs
NTaylorMullen: It's deffffinitely annoying. The only reason why this descriptor uses the ctor to initialize everything is because the `TagHelperDescriptor` and `TagHelperAttributeDescriptor` do; aka for consistency. 

At the initial creation of the descriptors I had them all constructorless but after a while the constructors were added and have stuck around. @dougbu If I remember correctly you showed the interest in adding the constructors initially which made sense at the time. Now that there's a boat load of properties on all of our descriptors how do you feel about nuking the ctors?

Will file an issue if @dougbu agrees (changing from ctors to no ctors will be a lot of noise + require changes to all of our comparers).

Eilon: Not only are there a boat load of properties, they're not all required.

I'm fine treating that as a separate issue.

dougbu: We moved toward constructor parameters to make the descriptors immutable while allowing serialization / deserialization. Comparisons, hash codes, and so on get messy if the descriptors can change. I know _we_ wouldn't change the descriptors but we do store some of them in `HashSet`s and so on.

So fine to open an issue but I'm not sure we should ever fix it :smiling_imp:

Eilon: What value does immutability add? Also, why is hashing these important? Are these ever compared?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: What was the decision here?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDesignTimeDescriptor.cs
dougbu: The containing descriptors are compared and / or stuck into `HashSet`s.

NTaylorMullen: https://github.com/aspnet/Razor/issues/441

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -14,40 +14,69 @@
...
+        public static TagHelperDesignTimeDescriptor CreateDescriptor([NotNull] Type type)
         {
             var id = XmlDocumentationProvider.GetId(type);
+            var documentationDescriptor = CreateDocumentationDescriptor(type.Assembly, id);
 
-            return CreateDescriptorCore(type.Assembly, id);
+            // Purposefully not using the TypeInfo.GetCustomAttributes method here to enable mockable unit testing.
+            var outputElementHintAttribute = type
+                .GetCustomAttributes(inherit: false)?
+                .OfType<OutputElementHintAttribute>()
```
dougbu: Keep the `?.` operator together

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -14,40 +14,69 @@
...
+        /// <returns>A <see cref="TagHelperDesignTimeDescriptor"/> that describes design time specific information
         /// for the given <paramref name="type"/>.</returns>
-        public static TagHelperUsageDescriptor CreateDescriptor([NotNull] Type type)
+        public static TagHelperDesignTimeDescriptor CreateDescriptor([NotNull] Type type)
         {
             var id = XmlDocumentationProvider.GetId(type);
+            var documentationDescriptor = CreateDocumentationDescriptor(type.Assembly, id);
 
-            return CreateDescriptorCore(type.Assembly, id);
+            // Purposefully not using the TypeInfo.GetCustomAttributes method here to enable mockable unit testing.
```
dougbu: Think you mean `.. to make it easier to mock the Type.`

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDesignTimeDescriptor.cs
dougbu: If others feel strongly about changing `TagHelperAttributeDesignTimeDescriptor` and `TagHelperDesignTimeDescriptor` to be mutable, suggest adding the `public` setters in this PR. Leave #441 to cover the existing `TagHelperAttributeDescriptor` and `TagHelperDescriptor`. This is possible / easy because the usage descriptors are only compared in test code.

Eilon: I'm presuming that the descriptors are never, ever mutated, so in practice this isn't an issue.

Eilon: This PR is 7 days old. Let's get it in and change the classes later.

dougbu:
We've got a buried discussion starting [here](https://github.com/aspnet/Razor/pull/430#discussion-diff-33269494)

dougbu:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: Consume them and let the editor decide what to do.

NTaylorMullen:
Updated `?.` spacing and comment.

dougbu:
:shipit: (I wasn't waiting on those minor comments :smiling_imp:)

Issue #418, opened 2015-06-04
Duplicate bound attributes should flow through as unbound attributes.


NTaylorMullen:
If you have a `TagHelper` with duplicate (same name) attributes only the first attribute gets added to `TagHelperContext.AllAttributes`. Ex:

```HTML
<myTagHelper abc="..." ABC="..." />
```

Today the second one `ABC` is ignored completely. Instead it should flow through to the output as if it were unbound.

dougbu:
Same thing happens even when the attribute names have identical case e.g.
``` html
 <myTagHelper name="value1" name="value2" />
```

Only the first name / value pair (containing `"value1"`) will be in `AllAttributes`.

NTaylorMullen:
0bb9457fd97c81c2e1a285dd9825b618f2755272

PR #432, opened 2015-06-23
Change duplicate TagHelper bound attributes to flow to output.

NTaylorMullen:
- Duplicate TagHelper bound attributes used to be ignored entirely; they now flow to the output as if they were unbound.
- Added code generation test to verify duplicate attributes. Added runtime and design time versions.
- Updated existing tests that happened to have duplicate bound attributes.

#418

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
@@ -229,15 +229,10 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attributeName)));
 
 
-                // Bound attributes have associated descriptors.
-                if (associatedDescriptors.Any())
+                // Bound attributes have associated descriptors. First attribute value wins if there are duplicates;
+                // later values of duplicate bound attributes are treated as if they were unbound.
+                if (associatedDescriptors.Any() && renderedBoundAttributeNames.Add(attributeName))
                 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -229,15 +229,9 @@ protected virtual string GenerateUniqueId()
                     descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attributeName)));
 
 
-                // Bound attributes have associated descriptors.
-                if (associatedDescriptors.Any())
+                // Bound attributes have associated descriptors. First attribute value wins if there are duplicates.
```
dougbu: Add enough to this comment to make it obvious the `else` behaviour is intended. `RenderUnboundAttribute()` method name raises questions otherwise. For example ... "Treat later values of bound attributes as if they were unbound."

dougbu:
Experience when editing is going to be strange. For example, user types in
``` html
  <element int-attribute="@joe + @john" />
```
then adds
``` html
  <element int-attribute="23" int-attribute="@joe + @john" />
```

Is switch from non-`string` behaviour to `string` behaviour (or visa versa) a special case where we want an error?

NTaylorMullen:
Ran that by @DamianEdwards and he wanted to keep it as flexible as possible without restricting users.

dougbu:
:watch:

NTaylorMullen:
What's the :watch: for?

dougbu:
> What's the :watch: for?

The answer to my [question](https://github.com/aspnet/Razor/pull/432#issuecomment-115308629).

dougbu:
:shipit: with added comment

Issue #387, opened 2015-05-13
Incorrect classification (and therefore IntelliSense) for expressions containing transition characters (ex @)


dougbu:
When providing values for non-`string` tag helper properties, Razor authors lose IntelliSense and other C# support when the expression contains transitions.  A simple example would be `<element non-string="@object + value" />`.

Using existing MVC tag helpers in a contrived way, the following illustrates the issue.  Note only `@ValidationSummary.All` is highlighted in the second case.  Even the attribute name gets the wrong classification.
![highlighting](https://cloud.githubusercontent.com/assets/6431421/7604287/ecef64b8-f8f4-11e4-9de1-3c7af7a025f8.png)

Problem is likely related to fact not all `Span` components of the more-complex expression get `SpanKind.Code`.  In particular `TagHelperBlockRewriter.CreateMarkupAttribute()` isn't called on every code path through `TagHelperBlockRewriter.TryParseBlock()`.

NTaylorMullen:
039062c5eb1e3c2f6f5320995406a40fb448b250

PR #435, opened 2015-06-24
Fix C# design time projections for complex TagHelper attributes.

NTaylorMullen:
- TagHelper attributes that have expressions intermingled within them (resulting in Block elements) resulted in Spans in the attribute being falsely marked as SpanKind.Markup.
- Updated tests to account for new SpanKind.Code behavior.
- Added complex scenario to validate SpanKind.Code is flowed through to surround attributes.

#387

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
@@ -254,7 +254,7 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
             // If we're not after an equal then we should treat the value as if it were a minimized attribute.
             var attributeValueBuilder = afterEquals ? builder : null;
             result.AttributeValueNode =
-                CreateMarkupAttribute(name, attributeValueBuilder, result.IsBoundNonStringAttribute);
+                CreateMarkupAttribute(attributeValueBuilder, result.IsBoundNonStringAttribute);
 
             return result;
         }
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -451,9 +476,9 @@ private static SourceLocation GetAttributeNameStartLocation(SyntaxTreeNode node)
             if (builder != null)
             {
                 // If the attribute was requested by a tag helper but the corresponding property was not a string,
-                // then treat its value as code. A non-string value can be any C# value so we need to ensure the
-                // SyntaxTreeNode reflects that.
-                if (isBoundNonStringAttribute)
+                // then treat its value as code if it was previously markup. A non-string value can be any C# value so
+                // we need to ensure the SyntaxTreeNode reflects that.
+                if (isBoundNonStringAttribute && builder.Kind == SpanKind.Markup)
```
dougbu: Are Razor comments allowed inside an attribute?
Think meta code is not legal here. Right?
Why leave transitions as-is?

NTaylorMullen: Could go either way on the Razor comments. Today they generate nothing and therefore have no impact but definitely break up the consistency of being in C#. As for metacode: Metacode spans can only occur in invalid Razor blocks within the attributes (errors are created). 

I like the option of enforcing only C# in the attributes and that does flow similar to what #436 is getting at. Will remove the restriction on `.Kind`.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -675,12 +727,12 @@ public static TheoryData CodeTagHelperAttributesData
...
                                             new MarkupBlock(
-                                                factory.Markup("@").Accepts(AcceptedCharacters.None),
-                                                factory.Markup("@")
+                                                factory.CodeMarkup("@").Accepts(AcceptedCharacters.None),
+                                                factory.CodeMarkup("@")
                                                     .With(SpanChunkGenerator.Null)
                                                     .Accepts(AcceptedCharacters.None)),
                                             new MarkupBlock(
-                                                factory.EmptyHtml(),
+                                                factory.EmptyHtml(SpanKind.Code),
```
dougbu: Rather odd. Could this instead be `factory.EmptyCode()` and hide `private static Span EmptySpan(SpanFactory factory, SpanKind spanKind)` inside `TestSpanBuilder`?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: :+1:

dougbu:
:shipit: after improving test helpers slightly

dougbu:
.. and removing that one condition

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
NTaylorMullen: Added `.As(SpanKind)` to `SpanConstructor` so it can be applied to any `Span`.

Issue #399, opened 2015-05-19
Support indexer assignments with get-only dictionary properties


dougbu:
This is a follow up to the fix for #89. Once that goes in, we'll support indexer assignments (i.e. `tagHelper.Property["{dictionary key}"] = value;`) for any `public` tag helper property of the appropriate type.

However the `public` requirement is currently identical to that for any other tag helper property: Must have a `public` setter in addition to a `public` getter. That `public` setter should be unnecessary for this feature. In addition settable collection properties are something of an anti-pattern.

We should separate setting a dictionary property directly (1) completely from supporting indexer assignments to the same property (2). After #89 is fixed, (1) will work without (2) but not (2) without (1).

dougbu:
69d8e52

PR #440, opened 2015-06-27
Add support for get-only indexer properties

dougbu:
- #399

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
@@ -14,22 +14,39 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         private string _dictionaryAttributePrefix;
 
         /// <summary>
+        /// Instantiates a new instance of the <see cref="HtmlAttributeNameAttribute"/> class with <c>Name==null</c>.
+        /// </summary>
+        /// <remarks>
+        /// Associated property must not have a public setter and must be compatible with
+        /// <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/> where <c>TKey</c> is
+        /// <see cref="string"/>.
+        /// </remarks>
+        public HtmlAttributeNameAttribute()
+        {
+        }
+
+        /// <summary>
         /// Instantiates a new instance of the <see cref="HtmlAttributeNameAttribute"/> class.
         /// </summary>
-        /// <param name="name">HTML attribute name for the associated property.</param>
+        /// <param name="name">
+        /// HTML attribute name for the associated property. Must be <c>null</c> or empty if associated property does
+        /// not have a public setter and is compatible with
+        /// <see cref="System.Collections.Generic.IDictionary{TKey, TValue}"/> where <c>TKey</c> is
+        /// <see cref="string"/>. Otherwise must not be <c>null</c> or empty.
+        /// </param>
         public HtmlAttributeNameAttribute(string name)
         {
-            if (string.IsNullOrEmpty(name))
-            {
-                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(name));
-            }
-
             Name = name;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1755,6 +1779,14 @@ private class MultipleValidHtmlAttributePrefix : TagHelper
 
             [HtmlAttributeName("valid-name6")]
             public string StringProperty { get; set; }
+
+            public IDictionary<string, int> GetOnlyDictionaryProperty { get; }
+
+            [HtmlAttributeName("valid-name8")]
+            public IDictionary<string, object> GetOnlyDictionaryPropertyWithAttributeName { get; }
+
+            [HtmlAttributeName("ignored-name9", DictionaryAttributePrefix = "valid-prefix9")]
```
NTaylorMullen: This is a realllly weird case. I can see users providing `ignored-name9` and `valid-prefix9` and trying to use `ignored-name9` and having no clue why it's not binding. Suggest modifying `HtmlAttributeName` to provide a `public string Name { get; set; }` and a 0 and 1 ctor overload option. That then moves the `isNullOrEmpty` check in the 1 arg `name` ctor over to the `TagHelperDescriptorFactory` or a helper method on `HtmlAttributeName`; whatever works. 

Nice part of this is it'll let us get rid of the ugliness of `string.IsNullOrEmpty` in the `HtmlAttributeName` ctor. We were doing a bad practice since it's an attribute :cry:

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: Should also make the case of a `{ get; }` attribute with a non-null (or empty if you want, either way is fine with me) `HtmlAttributeName.Name` an error.

dougbu: Discussed w/ @Eilon. I'll do something similar, adding and moving errors as appropriate. Will do another iteration because I expect this will affect at least a few tests.

dougbu:
@NTaylorMullen updated

/cc @Eilon lots of new error cases as we discussed
/cc @DamianEdwards this changes the `[HtmlAttributeName]` semantics a bit. Existing cases are almost-all still supported and the attribute constructors no longer `throw`.

Eilon:
Looks :shipit: to me. But we should consider splitting the attribute into two: one for dictionaries, and one for non-dictionaries. Or something.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
```diff
@@ -14,22 +14,39 @@ public sealed class HtmlAttributeNameAttribute : Attribute
         private string _dictionaryAttributePrefix;
 
         /// <summary>
+        /// Instantiates a new instance of the <see cref="HtmlAttributeNameAttribute"/> class with <c>Name==null</c>.
```
NTaylorMullen: `with <see cref="Name" /> set to <c>null</c>.` (so it can be refactored effectively).

NTaylorMullen:
Missing error tests for newly added errors (you said this would be the case so :+1:, I can add them while you're on vaca).

:shipit:, i'll push the error test cases later.

dougbu:
> Missing error tests

Yup. Hit one or two but a number of the cases aren't yet covered. Thanks very much.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlAttributeNameAttribute.cs
dougbu: done

dougbu:
69d8e52

PR #442, opened 2015-06-30
Making CreateCodeWriter protected so that it can be overridden in Mvc tests.

sornaks:


src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -24,8 +24,8 @@ public CSharpCodeGenerator(CodeGeneratorContext context)
         private ChunkTree Tree { get { return Context.ChunkTreeBuilder.ChunkTree; } }
         public RazorEngineHost Host { get { return Context.Host; } }
 
-        // Internal for testing
-        internal virtual CSharpCodeWriter CreateCodeWriter()
+        // Protected for testing
```
NTaylorMullen: Put this in a `<summary>` in-case someone tries to override it.

NTaylorMullen:
:shipit:

sornaks:
Thanks! https://github.com/aspnet/Razor/commit/1f480386f4fd36aacb4e0f3959469901aab67ed9

Issue #247, opened 2014-11-26
Auto Attribute removal in TagHelpers


yishaigalatzer:
Attributes that in normal razor code gets automatically removed gets special treatment with TagHelpers.

1. Only attributes that are not bound to tag helper properties will keep the existing razor behavior.
1. These attributes will generate the "naive" string and populate the TagHelperContext.AllAttributes
1. Expression Values that are result in false, null or string.Empty are stripped from TagHelperOutput.Attributes.
  1. Example input: <input class="@false" />
     1. TagHelperContext.AllAttributes will include [class, "false"]
     1. TagHelperOutput.Attributes will not include class.
  1. For other examples follow unit tests for WriteAttribute.

dougbu:
think we tested these scenarios and found `attribute="@string.Empty"` resulted in `attribute=""` with a regular (non-tag helper) attribute.  only `attribute="@null"` and `attribute="@false"` lead to no generation for the attribute.  tag helper attributes should do the same thing.

FYI, see special-cases at:
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L336 (for `null`)
- https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Razor/RazorPage.cs#L358 (for `false`)

NTaylorMullen:
4604c807c49daa560b92868503922bb89096a40d
https://github.com/aspnet/Mvc/commit/a0da6ec19f17dd0579ab0424044778608e483997

PR #446, opened 2015-07-09
[Design] Add ability for unbound complex `TagHelper` attributes to flow through MVC attribute resolution system.

NTaylorMullen:
- Updated `CSharpCodeVisitor` and `CSharpTagHelperCodeRenderer` to re-use more code.
- Added custom attribute visitor for complex unbound `TagHelper` attributes.
- Added an additional three variables used for `TagHelper` codegen.
- Expanded `GeneratedTagHelperContext` to contain more information about the final `TagHelper` code gen output for complex attributes.
- Updated existing test files.
- Added a new code gen test case to showcase the various use-cases of unbound dynamic `TagHelper` attributes.

**Expected Behavior:**
`TagHelper` attributes that are unbound behave identically to non-`TagHelper` attributes. This means that `checked="@false"` results in the attribute not being rendered. `checked="@true"` results in `checked="checked"` and `checked="hello @null world"` results in `checked="hello world"`.

When an attribute value is modified due to these rules a `TagHelper` author can see the unchanged value via `TagHelperContext.AllAttributes` and the changed value will show up in `TagHelperOutput.Attributes`. Therefore `checked="@true"` results in `checked="checked"` in `TagHelperContext.AllAttributes` and `checked="True"` in `TagHelperOutput.Attributes`. In the case that the attribute is entirely removed, it will only exist in `TagHelperContext.AllAttributes`.

#247

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -63,6 +63,11 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
             return (CSharpCodeWriter)base.WriteLine();
         }
 
+        public CSharpCodeWriter WriteEndLine()
+        {
+            return WriteLine(";");
+        }
+
         public CSharpCodeWriter WriteVariableDeclaration(string type, string name, string value)
         {
             Write(type).Write(" ").Write(name);
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
+using System.IO;
 using System.Linq;
 using Microsoft.AspNet.Razor.Chunks;
 using Microsoft.AspNet.Razor.CodeGenerators.Visitors;
@@ -21,6 +23,11 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         internal static readonly string StringValueBufferVariableName = "__tagHelperStringValueBuffer";
         internal static readonly string ScopeManagerVariableName = "__tagHelperScopeManager";
         internal static readonly string RunnerVariableName = "__tagHelperRunner";
+        internal static readonly string RawAttributeValueVariableName = "__rawTagHelperAttributeValue";
+        internal static readonly string OriginalTagHelperAttributeValueVariableName =
+            "__originalTagHelperAttributeValue";
+        internal static readonly string ShouldRenderTagHelperAttributeVariableName =
+            "__shouldRenderTagHelperAttribute";
 
         private readonly CSharpCodeWriter _writer;
         private readonly CodeGeneratorContext _context;
@@ -304,12 +311,12 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                 }
 
                 // We need to inform the context of the attribute value.
+                RenderStartAddAttributes(
+                    _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName,
+                    attributeName,
+                    encodedValue: false);
+
                 _writer
-                    .WriteStartInstanceMethodInvocation(
-                        ExecutionContextVariableName,
-                        _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName)
-                    .WriteStringLiteral(attributeName)
-                    .WriteParameterSeparator()
                     .Write(currentValueAccessor)
                     .WriteEndMethodInvocation();
 
@@ -323,7 +330,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                 _writer
                     .WriteStartAssignment(currentValueAccessor)
                     .Write(previousValueAccessor)
-                    .WriteLine(";");
+                    .WriteEndLine();
 
                 return previousValueAccessor;
             }
@@ -348,7 +355,10 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                 {
                     // If we haven't recorded a value and we need to buffer an attribute value and the value is not
                     // plain text then we need to prepare the value prior to setting it below.
-                    BuildBufferedWritingScope(attributeValueChunk, htmlEncodeValues: false);
+                    BuildBufferedWritingScope(renderValue: () =>
+                    {
+                        _literalBodyVisitor.Accept(attributeValueChunk);
+                    });
                 }
 
                 _writer.WriteStartAssignment(valueAccessor);
@@ -365,7 +375,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     RenderBufferedAttributeValue(attributeDescriptor);
                 }
 
-                _writer.WriteLine(";");
+                _writer.WriteEndLine();
             }
             else
             {
@@ -397,7 +407,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
 
                     // End the assignment to the attribute.
                     lineMapper.MarkLineMappingEnd();
-                    _writer.WriteLine(";");
+                    _writer.WriteEndLine();
                 }
             }
         }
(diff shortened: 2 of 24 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -63,6 +63,11 @@ public new CSharpCodeWriter WriteLine()
             return (CSharpCodeWriter)base.WriteLine();
         }
 
+        public CSharpCodeWriter WriteEndLine()
```
dougbu: Propagates lousy `bool endLine` parameter naming. Should be `WriteEndStatement()` and should probably be used in this class as well.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -416,7 +428,54 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
                 // C# code, then we need to buffer it.
                 if (!isPlainTextValue)
                 {
-                    BuildBufferedWritingScope(attributeValueChunk, htmlEncodeValues: true);
+                    BuildBufferedWritingScope(renderValue: () =>
+                    {
+                        IChunkVisitor visitor;
+                        var parentChunk = attributeValueChunk as ParentChunk;
+                        containsDynamicCodeAttributeChunk = parentChunk
+                            .Children
```
dougbu: What's the point of the conditional cast if you're going to dereference the result immediately?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -416,7 +428,54 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                            .Any(child => child is DynamicCodeAttributeChunk);
+
+                        if (!_designTimeMode && containsDynamicCodeAttributeChunk)
+                        {
+                            // If there are any non-dynamic code attribute chunks at the top level of the attribute
+                            // value's chunk it means there's static content in the value, the attribute will always
+                            // render.
```
dougbu: Comment describes the opposite case of this block. Reword to cover what's important when `containsDynamicCodeAttributeChunk` is `true`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -416,7 +428,54 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                            forceRenderAttribute = parentChunk
+                                .Children
+                                .Any(child => !(child is DynamicCodeAttributeChunk));
+
+                            if (!forceRenderAttribute)
+                            {
+                                _writer
```
dougbu: A comment here about the case where the attribute contains both `DynamicCodeAttributeChunk` and something else might help. Might even be the comment above.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -416,7 +428,54 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                                .WriteStartAssignment(OriginalTagHelperAttributeValueVariableName)
+                                .WriteStartNewObject(_tagHelperContext.UnchangedTagHelperAttributeValueBufferTypeName)
+                                .Write(_tagHelperContext.OutputTextWriterPropertyName)
+                                .Write(".")
+                                .Write(nameof(TextWriter.Encoding))
+                                .WriteLine(");");
+
+                            visitor = new UnboundDynamicAttributeValueVisitor(
+                                attributeName,
+                                forceRenderAttribute,
```
dougbu: Since you're passing this in, moving lines [448](https://github.com/aspnet/Razor/pull/446/files#diff-6f3af99214f3f3e7141c9d7bd2aea6c0R448) through [462](https://github.com/aspnet/Razor/pull/446/files#diff-6f3af99214f3f3e7141c9d7bd2aea6c0R462) into `UnboundDynamicAttributeValueVisitor`. This block would need to be refactored otherwise.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -436,32 +495,89 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
-                    .WriteStringLiteral(attributeName)
-                    .WriteParameterSeparator()
-                    .WriteStartMethodInvocation(_tagHelperContext.MarkAsHtmlEncodedMethodName);
+                    .WriteStringLiteral(textValue)
+                    .WriteEndMethodInvocation(endLine: false)
+                    .WriteEndMethodInvocation();
+            }
+            else if (containsDynamicCodeAttributeChunk)
+            {
+                // Complex attribute value that had to be buffered.
```
dougbu: Confusing. Don't all complex attribute values need to be buffered?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -436,32 +495,89 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
 
+                RenderStartAddAttributes(
+                    _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName,
+                    attributeName,
+                    encodedValue: true);
+
+                _writer
+                    .Write(OriginalTagHelperAttributeValueVariableName)
+                    .WriteEndMethodInvocation(endLine: false)
+                    .WriteEndMethodInvocation();
```
dougbu: Rather a lot going on in this block. Suggest refactoring into another method.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -436,32 +495,89 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                    encodedValue: true);
+
                 _writer
+                    .Write(StringValueBufferVariableName)
                     .WriteEndMethodInvocation(endLine: false)
                     .WriteEndMethodInvocation();
             }
         }
 
+        private void RenderStartAddAttributes(string methodName, string attributeName, bool encodedValue)
```
dougbu: Name should not be plural.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -639,6 +750,126 @@ private static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
...
+                string attributeName,
+                bool forceRenderAttribute,
+                CSharpTagHelperCodeRenderer tagHelperRenderer,
+                CSharpCodeWriter writer,
+                CodeGeneratorContext context)
+                : base(writer, context)
+            {
+                Debug.Assert(
+                    !Context.Host.DesignTimeMode,
+                    $"{nameof(UnboundDynamicAttributeValueVisitor)} should not be used during design time.");
```
dougbu: minor: `typeof(UnboundDynamicAttributeValueVisitor).FullName` will make it easier to get here if the assertion fails.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -639,6 +750,126 @@ private static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
...
+            protected override void Visit(LiteralChunk chunk)
+            {
+                base.Visit(chunk);
+
+                if (!_renderingDynamicValue)
+                {
+                    var currentWriter = Context.TargetWriterName;
+                    Context.TargetWriterName = OriginalTagHelperAttributeValueVariableName;
+
+                    var currentInstrumentation = Context.Host.EnableInstrumentation;
```
dougbu: Suggest `current` -> `original` for both of these variable names.

NTaylorMullen:
After the new WriteAttribute changes need to re-work how this is done. Closing for now, will re-open fresh.

Issue #447, opened 2015-07-10
Consume EditorBrowsableAttribute on TagHelpers.


NTaylorMullen:
By applying `[EditorBrowsable(EditorBrowsableState.Never)]` to your `TagHelper`s class and adding the metadata to `TagHelperDescriptor`s tooling can consume the metadata to hide `TagHelper` highlighting on HTML elements affected only by descriptors with `EditorBrowsableState.Never`. We can also add this functionality for bound attributes.

We should add this metadata to `TagHelperDesignTimeDescriptor.EditorBrowsable` and `TagHelperAttributeDesignTimeDescriptor.EditorBrowsable`.

NTaylorMullen:
b762830a437716b28b35bcc8d821dffe7267a47d

PR #448, opened 2015-07-11
Consume `EditorBrowsableAttribute` on `TagHelper`s.

NTaylorMullen:
- Changed `TagHelperDescriptorFactory` to not create individual descriptors when `EditorBrowsableAttribute` is present and set to `EditorBrowsableState.Never`.
- Added tests to validate the `TagHelperDescriptorFactory` creates the attribute correctly.
- Did not look down the inheritance chain for `EditorBrowsableAttribute` because `TargetElement` is not inherited.

#447

### DIFF
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 using Microsoft.AspNet.Razor.TagHelpers;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -38,13 +39,20 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor([NotNull] Type type
                 ?.OfType<OutputElementHintAttribute>()
                 .FirstOrDefault();
             var outputElementHint = outputElementHintAttribute?.OutputElement;
+            var editorBrowsableAttribute = type
+                .GetCustomAttributes(inherit: false)
```
NTaylorMullen: Not looking down the inheritance chain here because `TargetElement` isn't inherited. However, the default inheritance behavior of `EditorBrowsableAttribute` is `true`. I'm open to opinions :smile:

NTaylorMullen:
Changed behavior to not create `TagHelperDescriptor`s at design time.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1564,6 +1796,59 @@ public void ValidateTagHelperAttributeDescriptor_WithValidPrefix_ReturnsTrue(str
...
+            public int Property { get; set; }
+        }
+
+        private class HiddenPropertyEditorBrowsableTagHelper : TagHelper
+        {
+            [EditorBrowsable(EditorBrowsableState.Never)]
+            public int Property { get; set; }
+        }
+
+        private class MultiHiddenPropertyEditorBrowsableTagHelper : TagHelper
```
dougbu: This class has more than one property but how does "MultiHidden" make sense?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1564,6 +1796,59 @@ public void ValidateTagHelperAttributeDescriptor_WithValidPrefix_ReturnsTrue(str
...
+        private class OverriddenEditorBrowsableTagHelper : EditorBrowsableTagHelper
+        {
+            [EditorBrowsable(EditorBrowsableState.Advanced)]
+            public override int Property { get; set; }
+        }
+
+        [TargetElement("p")]
+        [TargetElement("div")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        private class MultiEditorBrowsableTagHelper : TagHelper
```
dougbu: This class has more than one `[TargetElement]` but how does "EditorBrowsable" make sense?

dougbu:
Implementation looks fine but please test end-to-end behaviour with `[EditorBrowsable]` on a tag helper class. In particular does Visual Studio provide IntelliSense (for properties of this class) when typing within a `services.InitializeTagHelper<NonBrowsableTagHelper>((h, vc) => ...)` delegate?

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: More than 1 property on the class.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: Hu? It's just the name of the test class.

- Multi = more than 1 element
- EditorBrowsable = Has the `EditorBrowsableAttribute`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: Will remove the "Hidden" part of the name.

NTaylorMullen:
@dougbu manually tested the comment here: https://github.com/aspnet/Razor/pull/448#issuecomment-121129218. As for end-to-end testing of this behavior. Can't really validate the design time portion of that; however, the runtime portion will be validated via https://github.com/aspnet/Mvc/issues/2807 since it will be `[EditorBrowsable(EditorBrowsableState.Never)]`.

dougbu:
Not a fan of the ambiguous "Multi" test class names...

dougbu:
:shipit:

Issue #454, opened 2015-07-23
[TagHelper] Setting EditorBrowsableState.Never for a single bound attribute affects all attributes


ajaybhargavb:
Example:
```c#
private class MultiPropertyEditorBrowsableTagHelper : TagHelper 
{ 
    [EditorBrowsable(EditorBrowsableState.Never)] 
    public int Property { get; set; } 

    public virtual int Property2 { get; set; } 
 } 
```
Both the properties are not highlighted.

cc' @NTaylorMullen

NTaylorMullen:
c87dd8d9de3eebd05e14745bf340b04f0fd7c181

PR #455, opened 2015-07-23
Fix EditorBrowsableAttribute detection to apply to individual attributes.

NTaylorMullen:
- Prior to this change we'd return early and not generate any found descriptors if ANY property on a `TagHelper` had editor browsable never.
- This issue was hidden from tests due to us using the wrong comparer. Updated the comparer and failures occurred without the `TagHelperDescriptorFactory` change (yay).

#454


Eilon:
@ajaybhargavb please review.

ajaybhargavb:
:shipit:

Issue #461, opened 2015-07-27
Expose test comparers so other Razor projects can use them.


NTaylorMullen:
Current `Microsoft.AspNet.Tooling.Razor` is using non-test variations of Razor's `TagHelperDescriptor` comparers. Instead of duplicating all the code for these comparers we should move the current comparers to a `Microsoft.AspNet.Razor.Test.Sources` project in `/src` (so a package is created). `Microsoft.AspNet.Razor.Test.Sources` should be a sources package that does not publish packages externally; this way we can use the common Razor test utilities across our repos but not impact external consumers.

NTaylorMullen:
4d0821f953ded0e7595591ed45b78a7bdb44df7a

PR #462, opened 2015-07-27
Transition Razor test comparers to new project.

NTaylorMullen:
- Changed namespaces for test comparers.
- Did project.json format change in separate commit.

#461


test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDesignTimeDescriptorComparer.cs
```diff
@@ -1,41 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Generic;
-using Microsoft.Internal.Web.Utils;
-
-namespace Microsoft.AspNet.Razor.TagHelpers
-{
-    public class TagHelperDesignTimeDescriptorComparer : IEqualityComparer<TagHelperDesignTimeDescriptor>
```
dougbu: Nice to get rid of this duplicate class.

src/Microsoft.AspNet.Razor.Test.Sources/CaseSensitiveTagHelperDescriptorComparer.cs
```diff
@@ -6,9 +6,9 @@
 using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.Internal.Web.Utils;
 
-namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+namespace Microsoft.AspNet.Razor.Test.Internal
```
dougbu: Shouldn't have `.Test` in the namespace except for test infrastructure classes. So `.Internal` seems redundant. Why not just `Microsoft.AspNet.Razor.Test`?

src/Microsoft.AspNet.Razor.Test.Sources/project.json
```diff
@@ -0,0 +1,13 @@
+﻿{
+  "version": "1.0.0-*",
```
dougbu: Odd for this project to be inconsistent with the rest of the repo. Suggest `4.0.0-*`

dougbu:
In the aspnet/Common repo, we've required a test class for every shared class. Probably useful to do that here as well.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Test.Sources/CaseSensitiveTagHelperDescriptorComparer.cs
NTaylorMullen: `.Internal` is consistent with `aspnet/Common`

src/Microsoft.AspNet.Razor.Test.Sources/project.json
NTaylorMullen: I thought of that as well, but this isn't specific to the other Razor dependencies. Don't feel strongly about it though. /cc @Eilon

NTaylorMullen:
@dougbu these are internal utilities for tests. We don't typically test test utilities.

src/Microsoft.AspNet.Razor.Test.Sources/project.json
Eilon: Meh don't care :smile:

Eilon:
This is just shuffling code around, so :shipit:

Issue #449, opened 2015-07-13
HashCodeCombiner should not be public


anurse:
The [HashCodeCombiner](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor/Common/HashCodeCombiner.cs) class is a purely internal helper class. If it's shared, it should be shared via source code sharing (using a shared-source package or a reference in the project.json `compile` project).

It's extremely unrelated to the project and collides heavily with other types, causing weird things like this:

![image](https://cloud.githubusercontent.com/assets/7574/8661621/c2564f10-296d-11e5-8770-2dab712160a2.png)

Above I was pressing Ctrl-. to see if the project I was working in had HashCodeCombiner lying around and saw this :).

@NTaylorMullen and I chatted and suggested just making copies of the class with different namespaces. It's not pretty but it works. This isn't a huge deal, but it is weird that Razor would be leaking such an internal infrastructure class to the public.

dougbu:
How is this an issue for our customers? `HashCodeCombiner` is hidden away in the `Microsoft.Internal.Web.Utils` namespace.

anurse:
As I said, it's probably not a big issue. Unfortunately, Ctrl-. and package search means it isn't fully hidden as it can easily appear in those lists. You do have to know what you're looking for (at least partly, in the case of package search) but it still seems like we're "productizing" a hack used to work around compiler issues :)

dougbu:
I'm missing something. Similar things could be said about every `public` class in one of our `.Internal` namespaces. Why is `HashCodeCombiner` different?

anurse:
[...]

PR #463, opened 2015-07-27
Transition `HashCodeCombiner` to use the `aspnet/Common` variation.

NTaylorMullen:
- Updated `LocationTagged<TValue>` to handle `null` implicit values.
- Removed `InternalsVisibleTo` declaration on `Microsoft.AspNet.Razor` to the runtime test project.
- Updated `TagHelperDescriptor` tests to utilize helper methods to construct `TagHelperDescriptor`s. This was needed since the `InternalsVisibleTo` declaration was removed.
- Formatted `project.json`s in different commit.

#449


src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
```diff
@@ -5,5 +5,4 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.AspNet.Razor.Test")]
-[assembly: InternalsVisibleTo("Microsoft.AspNet.Razor.Runtime.Test")]
```
anurse: One down! Yay :)

src/Microsoft.AspNet.Razor.Runtime/project.json
```diff
@@ -1,36 +1,52 @@
...
-            }
-        }
+    "Microsoft.Framework.CopyOnWriteDictionary.Sources": {
+      "type": "build",
+      "version": "1.0.0-*"
+    },
+    "Microsoft.Framework.HashCodeCombiner.Sources": {
+      "type": "build",
+      "version": "1.0.0-*"
+    },
```
dougbu: What needs this? The Runtime project isn't currently using a `HashCodeCombiner`

NTaylorMullen: :+1:

anurse:
:shipit:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -717,9 +717,7 @@ public void CreateDescriptors_BuildsDescriptorsWithInheritedProperties()
         {
             // Arrange
             var errorSink = new ErrorSink();
-
-            // Also confirm isStringProperty is calculated correctly.
```
dougbu: This was important when I added `isStringProperty`. Why isn't it still?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -1883,6 +1856,35 @@ public void ValidateTagHelperAttributeDescriptor_WithValidPrefix_ReturnsTrue(str
             return data;
         }
 
+        private static TagHelperDescriptor CreateTagHelperDescriptor(
```
dougbu: This refactoring removes coverage of a number of `TagHelperDescriptor` constructors. What is the upside?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -25,9 +25,14 @@ private static TagHelperDescriptor Valid_PlainTagHelperDescriptor
...
-                                               Valid_PlainTagHelperType.FullName,
-                                               AssemblyName);
+                return new TagHelperDescriptor(
+                    prefix:  string.Empty,
+                    tagName: "valid_plain",
+                    typeName: Valid_PlainTagHelperType.FullName,
+                    assemblyName: AssemblyName,
+                    attributes: Enumerable.Empty<TagHelperAttributeDescriptor>(),
+                    requiredAttributes: Enumerable.Empty<string>(),
+                    designTimeDescriptor: null);
```
dougbu: Ick. These constructor overloads exist in part to avoid the extra verbiage in our tests.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -1402,6 +1422,13 @@ private static TagHelperDescriptor CreatePrefixedStringDescriptor(string prefix)
                 assemblyName: stringType.GetTypeInfo().Assembly.GetName().Name);
         }
 
+        private static TagHelperDirectiveDescriptor CreateTagHelperDirectiveDescriptor(
+            string directiveText,
+            TagHelperDirectiveType directiveType)
+        {
+            return new TagHelperDirectiveDescriptor(directiveText, SourceLocation.Zero, directiveType);
+        }
```
dougbu: Again, what is the upside of this refactoring?

src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
dougbu: Hmm, what is the upside of removing this IVF? We have no restrictions against making `internal`s visible to test projects. This change is not needed at all, even if we need `HashCodeCombiner` in the test project (because the Runtime product code doesn't use that class).

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: There's only 1 `TagHelperDescriptor` constructor used at runtime, the public one. The other internal ones were there for testing convenience.

src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
NTaylorMullen: With the `InternalsVisibleTo` `Microsoft.AspNet.Razor.Runtime` would never be able to use an `aspnet/Common` package that `Microsoft.AspNet.Razor` was using. 

Currently not an issue but definitely an unneeded restriction.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
NTaylorMullen: This value is determined by the single public `TagHelperDescriptor` constructor. We only use the public ctor in src projects.

src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
Eilon: Why would we *keep* an unnecessary IVT?

dougbu: 1. @NTaylorMullen's comment just above (about using `aspnet/Common` packages) is not correct. `Microsoft.AspNet.Razor.Runtime` and `Microsoft.AspNet.Razor` are already using the same `aspnet/Common` packages e.g. `Microsoft.Framework.NotNullAttribute.Sources`. Problems online arise if a test project references both assemblies (i.e. must be the `Microsoft.AspNet.Razor.Runtime.Test` project) _and_ the test project itself uses something from the duplicate `aspnet/Common` packages. Remote case and not a problem we need to fix today.
2. Because this part of the change represents churn for churn's sake.
3. We don't make changes based on personal preferences which seems to be the real issue here.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
dougbu: The test was previously explicit about all properties. Now it's not. :-1:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
dougbu: Right. Now things are less convenient.

In addition, why bother wrapping a single constructor call with a pseudo-helper? This method and the next one don't have a reason for being. Certainly aren't necessary with the existing `internal` constructors.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
dougbu: Realized this one doesn't even have an optional parameter. One argument less isn't much of an advantage over calling the constructor directly, even if you insist on not using the `internal` one.

dougbu:
:watch:

Suggestion: If we're going to rewrite tests to avoid a few `internal` "for testing only" constructors, it should be done separately, because it's necessary or at least because we've discussed the point and agreed (grudgingly is fine of course :smiley_cat:). Further, we should completely remove the constructors; not update one of the two test projects using them to avoid a harmless IVT attribute.

src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
Eilon: IVT is allowed only from product assembly X to unit test assembly X.Tests. This isn't personal preference.

dougbu: @Eilon both of these IVT attributes are from this product assembly to the unit test assemblies. Are you saying the relationship is required to be 1:1?

dougbu: The MVC repo also contains a number of product projects with IVT attributes for more than one test project. If 1:1 is a requirement, is this new?

Eilon: Nothing new: https://github.com/aspnet/Home/wiki/Engineering-guidelines#when-to-use-internals-vs-public-and-when-to-use-internalsvisibleto

This seems like an opportunistic fix.

dougbu: These IVT attributes are about unit testing. Where is the 1:1 requirement in the guidelines?

This seems like an unrelated change mixed into an actual fix.

Eilon: It says:

>  `InternalsVisibleTo` is used only to allow a unit test to test internal types and members of its runtime assembly.

Because there is a 1:1 mapping between a unit test assembly and a runtime assembly<sup>[citation available](https://github.com/aspnet/Home/wiki/Engineering-guidelines#assembly-naming)</sup>, the IVT restriction is the logical conclusion of those two guidelines.

Let's please table this discussion now.

NTaylorMullen:
@dougbu The internal ctors are just one of the things used by `Microsoft.AspNet.Razor.Tests`. The IVT to the `Microsoft.AspNet.Razor.Test` project is used for muchhh more than internal ctors (method calls, classes etc.). Not seeing a compelling reason to break this PR apart, it's relatively small.

src/Microsoft.AspNet.Razor/Properties/AssemblyInfo.cs
dougbu: :ok:

dougbu:
> Not seeing a compelling reason to break this PR apart

We do individual commits / PRs for individual issues. My suggestion is to at least split the commits to match the two unrelated things in this PR::
- Use the new aspnet/Common `HashCodeCombiner`
- Remove IVT from Razor to Razor.Runtime.Test

NTaylorMullen:
Closing out this PR since the HashCodeCombiner bit was signed off on. WIll open another one for the IVT since that seems to have caused the most bit of conflict.

Issue #464, opened 2015-07-28
Remove extra IVT in Microsoft.AspNet.Razor


NTaylorMullen:
Today `Microsoft.AspNet.Razor` has an IVT for the `Microsoft.AspNet.Razor.Test` and `Microsoft.AspNet.Razor.Runtime.Test` projects. This breaks the 1:1 mapping referenced in the guidelines and more specifically limits what common packages can be used in the projects.

NTaylorMullen:
3b5692a5a5f20b7367ca8f9ba83aebe94addaef9

PR #465, opened 2015-07-28
Removed extra IVT from `Microsoft.AspNet.Razor`.

NTaylorMullen:
- Added tests to validate `IsStringProperty` on `TagHelperAttributeDescriptor`s.

#464

### DIFF
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
@@ -114,19 +114,19 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         true,
                         new[]
                         {
-                            new TagHelperDescriptor(
-                            tagName: "inherited-editor-browsable",
-                            typeName: typeof(InheritedEditorBrowsableTagHelper).FullName,
-                            assemblyName: AssemblyName,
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -717,9 +717,7 @@ public void CreateDescriptors_BuildsDescriptorsWithInheritedProperties()
         {
             // Arrange
             var errorSink = new ErrorSink();
-
-            // Also confirm isStringProperty is calculated correctly.
```
dougbu: side note: My earlier concern was around "losing" explicit checks of all `TagHelperDescriptor` properties against our expectations. That's covered now (slightly less explicitly but that's :ok:) between the test descriptor comparers and the new Razor unit tests. Note as well that without the descriptor comparer fixes, we weren't actually checking `IsStringProperty` at all :frowning:

dougbu:
Looks as if an earlier commit broke `AttributeBlockChunkGeneratorTest.GetHashCode_ReturnsSameValue_WhenEqual` on Mono. Should probably get fix for that in first.

dougbu:
:shipit:

NTaylorMullen:
Talked in person about https://github.com/aspnet/Razor/pull/465#issuecomment-125729471. (CI server hasn't built the fix yet which is contained in `aspnet/Common`)

PR #467, opened 2015-07-29
Cleanup `project.json` files

dougbu:
- react to updated Core package versions
 - remove `-beta`
 - led to warnings but only in VS builds
- remove some package references that are found transitively


Eilon:
Hold off on this please. @troydai is sending mail about this. This change is unlikely to be correct.

dougbu:
> This change is unlikely to be correct.

PR #473, opened 2015-08-03
Replacing BufferedEntryCollection with BufferedHtmlContent.

sornaks:
Now any IHtmlContent can be added to a TagHelperContent.

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -2,8 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.IO;
+using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.WebEncoders;
 
@@ -14,14 +14,20 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// </summary>
     public class DefaultTagHelperContent : TagHelperContent
     {
-        private readonly BufferEntryCollection _buffer;
+        private BufferedHtmlContent _buffer;
 
-        /// <summary>
-        /// Instantiates a new instance of <see cref="DefaultTagHelperContent"/>.
-        /// </summary>
-        public DefaultTagHelperContent()
+        // internal for testing
+        internal BufferedHtmlContent Buffer
         {
-            _buffer = new BufferEntryCollection();
+            get
+            {
+                if (_buffer == null)
+                {
+                    _buffer = new BufferedHtmlContent();
+                }
+
+                return _buffer;
+            }
         }
 
         /// <inheritdoc />
@@ -29,7 +35,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             get
             {
-                return _buffer.IsModified;
+                return (_buffer != null);
             }
         }
 
@@ -38,15 +44,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             get
             {
-                foreach (var value in _buffer)
+                if (_buffer == null)
                 {
-                    if (!string.IsNullOrWhiteSpace(value))
-                    {
-                        return false;
-                    }
+                    return true;
                 }
 
-                return true;
+                return CheckIfBufferedHtmlContentIsWhiteSpace(Buffer);
             }
         }
 
@@ -55,67 +58,47 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         {
             get
             {
-                foreach (var value in _buffer)
+                if (_buffer == null)
                 {
-                    if (!string.IsNullOrEmpty(value))
-                    {
-                        return false;
-                    }
+                    return true;
                 }
 
-                return true;
+                return CheckIfBufferedHtmlContentIsEmpty(Buffer);
             }
         }
 
         /// <inheritdoc />
-        public override TagHelperContent SetContent(string value)
-        {
-            Clear();
-            Append(value);
-            return this;
-        }
-
-        /// <inheritdoc />
-        public override TagHelperContent SetContent(TagHelperContent tagHelperContent)
-        {
-            Clear();
-            Append(tagHelperContent);
-            return this;
-        }
-
-
-        /// <inheritdoc />
         public override TagHelperContent Append(string value)
         {
-            _buffer.Add(value);
+            Buffer.Append(value);
             return this;
         }
 
         /// <inheritdoc />
         public override TagHelperContent AppendFormat([NotNull] string format, object arg0)
         {
-            _buffer.Add(string.Format(format, arg0));
+            Buffer.Append(string.Format(format, arg0));
             return this;
         }
 
         /// <inheritdoc />
         public override TagHelperContent AppendFormat([NotNull] string format, object arg0, object arg1)
         {
-            _buffer.Add(string.Format(format, arg0, arg1));
+            Buffer.Append(string.Format(format, arg0, arg1));
             return this;
         }
 
         /// <inheritdoc />
         public override TagHelperContent AppendFormat([NotNull] string format, object arg0, object arg1, object arg2)
         {
-            _buffer.Add(string.Format(format, arg0, arg1, arg2));
+            Buffer.Append(string.Format(format, arg0, arg1, arg2));
             return this;
         }
 
         /// <inheritdoc />
         public override TagHelperContent AppendFormat([NotNull] string format, params object[] args)
         {
-            _buffer.Add(string.Format(format, args));
+            Buffer.Append(string.Format(format, args));
             return this;
         }
 
@@ -125,7 +108,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             [NotNull] string format,
             object arg0)
         {
-            _buffer.Add(string.Format(provider, format, arg0));
+            Buffer.Append(string.Format(provider, format, arg0));
             return this;
         }
 
@@ -136,7 +119,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             object arg0,
             object arg1)
         {
-            _buffer.Add(string.Format(provider, format, arg0, arg1));
+            Buffer.Append(string.Format(provider, format, arg0, arg1));
             return this;
         }
 
@@ -148,7 +131,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             object arg1,
             object arg2)
         {
-            _buffer.Add(string.Format(provider, format, arg0, arg1, arg2));
+            Buffer.Append(string.Format(provider, format, arg0, arg1, arg2));
             return this;
         }
 
@@ -158,66 +141,119 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             [NotNull] string format,
             params object[] args)
         {
-            _buffer.Add(string.Format(provider, format, args));
+            Buffer.Append(string.Format(provider, format, args));
             return this;
         }
 
         /// <inheritdoc />
-        public override TagHelperContent Append(TagHelperContent tagHelperContent)
+        public override TagHelperContent Append(IHtmlContent htmlContent)
         {
-            if (tagHelperContent != null)
-            {
-                foreach (var value in tagHelperContent)
-                {
-                    Append(value);
-                }
-            }
-
-            // If Append() was called with an empty TagHelperContent IsModified should
-            // still be true. If the content was not already modified, it means it is empty.
-            // So the Clear() method can be used to indirectly set the IsModified.
-            if (!IsModified)
-            {
-                Clear();
-            }
-
+            Buffer.Append(htmlContent);
             return this;
         }
 
         /// <inheritdoc />
         public override TagHelperContent Clear()
         {
-            _buffer.Clear();
+            Buffer.Clear();
             return this;
         }
 
         /// <inheritdoc />
         public override string GetContent()
         {
-            return string.Join(string.Empty, _buffer);
+            if (_buffer == null)
+            {
+                return string.Empty;
+            }
+
+            return string.Join(string.Empty, Buffer);
         }
 
         /// <inheritdoc />
         public override string ToString()
         {
-            return GetContent();
-        }
-
-        /// <inheritdoc />
-        public override IEnumerator<string> GetEnumerator()
-        {
-            // The enumerator is exposed so that SetContent(TagHelperContent) and Append(TagHelperContent)
-            // can use this to iterate through the values of the buffer.
-            return _buffer.GetEnumerator();
-        }
-
-        /// <inheritdoc />
-        public override void WriteTo(TextWriter writer, IHtmlEncoder encoder)
-        {
-            foreach (var entry in _buffer)
+            using (var writer = new StringWriter())
             {
-                writer.Write(entry);
+                WriteTo(writer, new HtmlEncoder());
+                return writer.ToString();
             }
+        }
+
+        /// <inheritdoc />
+        public override void WriteTo([NotNull] TextWriter writer, [NotNull] IHtmlEncoder encoder)
+        {
+            Buffer.WriteTo(writer, encoder);
+        }
+
+        private bool CheckIfBufferedHtmlContentIsEmpty(BufferedHtmlContent buffer)
+        {
+            foreach (var value in buffer)
+            {
+                var valueAsString = value as string;
+                if (valueAsString != null)
+                {
+                    if (!string.IsNullOrEmpty(valueAsString))
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    var valueAsBufferedHtmlContent = value as BufferedHtmlContent;
+                    if (valueAsBufferedHtmlContent != null)
+                    {
+                        if (!CheckIfBufferedHtmlContentIsEmpty(valueAsBufferedHtmlContent))
+                        {
+                            return false;
+                        }
+                    }
+                    else
+                    {
+                        if (!string.IsNullOrEmpty(value.ToString()))
+                        {
+                            return false;
+                        }
+                    }
+                }
+            }
+
+            return true;
+        }
+
+        private bool CheckIfBufferedHtmlContentIsWhiteSpace(BufferedHtmlContent buffer)
+        {
+            foreach (var value in buffer)
+            {
+                var valueAsString = value as string;
+                if (valueAsString != null)
+                {
+                    if (!string.IsNullOrWhiteSpace(valueAsString))
+                    {
+                        return false;
+                    }
+                }
+                else
+                {
+                    var valueAsBufferedHtmlContent = value as BufferedHtmlContent;
+                    if (valueAsBufferedHtmlContent != null)
+                    {
+                        if (!CheckIfBufferedHtmlContentIsWhiteSpace(valueAsBufferedHtmlContent))
+                        {
+                            return false;
+                        }
+                    }
+                    else
+                    {
+                        if (!string.IsNullOrWhiteSpace(value.ToString()))
+                        {
+                            return false;
+                        }
+                    }
+                }
+            }
+
+            return true;
         }
     }
 }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Text;
 using Microsoft.AspNet.Html.Abstractions;
 using Microsoft.Framework.Internal;
 using Microsoft.Framework.WebEncoders;
@@ -31,15 +32,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <inheritdoc />
-        public override bool IsModified
-        {
-            get
-            {
-                return (_buffer != null);
-            }
-        }
+        public override bool IsModified => _buffer != null;
 
         /// <inheritdoc />
+        /// <remarks>Returns <c>true</c> for a cleared <see cref="TagHelperContent"/>.</remarks>
         public override bool IsWhiteSpace
         {
             get
@@ -49,7 +45,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                     return true;
                 }
 
-                return CheckIfBufferedHtmlContentIsWhiteSpace(Buffer);
+                using (var writer = new EmptyOrWhitespaceWriter())
+                {
+                    WriteTo(writer, HtmlEncoder.Default);
+                    return writer.IsWhitespace;
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -29,7 +35,7 @@ public override bool IsModified
         {
             get
             {
-                return _buffer.IsModified;
+                return (_buffer != null);
```
NTaylorMullen: no need for parens.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -38,15 +44,12 @@ public override bool IsWhiteSpace
         {
             get
             {
-                foreach (var value in _buffer)
+                if (_buffer == null)
```
NTaylorMullen: `if (!IsModified)` so what you're protecting against is clear.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -38,15 +44,12 @@ public override bool IsWhiteSpace
...
                 {
-                    if (!string.IsNullOrWhiteSpace(value))
-                    {
-                        return false;
-                    }
+                    return true;
                 }
 
-                return true;
+                return CheckIfBufferedHtmlContentIsWhiteSpace(Buffer);
```
dougbu: No need for the indirection through the property here or in a few other places in this class. Suggest sticking with the field (`_buffer`) after checking it for null.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -158,66 +141,119 @@ public override TagHelperContent AppendFormat([NotNull] string format, params ob
...
+            return string.Join(string.Empty, Buffer);
         }
 
         /// <inheritdoc />
         public override string ToString()
         {
-            return GetContent();
+            using (var writer = new StringWriter())
+            {
+                WriteTo(writer, new HtmlEncoder());
```
NTaylorMullen: This is worrys me. It means it can't use the encoder that's in the system.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -38,15 +44,12 @@ public override bool IsWhiteSpace
...
             get
             {
-                foreach (var value in _buffer)
+                if (_buffer == null)
                 {
-                    if (!string.IsNullOrWhiteSpace(value))
-                    {
-                        return false;
-                    }
+                    return true;
```
dougbu: Might be worth `<remarks>` that `IsWhiteSpace` is `true` for a `null` or cleared `DefaultTagHelperContent`. This is an existing issue related to the property name but worth fixing.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: @NTaylorMullen huh? `IsModified` muddies the protections here. This really is a `null` check to avoid NREs in `CheckIfBufferedHtmlContentIsWhiteSpace()`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: Why encode at all? Return value wasn't before.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -158,66 +141,119 @@ public override TagHelperContent AppendFormat([NotNull] string format, params ob
...
         {
-            // The enumerator is exposed so that SetContent(TagHelperContent) and Append(TagHelperContent)
-            // can use this to iterate through the values of the buffer.
-            return _buffer.GetEnumerator();
+            Buffer.WriteTo(writer, encoder);
         }
 
-        /// <inheritdoc />
-        public override void WriteTo(TextWriter writer, IHtmlEncoder encoder)
+        private bool CheckIfBufferedHtmlContentIsEmpty(BufferedHtmlContent buffer)
```
dougbu: minor but you don't need "CheckIf" in these method names e.g. `IsBufferedHtmlContentEmpty`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: If we can remove this method that'd be best. We already have a way to return the raw version and encoded version ofthe string. Don't want to make `ToString` magic.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -13,7 +11,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// <summary>
     /// Abstract class used to buffer content returned by <see cref="ITagHelper"/>s.
     /// </summary>
-    public abstract class TagHelperContent : IEnumerable<string>, IHtmlContent
+    public abstract class TagHelperContent : IHtmlContent
```
dougbu: Should not change the public surface of this class or its subclasses.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -527,23 +526,6 @@ public void ToString_ReturnsExpectedValue()
         }
 
         [Fact]
-        public void GetEnumerator_EnumeratesThroughBuffer()
```
dougbu: Restore test when you restore `IEnumerable<string>` implementation.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -14,14 +14,20 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
     {
-        private readonly BufferEntryCollection _buffer;
+        private BufferedHtmlContent _buffer;
 
-        /// <summary>
-        /// Instantiates a new instance of <see cref="DefaultTagHelperContent"/>.
-        /// </summary>
-        public DefaultTagHelperContent()
+        // internal for testing
+        internal BufferedHtmlContent Buffer
```
dougbu: Won't need this property to be `internal` after restoring `IEnumerable<string>` implementation.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -301,7 +301,6 @@ public void Append_WithTagHelperContent_MultipleAppends()
 
             // Assert
             Assert.Equal(expected, copiedTagHelperContent.GetContent());
-            Assert.Equal(new[] { text1, text2 }, copiedTagHelperContent.ToArray());
```
dougbu: Restore

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: This method can be: 
```C#
private bool CheckIfBufferedHtmlContentIsEmpty(BufferedHtmlContent buffer)
{
	foreach (var value in buffer)
	{
		var valueAsBufferedHtmlContent = value as BufferedHtmlContent;
		if (valueAsBufferedHtmlContent != null)
		{
			if (!CheckIfBufferedHtmlContentIsEmpty(valueAsBufferedHtmlContent))
			{
				return false;
			}
		}
		else
		{
			var stringValue = value as string ?? value.ToString();
			
			if (!string.IsNullOrEmpty(stringValue))
			{
				return false;
			}
		}
	}
	
	return true;
}
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -158,66 +141,119 @@ public override TagHelperContent AppendFormat([NotNull] string format, params ob
...
+                            return false;
+                        }
+                    }
+                }
             }
+
+            return true;
+        }
+
+        private bool CheckIfBufferedHtmlContentIsWhiteSpace(BufferedHtmlContent buffer)
```
NTaylorMullen: Only difference between this method and the one above it is the predicate `string.IsNullOrWhiteSpace` vs `string.IsNullOrEmpty`. Should combine into a tri method group where the third method takes in the predicate and recurses into itself.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
pranavkm: or braces 
`public override bool IsModified => _buffer != null;`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
Eilon: It certainly *could* be refactored, but why?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
Eilon: Why not?

dougbu: Because the `IEnumerable<string>` implementation was useful and used.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Don't see the value in having two nearly identical methods for the same behavior. Less opportunity for bugs/single source of truth :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
Eilon: A `TagHelperContent` being a list of strings seems odd to me, given that we have a new type that represents this concept better. No? HTML content is more complex than simply being a list of strings...

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: However we discussed this method a great deal when creating `TagHelperContent` in the first place. Decision was to provide a useful `ToString()` method specifically for debugging but use `GetContent()` in all product code. Previous one-line implementation of `ToString()` remains fine.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
dougbu: @Eilon what new type are you talking about?

The `IEnumerable<string>` implementation remains useful in e.g. tag helpers that need to check _raw_ content i.e. aren't worried about HTML encoding.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
Eilon: Unit tests will verify correctness :smile: I get worried that by adding tons more method calls that's making things even more complicated. This whole change is about perf.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
Eilon: I talked to @dougbu and @rynowak and where I stand now is that this indeed shouldn't implement `IEnumerable<anything>`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -158,66 +141,119 @@ public override TagHelperContent AppendFormat([NotNull] string format, params ob
...
+                    if (valueAsBufferedHtmlContent != null)
+                    {
+                        if (!CheckIfBufferedHtmlContentIsEmpty(valueAsBufferedHtmlContent))
+                        {
+                            return false;
+                        }
+                    }
+                    else
+                    {
+                        if (!string.IsNullOrEmpty(value.ToString()))
```
dougbu: Does this correctly handle MVC's (distinct) `BufferedHtmlContent` class, `HtmlString` or `StringHtmlContent`? Further nothing in the `IHtmlContent` interface guarantees an implementation has overridden `ToString()` in a manner that's useful here. Probably need to add something like `GetRawString()` to that interface.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: Ignore given counter-arguments elsewhere.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
dougbu: The test code for the MVC tag helpers is more convoluted given this change. This may be an indication we've made the `TagHelperContent` API harder to use. But :ok: until and unless some external tag helper author submits an issue.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -35,14 +33,24 @@ public abstract class TagHelperContent : IEnumerable<string>, IHtmlContent
...
-        /// <param name="tagHelperContent">The <see cref="TagHelperContent"/> that replaces the content.</param>
+        /// <param name="htmlContent">The <see cref="IHtmlContent"/> that replaces the content.</param>
         /// <returns>A reference to this instance after the set operation has completed.</returns>
-        public abstract TagHelperContent SetContent(TagHelperContent tagHelperContent);
+        public virtual TagHelperContent SetContent(IHtmlContent htmlContent)
+        {
+            Clear();
+            Append(htmlContent);
+            return this;
+        }
```
dougbu: What is the use case for overriding either `SetContent()` method?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
dougbu: Note this can still be done using the `Buffer` property.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
sornaks: I can't think of one. It was overridable before. Hence left it as is.

sornaks:
Updated..

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: thx

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
dougbu: Why didn't you restore this i.e. `Assert.Equal(..., copiedTagHelperContent.Buffer.Entries)`?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
dougbu: Why hasn't this test been restored?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -562,17 +544,14 @@ public void Fluent_SetContent_AppendFormat_WritesExpectedContent()
...
 
             // Act
             tagHelperContent.SetContent("First ").AppendFormat("{0} Third", "Second");
 
             // Assert
-            foreach (var value in tagHelperContent)
-            {
-                Assert.Equal(expected[i++], value);
-            }
+            Assert.Equal(expected, tagHelperContent.GetContent());
```
dougbu: Less important than earlier (lower-level) tests but old approach works fine still -- `foreach (var value in tagHelperContent.Buffer.Entries)`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
dougbu: This class is `abstract` to clean up the user-visible interface, not for extensibility. Remove `virtual`.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -14,39 +15,41 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
-                    {
-                        return false;
-                    }
+                    return true;
                 }
 
-                return true;
+                using (var writer = new EmptyOrWhitespaceWriter())
+                {
+                    WriteTo(writer, HtmlEncoder.Default);
```
dougbu: What we discussed earlier for this and `IsEmpty` would have avoided running through the whole buffer in common cases:
``` c#
    foreach (var entry in _buffer.Entries)
    {
        if (entry == null)
        {
            continue;
        }

        var stringValue = entry as string;
        if (stringValue != null)
        {
            if (!string.IsNullOrWhitespace(stringValue))
            {
                return false;
            }
        }
        else
        {
            ((IHtmlContent)entry).WriteTo(writer, HtmlEncoder.Default);
            if (!writer.IsWhiteSpace)
            {
                return false;
            }
        }
    }
```

Then `return true;` after the `using`.

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
sornaks: Because it is an implementation detail of `BufferedHtmlContent` which has been verified in its own tests. Since `TagHelperContent` is no longer an `IEnumerable` I have removed it.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
sornaks: Because `TagHelperContent` is not an `IEnumerable`.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
sornaks: Same reasoning as above. Since this is an implementation detail of `BufferedHtmlContent` I am not verifying it here. `BufferedHtmlContent` has its own tests to verify that it is stored as a list of entries.

dougbu: No tests confirm `DefaultTagHelperContent` has anything to do with `BufferedHtmlcontent` or even that it maintains added content separately. In particular, `DefaultTagHelperContent` could concatenate all strings as content is added without breaking tests left in this class.

Put another way the `DefaultTagHelperContent.Buffer` property exists so that internal details can be checked (these are unit tests after all). But no tests use that "internal for testing" property.

Eilon: @dougbu I'm not sure that's necessary. That really is an implementation detail. The rest is left to perf tests.

dougbu: Bit surprised but :ok: @Eilon.

@sornaks please make the `Buffer` property `private` and remove the "internal for testing" comment.

Eilon: The BufferEntries property is used in BufferEntryCollection's tests - it can't be made private.

sornaks: I think what @dougbu meant was to make `Buffer` private in `DefaultTagHelperContent`.

dougbu: @Eilon that's in the Common repo. Here `DefaultTagHelperContent` has a `Buffer` property that's "internal for testing" but not used in tests.

Eilon: Ah, got it.

sornaks:
Updated..

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -158,65 +192,85 @@ public override TagHelperContent AppendFormat([NotNull] string format, params ob
...
         /// <inheritdoc />
         public override TagHelperContent Clear()
         {
-            _buffer.Clear();
+            Buffer.Clear();
             return this;
         }
 
         /// <inheritdoc />
         public override string GetContent()
```
dougbu: This method is likely going to be used less than it was. But it'll be very useful when user code needs to check existing content e.g. where the `<option/>` tag helper compares attribute values against known values.

To be truly useful however, should allow caller to pass an `IHtmlEncoder` instance e.g. something analogous to the `NullTestEncoder`. This would ease comparisons that HTML encoding currently messes up.

But I think this is a follow-up issue. If you agree, please file the issue.

dougbu:
:shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: :+1:

sornaks:
Checked in. Thanks! https://github.com/aspnet/Razor/commit/168fcbb9c13747da58dbda8a501888c2393d04f8

PR #476, opened 2015-08-06
Microsoft.AspNet.Razor.Test -> Microsoft.AspNet.Razor.Tests & Microsoft.AspNet.Razor.Runtime.Test -> Microsoft.AspNet.Razor.Runtime.Tests

hishamco:
According to the guidelines https://github.com/aspnet/Home/wiki/Engineering-guidelines#solution-and-project-folder-structure-and-naming, we should rename the test project to `Microsoft.AspNet.Razor.Tests` instead of `Microsoft.AspNet.Razor.Test` and `Microsoft.AspNet.Razor.Runtime.Tests` instead of `Microsoft.AspNet.Razor.Runtime.Test`


test/Microsoft.AspNet.Razor.Runtime.Tests/TagHelpers/TagHelperScopeManagerTest.cs
```diff
@@ -7,7 +7,7 @@
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
 using Xunit;
 
-namespace Microsoft.AspNet.Razor.Runtime.Test.TagHelpers
+namespace Microsoft.AspNet.Razor.Runtime.Tests.TagHelpers
```
dougbu: The [guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines#unit-test-class-naming) also say a test of the `TagHelperScopeManager` class should be in the same namespace -- `Microsoft.AspNet.Razor.Runtime.TagHelpers`. Please either leave this alone for another "Engineering Day" activity or change it to the correct namespace.

Same for the rest of the non-infrastructure test namespace changes.

dougbu:
:watch:

Eilon:
See https://github.com/aspnet/EventNotification/pull/12#issuecomment-128427434.

Issue #450, opened 2015-07-14
Add ability for a TagHelper to state that it doesn't expect content


NTaylorMullen:
This would enable `TagHelper` authors to specify how a `TagHelper` can be used when writing HTML. More importantly it would optionally allow void elements to be used with `TagHelper`s. It'd be specified via a `[TargetElement]`'s `CloseAfterOpeningTag` property.

Ex:

```C#
[TargetElement("input", CloseAfterOpeningTag = true)]
public class MyTagHelper : TagHelper
{
...
}
```

The default value of `CloseAfterOpeningTag` will be `false` with the special behavior of IF there are multiple `TagHelper`s working on a single HTML element which don't conflict then the `CloseAfterOpeningTag` value of the `TagHelper` that specifies its value will be used.

An invalid value combination is `CloseAfterOpeningTag = true` and `(user specified value to) CloseAfterOpeningTag = false` when multiple `TagHelper`s are targeting a single tag.

danroth27:
@DamianEdwards

GuardRex:
@NTaylorMullen I brought up a related issue in https://github.com/aspnet/Razor/issues/393 ... the problem there was that a Gulp task runner, like `gulp-minify-html`, to minify markup strips out void element's closing /, such that
```
<input asp-for="name" type="text"/>
```
becomes
```
<input asp-for="name" type="text">
```
which chokes the Razor parser and you get
```
Found a malformed 'input' tag helper.
Tag helpers must have a start and end tag or be self closing.
```
@dougbu explained
> Razor needs explicit (XHTML-ish) tag closure to determine the tag helper element content. Razor would otherwise get nearly as complicated as your browser. To avoid a hard-coded list of void elements (for example), elements that tag helpers target must either be self-closed or have an end tag. But prefixes are a great workaround when using gulp-minify-html or other preprocessors that confuse Razor.

... but that still doesn't sit well with me. I'd still greatly prefer not to have to use tag helpers to get around this. As it stands, we're going have a bazillion views with tag helper prefixes all over the place when going for minfied markup just because the Razor parser can't deal with unclosed void elements.

You are suggesting this feature for tag helper authors and that would seem to extend to solve my problem and allow the Razor parser to work with unclosed void elements. Is that a correct reading of your feature suggestion?

NTaylorMullen:
[...]

PR #480, opened 2015-08-10
Add ability to write start tag only `TagHelper`  elements.

NTaylorMullen:
- To write a start tag only `TagHelper` you can now utilize the `TagStructure` property on the `TargetElement` attribute. If none is specified it'll be treated as unspecified and default to old behavior of being start/end tag or self-closing.
- Added `TagMode` to showcase what the user initially wrote in their Razor document. This way `TagHelper`s can flow end-to-end in thesame format as they were written with.
- Updated code generation to specify `TagMode` instead of the boolean self-closing.
- Updated existing tests to move from `SelfClosing` => `TagMode`.
- Added `TagStructure` related tests to the set of tests that we currently have for `TagHelperBlockRewriter` and `TagHelperParseTreeRewriter`.

#450

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -22,9 +22,9 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// Internal for testing purposes only.
         /// </summary>
-        internal TagHelperExecutionContext(string tagName, bool selfClosing)
+        internal TagHelperExecutionContext(string tagName, TagMode tagMode)
             : this(tagName,
-                   selfClosing,
+                   tagMode,
                    items: new Dictionary<object, object>(),
                    uniqueId: string.Empty,
                    executeChildContentAsync: async () => await Task.FromResult(result: true),
@@ -37,8 +37,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Instantiates a new <see cref="TagHelperExecutionContext"/>.
         /// </summary>
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
-        /// <param name="selfClosing">
-        /// <see cref="bool"/> indicating whether or not the tag in the Razor source was self-closing.</param>
+        /// <param name="tagMode">HTML syntax of the element in the Razor source.</param>
         /// <param name="items">The collection of items used to communicate with other
         /// <see cref="ITagHelper"/>s</param>
         /// <param name="uniqueId">An identifier unique to the HTML element this context is for.</param>
@@ -47,7 +46,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <param name="endTagHelperWritingScope">A delegate used to end a writing scope in a Razor page.</param>
         public TagHelperExecutionContext(
             [NotNull] string tagName,
-            bool selfClosing,
+            TagMode tagMode,
             [NotNull] IDictionary<object, object> items,
             [NotNull] string uniqueId,
             [NotNull] Func<Task> executeChildContentAsync,
@@ -59,7 +58,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             _startTagHelperWritingScope = startTagHelperWritingScope;
             _endTagHelperWritingScope = endTagHelperWritingScope;
 
-            SelfClosing = selfClosing;
+            TagMode = tagMode;
             HTMLAttributes = new TagHelperAttributeList();
             AllAttributes = new TagHelperAttributeList();
             TagName = tagName;
@@ -68,9 +67,9 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// Gets a value indicating whether or not the tag in the Razor source was self-closing.
+        /// Gets the HTML syntax of the element in the Razor source.
         /// </summary>
-        public bool SelfClosing { get; }
+        public TagMode TagMode { get; }
 
         /// <summary>
         /// Indicates if <see cref="GetChildContentAsync"/> has been called.
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Collections.Generic;
 using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
@@ -82,9 +80,9 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// Indicates whether or not the tag is self-closing.
+        /// Syntax of the element in the generated HTML.
         /// </summary>
-        public bool SelfClosing { get; set; }
+        public TagMode TagMode { get; set; }
 
         /// <summary>
         /// The HTML element's attributes.
@@ -100,7 +98,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Changes <see cref="TagHelperOutput"/> to generate nothing.
         /// </summary>
         /// <remarks>
-        /// Sets <see cref="TagName"/> to <c>null</c>, and clears <see cref="PreElement"/>, <see cref="PreContent"/>, 
+        /// Sets <see cref="TagName"/> to <c>null</c>, and clears <see cref="PreElement"/>, <see cref="PreContent"/>,
         /// <see cref="Content"/>, <see cref="PostContent"/>, and <see cref="PostElement"/> to suppress output.
         /// </remarks>
         public void SuppressOutput()
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
@@ -27,9 +27,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// Starts a <see cref="TagHelperExecutionContext"/> scope.
         /// </summary>
         /// <param name="tagName">The HTML tag name that the scope is associated with.</param>
-        /// <param name="selfClosing">
-        /// <see cref="bool"/> indicating whether or not the tag of this scope is self-closing.
-        /// </param>
+        /// <param name="tagMode">HTML syntax of the element in the Razor source.</param>
         /// <param name="uniqueId">An identifier unique to the HTML element this scope is for.</param>
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
         /// <param name="startTagHelperWritingScope">A delegate used to start a writing scope in a Razor page.</param>
@@ -37,7 +35,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <returns>A <see cref="TagHelperExecutionContext"/> to use.</returns>
         public TagHelperExecutionContext Begin(
             [NotNull] string tagName,
-            bool selfClosing,
+            TagMode tagMode,
             [NotNull] string uniqueId,
             [NotNull] Func<Task> executeChildContentAsync,
             [NotNull] Action startTagHelperWritingScope,
@@ -59,7 +57,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
             var executionContext = new TagHelperExecutionContext(
                 tagName,
-                selfClosing,
+                tagMode,
                 items,
                 uniqueId,
                 executeChildContentAsync,
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
@@ -53,5 +53,31 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <c>*</c> at the end of an attribute name acts as a prefix match.
         /// </remarks>
         public string Attributes { get; set; }
+
+        /// <summary>
+        /// The expected tag structure.
+        /// </summary>
+        /// <remarks>
+        /// If <see cref="TagStructure.Unspecified"/> and no other tag helpers applying to the same element specify
+        /// their <see cref="TagStructure"/> the <see cref="TagStructure.NormalOrSelfClosing"/> behavior is used:
+        /// <para>
+        /// <code>
+        /// &lt;my-tag-helper&gt;&lt;/my-tag-helper&gt;
+        /// &lt;!-- OR --&gt;
+        /// &lt;my-tag-helper /&gt;
+        /// </code>
+        /// Otherwise, if another tag helper applying to the same element does specify their behavior, that behavior
+        /// is used.
+        /// </para>
+        /// <para>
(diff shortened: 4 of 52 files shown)


src/Microsoft.AspNet.Razor/TagHelpers/TagMode.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.TagHelpers
```
NTaylorMullen: Thoughts on throwing this in the `Microsoft.AspNet.Razor.Runtime.TagHelpers` namespace despite its location? That way when writing your `TagHelper`s/`TargetElement`s you wont need to bring in any additional usings. Will also make codegen simpler.

src/Microsoft.AspNet.Razor/TagHelpers/TagStructure.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.TagHelpers
```
NTaylorMullen: Thoughts on throwing this in the `Microsoft.AspNet.Razor.Runtime.TagHelpers` namespace despite its location? That way when writing your `TagHelper`s/`[TargetElement]`s you wont need to bring in any additional usings.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -108,11 +108,50 @@ public static class TagHelperBlockRewriter
...
+            Block beginTagBlock,
+            IEnumerable<TagHelperDescriptor> descriptors,
+            ErrorSink errorSink)
         {
+            TagHelperDescriptor baseDescriptor = null;
+            foreach (var descriptor in descriptors)
+            {
+                if (descriptor.TagStructure != TagStructure.Unspecified)
+                {
+                    // Can't have a set of TagHelpers that expect different structures.
```
dougbu: Other errors in this file relate to the syntax of the Razor file. This class should continue to assume the descriptors are consistent.

Suggest this code should use the `FirstOrDefault()` descriptor with `TagStructure != TagStructure.Unspecified`. The error about conflicting descriptors for the current element should be in `TgHelperParseTreeRewriter`, just after determining `GetDescriptors()` returned a non-empty set. Might be a bit more complex but that location makes more sense and we could eventually optimize things e.g. tracking descriptor sets and their associated errors.

src/Microsoft.AspNet.Razor/TagHelpers/TagMode.cs
dougbu: As discussed offline, we've got a few choices and none of them are perfect
1. Leave as-is which leads to additional `using` statements in tag helper classes that need to specify the requested structure or override the final mode.
2. Move to the `Runtime` namespace which creates a strange overlap between this project and the `Runtime` project.
3. Create separate `enum` declarations in the `Runtime` project and namespace. Might need a couple of casts to convert one into the other. But more importantly leads to code duplication, fortunately of code that should almost never change.

NTaylorMullen: Decided to go with 2.

src/Microsoft.AspNet.Razor/TagHelpers/TagStructure.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/480#discussion_r36782894

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -257,7 +259,7 @@ public void AddMinimizedHtmlAttribute_MaintainsHTMLAttributes()
         public void AddMinimizedHtmlAttribute_MaintainsHTMLAttributes_SomeMinimized()
         {
             // Arrange
-            var executionContext = new TagHelperExecutionContext("input", selfClosing: true);
+            var executionContext = new TagHelperExecutionContext("input", tagMode: TagMode.SelfClosing);
```
dougbu: Should use `StartTagOnly` in at least one test of this type per test class. That is, don't always change `selfClosing: true` to `tagMode: TagMode.SelfClosing` or the shorter `TagMode.SelfClosing` in the updated test classes.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -37,8 +38,7 @@ internal TagHelperExecutionContext(string tagName, bool selfClosing)
         /// Instantiates a new <see cref="TagHelperExecutionContext"/>.
         /// </summary>
         /// <param name="tagName">The HTML tag name in the Razor source.</param>
-        /// <param name="selfClosing">
-        /// <see cref="bool"/> indicating whether or not the tag in the Razor source was self-closing.</param>
+        /// <param name="tagMode">A mode indicating how the tag was written in the Razor source.</param>
```
dougbu: For this and the associated property here and through classes such as `TagMode` (though that one needs to be ambiguous w.r.t. input and output) and so on, suggest `HTML syntax of the element in the Razor source.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -82,9 +83,9 @@ public bool IsContentModified
         }
 
         /// <summary>
-        /// Indicates whether or not the tag is self-closing.
+        /// The mode of the HTML element.
```
dougbu: Use comments more like those in `TagHelperExecutionContext` e.g. `Syntax of the element in the generated HTML.`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -27,8 +28,9 @@ public TagHelperScopeManager()
         /// Starts a <see cref="TagHelperExecutionContext"/> scope.
         /// </summary>
         /// <param name="tagName">The HTML tag name that the scope is associated with.</param>
-        /// <param name="selfClosing">
-        /// <see cref="bool"/> indicating whether or not the tag of this scope is self-closing.
+        /// <param name="tagMode">
+        /// <see cref="TagMode"/> indicating the mode the tag is in: <see cref="TagMode.StartTagAndEndTag"/>,
+        /// <see cref="TagMode.SelfClosing"/> or <see cref="TagMode.StartTagOnly"/>.
```
dougbu: Align these comments with those in `TagHelperExecutionContext`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TargetElementAttribute.cs
```diff
@@ -53,5 +53,29 @@ public TargetElementAttribute(string tag)
...
+        /// element specify their <see cref="TagStructure"/> the <see cref="TagStructure.NormalOrSelfClosing"/>
+        /// behavior is used:
+        /// <code>
+        /// &lt;my-tag-helper&gt;&lt;/my-tag-helper&gt;
+        /// &lt;!-- OR --&gt;
+        /// &lt;my-tag-helper /&gt;
+        /// </code>
+        /// Otherwise, if another tag helper applying to the same element does specify their behavior, that behavior
+        /// is used.
+        ///
```
dougbu: Here and in `TagHelperDescriptor`: Blank lines don't change the final help. Instead use `<para>` elements to wrap the separate bits.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -105,7 +106,9 @@ private void RenderBeginTagHelperScope(string tagName, bool selfClosing, IList<C
             // per call site, e.g. if the tag is on the view twice, there should be two IDs.
             _writer.WriteStringLiteral(tagName)
                    .WriteParameterSeparator()
-                   .WriteBooleanLiteral(selfClosing)
+                   .Write(nameof(TagMode))
```
dougbu: Does MVC generate a `using` for the `Microsoft.AspNet.Razor.Runtime.TagHelpers` namespace? If not use `typeof(TagMode).FullName`.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -413,4 +413,10 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="SourceLocationFilePathDoesNotMatch" xml:space="preserve">
     <value>Cannot perform '{1}' operations on '{0}' instances with different file paths.</value>
   </data>
+  <data name="TagHelperParseTreeRewriter_InconsistentTagStructure" xml:space="preserve">
+    <value>Tag helpers '{0}' and '{1}' targeting element '{2}' must not expect different tag structures. '{0}' expects tag structure '{3}' and '{1}' expects tag structure '{4}'.</value>
+  </data>
+  <data name="TagHelperParseTreeRewriter_EndTagTagHelperMustNotHaveAnEndTag" xml:space="preserve">
+    <value>Found an end '{0}' tag for tag helper '{1}' with tag structure '{2}'. Cannot parse end tags for '{2}' tag structure tag helpers.</value>
```
dougbu: First sentence looks sufficient for both of these messages. Suggest
```
Tag helpers '{0}' and '{1}' targeting element '{2}' must not expect different {3} values.
```
where caller passes in "TagStructure" because `nameof` won't work here -- the attribute isn't accessible.

And
```
Found an end tag (&lt;/{0}&gt;) for tag helper '{1}' with tag structure that disallows an end tag ('{2}').
```

src/Microsoft.AspNet.Razor/Runtime/TagHelpers/TagStructure.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// Tag can be written as &lt;my-tag-helper&gt;&lt;/my-tag-helper&gt; or &lt;my-tag-helper /&gt;.
+        /// </summary>
+        NormalOrSelfClosing,
+
+        /// <summary>
+        /// Tag can be written as &lt;my-tag-helper&gt; or &lt;my-tag-helper /&gt;.
+        /// </summary>
+        /// <remarks>Tags with a <see cref="WithoutEndTag"/> structure will never have any content.</remarks>
```
dougbu: Try to stick to usual meanings of "tag" and "element". This structure, for example, requires a single tag for the entire element. So it's the _element_ that can be written these ways and that will always be empty.

dougbu:
Note this PR introduces a breaking change in `TagHelperDescriptor` which will impact tooling and `RazorTooling`. How will one VS instance handle Beta 6 and Beta 7 projects?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Razor does if it finds `TagHelper`s

NTaylorMullen:
@dougbu how does it introduce a breaking change in `TagHelperDescriptor`?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
NTaylorMullen: Updated the other instance in this test class to go from `SelfClosing` => `StartTagOnly`. For other test classes I added specific tests that utilize `StartTagOnly`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Going to leave `TagHelperOutput` as is since it's not specifically tied to the Razor source's syntax.

dougbu:
> as far as tooling is concerned

It adds a required parameter to the only `public` constructor. Old code won't see `TagStructure` values (meaning tooling can't base dynamic schema on the information provided) and new code likely won't be able to deserialize JSON from old code.

src/Microsoft.AspNet.Razor/RazorResources.resx
NTaylorMullen: Will do `nameof(TagHelperDescriptor.TagStructure)`. It's name will stay consistent with `TargetElement`'s

NTaylorMullen:
Ah, that wont break tooling. It'll see its own version of `TagHelperDescriptor` which will be outdated and not have the `TagStructure` parameter. Therefore it'll just work. 

As for the Razor plugin, it doesn't care about self-closing and delegate all `TagHelperDescriptor` creation to the currently loaded Razor assembly, therefore it shouldn't break.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -413,4 +413,10 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="SourceLocationFilePathDoesNotMatch" xml:space="preserve">
     <value>Cannot perform '{1}' operations on '{0}' instances with different file paths.</value>
   </data>
+  <data name="TagHelperParseTreeRewriter_InconsistentTagStructure" xml:space="preserve">
+    <value>Tag helpers '{0}' and '{1}' targeting element '{2}' must not expect different {3} values.</value>
+  </data>
+  <data name="TagHelperParseTreeRewriter_EndTagTagHelperMustNotHaveAnEndTag" xml:space="preserve">
+    <value>Found an end '{0}' tag for tag helper '{1}' with tag structure '{2}'. Cannot parse end tags for '{2}' tag structure tag helpers.</value>
```
dougbu: First sentence says everything here too. And we tend to use angle brackets around element names. Suggest
```
Found an end tag (&lt;/{0}&gt;) for tag helper '{1}' with tag structure that disallows an end tag ('{2}').
```

The last parenthetical bit also works after "tag structure" if you prefer.

src/Microsoft.AspNet.Razor/Runtime/TagHelpers/TagMode.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// The mode in which a tag should render.
```
dougbu: Like the `TagStructure` comment changes. My comment applied here as well e.g. `The mode in which an element should render.` or `The tags included when rendering an element.` here and `Include both start and end tags.` for `StartTagAndEndTag`.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: Fine for this class.

Clarifying: Are you saying `TagHelperBlockRewriterTest` has sufficient coverage of `StartTagOnly`? Seems to be a great deal more `SelfClosing` despite the new `Rewrite_AllowsCompatibleTagStructures()` test.

dougbu:
:shipit: with a few more doc changes and maybe a couple more start tag only tests. Of course don't YOLO push :smiling_imp:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperExecutionContextTest.cs
NTaylorMullen: Yes

Issue #428, opened 2015-06-23
Razor Comments Introduce Carriage Returns


RehanSaeed:
Using Razor comments, can cause extra carriage returns to appear in the generated HTML. This has been a problem in MVC 5 also.

```
<p>Paragraph 1</p>
@* Razor Comment *@
<p>Paragraph 2</p>
```

The above code generates the following HTML. You can imagine that with a lot of comments you get a lot of blank lines appearing in your code. If you really wanted a blank line to appear, you could add one yourself before the comment.

```
<p>Paragraph 1</p>

<p>Paragraph 2</p>
```

Ideally it should generate this HTML:

```
<p>Paragraph 1</p>
<p>Paragraph 2</p>
```

NTaylorMullen:
@RehanSaeed The reason why you're seeing the extra carriage returns is due to the newline following the `*@`. To work around the issue you can do:

```
<p>Paragraph 1</p>
@* Razor Comment 
*@<p>Paragraph 2</p>
```

As for if this is right or not: Currently the only Razor comment we have only consumes the content it surrounds. `@*` to denote the begin of the comment and `*@` to denote the end. Therefore the `@*` comment is working as intended.

You're definitely not the first person to bring up such an issue however. Maybe filing an issue to add the ability to provide single line Razor comments (like C#'s `//`) would work well.

RehanSaeed:
Adding single line comments sounds nice but that will still mean multi-line comments does not do what most developers expect. I can think of no reason when a developer would want it to add blank lines. The only way to get it to behave as expected is to use your workaround which is pretty ugly.

NTaylorMullen:
@RehanSaeed The Razor comment clearly distinguishes were it starts and where it ends. HTML is white-space sensitive so adding newline consumption to a type of comment that has a start/end would lead to other issues. Take the following examples:

```
<p>
Hello World @* abc *@
</p>
<p>
Hello 
World @* abc *@
</p>
```

I'd definitely expect to see a newline in each of these examples. 

You could say that if `@* ... *@` is on its own line that it could then and only then consume the newline after it however that's why languages like C# invented things like `//` to deal with comments that consume a line to avoid magicish behavior.

NTaylorMullen:
Created a "Create Razor single line comment alternative" issue: https://github.com/aspnet/Razor/issues/451

danroth27:
[...]

PR #482, opened 2015-08-12
Removed newlines at the end of razor comments

ajaybhargavb:
Issue - #428 

This does not include the fix for the issue mentioned here https://github.com/aspnet/Razor/issues/428#issuecomment-123868426.
Will add and update tests once the design is finalized.

@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
@@ -134,12 +134,35 @@ namespace Microsoft.AspNet.Razor.Parser
                 {
                     if (last != null)
                     {
+                        // Don't render the whitespace between the start of the line and the razor comment.
+                        if (startOfLine && last.Type == HtmlSymbolType.WhiteSpace)
+                        {
+                            AddMarkerSymbolIfNecessary();
+                            // Output the symbols that may have been accepted prior to the whitespace.
+                            Output(SpanKind.Markup);
+
+                            Span.ChunkGenerator = SpanChunkGenerator.Null;
+                        }
+
                         Accept(last);
                         last = null;
                     }
+
                     AddMarkerSymbolIfNecessary();
                     Output(SpanKind.Markup);
+
                     RazorComment();
+
+                    // Handle the whitespace and newline at the end of a razor comment.
+                    if (startOfLine &&
+                        (At(HtmlSymbolType.NewLine) ||
+                        (At(HtmlSymbolType.WhiteSpace) && NextIs(HtmlSymbolType.NewLine))))
+                    {
+                        AcceptWhile(IsSpacingToken(includeNewLines: false));
+                        AcceptAndMoveNext();
+                        Span.ChunkGenerator = SpanChunkGenerator.Null;
+                        Output(SpanKind.Markup);
+                    }
                 }
                 else
                 {
--- a/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Razor.Chunks.Generators;
 using Microsoft.AspNet.Razor.Parser;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Test.Framework;
@@ -128,7 +129,8 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
                         new MarkupBlock(
                             new MarkupTagBlock(
                                 Factory.MarkupTransition("<text").Accepts(AcceptedCharacters.Any)),
-                            Factory.Markup(Environment.NewLine + "    ").Accepts(AcceptedCharacters.None),
+                            Factory.Markup(Environment.NewLine).Accepts(AcceptedCharacters.None),
+                            Factory.Markup("    ").With(SpanChunkGenerator.Null),
                             new CommentBlock(
                                 Factory.MarkupTransition(HtmlSymbolType.RazorCommentTransition)
                                        .Accepts(AcceptedCharacters.None),
@@ -143,7 +145,8 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
                                        .Accepts(AcceptedCharacters.None),
                                 Factory.MarkupTransition(HtmlSymbolType.RazorCommentTransition)
                                        .Accepts(AcceptedCharacters.None)),
-                            Factory.Markup(Environment.NewLine + "}")))),
+                            Factory.Markup(Environment.NewLine).With(SpanChunkGenerator.Null),
+                            Factory.Markup("}")))),
                 new RazorError(RazorResources.ParseError_TextTagCannotContainAttributes, 6 + Environment.NewLine.Length, 1, 4),
                 new RazorError(RazorResources.FormatParseError_MissingEndTag("text"), 6 + Environment.NewLine.Length, 1, 4),
                 new RazorError(RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(RazorResources.BlockName_Code, "}", "{"), 1, 0, 1));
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
                 {
                     if (last != null)
                     {
-                        Accept(last);
+                        if (!startOfLine || last.Type != HtmlSymbolType.WhiteSpace)
```
NTaylorMullen: What's this for?

ajaybhargavb: If not for this condition, the whitespace preceding the razor comment will show up as part of the next line.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
...
                         last = null;
                     }
+
                     AddMarkerSymbolIfNecessary();
                     Output(SpanKind.Markup);
+
                     RazorComment();
+
+                    // Handle the whitespace and newline at the end of a razor comment.
+                    if (startOfLine &&
```
NTaylorMullen: It'd be worth it to move this bit of logic into `RazorComment` so it can be used/consistent in its other locations (`CSharpCodeParser.cs` and `CSharpCodeParser.Statement.cs`).

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
...
+                        (At(HtmlSymbolType.NewLine) ||
+                        (At(HtmlSymbolType.WhiteSpace) && NextIs(HtmlSymbolType.NewLine))))
+                    {
+                        AcceptWhile(IsSpacingToken(includeNewLines: false));
```
NTaylorMullen: This should consume the newline instead of skipping it below (if it's skipped it'll mess up tooling).

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
NTaylorMullen: Ahh. Instead of skipping the `last` we should conditionally accept it, mark its code generator as `null` (if there wasn't something prior to it like `<p>hi</p>   @* abcd *@`) to not mess up tooling.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
ajaybhargavb: Just checked. The razor comments in the other two places are working fine as is. No need to add this logic there.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
NTaylorMullen: gracias

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
                 {
                     if (last != null)
                     {
+                        if (startOfLine && last.Type == HtmlSymbolType.WhiteSpace)
```
NTaylorMullen: Should add a comment explaining this one :smile:. This parser loop can be confusing.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
...
+                        {
```
NTaylorMullen: `AddMarkerSymbolIfNecessary();`

NTaylorMullen:
Design looks good. Ensure that this works (pretty sure it will, but want to double check):

```
<p>
  @* Hello *@    @* World *@
</p>
```

NTaylorMullen:
Just to note, you may want to label this PR as part 1/X where X is how many parts you want to take to apply the same niceness to other block level elements in Razor :smile:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
dougbu: Any chance there's more than one symbol waiting to be output? Concern is using the wrong generator for non-whitespace.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
...
                         Accept(last);
                         last = null;
                     }
+
                     AddMarkerSymbolIfNecessary();
                     Output(SpanKind.Markup);
+
                     RazorComment();
+
+                    // Handle the whitespace and newline at the end of a razor comment.
```
dougbu: Why does this only consume the whitespace if `startOfLine` is `true`? Shouldn't the final newline be removed in (less common I'm sure but possible) cases like `@* hello *@      @* world *@`?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,31 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
                 {
                     if (last != null)
                     {
+                        if (startOfLine && last.Type == HtmlSymbolType.WhiteSpace)
+                        {
+                            Output(SpanKind.Markup);
+                            Span.ChunkGenerator = SpanChunkGenerator.Null;
```
dougbu: Shouldn't these two lines be in the same order as lines 159 and 160?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
dougbu: Talked offline; I got it.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -134,12 +134,33 @@ protected void SkipToAndParseCode(Func<HtmlSymbol, bool> condition)
                 {
                     if (last != null)
                     {
+                        // Don't render the whitespace between the start of the line and the razor comment.
+                        if (startOfLine && last.Type == HtmlSymbolType.WhiteSpace)
+                        {
+                            AddMarkerSymbolIfNecessary();
+                            Output(SpanKind.Markup);
```
dougbu: Talked offline; worth a comment reminding reader that `last` hasn't been accepted so this `Output()` handles what came _before_ the whitespace symbol.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
```diff
@@ -173,5 +176,36 @@ public void UnterminatedRazorCommentInVerbatimBlock()
...
                 new RazorError(RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(RazorResources.BlockName_Code, "}", "{"), 1, 0, 1));
         }
+
+        [Fact]
+        public void RazorCommentInMarkup()
+        {
+            ParseDocumentTest(
+                "<p>" + Environment.NewLine
+                + "@**@" + Environment.NewLine
+                + "</p>",
```
dougbu: Suggest adding a couple of variants with whitespace after the newline and before the newline but after a comment, more than one comment inside the `<p>` element, and more than one comment on a line.

dougbu:
:watch:

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
ajaybhargavb: Added.

dougbu: Like the new test. Does anything cover more than one comment on a line e.g. `"  @* hello *@   @* world *@" + Environment.Newline`?

dougbu: Hmm, this matches @NTaylorMullen's [earlier](https://github.com/aspnet/Razor/pull/482#issuecomment-130463445) example:
``` html
<p>
  @* Hello *@    @* World *@
</p>
```

dougbu:
Very close

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
```diff
@@ -173,5 +176,131 @@ public void UnterminatedRazorCommentInVerbatimBlock()
...
+                               .Accepts(AcceptedCharacters.Any),
+                        Factory.MetaMarkup("*", HtmlSymbolType.RazorCommentStar)
+                               .Accepts(AcceptedCharacters.None),
+                        Factory.MarkupTransition(HtmlSymbolType.RazorCommentTransition)
+                               .Accepts(AcceptedCharacters.None)),
+                    Factory.Markup(Environment.NewLine).With(SpanChunkGenerator.Null),
+                    new MarkupTagBlock(
+                        Factory.Markup("</p>"))
+                    ));
+        }
```
NTaylorMullen:
```
<p>
@* abcd *@

</p>
```
```
<p>

@* abcd *@
</p>
```

```
<p>
@* abcd *@ Extra @* efgh *@
</p>
```

```
<p>
@* 
abcd 
*@
</p>
```

Intermingle some content in the above Razor comments just for variation instead of always doing `@**@`.

NTaylorMullen:
:watch: for extra test (https://github.com/aspnet/Razor/pull/482#discussion_r36941987)

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
ajaybhargavb: The second test I added covers it.
https://github.com/aspnet/Razor/pull/482/files#diff-a98a97bbee42620dfdc3826ce1d0192aR264

dougbu: thx

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpRazorCommentsTest.cs
ajaybhargavb: Added.

NTaylorMullen:
:shipit:

dougbu:
:shipit: if we decide to change things related to my new question, can be done later.

PR #483, opened 2015-08-12
Create an abstraction model for TypeInfo

pranavkm:
cc @davidfowl \ @NTaylorMullen  \ @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IMemberInfo.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IMemberInfo.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Metadata common to types and properties.
+    /// </summary>
+    public interface IMemberInfo
+    {
+        /// <summary>
+        /// Gets the name.
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Retrieves a collection of custom <see cref="Attribute"/>s of type <typeparamref name="TAttribute"/> applied
+        /// to this instance of <see cref="IMemberInfo"/>.
+        /// </summary>
+        /// <typeparam name="TAttribute">The type of <see cref="Attribute"/> to search for.</typeparam>
+        /// <returns>A sequence of custom <see cref="Attribute"/>s of type
+        /// <typeparamref name="TAttribute"/>.</returns>
+        /// <remarks>Result not include inherited <see cref="Attribute"/>s.</remarks>
+        IEnumerable<TAttribute> GetCustomAttributes<TAttribute>()
+            where TAttribute : Attribute;
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
@@ -1,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Contains type metadata.
+    /// </summary>
+    public interface ITypeInfo : IMemberInfo
+    {
+        /// <summary>
+        /// Fully qualified name of the type.
+        /// </summary>
+        string FullName { get; }
+
+        /// <summary>
+        /// Gets <see cref="IPropertyInfo"/>s for all properties of the current type excluding indexers.
+        /// </summary>
+        /// <remarks>
+        /// Indexers in this context refer to the CLR notion of an indexer (<c>this [string name]</c> 
+        /// and does not overlap with the semantics of 
+        /// <see cref="Razor.TagHelpers.TagHelperAttributeDescriptor.IsIndexer"/>.
+        /// </remarks>
+        IEnumerable<IPropertyInfo> Properties { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the type is public.
+        /// </summary>
+        bool IsPublic { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the type is abstract or an interface.
+        /// </summary>
+        bool IsAbstract { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the type is generic.
+        /// </summary>
+        bool IsGenericType { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the type implements the <see cref="ITagHelper"/> interface.
+        /// </summary>
+        bool IsTagHelper { get; }
+
+        /// <summary>
+        /// Gets the full names of the parameter types if the type implements
+        /// <see cref="IDictionary{TKey, TValue}"/>.
+        /// </summary>
+        /// <returns>
+        /// The full type names (<seealso cref="System.Type.FullName"/>) of <c>TKey</c> and <c>TValue</c>
+        /// parameters if the type implements <see cref="IDictionary{TKey, TValue}"/>, otherwise <c>null</c>.
+        /// </returns>
+        /// <remarks>
+        /// For open generic types, full type names for generic type parameters is <c>null</c>. 
+        /// </remarks>
+        string[] GetGenericDictionaryParameterNames();
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
-using System.Reflection;
 using System.Text.RegularExpressions;
 using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.Framework.Internal;
@@ -13,7 +12,7 @@ using Microsoft.Framework.Internal;
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
     /// <summary>
-    /// Factory for <see cref="TagHelperDescriptor"/>s from <see cref="Type"/>s.
+    /// Factory for <see cref="TagHelperDescriptor"/>s from <see cref="ITypeInfo"/>s.
     /// </summary>
     public static class TagHelperDescriptorFactory
     {
@@ -35,31 +34,30 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'', '*' });
 
         /// <summary>
-        /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="type"/>.
+        /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="typeInfo"/>.
         /// </summary>
         /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
-        /// <param name="type">The type to create a <see cref="TagHelperDescriptor"/> from.</param>
+        /// <param name="typeInfo">The <see cref="ITypeInfo"/> to create a <see cref="TagHelperDescriptor"/> from.
+        /// </param>
         /// <param name="designTime">Indicates if the returned <see cref="TagHelperDescriptor"/>s should include
         /// design time specific information.</param>
         /// <param name="errorSink">The <see cref="ErrorSink"/> used to collect <see cref="RazorError"/>s encountered
-        /// when creating <see cref="TagHelperDescriptor"/>s for the given <paramref name="type"/>.</param>
+        /// when creating <see cref="TagHelperDescriptor"/>s for the given <paramref name="typeInfo"/>.</param>
         /// <returns>
-        /// A collection of <see cref="TagHelperDescriptor"/>s that describe the given <paramref name="type"/>.
+        /// A collection of <see cref="TagHelperDescriptor"/>s that describe the given <paramref name="typeInfo"/>.
         /// </returns>
         public static IEnumerable<TagHelperDescriptor> CreateDescriptors(
             string assemblyName,
-            [NotNull] Type type,
+            [NotNull] ITypeInfo typeInfo,
             bool designTime,
             [NotNull] ErrorSink errorSink)
         {
-            var typeInfo = type.GetTypeInfo();
-
             if (ShouldSkipDescriptorCreation(designTime, typeInfo))
             {
                 return Enumerable.Empty<TagHelperDescriptor>();
             }
 
-            var attributeDescriptors = GetAttributeDescriptors(type, designTime, errorSink);
+            var attributeDescriptors = GetAttributeDescriptors(typeInfo, designTime, errorSink);
             var targetElementAttributes = GetValidTargetElementAttributes(typeInfo, errorSink);
             var allowedChildren = GetAllowedChildren(typeInfo, errorSink);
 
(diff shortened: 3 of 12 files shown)


pranavkm:
MVC PR: https://github.com/aspnet/Mvc/pull/2961

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -94,7 +91,12 @@ public static class TagHelperDescriptorFactory
 #if !DNXCORE50
             if (designTime)
```
NTaylorMullen: Just to clarify. With pre-compilation `designTime` will always be `false` right?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -388,9 +394,24 @@ private static bool ShouldSkipDescriptorCreation(bool designTime, MemberInfo mem
             return false;
         }
 
+        private static bool ShouldSkipDescriptorCreation(bool designTime, IPropertyInfo property)
```
NTaylorMullen: Makes me feel like we should have a common interface of the two `ITypeInfo` and `IPropertyInfo` (much like `MemberInfo`). Today `EditorBrowsableAttribute` is the only one that applies to both classes and properties but it'd be nice to have the foundation laid out for other similar properties (since we already have one anyways that'd use it).

NTaylorMullen:
Overall looks good.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
pranavkm: Yup

pranavkm:
Updates. No longer in design.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IMemberInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// Gets the name.
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Retrieves a collection of custom <see cref="Attribute"/>s of type <typeparamref name="TAttribute"/> applied
+        /// to this instance of <see cref="IMemberInfo"/>.
+        /// </summary>
+        /// <typeparam name="TAttribute">The type of <see cref="Attribute"/> to search for.</typeparam>
+        /// <returns>A sequence of custom <see cref="Attribute"/> of type
```
NTaylorMullen: `<see cref="Attribute"/>s`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -35,31 +34,30 @@ public static class TagHelperDescriptorFactory
...
-        /// when creating <see cref="TagHelperDescriptor"/>s for the given <paramref name="type"/>.</param>
+        /// when creating <see cref="TagHelperDescriptor"/>s for the given <paramref name="typeInfo"/>.</param>
         /// <returns>
-        /// A collection of <see cref="TagHelperDescriptor"/>s that describe the given <paramref name="type"/>.
+        /// A collection of <see cref="TagHelperDescriptor"/>s that describe the given <paramref name="typeInfo"/>.
         /// </returns>
         public static IEnumerable<TagHelperDescriptor> CreateDescriptors(
             string assemblyName,
-            [NotNull] Type type,
+            [NotNull] ITypeInfo typeInfo,
```
NTaylorMullen: This part will also require a change in razor tooling.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,366 @@
...
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class RuntimeTypeInfoTest
+    {
+        [Theory]
+        [InlineData(typeof(int))]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(Tuple<,>))]
+        [InlineData(typeof(IDictionary<string, string>))]
```
NTaylorMullen: Add a more complex version: `IDictionary<string, IDictionary<string, CustomType>>`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        [Theory]
+        [InlineData(typeof(ITagHelper), typeof(ITagHelper))]
+        [InlineData(typeof(TagHelper), typeof(ITagHelper))]
+        [InlineData(typeof(DerivesFromITagHelper), typeof(ITagHelper))]
+        [InlineData(typeof(DerivesFromTagHelper), typeof(ITagHelper))]
+        [InlineData(typeof(string), typeof(IEnumerable<char>))]
+        [InlineData(typeof(Dictionary<string, string>), typeof(IDictionary<string, string>))]
+        [InlineData(typeof(Dictionary<string, string>), typeof(IDictionary))]
```
NTaylorMullen: Add a complex variation.

NTaylorMullen:
Looks really good. :watch: for complex test variations.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+    /// </summary>
+    public interface ITypeInfo : IMemberInfo
+    {
+        /// <summary>
+        /// Fully qualified name of the type.
+        /// </summary>
+        string FullName { get; }
+
+        /// <summary>
+        /// Gets <see cref="IPropertyInfo"/>s for all properties of the current type excluding indexers.
```
dougbu: Though "indexer" has a well-defined meaning here, we confuse things a bit elsewhere in the `TagHelper` world. For example `TagHelperAttributeDescriptor.IsIndexer` is about how the property is treated when used as an attribute in Razor files. Perhaps useful to distinguish this case more completely.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Gets the fully qualified names of the parameter types if the type implements
+        /// <see cref="IDictionary{TKey, TValue}"/>.
+        /// </summary>
+        /// <returns>
+        /// The fully qualified type names of <c>TKey</c> and <c>TValue</c> parameters if the type
+        /// implements <see cref="IDictionary{TKey, TValue}"/>, otherwise <c>null</c>.
+        /// </returns>
+        string[] GetGenericDictionaryParameterNames();
```
dougbu: What does this look like when the type is an open generic such as `public class MyDictionary<TValue> : IDictionary<int, TValue> { }`?

I suspect this interface cannot describe any open generic type. If that's the case, make it more clear in the doc comments.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// Gets a value indicating whether the type is public.
+        /// </summary>
+        bool IsPublic { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether the type is abstract.
+        /// </summary>
+        bool IsAbstract { get; }
```
dougbu: Is this property also `true` for interfaces? If not, what (if anything) must be special about the `TypeInfo` instances passed into `ImplementsInterface()`?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// Gets a value indicating whether the type is generic.
+        /// </summary>
+        bool IsGenericType { get; }
+
+        /// <summary>
+        /// Determines whether the current type implements the specified interface.
+        /// </summary>
+        /// <param name="interfaceType">The <see cref="TypeInfo"/> of the interface.</param>
+        /// <returns><c>true</c> if the type implements the interface, otherwise <c>false</c>.</returns>
+        bool ImplementsInterface(TypeInfo interfaceType);
```
dougbu: Seems to break the abstraction. Why doesn't this take an `ITypeInfo` parameter instead?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,366 @@
...
+
+        [Fact]
+        public void Properties_ReturnsPublicPropertiesOfAdaptingType()
+        {
+            // Arrange
+            var typeInfo = typeof(SubType).GetTypeInfo();
+            var runtimeTypeInfo = new RuntimeTypeInfo(typeInfo);
+
+            // Act and Assert
+            Assert.Collection(runtimeTypeInfo.Properties,
```
dougbu: This asserts a particular order. Even if the test works fine on all platforms today, would anything break if the order changed in the future? If not, explicitly order `Properties` to ignore their original order.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,366 @@
...
+        [TargetElement("test2")]
+        private class TypeWithAttributes
+        {
+        }
+
+        private class DerivesFromTagHelper : TagHelper
+        {
+        }
+
+        private class DerivesFromITagHelper : ITagHelper
```
dougbu: For consistency with `ImplementsIDictionary`, suggest `ImplementsITagHelper`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
pranavkm: In precompilation, we might eventually be able to map compilation symbols to `Type`s (rather than looking them up by names). This would allow for such a case.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
pranavkm: It assumes `TypeInfo.GetRuntimeProperties` returns properties in some fixed order. Seems to be the case in all platforms so far.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
pranavkm: It returns `new[] { "System.String", null }`. Added test cases to cover this.

pranavkm:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/IMemberInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+        string Name { get; }
+
+        /// <summary>
+        /// Retrieves a collection of custom <see cref="Attribute"/>s of type <typeparamref name="TAttribute"/> applied
+        /// to this instance of <see cref="IMemberInfo"/>.
+        /// </summary>
+        /// <typeparam name="TAttribute">The type of <see cref="Attribute"/> to search for.</typeparam>
+        /// <returns>A sequence of custom <see cref="Attribute"/>s of type
+        /// <typeparamref name="TAttribute"/>.</returns>
+        IEnumerable<TAttribute> GetCustomAttributes<TAttribute>()
```
NTaylorMullen: Should state in one of the docs that it doesn't retrieve inherited attributes (the non `ITypeInfo` default behavior of `GetCustomAttributes<T>`)

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
dougbu: If it's eventually going to be possible to map compilation symbols to `Type`s, what do we need this abstraction for? In any case the future mapping you mention could also be to `ITypeInfo` instances.

davidfowl:
Aren't TypeInfo and friends implementable? Or is it just too hard because there are too many things?

/cc @nguerrera

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
dougbu: Also suggest `Dictionary<string, int>` and something similar to VDD -- `class PartiallyClosedGenericDictionary<TValue> : Dictionary<string, TValue>`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
pranavkm: See - https://github.com/aspnet/Mvc/pull/2961#discussion_r37934560. Regardless, we'll need a `Type` \ `TypeInfo` which Roslyn can understand.

pranavkm:
@davidfowl, `TypeInfo` has too many things we don't care about. Additionally `TypeInfo.Properties[0].PropertyType` is an instance of `Type` which just makes the API quite asymmetric and consequently more work to implement.

nguerrera:
@davidfowl @pranavkm At this time, TypeInfo et al cannot be implemented by user code on System.Runtime-based platforms. This will change, but right now that's where we're at now. Plus, I agree that the surface is quite large if you only need to model a small subset.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
NTaylorMullen: `[NotNull]`

pranavkm:
@dougbu updated

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Fully qualified name of the type.
+        /// </summary>
+        string FullName { get; }
+
+        /// <summary>
+        /// Gets <see cref="IPropertyInfo"/>s for all properties of the current type excluding indexers.
+        /// </summary>
+        /// <remarks>
+        /// Indexers in this context refer to the CLR notion of an indexer (<c>this [string name]</c>
```
dougbu: Add close paren.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
dougbu: thx

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
dougbu: thx

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
dougbu: thx

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,403 @@
...
+        {
+        }
+
+        private class CustomType
+        {
+        }
+    }
+}
+
+namespace Microsoft.AspNet.Razor.Fake
```
dougbu: Move to separate files. Everything here is at the top level and visible throughout the project. So there's no upside to burying the definitions here.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -0,0 +1,403 @@
...
+                throw new NotImplementedException();
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                throw new NotImplementedException();
+            }
+        }
+
+        public class KnownKeyDictionary<TValue> : Dictionary<string, TValue>
```
dougbu: FYI could keep these private if you used `TheoryData` instead. Might not be worth changing anything...

dougbu:
:shipit:

Issue #255, opened 2014-12-11
Allow Tag Helpers to restrict child elements


DamianEdwards:
It can be desirable to design a number of Tag Helpers that work in conjunction with each other, e.g. aspnet/MVC#1688. To better enable this and provide a more guided IntelliSense experience, Tag Helpers should be able to declare that they expect only certain types of elements (usually other Tag Helpers) as their children, by way of  the `RestrictChildrenAttribute`.

The parser should throw an exception if it encounters children of an element with a Tag Helper attached where those children don't match those specified by the Tag Helper. This would show up as an error in Visual Studio with suitably located squiggly.

### Declaration
``` C#
[System.AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
public class RestrictChildrenAttribute : Attribute {
    public RestrictChildrenAttribute(params string[] tagNames) {
        ChildTagNames = tagNames;
    }

    public string[] ChildTagNames { get; private set; }
}
```

### Usage
``` C#
[RestrictChildren("source")]
public class TraceTagHelper : TagHelper {
    ...
}

public class SourceTagHelper : TagHelper {
    ...
}
```
``` HTML
<trace output="TraceOutput.Console">
    <source name="Microsoft.AspNet.Mvc.TagHelpers" level="LogLevel.Warning" />
</trace>
```

NTaylorMullen:
Been thinking this issue through a bit and came up with two semi-edge cases:

---
**Issue:** Multiple `TagHelper`s targeting same element with different `RestrictedChildren`.  
**Solution**: Concatenate all `TagHelper`s restricted children. This also makes https://github.com/aspnet/Razor/issues/474 possible through the use of multiple `TagHelper`s. Still might be worth adding the feature however.

---
**Issue:** Non tag related content inside of `TagHelper`. Aka:
```HTML
<trace output="TraceOutput.Console">
    Hello World
    @if(...)
    {
        <source name="Microsoft.AspNet.Mvc.TagHelpers" level="LogLevel.Warning" />
    }
    else
    {
        <super-source name="SomethingElse" level="LogLevel.Error" />
    }
</trace>
```  
**Solution**: Allow non tag related content. Possibly create another issue to add a property to `RestrictChildrenAttribute` to disallow it.

DamianEdwards:
The first is fine. I think for the second we should just simplify for now to strictly allow only what was specified. We can always open up ability to allow non-element content later based on usage and feedback.

[...]

PR #486, opened 2015-08-13
Add ability for `TagHelper`s to specify restricted children.

NTaylorMullen:
- Specifying the `RestrictChildrenAttribute` enables `TagHelper`s to only allow other `TagHelper`s targeting specified names to be in the children.
- Used the `null` value to indicate that `AllowedChildren` was not specified and therefore everything is allowed. This is the default.
- Added name verification to name values to ensure that no bad values pass through the system.
- Added parsing tests to validate a mixture of content generates errors when expected.

#255

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
+++ b/src/Microsoft.AspNet.Razor.Runtime/Resources.resx
@@ -189,4 +189,10 @@
   <data name="TagHelperAttributeList_CannotAddAttribute" xml:space="preserve">
     <value>Cannot add a {0} with inconsistent names. The {1} property '{2}' must match the location '{3}'.</value>
   </data>
+  <data name="TagHelperDescriptorFactory_InvalidRestrictChildrenAttributeName" xml:space="preserve">
+    <value>Invalid '{0}' tag name '{1}' for tag helper '{2}'. Tag helpers cannot restrict child elements that contain a '{3}' character.</value>
+  </data>
+  <data name="TagHelperDescriptorFactory_InvalidRestrictChildrenAttributeNameNullWhitespace" xml:space="preserve">
+    <value>Invalid '{0}' tag name for tag helper '{1}'. Name cannot be null or whitespace.</value>
+  </data>
 </root>
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Restricts children of the <see cref="ITagHelper"/>'s element.
+    /// </summary>
+    /// <remarks>Combining this attribute with a <see cref="TargetElementAttribute"/> that specifies its
+    /// <see cref="TargetElementAttribute.TagStructure"/> as <see cref="TagStructure.WithoutEndTag"/> will result in
+    /// this attribute being ignored.</remarks>
+    [AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+    public class RestrictChildrenAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="RestrictChildrenAttribute"/> class.
+        /// </summary>
+        /// <param name="tagName">
+        /// The tag name of an element allowed as a child. Tag helpers must target the element.
+        /// </param>
+        /// <param name="tagNames">
+        /// Additional names of elements allowed as children. Tag helpers must target all such elements.
+        /// </param>
+        public RestrictChildrenAttribute(string tagName, params string[] tagNames)
+        {
+            var concatenatedNames = new string[1 + tagNames.Length];
+            concatenatedNames[0] = tagName;
+
+            tagNames.CopyTo(concatenatedNames, 1);
+
+            ChildTagNames = concatenatedNames;
+        }
+
+        /// <summary>
+        /// Get the names of elements allowed as children. Tag helpers must target all such elements.
+        /// </summary>
+        public IEnumerable<string> ChildTagNames { get; }
+    }
+}
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -61,6 +61,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
             var attributeDescriptors = GetAttributeDescriptors(type, designTime, errorSink);
             var targetElementAttributes = GetValidTargetElementAttributes(typeInfo, errorSink);
+            var allowedChildren = GetAllowedChildren(typeInfo, errorSink);
 
             var tagHelperDescriptors =
                 BuildTagHelperDescriptors(
@@ -68,6 +69,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                     assemblyName,
                     attributeDescriptors,
                     targetElementAttributes,
+                    allowedChildren,
                     designTime);
 
             return tagHelperDescriptors.Distinct(TagHelperDescriptorComparer.Default);
@@ -87,6 +89,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             IEnumerable<TargetElementAttribute> targetElementAttributes,
+            IEnumerable<string> allowedChildren,
             bool designTime)
         {
             TagHelperDesignTimeDescriptor typeDesignTimeDescriptor = null;
@@ -118,6 +121,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         assemblyName,
                         attributeDescriptors,
                         requiredAttributes: Enumerable.Empty<string>(),
+                        allowedChildren: allowedChildren,
                         tagStructure: default(TagStructure),
                         designTimeDescriptor: typeDesignTimeDescriptor)
                 };
@@ -130,7 +134,62 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         assemblyName,
                         attributeDescriptors,
                         attribute,
+                        allowedChildren,
                         typeDesignTimeDescriptor));
+        }
+
+        private static IEnumerable<string> GetAllowedChildren(TypeInfo typeInfo, ErrorSink errorSink)
+        {
+            var restrictChildrenAttribute = typeInfo.GetCustomAttribute<RestrictChildrenAttribute>(inherit: false);
+            if (restrictChildrenAttribute == null)
+            {
+                return null;
+            }
+
+            var allowedChildren = restrictChildrenAttribute.ChildTagNames;
+            var validAllowedChildren = GetValidAllowedChildren(allowedChildren, typeInfo.FullName, errorSink);
+
+            if (validAllowedChildren.Any())
+            {
+                return validAllowedChildren;
+            }
+            else
+            {
+                // All allowed children were invalid, return null to indicate that any child is acceptable.
+                return null;
+            }
+        }
+
+        // Internal for unit testing
+        internal static IEnumerable<string> GetValidAllowedChildren(
+            IEnumerable<string> allowedChildren,
+            string tagHelperName,
+            ErrorSink errorSink)
+        {
+            var validAllowedChildren = new List<string>();
+
+            foreach (var name in allowedChildren)
+            {
+                var valid = TryValidateName(
+                    name,
+                    whitespaceError: Resources.FormatTagHelperDescriptorFactory_InvalidRestrictChildrenAttributeNameNullWhitespace(
+                        nameof(RestrictChildrenAttribute),
+                        tagHelperName),
+                    characterErrorBuilder: (invalidCharacter) =>
+                        Resources.FormatTagHelperDescriptorFactory_InvalidRestrictChildrenAttributeName(
+                            nameof(RestrictChildrenAttribute),
+                            name,
+                            tagHelperName,
+                            invalidCharacter),
+                    errorSink: errorSink);
+
+                if (valid)
+                {
+                    validAllowedChildren.Add(name);
+                }
+            }
+
+            return validAllowedChildren;
         }
 
         private static TagHelperDescriptor BuildTagHelperDescriptor(
@@ -138,6 +197,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             TargetElementAttribute targetElementAttribute,
+            IEnumerable<string> allowedChildren,
             TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
             var requiredAttributes = GetCommaSeparatedValues(targetElementAttribute.Attributes);
@@ -148,6 +208,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 assemblyName,
                 attributeDescriptors,
                 requiredAttributes,
+                allowedChildren,
                 targetElementAttribute.TagStructure,
                 designTimeDescriptor);
         }
@@ -158,6 +219,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             IEnumerable<string> requiredAttributes,
+            IEnumerable<string> allowedChildren,
             TagStructure tagStructure,
             TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
@@ -168,6 +230,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 assemblyName: assemblyName,
                 attributes: attributeDescriptors,
                 requiredAttributes: requiredAttributes,
+                allowedChildren: allowedChildren,
                 tagStructure: tagStructure,
                 designTimeDescriptor: designTimeDescriptor);
         }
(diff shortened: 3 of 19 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Restricts children of the <see cref="ITagHelper"/>'s element.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
+    public class RestrictChildrenAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="RestrictChildrenAttribute"/> class.
+        /// </summary>
+        /// <param name="tagName">The tag name of the HTML element to allow as children.</param>
+        /// <param name="tagNames">Additional tag names of HTML elements to restrict as children.</param>
```
ajaybhargavb: Confusing as this line says `restrict as children` and the previous one says `allow as children`.

src/Microsoft.AspNet.Razor.Runtime/Resources.resx
```diff
@@ -189,4 +189,10 @@
   <data name="TagHelperAttributeList_CannotAddAttribute" xml:space="preserve">
     <value>Cannot add a {0} with inconsistent names. The {1} property '{2}' must match the location '{3}'.</value>
   </data>
+  <data name="TagHelperDescriptorFactory_InvalidRestrictChildrenAttributeName" xml:space="preserve">
+    <value>Invalid child restriction '{0}' for tag helper '{1}'. Tag helpers cannot restrict child elements that contain a '{2}' character.</value>
+  </data>
+  <data name="TagHelperDescriptorFactory_InvalidRestrictChildrenAttributeNameNullWhitespace" xml:space="preserve">
+    <value>Invalid child restriction tag name for tag helper '{0}'. Name cannot be null or whitespace.</value>
```
dougbu: For both of these, suggest referencing the `RestrictChildrenAttribute` type and its property rather than "child restriction". Generally make these resources more like `TagHelperDescriptorFactory_InvalidAttributePrefixNotNull` and the other invalid XYZ resources in this file.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Restricts children of the <see cref="ITagHelper"/>'s element.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
```
dougbu: Why is `Inherited` `true` when it's `false` for `TargetElementAttribute`? (Get the different `AllowMultiple` choice but not this one.)

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class RestrictChildrenAttribute : Attribute
+    {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="RestrictChildrenAttribute"/> class.
+        /// </summary>
+        /// <param name="tagName">The tag name of the HTML element to allow as children.</param>
+        /// <param name="tagNames">Additional tag names of HTML elements to restrict as children.</param>
+        public RestrictChildrenAttribute(string tagName, params string[] tagNames)
```
dougbu: Suggest dropping the first parameter and just using `params string[] tagNames`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+            var concatenatedNames = new string[1 + tagNames.Length];
+            concatenatedNames[0] = tagName;
+
+            tagNames.CopyTo(concatenatedNames, 1);
+
+            ChildTagNames = concatenatedNames;
+        }
+
+        /// <summary>
+        /// HTML element tag names targeted by tag helpers allowed as children.
```
dougbu: Suggest `Get the names of elements allowed as children. Tag helpers must target all such elements.` and everything but `Get` in the parameter description.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
dougbu: Unless of course you did this to have the compiler enforce the "at least one" restriction...

src/Microsoft.AspNet.Razor.Test.Sources/CaseSensitiveTagHelperDescriptorComparer.cs
```diff
@@ -60,6 +65,14 @@ public override int GetHashCode(TagHelperDescriptor descriptor)
                 hashCodeCombiner.Add(requiredAttribute, StringComparer.Ordinal);
             }
 
+            if (descriptor.AllowedChildren != null)
+            {
+                foreach (var child in descriptor.AllowedChildren)
```
dougbu: Just realized this and other sorted collections should be sorted when adding to the hash code. Otherwise end up with equal descriptors that have different hash codes. Either fix that for this descriptor and file a follow-up bug for other affected `GetHashCode()` implementations or leave as-is in this PR and file the follow-up for the whole deal.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -267,6 +279,65 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
...
+                    errorSink);
+            }
+        }
+
+        private void ValidateAllowedChildren(string errorMessageFormat, SyntaxTreeNode node, ErrorSink errorSink)
+        {
+            var currentTracker = _trackerStack.Count > 0 ? _trackerStack.Peek() : null;
+            var currentlyAllowedChildren = currentTracker?.AllowedChildren;
+
+            if (currentlyAllowedChildren != null)
```
dougbu: This gets called a few times per encountered tag name and numerous times per parent tag. Suggest maintaining a `_currentlyAllowedChildren` field when the `_trackerStack` is changed would be more straightforward.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -419,4 +419,10 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="TagHelperParseTreeRewriter_EndTagTagHelperMustNotHaveAnEndTag" xml:space="preserve">
     <value>Found an end tag (&amp;lt;/{0}&amp;gt;) for tag helper '{1}' with tag structure that disallows an end tag ('{2}').</value>
   </data>
+  <data name="TagHelperParseTreeRewriter_CannotHaveNonTagContent" xml:space="preserve">
+    <value>The parent '&amp;lt;{0}&amp;gt;' tag helper does not allow non-tag content. Only child tag helper(s) targeting tag name(s) '{1}' are allowed.</value>
+  </data>
+  <data name="TagHelperParseTreeRewriter_InvalidNestedTag" xml:space="preserve">
+    <value>The parent '&amp;lt;{0}&amp;gt;' tag helper only allows child tag helper(s) targeting tag name(s) '{1}'.</value>
```
dougbu: This resource should let the user know what tag is invalid.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -419,4 +419,10 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="TagHelperParseTreeRewriter_EndTagTagHelperMustNotHaveAnEndTag" xml:space="preserve">
     <value>Found an end tag (&amp;lt;/{0}&amp;gt;) for tag helper '{1}' with tag structure that disallows an end tag ('{2}').</value>
   </data>
+  <data name="TagHelperParseTreeRewriter_CannotHaveNonTagContent" xml:space="preserve">
+    <value>The parent '&amp;lt;{0}&amp;gt;' tag helper does not allow non-tag content. Only child tag helper(s) targeting tag name(s) '{1}' are allowed.</value>
```
dougbu: This resource should let the user know what content is invalid.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -67,6 +73,10 @@ private void RewriteTags(Block input, RewritingContext context)
                         continue;
                     }
                 }
+                else
+                {
+                    ValidateParentTagHelperAllowsContent(child, context.ErrorSink);
```
dougbu: Suspect this will result in errors about whitespace e.g. the following won't be valid but should be
``` html
  <parent>
    <allowedChild />
  </parent>
```
Is that correct?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
NTaylorMullen: Yup, did it to enforce at least one :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -130,14 +134,85 @@ public static class TagHelperDescriptorFactory
                         assemblyName,
                         attributeDescriptors,
                         attribute,
+                        allowedChildren,
                         typeDesignTimeDescriptor));
         }
 
+        private static IEnumerable<string> GetAllowedChildren(TypeInfo typeInfo, ErrorSink errorSink)
+        {
+            var restrictChildrenAttribute = typeInfo.GetCustomAttribute<RestrictChildrenAttribute>(inherit: true);
```
dougbu: Method should always return `null` if `TagStructure` is `WithoutEndTag` in this descriptor. Also need an error if that tag structure is mixed with the `RestrictChildrenAttribute`. Main question is whether the error should be displayed once per tag helper or once per conflicting `TargetElementAttribute`.

src/Microsoft.AspNet.Razor.Test.Sources/CaseSensitiveTagHelperDescriptorComparer.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/489

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -130,14 +134,85 @@ public static class TagHelperDescriptorFactory
...
+                        {
+                            errorSink.OnError(
+                                SourceLocation.Zero,
+                                Resources.FormatTagHelperDescriptorFactory_InvalidRestrictChildrenAttributeName(
+                                    name,
+                                    tagHelperName,
+                                    character));
+                            valid = false;
+                        }
+                    }
```
dougbu: Suggest refactoring slightly so that this bit is shared with `ValidateName()`. Also the resource should be similar to `FormatTargetElementAttribute_InvalidName`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -148,6 +157,12 @@ public class TagHelperDescriptor
         public IReadOnlyList<string> RequiredAttributes { get; }
 
         /// <summary>
+        /// HTML element tag names targeted by tag helpers allowed as children.
+        /// </summary>
+        /// <remarks><c>null</c> indicates all tag helper children are allowed.</remarks>
+        public IReadOnlyList<string> AllowedChildren { get; }
```
dougbu: No need for more than `IEnumerable<string>`.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Will maintain the tracker instead.

src/Microsoft.AspNet.Razor/RazorResources.resx
NTaylorMullen: Can't do that. The content can be anything. It'd be extremely difficult to tostring its value.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -19,6 +19,367 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
     public class TagHelperParseTreeRewriterTest : TagHelperRewritingTestBase
     {
         [Fact]
+        public void Rewrite_RecoversWhenRequiredAttributeMismatchAndRestrictedChildren()
```
dougbu: Not seeing any multi-level allowed cases in here. Am I missing something like `<select><option/><optgroup><option/></optgroup></select>` where `<select>` allows either `<option>` or `<optgroup>` and `<optgroup>` allows only `<option>`?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: This is consistent with the rest of the properties.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Yup down in the Theory test the p and strong tags both have several levels of `allowedChildren`

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Yup, great catch :+1: .

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Restricts children of the <see cref="ITagHelper"/>'s element.
+    /// </summary>
+    /// <remarks>Combining this attribute with a <see cref="TargetElementAttribute"/> that specifies its
+    /// <see cref="TargetElementAttribute.TagStructure"/> as <see cref="TagStructure.WithoutEndTag"/> will result in
+    /// this attribute being ignored.</remarks>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
```
dougbu: As mentioned in commit (sorry), this `Inherited` value (`true`) is inconsistent with `TargetElementAttribute`.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -267,6 +280,64 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
             return attributeNames;
         }
 
+        private void ValidateParentTagHelperAllowsContent(SyntaxTreeNode child, ErrorSink errorSink)
+        {
+            var allowedChildren = _currentTagHelperTracker?.AllowedChildren;
+            if (allowedChildren != null)
+            {
+                var childSpan = (Span)child;
```
dougbu: Unconditional cast looks dangerous. Either pass in `Span` (i.e. confirm type in caller) or check `IsBlock` and use `FindFirstDescendentSpan((Block)child)` if that's `true`.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -419,4 +419,10 @@ Instead, wrap the contents of the block in "{{}}":
...
+  </data>
+  <data name="TagHelperParseTreeRewriter_InvalidNestedTag" xml:space="preserve">
+    <value>The &amp;lt;'{0}'&amp;gt; tag is not allowed by parent '&amp;lt;{1}&amp;gt;' tag helper. Only child tag helper(s) targeting tag name(s) '{2}' are allowed.</value>
```
dougbu: Single quotes are not useful mixed with angle brackets. Expect `<disallowed> tag` here and `parent <allower> tag helper` here and above.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -148,6 +159,12 @@ public class TagHelperDescriptor
         public IReadOnlyList<string> RequiredAttributes { get; }
 
         /// <summary>
+        /// Get the names of elements allowed as children. Tag helpers must target all such elements.
+        /// </summary>
+        /// <remarks><c>null</c> indicates all tag helper children are allowed.</remarks>
```
dougbu: Perhaps worth mentioning _all_ children are allowed.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -32,6 +32,7 @@ private static TagHelperDescriptor Valid_PlainTagHelperDescriptor
                     assemblyName: AssemblyName,
                     attributes: Enumerable.Empty<TagHelperAttributeDescriptor>(),
                     requiredAttributes: Enumerable.Empty<string>(),
+                    allowedChildren: null,
```
dougbu: Lots of lovely conflicts with #490 expected :smiling_imp:

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
NTaylorMullen: He based his change off of this :smile:. No conflicts yay!

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs
NTaylorMullen: Oh really? Interesting, wasn't aware.

NTaylorMullen:
@dougbu Updated.

dougbu:
:shipit:

PR #487, opened 2015-08-14
Remove ability to specify ignored '@' in the middle of non-string TagHelper bound attributes.

### LINKED ISSUE #436: Remove ability to specify ignored '@' in the middle of non-string TagHelper bound attributes.
Today you're able to write the following (where `foo` is bound to an `int`):

```C#
@{
    var num = 2;
}
<my-tag-helper foo="1 + @num + 3" />
```

Which then results in: `MyTagHelper.Foo = 1 + num + 3;` *Note the missing `@`*.

Ran this by @DamianEdwards and the behavior is currently a bit wonky since you're in C# but not really because the `@` is removed (and not a transition). It gets extra weird when you get into situations where the variable `num` is a C# reserved keyword like `int` and you want the `@`. 

After re-evaluating the feature we should allow an ignored `@` symbol (which shows up like a transition) **only** at the beginning of a non-string `TagHelper` bound attribute. If any `@` symbols occur after the beginning they are flowed through into the generated C#.

pranavkm:
Fixes #436


test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/TransitionsInTagHelperAttributes.cs
```diff
@@ -0,0 +1,148 @@
...
+            __tagHelperExecutionContext = __tagHelperScopeManager.Begin("p", false, "test", async() => {
+                Instrumentation.BeginContext(128, 11, true);
+                WriteLiteral("Body of Tag");
+                Instrumentation.EndContext();
+            }
+            , StartTagHelperWritingScope, EndTagHelperWritingScope);
+            __PTagHelper = CreateTagHelper<PTagHelper>();
+            __tagHelperExecutionContext.Add(__PTagHelper);
+            AddHtmlAttributeValues("class", __tagHelperExecutionContext, 
+            Tuple.Create(Tuple.Create("", 109), Tuple.Create<System.Object, System.Int32>(new Template((__razor_attribute_value_writer) => {
```
pranavkm: This is the behavior on the dev branch but it doesn't look correct? It's the generated output of:
`<p class="@(@class)">Para contents</p>`

where `class` is a string bound attribute.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -343,12 +347,52 @@ private static bool IsSelfClosing(Block beginTagBlock)
...
+                    if (result.IsBoundNonStringAttribute)
+                    {
+                        // For bound non-string attributes, we'll only allow a transition span to appear at the very
+                        // beginning of the attribute expression. All later transitions would appear as code so that
+                        // they are part of the generated output. E.g.
+                        // key="@value" -> MyTagHelper.key = value
+                        // key=" @value" -> MyTagHelper.key =  @value
+                        // key="1 + @case" -> MyTagHelper.key = 1 + @case
+                        // key="@int + @case" -> MyTagHelper.key = int + @case
+                        // key="4 + @(a + b)" -> MyTagHelper.key = 4 + @a + b
```
pranavkm: Also a bit strange. Fixing this is a bit more involved:

|  Attribute | Generated |
|----------------------------|--------------------------|
| `key="@(a + b)"`            | `key = a + b`            |
| `key="@(a + @(c + d) + e)"` | `key = a + @(c + d) + e` |
| `key="@(a + @c + d) + e"`   | `key = a + @c + d + e`   |

pranavkm:
cc @NTaylorMullen \ @dougbu

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -252,10 +253,13 @@ private static bool IsSelfClosing(Block beginTagBlock)
             var result = CreateTryParseResult(name, descriptors);
 
             // If we're not after an equal then we should treat the value as if it were a minimized attribute.
-            var attributeValueBuilder = afterEquals ? builder : null;
-            result.AttributeValueNode =
-                CreateMarkupAttribute(attributeValueBuilder, result.IsBoundNonStringAttribute);
+            Span attributeValue = null;
+            if (afterEquals)
+            {
+                attributeValue = CreateMarkupAttribute(builder, result.IsBoundNonStringAttribute);
```
NTaylorMullen: What'd this fix?

pranavkm:
Hold off on PR until @DamianEdwards has had a look at https://github.com/aspnet/Razor/pull/487#discussion_r37099005

pranavkm:
Updated based on discussions with @Eilon  \ @DamianEdwards

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -252,10 +253,13 @@ private static bool IsSelfClosing(Block beginTagBlock)
...
+            }
 
+            result.AttributeValueNode = attributeValue;
```
dougbu: Why did this block (and `CreateMarkupAttribute()` change at all?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
dougbu: Does the comment correctly describe the current code? Looks as if the last case results in `4 + @(a + b)`. May also be worth mentioning `@(a + b)` -> `(a + b)`.

test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpTagHelperRenderingTest.cs
```diff
@@ -404,7 +404,7 @@ public static TheoryData DesignTimeTagHelperTestData
             get
             {
                 // Test resource name, baseline resource name, expected TagHelperDescriptors, expected LineMappings
-                return new TheoryData<string, string, IEnumerable<TagHelperDescriptor>, List<LineMapping>>
+                return new TheoryData<string, string, IEnumerable<TagHelperDescriptor>, IList<LineMapping>>
```
dougbu: Realize the Razor data sets are inconsistent between lists and arrays. But why introduce an inconsistency within one data set? Just use another `List<LineMapping>`. If we decide to make everything consistent one way or another, we can do that in one pass.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
pranavkm: The method this was calling into was refactored. Consequently this changed

test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpTagHelperRenderingTest.cs
pranavkm: This is far too trivial a change to have a discussion about.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
pranavkm: Oops, the last comment needs to be updated. I'll include the extra case.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/TransitionsInTagHelperAttributes.cs
dougbu: `class` is not a bound attribute from what I can see. But the template treatment does look odd.

dougbu:
:shipit:

NTaylorMullen:
:shipit:

Issue #441, opened 2015-06-30
Change all TagHelper related descriptors to not use constructors.


NTaylorMullen:
Related conversation: https://github.com/aspnet/Razor/pull/430#discussion_r33592744

Today all of our descriptors: `TagHelperDescriptor`, `TagHelperAttributeDescriptor`, `TagHelperDesignTimeDescriptor` and `TagHelperAttributeDesignTimeDescriptor` are created via constructors. At first this was fine, however they're getting quite large in terms of property count. Therefore, for maintainability removing constructors from these types and creating them with auto properties would be a more maintainable approach.

This also involves updating all of the `TagHelper` comparers that use them to account for their new mutability.

/cc @Eilon @dougbu

PR #488, opened 2015-08-14
change all tag helper descriptors to not use constructors - part 1

ajaybhargavb:
#441 
@NTaylorMullen

### DIFF
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
@@ -138,10 +138,10 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                                 attributes: Enumerable.Empty<TagHelperAttributeDescriptor>(),
                                 requiredAttributes: Enumerable.Empty<string>(),
                                 tagStructure: default(TagStructure),
-                                designTimeDescriptor: new TagHelperDesignTimeDescriptor(
-                                    summary: null,
-                                    remarks: null,
-                                    outputElementHint: "strong"))
+                                designTimeDescriptor: new TagHelperDesignTimeDescriptor
+                                {
+                                    OutputElementHint = "strong"
+                                })
                         }
                     },
                     {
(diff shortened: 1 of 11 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -138,10 +138,12 @@ public static TheoryData OutputElementHintData
                                 attributes: Enumerable.Empty<TagHelperAttributeDescriptor>(),
                                 requiredAttributes: Enumerable.Empty<string>(),
                                 tagStructure: default(TagStructure),
-                                designTimeDescriptor: new TagHelperDesignTimeDescriptor(
-                                    summary: null,
-                                    remarks: null,
-                                    outputElementHint: "strong"))
+                                designTimeDescriptor: new TagHelperDesignTimeDescriptor
+                                {
+                                    Summary = null,
```
NTaylorMullen: No need to set these values to `null`. Just don't set them. Same applies for other instances.

ajaybhargavb: I just thought that explicitly setting it is better in a test.

NTaylorMullen: Nah. Just noise :smile:

NTaylorMullen:
:shipit: once https://github.com/aspnet/Razor/pull/488#discussion_r37103379 is completed.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
ajaybhargavb: Okay. Will change

dougbu:
`Equals()` and `GetHashCode()` implementations in these descriptors (if any) plus `TagHelperAttributeDescriptorComparer` and `TagHelperDesignTimeDescriptorComparer` should change now that all properties of these descriptors are settable. Shouldn't really include mutable values in any of that.

ajaybhargavb:
Spoke offline; Decided not to change the comparers.

PR #490, opened 2015-08-14
Change all TagHelper related descriptors to not use constructors

ajaybhargavb:
Issue - #441 
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -223,16 +223,17 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             TagStructure tagStructure,
             TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
-            return new TagHelperDescriptor(
-                prefix: string.Empty,
-                tagName: tagName,
-                typeName: typeName,
-                assemblyName: assemblyName,
-                attributes: attributeDescriptors,
-                requiredAttributes: requiredAttributes,
-                allowedChildren: allowedChildren,
-                tagStructure: tagStructure,
-                designTimeDescriptor: designTimeDescriptor);
+            return new TagHelperDescriptor
+            {
+                TagName = tagName,
+                TypeName = typeName,
+                AssemblyName = assemblyName,
+                Attributes = attributeDescriptors,
+                RequiredAttributes = requiredAttributes,
+                AllowedChildren = allowedChildren,
+                TagStructure = tagStructure,
+                DesignTimeDescriptor = designTimeDescriptor
+            };
         }
 
         /// <summary>
@@ -647,12 +648,14 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             }
 #endif
 
-            return new TagHelperAttributeDescriptor(
-                attributeName,
-                property.Name,
-                typeName,
-                isIndexer,
-                propertyDesignTimeDescriptor);
+            return new TagHelperAttributeDescriptor
+            {
+                Name = attributeName,
+                PropertyName = property.Name,
+                TypeName = typeName,
+                IsIndexer = isIndexer,
+                DesignTimeDescriptor = propertyDesignTimeDescriptor
+            };
         }
 
         private static bool IsAccessibleProperty(PropertyInfo property)
--- a/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
+++ b/src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
@@ -12,70 +12,21 @@ namespace Microsoft.AspNet.Razor.TagHelpers
     /// </summary>
     public class TagHelperAttributeDescriptor
     {
-        // Internal for testing i.e. for easy TagHelperAttributeDescriptor creation when PropertyInfo is available.
-        internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyInfo propertyInfo)
-            : this(
-                  name,
-                  propertyInfo.Name,
-                  propertyInfo.PropertyType.FullName,
-                  isIndexer: false,
-                  isStringProperty: propertyInfo.PropertyType == typeof(string),
-                  designTimeDescriptor: null)
-        {
-        }
+        private string _typeName;
 
         /// <summary>
         /// Instantiates a new instance of the <see cref="TagHelperAttributeDescriptor"/> class.
         /// </summary>
-        /// <param name="name">
-        /// The HTML attribute name or, if <paramref name="isIndexer"/> is <c>true</c>, the prefix for matching
-        /// attribute names.
-        /// </param>
-        /// <param name="propertyName">The name of the CLR property that corresponds to the HTML attribute.</param>
-        /// <param name="typeName">
-        /// The full name of the named (see <paramref name="propertyName"/>) property's <see cref="Type"/> or,
-        /// if <paramref name="isIndexer"/> is <c>true</c>, the full name of the indexer's value <see cref="Type"/>.
-        /// </param>
-        /// <param name="isIndexer">
-        /// If <c>true</c> this <see cref="TagHelperAttributeDescriptor"/> is used for dictionary indexer assignments.
-        /// Otherwise this <see cref="TagHelperAttributeDescriptor"/> is used for property assignment.
-        /// </param>
-        /// <param name="designTimeDescriptor">The <see cref="TagHelperAttributeDesignTimeDescriptor"/> that contains
-        /// design time information about this attribute.</param>
-        /// <remarks>
-        /// HTML attribute names are matched case-insensitively, regardless of <paramref name="isIndexer"/>.
-        /// </remarks>
-        public TagHelperAttributeDescriptor(
-            [NotNull] string name,
-            [NotNull] string propertyName,
-            [NotNull] string typeName,
-            bool isIndexer,
-            TagHelperAttributeDesignTimeDescriptor designTimeDescriptor)
-            : this(
-                  name,
-                  propertyName,
-                  typeName,
-                  isIndexer,
-                  isStringProperty: string.Equals(typeName, typeof(string).FullName, StringComparison.Ordinal),
-                  designTimeDescriptor: designTimeDescriptor)
+        public TagHelperAttributeDescriptor()
         {
         }
 
-        // Internal for testing i.e. for confirming above constructor sets IsStringProperty as expected.
-        internal TagHelperAttributeDescriptor(
-            [NotNull] string name,
-            [NotNull] string propertyName,
-            [NotNull] string typeName,
-            bool isIndexer,
-            bool isStringProperty,
-            TagHelperAttributeDesignTimeDescriptor designTimeDescriptor)
+        // Internal for testing i.e. for easy TagHelperAttributeDescriptor creation when PropertyInfo is available.
+        internal TagHelperAttributeDescriptor(string name, PropertyInfo propertyInfo)
         {
             Name = name;
-            PropertyName = propertyName;
-            TypeName = typeName;
-            IsIndexer = isIndexer;
-            IsStringProperty = isStringProperty;
-            DesignTimeDescriptor = designTimeDescriptor;
+            PropertyName = propertyInfo.Name;
+            TypeName = propertyInfo.PropertyType.FullName;
         }
 
         /// <summary>
@@ -88,7 +39,10 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         /// <see cref="TagHelperAttributeDescriptor"/> is used for property assignment and is only associated with an
         /// HTML attribute that has the exact <see cref="Name"/>.
         /// </value>
-        public bool IsIndexer { get; }
+        /// <remarks>
+        /// HTML attribute names are matched case-insensitively, regardless of <see cref="IsIndexer"/>.
+        /// </remarks>
+        public bool IsIndexer { get; set; }
 
         /// <summary>
         /// Gets an indication whether this property is of type <see cref="string"/> or, if <see cref="IsIndexer"/> is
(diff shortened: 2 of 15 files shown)


test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs
```diff
@@ -3600,12 +3605,13 @@ public static TheoryData MinimizedAttributeData_PartialTags
...
-                                "BoundRequiredString",
-                                typeof(string).FullName,
-                                isIndexer: false,
-                                designTimeDescriptor: null)
+                            new TagHelperAttributeDescriptor
+                            {
+                                Name = "bound-required-string",
+                                PropertyName = "BoundRequiredString",
+                                TypeName = typeof(string).FullName,
+                                IsIndexer = false
```
dougbu: Don't bother to set `bool` properties to `false`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -88,7 +39,10 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
         /// <see cref="TagHelperAttributeDescriptor"/> is used for property assignment and is only associated with an
         /// HTML attribute that has the exact <see cref="Name"/>.
         /// </value>
-        public bool IsIndexer { get; }
+        /// <remarks>
+        /// HTML attribute names are matched case-insensitively, regardless of <see cref="IsIndexer"/>.
+        /// </remarks>
+        public bool IsIndexer { get; set; } = false;
```
dougbu: Don't initialize this to its default.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -99,30 +53,41 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
         /// to allow empty values for HTML attributes matching this <see cref="TagHelperAttributeDescriptor"/>. If
         /// <c>false</c> empty values for such matching attributes lead to errors.
         /// </value>
-        public bool IsStringProperty { get; }
+        public bool IsStringProperty { get; set; }
```
dougbu: Do not allow this to be publicly set. At most `internal set` for testing purposes; `private set` if we don't have any tests like that.

dougbu:
Shouldn't title be something specific like "Change `TagHelperAttributeDescriptor` to not use constructors"?

dougbu:
:watch:

ajaybhargavb:
I am currently changing `TagHelperDescriptor` and will push that to the same PR.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -12,70 +12,21 @@ namespace Microsoft.AspNet.Razor.TagHelpers
...
-        // Internal for testing i.e. for confirming above constructor sets IsStringProperty as expected.
-        internal TagHelperAttributeDescriptor(
-            [NotNull] string name,
-            [NotNull] string propertyName,
-            [NotNull] string typeName,
-            bool isIndexer,
-            bool isStringProperty,
-            TagHelperAttributeDesignTimeDescriptor designTimeDescriptor)
+        // Internal for testing i.e. for easy TagHelperAttributeDescriptor creation when PropertyInfo is available.
+        internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyInfo propertyInfo)
```
pranavkm: Shouldn't need `NotNull` on test-only code.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -12,70 +12,21 @@ namespace Microsoft.AspNet.Razor.TagHelpers
...
-            bool isIndexer,
-            TagHelperAttributeDesignTimeDescriptor designTimeDescriptor)
-            : this(
-                  name,
-                  propertyName,
-                  typeName,
-                  isIndexer,
-                  isStringProperty: string.Equals(typeName, typeof(string).FullName, StringComparison.Ordinal),
-                  designTimeDescriptor: designTimeDescriptor)
+        public TagHelperAttributeDescriptor()
```
NTaylorMullen: Do we need to have the ctor?

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
dougbu: @NTaylorMullen IIRC @ajaybhargavb did the right thing here. Would otherwise only be able to create instances w/ the `internal` constructor.

ajaybhargavb:
Updated.

NTaylorMullen:
Solid. :shipit: once the childrestrictions are signed off / in.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -99,30 +53,41 @@ internal TagHelperAttributeDescriptor([NotNull] string name, [NotNull] PropertyI
...
         /// names.
         /// </summary>
-        public string Name { get; }
+        public string Name { get; [param: NotNull] set; }
 
         /// <summary>
         /// The name of the CLR property that corresponds to the HTML attribute.
         /// </summary>
-        public string PropertyName { get; }
+        public string PropertyName { get; [param: NotNull] set; }
```
dougbu: Note that nothing guarantees `Name`, `PropertyName`, or `TypeName` are ever set. Fortunately these descriptors are originally created in `TagHelperDescriptorFactory` -- after checking source attribute values. But may be worth adding `Debug.Assert()` expressions checking these values in `TagHelperBlockRewriter` or some place similar.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -14,139 +14,43 @@ namespace Microsoft.AspNet.Razor.TagHelpers
...
-                AllowedChildren = new List<string>(allowedChildren);
-            }
-        }
-
-        /// <summary>
         /// Text used as a required prefix when matching HTML start and end tags in the Razor source to available
         /// tag helpers.
         /// </summary>
-        public string Prefix { get; }
+        public string Prefix { get; set; } = string.Empty;
```
dougbu: As mentioned in commit (sorry): Probably need a backing field and to store `string.Empty` even if explicitly set to `null`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -14,139 +14,43 @@ namespace Microsoft.AspNet.Razor.TagHelpers
...
         /// The full name of the tag helper class.
         /// </summary>
-        public string TypeName { get; }
+        public string TypeName { get; [param: NotNull] set; }
 
         /// <summary>
         /// The name of the assembly containing the tag helper class.
         /// </summary>
-        public string AssemblyName { get; }
+        public string AssemblyName { get; [param: NotNull] set; }
```
dougbu: Note that nothing guarantees `TagName`, `TypeName`, or `AssemblyName` are ever set. Fortunately these descriptors are originally created in TagHelperDescriptorFactory -- after checking source attribute values. But may be worth adding Debug.Assert() expressions checking these values in `TagHelperDescriptorResolver`.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
NTaylorMullen: Ehh no one uses these things anyways. Adding those would probably be too protective.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
NTaylorMullen: ditto

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
```diff
@@ -154,13 +58,13 @@ public class TagHelperDescriptor
...
         /// </remarks>
-        public IReadOnlyList<string> RequiredAttributes { get; }
+        public IEnumerable<string> RequiredAttributes { get; [param: NotNull] set; } = Enumerable.Empty<string>();
 
         /// <summary>
         /// HTML element tag names targeted by tag helpers allowed as children.
         /// </summary>
         /// <remarks><c>null</c> indicates all tag helper children are allowed.</remarks>
-        public IReadOnlyList<string> AllowedChildren { get; }
+        public IEnumerable<string> AllowedChildren { get; set; }
```
dougbu: None of these properties match the earlier implementation. Previously all descriptors had their own list instances, created using the `List<T>` copy constructor. Isolation responsibility has shifted here to caller which implies `TagHelperDescriptorResolver` and perhaps `TagHelperDescriptorFactory` need to change more.

(Side note: It's strange to have settable collection properties. But the `IEnumerable<T>` types help a bit. And `AllowedChildren` must have a setter; so this descriptor is consistent.)

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -28,68 +28,53 @@ public static TheoryData TagStructureData
...
+                                TypeName = typeof(MultiWithUnspecifiedTagStructureTagHelper).FullName,
+                                AssemblyName = AssemblyName,
+                                TagStructure = TagStructure.WithoutEndTag
+                            },
+                            new TagHelperDescriptor
+                            {
+                                TagName = "p",
+                                TypeName = typeof(MultiWithUnspecifiedTagStructureTagHelper).FullName,
+                                AssemblyName = AssemblyName,
+                                TagStructure = TagStructure.Unspecified
```
dougbu: No need to init `enum` property to its default value here either. Definitely no need when you're (unnecessarily) explicitly initializing the property in `TagHelperDescriptor` to the same value.

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDescriptorTest.cs
```diff
@@ -289,29 +290,31 @@ public void TagHelperDescriptor_WithAttributes_CanBeDeserialized()
             Assert.Equal(expectedDescriptor.TypeName, descriptor.TypeName, StringComparer.Ordinal);
             Assert.Equal(expectedDescriptor.AssemblyName, descriptor.AssemblyName, StringComparer.Ordinal);
 
-            Assert.Equal(2, descriptor.Attributes.Count);
-            Assert.Equal(expectedDescriptor.Attributes[0].IsIndexer, descriptor.Attributes[0].IsIndexer);
-            Assert.Equal(expectedDescriptor.Attributes[0].IsStringProperty, descriptor.Attributes[0].IsStringProperty);
-            Assert.Equal(expectedDescriptor.Attributes[0].Name, descriptor.Attributes[0].Name, StringComparer.Ordinal);
+            var descriptorAttributes = descriptor.Attributes.ToList();
```
dougbu: Probably time to simplify these ugly blocks of assertions e.g.
``` c#
Assert.Equal(expectedDescriptor.Attributes, descriptor.Attributes, TagHelperAttributeDescriptorComparer.Default);
```

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: Disagree on this one @NTaylorMullen because the `TagHelperDescriptorResolver` sets this to `null` in the normal case. We want this to be `string.Empty` in that case.

dougbu:
:watch:

NTaylorMullen:
Child restriction PR is in. Will probably need to update the 3 tests added via cf1355245f010193c4b77d5fd9a8345be56b6513 once you rebase on dev.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
ajaybhargavb: Not sure where the in `TagHelperDescriptorResolver` it is being set as `null`. The prefix is set only if it is not null or empty. https://github.com/aspnet/Razor/blob/taghelper-descriptor-fix/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs#L147.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
ajaybhargavb: Not sure what you mean here. Previously there was no constructor that accepted `List<T>` directly, it was always `IEnumerable<T>`. We created a list out of that and set it to the read only list. I made this change because we agreed that the `IReadOnlyList<T>` was unnecessary and there was no other change needed anywhere else except in a test here https://github.com/aspnet/Razor/blob/taghelper-descriptor-fix/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperDescriptorTest.cs#L387.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
ajaybhargavb: :+1: will remove.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: Good point @ajaybhargavb. Also @NTaylorMullen and I were both wishy-washy here -- words like "probably" and "may" appear in both our comments.

Just add `[param: NotNull]` above this setter and we should be fine forevermore.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -158,18 +234,21 @@ public static class TagHelperDescriptorFactory
...
+            return new TagHelperDescriptor
+            {
+                TagName = tagName,
+                TypeName = typeName,
+                AssemblyName = assemblyName,
+                Attributes = attributeDescriptors,
+                RequiredAttributes = requiredAttributes,
+                AllowedChildren = allowedChildren,
+                TagStructure = tagStructure,
+                DesignTimeDescriptor = designTimeDescriptor
```
dougbu: Side note: Not for this PR but please include the final comma when using multi-line initializers. Results in less noise in PRs that add / remove properties later.

dougbu:
:shipit: with one more `[param: NotNull] `

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
dougbu: @ajaybhargavb ignore the specific collection types used and focus on the separation between these descriptors and the calling code. Each `TagHelperDescriptor` had unique instances of all exposed collections. We had this isolation in place because it's the normal guarantee for classes exposing collections i.e. they "own" the collection instances.

Now for example each instance of this class that `TagHelperDescriptorResolver` creates from its cache shares the collections with the original. We might get away with it if neither `TagHelperDescriptorResolver` nor `TagHelperDescriptorFactory` reuse the collection instances they're setting these properties to but it's not a great policy.

Double-check the code that creates `TagHelperDescriptor` instances to confirm we're :ok:.

dougbu:
Oops, :watch: for end of hidden discussion [here](https://github.com/aspnet/Razor/pull/490#discussion-diff-37133205).

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptor.cs
ajaybhargavb: Ah, I understand. I checked and It looks like the instances are used only to create the instance of the `TagHelperDescriptor`. So, we're okay.

ajaybhargavb:
dad3e3ade33e9b0f9914af4e4fa3f73f79c77eae

ajaybhargavb:
Thanks @dougbu and @NTaylorMullen

Issue #492, opened 2015-08-17
Add copy constructor for IReadonlyTagHelperAttribute => TagHelperAttribute


NTaylorMullen:
Today in a `TagHelper`s `Process` method you're unable to copy values from `TagHelperContext.AllAttributes` => `TagHelperOutput` effectively. Aka this results in an error:

```C#
public override void Process(TagHelperContext context, TagHelperOutput output)
{
    output.Attributes.Insert(0, context.AllAttributes[0]);
}
```

Because `context.AllAttributes[0]` is an `IReadonlyTagHelperAttribute` and `output.Attributes.Insert` only takes `TagHelperAttribute`s (as expected). To make this work you need to re-create the `TagHelperAttribute` via`

```C#
var existingAttribute = context.AllAttributes[0];
var copiedAttribute = new TagHelperAttribute
{
    Name = existingAttribute.Name,
    Value = existingAttribute.Value,
    Minimized = existingAttribute.Minimized
};
```

It'd be nice to make this slightly easier by providing a semi-copy ctor for `TagHelperAttribute` so  it can take in an `IReadonlyTagHelperAttribute` and construct itself accordingly. Would let us do:

```C#
var existingAttribute = context.AllAttributes[0];
var copiedAttribute = new TagHelperAttribute(existingAttribute);
```

dougbu:
Alternatively, could add an `Insert(int index, [NotNull] IReadOnlyTagHelperAttribute attribute)` or `Insert(int index, [NotNull] string name, object value)` method to `TagHelperAttributeList`. The second is a nice match for the existing `Add([NotNull] string name, object value)` method.

NTaylorMullen:
Going to take the copy ctor approach.

NTaylorMullen:
55760b4fcbb7285aa1b1d8f328378e8e72df6639

PR #495, opened 2015-08-21
Add copy constructor to `TagHelperAttribute`.

NTaylorMullen:
- Added test to validate copy constructor does its job.

#492

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 {
@@ -17,6 +18,17 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// </summary>
         public TagHelperAttribute()
         {
+        }
+
+        /// <summary>
+        /// Instantiates a new instance of <see cref="TagHelperAttribute" /> with values provided by the given
+        /// <paramref name="attribute"/>.
+        /// </summary>
+        /// <param name="attribute">A <see cref="IReadOnlyTagHelperAttribute"/> whose values should be copied.</param>
+        public TagHelperAttribute([NotNull] IReadOnlyTagHelperAttribute attribute)
+            : this (attribute?.Name, attribute?.Value)
+        {
+            Minimized = attribute.Minimized;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -20,6 +20,17 @@ public TagHelperAttribute()
         }
 
         /// <summary>
+        /// Instantiates a new instance of <see cref="TagHelperAttribute" /> with values provided by the given
+        /// <paramref name="attribute"/>.
+        /// </summary>
+        /// <param name="attribute">A <see cref="IReadOnlyTagHelperAttribute"/> whose values should be copied.</param>
+        public TagHelperAttribute(IReadOnlyTagHelperAttribute attribute)
+            : this (attribute.Name, attribute.Value)
```
pranavkm: `(attribute?.Name, attribute?.Value)` just because the `NotNull` check won't happen prior to the `this` call.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeTest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+{
+    public class TagHelperAttributeTest
+    {
+        public static TheoryData CopyConstructorData
+        {
+            get
+            {
+                // readOnlyTagHelperAttribute
```
pranavkm: ?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeTest.cs
NTaylorMullen: Just labeling the attribute name

NeelBhatt:
I guess Constructor is copying "Minimized" property but in assert you are checking for whole "tagHelperAttribute".

So, just curios to ask should not there something like 

var tagHelperAttribute = new TagHelperAttribute(readOnlyTagHelperAttribute);

 Assert.Equal(
                readOnlyTagHelperAttribute.Minimized,
               tagHelperAttribute.Minimized,
                CaseSensitiveTagHelperAttributeComparer.Default);

NTaylorMullen:
@NeelBhatt the comparer verifies all the properties: https://github.com/aspnet/Razor/blob/f7d42ebe803a0e5b87155d20646a6ba5546e0c0e/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs#L20-L29

NTaylorMullen:
Updated.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperAttributeTest.cs
pranavkm: Seems kinda hard to maintain if the test methods get refactored. If you really want to keep it, you could throw in a `<see cref="..."` instead. But for a single argument having any comment seems unnecessary.

pranavkm:
:shipit:

NeelBhatt:
That is something Awesome!

Issue #484, opened 2015-08-13
Attributes don't work in Tag Helper if newline precedes first attribute


Daniel15:
Tag helper:
```csharp
public class HelloTagHelper : TagHelper
{
	public string First { get; set; }
	public string Second { get; set; }

	public override void Process(TagHelperContext context, TagHelperOutput output)
	{
		output.TagName = "span";
		output.SelfClosing = false;
		output.Attributes.Add("data-first", First);
		output.Attributes.Add("data-second", Second);
	}
}
```

cshtml:
```
<hello first="1111" second="2222" />
<hello
	first="1111"
	second="2222"
/>
<hello first="1111"
       second="2222"
/>
```

Result:
```
<span data-first="1111" data-second="2222"></span>
<span data-first="" data-second=""></span>
<span data-first="1111" data-second="2222"></span>
```

The middle `<hello />` tag has a newline before the first attribute, which seems to break it (all attributes come through as empty). The final `<hello />` only has a newline before the second attribute, which seems to work.

Tested on beta 6.

dougbu:
Just did a quick repro with current Beta 7 bits. It looks as if Razor generates nothing at all for the attributes in the newline-before-attributes case. Likely something is causing the attributes to have a null generator or they're lost completely.

PR #499, opened 2015-08-24
Attributes parsed correctly when newline precedes attributes

ajaybhargavb:
Issue - #484 
@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -405,9 +405,9 @@ namespace Microsoft.AspNet.Razor.Parser
 
         private void TagContent()
         {
-            if (!At(HtmlSymbolType.WhiteSpace))
+            if (!At(HtmlSymbolType.WhiteSpace) && !At(HtmlSymbolType.NewLine))
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -405,9 +405,9 @@ private bool IsTagRecoveryStopPoint(HtmlSymbol sym)
 
         private void TagContent()
         {
-            if (!At(HtmlSymbolType.WhiteSpace))
+            if (!At(HtmlSymbolType.WhiteSpace) && !At(HtmlSymbolType.NewLine))
```
dougbu: Nice. Do we have other cases where a newline would confuse the parser? For example does `EndTagBlock()` in this class handle the following?
``` html
   <p
    >
```

dougbu: Suggest checking around for uses of `HtmlSymbolType.WhiteSpace` alone that seem suspect. Then test a bit.

dougbu:
:shipit: but please confirm no other evil is lurking in nearby code.

NTaylorMullen:
:shipit:

PR #503, opened 2015-08-29
Parse the whitespace surrounding equals in attribute correctly

ajaybhargavb:
Issue - #123 

@NTaylorMullen @dougbu 

Will add/update tests once the fix is accepted.


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -462,6 +463,11 @@ private void BeforeAttribute()
                                  sym.Type != HtmlSymbolType.CloseAngle &&
                                  sym.Type != HtmlSymbolType.OpenAngle &&
                                  (sym.Type != HtmlSymbolType.ForwardSlash || !NextIs(HtmlSymbolType.CloseAngle)));
+
+                // capture whitespace after attribute name (if any)
+                whitespaceAfterAttributeName = ReadWhile(
+                    sym => sym.Type == HtmlSymbolType.WhiteSpace ||
+                    sym.Type == HtmlSymbolType.NewLine);
```
NTaylorMullen: I'm pretty sure this is just fine but we should validate that this newline and the previous whitespace symbol types can handle all the valid whitespace/newline characters defined by the spec for attributes:

```
- U+0020 SPACE
- U+0009 CHARACTER TABULATION (tab)
- U+000A LINE FEED (LF)
- U+000C FORM FEED (FF)
- U+000D CARRIAGE RETURN (CR)
```

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -510,8 +520,13 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
             // Accept the whitespace and name
             Accept(whitespace);
             Accept(nameSymbols);
+            Accept(whitespaceAfterAttributeName);
             Assert(HtmlSymbolType.Equals); // We should be at "="
             AcceptAndMoveNext();
+
+            // Accept the whitespace after Equals
+            AcceptWhile(sym => sym.Type == HtmlSymbolType.WhiteSpace || sym.Type == HtmlSymbolType.NewLine);
```
NTaylorMullen: ditto

NTaylorMullen:
:watch: just for verification that the symbol types can handle the spec defined whtiespace.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: Just checked and yes, it handles all characters in list. I'll add them to some tests.

NTaylorMullen:
Design PR looks good then.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -527,10 +552,15 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
                 Span.ChunkGenerator = SpanChunkGenerator.Null; // The block chunk generator will render the prefix
                 Output(SpanKind.Markup);
 
-                // Read the values
-                while (!EndOfFile && !IsEndOfAttributeValue(quote, CurrentSymbol))
+                // Read the attribute value only if the value is quoted
+                // or if there is no whitespace between '=' and the unquoted value.
+                if (quote != HtmlSymbolType.Unknown || !whitespaceAfterEquals.Any())
```
NTaylorMullen: What happens now with `<input name=@value />` ?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -557,6 +587,11 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
                 // Output the attribute name, the equals and optional quote. Ex: foo="
                 Output(SpanKind.Markup);
 
+                if (whitespaceAfterEquals.Any())
```
NTaylorMullen: Doesn't returning early here mess up `data-` attributes with space around the equals?

test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpTagHelperRenderingTest.cs
```diff
@@ -848,32 +848,36 @@ public static TheoryData DesignTimeTagHelperTestData
...
-                                             characterOffsetIndex: 30,
-                                             contentLength: 0),
-                            BuildLineMapping(documentAbsoluteIndex: 88,
-                                             documentLineIndex: 4,
-                                             documentCharacterOffsetIndex: 12,
-                                             generatedAbsoluteIndex: 1789,
-                                             generatedLineIndex: 54,
-                                             generatedCharacterOffsetIndex: 19,
-                                             contentLength: 0)
+                            BuildLineMapping(
```
NTaylorMullen: Did these change at all or did you just format it? Want to make sure I'm not missing something.

NTaylorMullen:
Need a few more tests:
- Add a dynamic value one (`attrName  =  @attrValue`)
- Add a data- variation with and without a dynamic value.
- `TagHelper` test for bound/unbound variation of attribute. (Did the `TagHelper` bits just work?)

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: This condition will return `true`. No change in the behavior.

NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpTagHelperRenderingTest.cs
ajaybhargavb: Just formatting.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: Yup, good catch. :+1: Updating this condition to return only if there is no quote.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -557,6 +587,11 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
                 // Output the attribute name, the equals and optional quote. Ex: foo="
                 Output(SpanKind.Markup);
 
+                if (quote == HtmlSymbolType.Unknown && whitespaceAfterEquals.Any())
```
NTaylorMullen: What text input makes this evaluate to `true`?

ajaybhargavb: `<p data-foo= blah />`

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/TagHelpersWithWeirdlySpacedAttributes.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@addTagHelper "something, nice"
+
+<p
+class 
+   =
+"Hello World" age   =1337
+   >Body of Tag</p>
+
+<input type = 'text' data-content= "hello" />
```
NTaylorMullen: Add a `data-` attribute with a dynamic value.

NTaylorMullen:
:watch:

Need to add tests to validate following chars:

- U+0009 CHARACTER TABULATION (tab)
- U+000C FORM FEED (FF)

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: When would this evaluate to `false`? You're in an `else` block for `quote != HtmlSymbolType.Unknown || !whitespaceAfterEquals.Any()`. Do a bit of Boolean algebra on `!(A != B || !C)` and you get `A == B && C`.

Looks line [line 594](https://github.com/aspnet/Razor/pull/503/files#diff-58782e50bed88f4f2f70af4835346999R595) is unreachable. What is lost if that code is never evaluated?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -206,23 +206,32 @@ public static class TagHelperBlockRewriter
...
-                    // Spaces after/before the equal symbol are not yet supported:
-                    // https://github.com/aspnet/Razor/issues/123
 
                     // TODO: Handle malformed tags, if there's an '=' then there MUST be a value.
                     // https://github.com/aspnet/Razor/issues/104
 
                     SourceLocation symbolStartLocation;
 
+                    // Skip the whitespace preceding the start of the attribute value.
+                    var valueStartIndex = i + 1;
```
dougbu: Add a code comment explaining `+ 1` skips over the `=` sign.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
```diff
@@ -29,6 +29,39 @@ public void SimpleLiteralAttribute()
...
+        public void SimpleLiteralAttributeWithWhitespaceSurroundingEquals()
+        {
+            ParseBlockTest($"<a href \f{Environment.NewLine}= \t{Environment.NewLine}'Foo' />",
+                new MarkupBlock(
+                    new MarkupTagBlock(
+                        Factory.Markup("<a"),
+                        new MarkupBlock(new AttributeBlockChunkGenerator(name: "href", prefix: new LocationTagged<string>($" href \f{Environment.NewLine}= \t{Environment.NewLine}'", 2, 0, 2), suffix: new LocationTagged<string>("'", 20, 2, 4)),
+                            Factory.Markup($" href \f{Environment.NewLine}= \t{Environment.NewLine}'").With(SpanChunkGenerator.Null),
+                            Factory.Markup("Foo").With(new LiteralAttributeChunkGenerator(prefix: new LocationTagged<string>(string.Empty, 17, 2, 1), value: new LocationTagged<string>("Foo", 17, 2, 1))),
+                            Factory.Markup("'").With(SpanChunkGenerator.Null)),
```
dougbu: Adding clumps of the longest lines in this file. Please wrap.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
```diff
@@ -329,12 +362,12 @@ public void ConditionalAttributesDoNotCreateExtraDataForEntirelyLiteralAttribute
         [Fact]
         public void ConditionalAttributesAreDisabledForDataAttributesInBlock()
         {
-            ParseBlockTest("<span data-foo='@foo'></span>",
+            ParseBlockTest("<span data-foo  =  '@foo'></span>",
```
dougbu: Why not do both the old and new cases? Same for `ConditionalAttributesAreDisabledForDataAttributesInDocument()`.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/TagHelpersWithWeirdlySpacedAttributes.cshtml
```diff
@@ -0,0 +1,9 @@
+﻿@addTagHelper "something, nice"
+
+<p
+class 
+   =
+"Hello World" age   =1337
+   data-content= "@true">Body of Tag</p>
+
+<input type 	= 'text' data-content=	 "hello" />
```
dougbu: Suggest adding additional `type` and `age` variations. Not about all the scenarios with bound and unbound attributes. But good to cover more than just one example of each.

Separately use single quotes for the second `data-content` attribute to mix things up a bit more.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -527,10 +552,15 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
...
+                // Read the attribute value only if the value is quoted
+                // or if there is no whitespace between '=' and the unquoted value.
+                if (quote != HtmlSymbolType.Unknown || !whitespaceAfterEquals.Any())
                 {
-                    AttributeValue(quote);
+                    // Read the values
+                    while (!EndOfFile && !IsEndOfAttributeValue(quote, CurrentSymbol))
+                    {
+                        AttributeValue(quote);
+                    }
```
dougbu: "Values" for a single attribute i.e. name / value pair? Do you mean "symbols in the value"?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -510,14 +523,26 @@ private void AttributePrefix(IEnumerable<HtmlSymbol> whitespace, IEnumerable<Htm
             // Accept the whitespace and name
             Accept(whitespace);
             Accept(nameSymbols);
+            // Since this is not a minimized attribute, the whitespace after attribute name belongs to this attribute.
```
dougbu: Separate this comment from preceding lines with a blank line.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: Please unwrap this. Hard to read with expression split up but lambda parameter mixed in with one part.

dougbu:
Need to update tests with source locations to account for single-character `Environment.NewLine` before Travis will succeed.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Got it. @dougbu you may be misreading the diff. This is the `else` of `if (attributeCanBeConditional)`

NTaylorMullen:
:shipit: once travis tests pass by addressing @dougbu's [comment](https://github.com/aspnet/Razor/pull/503#issuecomment-137309161).

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: @NTaylorMullen yup, thought I expanded everything in between but noooo...

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
```diff
@@ -29,6 +29,39 @@ public void SimpleLiteralAttribute()
         }
 
         [Fact]
+        public void SimpleLiteralAttributeWithWhitespaceSurroundingEquals()
+        {
+            ParseBlockTest($"<a href \f{Environment.NewLine}= \t{Environment.NewLine}'Foo' />",
```
dougbu: Use of `Environment.NewLine` instead of `\r\n` means you're not testing `'\r'` on Linux. I.e. I suggest being explicit about the characters and dropping `Environment.NewLine`. That should also fix the problems testing on Linux because the source locations will be identical.

NeelBhatt: I agree with @dougbu because:

\r\n is Windows., \n is Unix and \r is Mac

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: Single attribute value. This will run for every part of the attribute value. `abc="foo bar baz"` will run thrice.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
ajaybhargavb: Seems redundant because we already have other tests covering those normal cases.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/TagHelpersWithWeirdlySpacedAttributes.cshtml
ajaybhargavb: Added.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: Missing a change here?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: Please update the comment so that it is clear. Wasn't sure what you meant.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
dougbu: `data-` attributes are special and the only tests covering them in this class are this method and the next one. Restore the previous cases and _add_ these new cases.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/TagHelpersWithWeirdlySpacedAttributes.cshtml
dougbu: thx

dougbu:
:watch: for remaining unaddressed feedback.

ajaybhargavb:
Updated.

dougbu:
:shipit:

PR #506, opened 2015-09-03
Change HTML in nested C# blocks to properly handle dots.

NTaylorMullen:
- Prior to this change adding a `.` after an implicit expression would result in compile errors due to Razor thinking the `.` was part of the C# (normally not the case).
- Added a code generation and unit test to validate behavior.

#491

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
@@ -631,6 +631,11 @@ namespace Microsoft.AspNet.Razor.Parser
 
         private void ParseWithOtherParser(Action<ParserBase> parseAction)
         {
+            // When transitioning to the HTML parser we no longer want to act as if we're in a nested C# state.
+            // For instance, if <div>@hello.</div> is in a nested C# block we don't want the trailing '.' to be handled
+            // as C#; it should be handled as a period because it's wrapped in markup.
+            var wasNested = IsNested;
+            IsNested = false;
             using (PushSpanConfig())
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -631,6 +631,8 @@ protected override void HandleEmbeddedTransition()
 
         private void ParseWithOtherParser(Action<ParserBase> parseAction)
         {
+            var wasNested = IsNested;
+            IsNested = false;
```
ajaybhargavb: Would be good to have a comment saying why this is needed

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpBlockTest.cs
```diff
@@ -16,6 +16,31 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
     public class CSharpBlockTest : CsHtmlCodeParserTestBase
     {
         [Fact]
+        public void NestedCodeBlockWithMarkupSetsDotAsMarkup()
+        {
+            ParseBlockTest("if (true) { @if(false) { <div>@something.</div> } }",
```
ajaybhargavb: Maybe add another case with `ParseDocumentTest` just to validate if it behaves the same way.

ajaybhargavb:
:watch:

NTaylorMullen:
Updated.

NTaylorMullen:
This good?

ajaybhargavb:
Yes, :shipit:

Issue #485, opened 2015-08-13
Remove unnecessary whitespace that is rendered for non markup content


ajaybhargavb:
Extra new lines from razor comments has been removed in #428. But we still need to handle unnecessary whitespace in other places.

From @danroth27:
We've decided to be smart about removing whitespace for any non-HTML content that occupies the entire line. So in addition to @RehanSaeed's example https://github.com/aspnet/Razor/issues/428#issue-90321772, the following example:

```html
@{
}
<!DOCTYPE html>
```
Should render just the doctype without any leading whitespace (which would be invalid). Also:

    @foreach (var item in items)
    {
        <li>@item</li>
    }
Should not render a line with four spaces before the list item.

cc' @NTaylorMullen

RehanSaeed:
@ajaybhargavb Please also consider this (https://github.com/aspnet/Razor/issues/423) related suggestion for minifying HTML at compile time.

ajaybhargavb:
e2881b0eca0d219539543128ee5c90aa24ba97f3

NickAb:
In ASP.NET Core project in my .cshtml when I use
```
Some text
@if (Model.BranchesSection.TotalCount == Model.BranchesSection.PreviousIssuedBranchesCount)
                    {
                        <text> same as before</text>
                    }.
```
I get following as a result:
```
Some text same as before .
```

There is a space before full stop symbol. How can I get rid of it? Is my only option to inline it like so:
```
Some text @if (Model.BranchesSection.TotalCount == Model.BranchesSection.PreviousIssuedBranchesCount) { <text> same as before</text> }.
```

NTaylorMullen:
@NickAb I was able to reproduce your issue with:

```
Some text
@if (true)
{
    <text> same as before</text>
}.
```
And even
```
@if (true)
{
    @: same as before
}.
```

Would you mind opening a separate issue with what you have?

Also, as a work around this works (still ugly :cry: ):
```
Some text
@if (true)
{
    <text> same as before</text>}.
```

NTaylorMullen:
Actually, nm @NickAb this issue should be reopened. The browser was being kind rendering a space, the HTML is actually a newline which should have been covered by this issue.

ajaybhargavb:
b520c9c15ef61f6e84c83d4a297449c87f6a74fa

PR #507, opened 2015-09-03
Extra newline no longer rendered at the end of a code block

ajaybhargavb:
Issue - #485 

This fixes the issue with the first example in https://github.com/aspnet/Razor/issues/485#issue-100841789.
The second example no longer seems to be an issue. 
I'm trying various scenarios to see if that issue exists and also to see if this change broke something else. (All tests passing other than the ones updated here)

@NTaylorMullen @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
@@ -316,12 +316,16 @@ namespace Microsoft.AspNet.Razor.Parser
                 Span.ChunkGenerator = SpanChunkGenerator.Null;
             }
 
-            if (!At(CSharpSymbolType.WhiteSpace) && !At(CSharpSymbolType.NewLine))
+            if (!IsNested)
             {
-                PutCurrentBack();
+                EnsureCurrent();
+                if (At(CSharpSymbolType.NewLine) ||
+                    (At(CSharpSymbolType.WhiteSpace) && NextIs(CSharpSymbolType.NewLine)))
+                {
+                    Context.NullGenerateWhitespaceAndNewLine = true;
+                }
             }
 
-            CompleteBlock(insertMarkerIfNecessary: false);
             Output(SpanKind.MetaCode);
         }
 
--- a/test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
@@ -291,6 +291,134 @@ namespace Microsoft.AspNet.Razor.Test.Parser.Html
         }
 
         [Fact]
+        public void ParseDocumentDoesNotRenderExtraNewLineAtTheEndOfVerbatimBlock()
+        {
+            ParseDocumentTest("@{\r\n}\r\n<html>",
+                new MarkupBlock(
+                    Factory.EmptyHtml(),
+                    new StatementBlock(
+                        Factory.CodeTransition(),
+                        Factory.MetaCode("{").Accepts(AcceptedCharacters.None),
+                        Factory.Code("\r\n").AsStatement().AutoCompleteWith(null, false),
+                        Factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                    Factory.Markup("\r\n").With(SpanChunkGenerator.Null),
+                    BlockFactory.MarkupTagBlock("<html>")));
+        }
+
+        [Fact]
+        public void ParseDocumentDoesNotRenderExtraWhitespaceAndNewLineAtTheEndOfVerbatimBlock()
+        {
+            ParseDocumentTest("@{\r\n} \t\r\n<html>",
+                new MarkupBlock(
+                    Factory.EmptyHtml(),
+                    new StatementBlock(
+                        Factory.CodeTransition(),
+                        Factory.MetaCode("{").Accepts(AcceptedCharacters.None),
+                        Factory.Code("\r\n").AsStatement().AutoCompleteWith(null, false),
+                        Factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                    Factory.Markup(" \t\r\n").With(SpanChunkGenerator.Null),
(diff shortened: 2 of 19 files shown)


test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -261,6 +261,20 @@ public void ParseDocumentIgnoresTagsInContentsOfScriptTag()
         }
 
         [Fact]
+        public void ParseDocumentDoesNotRenderExtraNewLine()
```
NTaylorMullen: Add `@{  } <html>\r\n` variation to ensure that the end newline isn't ignored.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -261,6 +261,20 @@ public void ParseDocumentIgnoresTagsInContentsOfScriptTag()
...
+        {
+            ParseDocumentTest("@{\r\n}\r\n<html>",
+                new MarkupBlock(
+                    Factory.EmptyHtml(),
+                    new StatementBlock(
+                        Factory.CodeTransition(),
+                        Factory.MetaCode("{").Accepts(AcceptedCharacters.None),
+                        Factory.Code("\r\n").AsStatement().AutoCompleteWith(null, false),
+                        Factory.MetaCode("}\r\n").Accepts(AcceptedCharacters.None)),
```
NTaylorMullen: `\r\n` isn't metacode. It'll be labeled as yellow in the Razor editor if it's left this way. It needs to be split out as a markup span without a `ChunkGenerator` if not nested in a C# block. Otherwise it needs it needs to be a code span.

NTaylorMullen:
:watch:

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
ajaybhargavb: Added.

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -316,12 +316,15 @@ private void VerbatimBlock()
                 Span.ChunkGenerator = SpanChunkGenerator.Null;
             }
 
-            if (!At(CSharpSymbolType.WhiteSpace) && !At(CSharpSymbolType.NewLine))
+            if (IsNested)
             {
-                PutCurrentBack();
+                CompleteBlock(insertMarkerIfNecessary: false);
```
NTaylorMullen: Shouldn't you always complete the block here whether it's nested or not?

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -316,12 +316,15 @@ private void VerbatimBlock()
...
+            }
+            else if (At(CSharpSymbolType.WhiteSpace) || At(CSharpSymbolType.NewLine))
```
NTaylorMullen: Not entirely true. What if we have:

```C#
@if (...)
{
}   <p>
```

In this case we'd want to render the whitespace.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
NTaylorMullen: In the nested situation we're not `null`ing out the `SpanKind.Code`'s `ChunkGenerator`. I think this should be fine but we should validate with debugging (make sure not excess whitespace is highlighted) once a WTE is produced.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
ajaybhargavb: Hmm. The name is misleading here. This `CompleteBlock` does nothing here other than `EnsureCurrent` and `PutCurrentBack`. Removing it affects no tests. I am thinking we should remove it completely.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.cs
ajaybhargavb: Valid point. Will handle.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -291,6 +291,72 @@ public void ParseDocumentIgnoresTagsInContentsOfScriptTag()
...
+                        Factory.Code("\r\n").AsStatement().AutoCompleteWith(null, false),
+                        new StatementBlock(
+                            Factory.CodeTransition(),
+                            Factory.Code("if(true){\r\n}").AsStatement()),
+                        Factory.Code(" \r\n").AsStatement(),
+                        Factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                    Factory.Markup("\r\n").With(SpanChunkGenerator.Null),
+                    BlockFactory.MarkupTagBlock("<html>")));
+        }
+
```
NTaylorMullen: Few more tests
- `@{\r\n}    \t\r\n<html>`
- `@{\r\n@if(true){\r\n}  <input>  }`
- `@{\r\n@if(true){\r\n}  \r\n<input>  \r\n}<html>`

NTaylorMullen:
:watch:

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlDocumentTest.cs
ajaybhargavb: Added.

NTaylorMullen:
:shipit:

PR #510, opened 2015-09-10
Build break: Mock `TypeInfo` as well to get 4 tests working

dougbu:
- as of aspnet/aspnet.xunit@5a12e89 xUnit uses `TypeInfo` more often


dougbu:
FYI only since this fixes a build break.

dougbu:
493e552

PR #514, opened 2015-09-11
React to BufferedHtmlContent changes

rynowak:
This changes all TagHelperContent methods to assume that input has NOT
YET been encoded.

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -125,6 +125,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             return this;
         }
 
+        public override TagHelperContent AppendEncoded(string value)
+        {
+            Buffer.AppendEncoded(value);
+            return this;
+        }
+
         /// <inheritdoc />
         public override TagHelperContent AppendFormat([NotNull] string format, object arg0)
         {
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
@@ -60,6 +60,14 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         public abstract TagHelperContent Append(string value);
 
         /// <summary>
+        /// Appends <paramref name="value"/> to the existing content. <paramref name="value"/> is assumed
+        /// to be an HTML encoded <see cref="string"/> and no further encoding will be performed.
+        /// </summary>
+        /// <param name="value">The <see cref="string"/> to be appended.</param>
+        /// <returns>A reference to this instance after the append operation has completed.</returns>
+        public abstract TagHelperContent AppendEncoded(string value);
+
+        /// <summary>
         /// Appends the specified <paramref name="format"/> to the existing content after
         /// replacing the format item with the <see cref="string"/> representation of the
         /// <paramref name="arg0"/>.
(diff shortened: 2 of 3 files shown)


rynowak:
Discussion here: https://github.com/aspnet/Common/issues/34

davidfowl:
Travis hates this

rynowak:
If only github had some way to inform of that fact...

rynowak:
FYI @davidfowl this is breaking behavior change coming from the common repo. Travis will fail until that part goes in. Same with the PR I'm about to open on MVC.

NTaylorMullen:
Just to clarify. Doing `tagHelperContent.Append("<strong></strong>")` will now have to be written as `tagHelperContent.Append(new HtmlString("<strong></strong>"))` to ensure the `<` and `>` characters are output as literals to render appropriate strong tags?

rynowak:
Correct.

rynowak:
We can add an `AppendEncoded` if that helps

NTaylorMullen:
I think adding `AppendEncoded` would be awesome. Should probably be an extension method in MVC though since Razor can't access `HtmlString`.

Also, this change warrants an announcement. We have several tutorials that do `SetContent`/`Append` of tag values. Chances are it'll disrupt some of our users.

dougbu:
Not sure about `AppendEncoded()`. Already have `new HtmlString()` and `Html.Raw()` to do exactly the same thing and they are well-known MVC gestures.

NTaylorMullen:
@dougbu `Hml.Raw` is a nice way to do `HtmlString`s in Razor pages (instead of doing `@(new HtmlString(....))`. `AppendEncoded` can be the nice way to do this in `TagHelper`s. Wouldn't want users to inject `HtmlHelper`s into their `TagHelper`s just to use `Html.Raw`.

dougbu:
@NTaylorMullen my point was the system already has multiple ways to do this. I wasn't suggesting grabbing an `HtmlHelper` from DI just to use _one_ of those ways.

What's wrong with the well-known `new HtmlString()` approach?

NTaylorMullen:
@dougbu Users can write `TagHelper`s in projects that don't reference MVC at all. Therefore, if their project only referrences **Microsoft.AspNet.Razor.Runtime** (this invalidates my comment about having `AppendEncoded` as an MVC extension, it'd have to exist in Razor.Runtime) they'd have to write their own `IHtmlContent` to prevent encoding :sob:

dougbu:
Let's discuss offline tomorrow.

NeelBhatt:
I guess Html.Raw uses new HtmlString() by default as below:

```
public IHtmlString Raw(object value)
{
    return new HtmlString(value == null ? null : value.ToString());
}
```

isn't it?
Not sure though!!

rynowak:
This would make THC consistent with what we're doing for TagBuilder. https://github.com/aspnet/Mvc/issues/3087#issuecomment-137501755

These are both builder style APIs, so ideally we give you sugar to stay in that flow. It's the same reason why we defined `AppendFormat(...)`. It's far from required, just nice to have.

I don't see any reason to rush this change, it probably will just shake out of the changes for #3087

Eilon:
I'm quite certain we discussed and agreed on having an `AppendEncoded()` method like a month ago...

dougbu:
> I'm quite certain we discussed and agreed on having an `AppendEncoded()` method like a month ago.

Entirely possible that I'm forgetting something and fine if so. But note this just got more expensive if the method has to live in Razor.

dougbu:
@rynowak I suggest including `AppendEncoded()` in the first pass. For one thing, makes similar sense to have `AppendEncoded()` in `TagHelperContent` (hmm, will that implement `IHtmlContentBuilder`?) as well.

rynowak:
updated

NTaylorMullen:
:shipit:

dougbu:
Perhaps as a followup, should code generation use `AppendEncoded()` some?

dougbu:
:shipit:

rynowak:
> Perhaps as a followup, should code generation use AppendEncoded() some? 

I'm not familiar enough with codegen to know exactly what this means. Let's chat on Monday

PR #515, opened 2015-09-11
Removing NotNullAttribute from Microsoft.AspNet.Razor

pranavkm:


pranavkm:
cc @NTaylorMullen

NTaylorMullen:
:shipit:

Issue #517, opened 2015-09-12
Rename RestrictChildrenAttribute properties/params to be `ChildTag` etc.


Eilon:
See https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/RestrictChildrenAttribute.cs

```c#
tagName --> childTag
tagNames --> childTags
ChildTagNames --> ChildTags
```

NTaylorMullen:
4fd866f340ef5c53432717fe4451be25ac2c71b7

PR #518, opened 2015-09-13
Update `RestrictChildrenAttribute`s properties and parameters to have consistent names.

NTaylorMullen:
- Other attributes such as `TargetElement` have names referring to tag names as "tags". This update makes both attributes more consistent.

#517


NTaylorMullen:
Just an FYI.

Issue #502, opened 2015-08-28
Tag Helpers not available within script tag


dgee2:
I'm trying to use a tag helper within a script tag (this is how knockoutjs does its templating) but this seems to be blocked from happening. This is the sort of thing that I'm wanting to do where bool is my own custom tag helper:
```
<script type="text/html" id="checkbox-template">
    <li>
        <bool ko-bind-checked="Checked" ko-bind-value="Value" ko-bind-label="Name" />
    </li>
</script>
```

dgee2:
Just had a further think about this and I have a further question. Are the contents of a tag helper processed before or after the outer tag helper is processed or are they processed at all?

e.g. if I had a tag helper for divs (very hypothetical here) would the form tag helper be processed.
```
<div>
    <form />
</div>
``

dougbu:
At present Razor does almost nothing with the body of a `<script>` tag. That's partially an optimization for the common `type="text/javascript"` case -- a syntax Razor can't parse. Agree it may make sense to handle at least `type="text/html"`. Just likely to be a lot of extra work...

dgee2:
Yes it potentially could make using templates very difficult for us due to the amount that we are proposing to produce with tag helpers. Are there any work arounds at present? Also what would need to be involved to make the contents of custom tag helpers be processed by Razor (and thus any contained tag helpers)?

dougbu:
>  what would need to be involved to make the contents of custom tag helpers be processed by Razor

Are you asking in general or specifically about within `<script>` elements? I ask because tag helpers work pretty much everywhere _but_ within `<script>` elements. (Fixed a few bugs related to templates and some helpers recently. But the basics work in released bits.)

dgee2:
Well the fact that it didn't work in script tags made me wonder if it worked in any at all (but they do). A workaround for the script tag would be useful though.

danroth27:
@DamianEdwards Thoughts on this?

Eilon:
We decided to special case `<script>` tags with `type="text/html"` and process everything inside it as HTML, and thus process tag helpers.

dgee2:
Thankyou. Will this be configurable (either on the fly or via startup)? Given that KO is using script tags in this way, I guess there's probably others as well that may or may not be using the same type.

Eilon:
@dgee2 We're currently planning on just special-casing `text/html` because Razor knows how to parse only HTML. If in the future that is not enough, we can look at making it fancier.

[...]

PR #522, opened 2015-09-15
Allow `TagHelper`s inside of text/html typed script tags.

NTaylorMullen:
- To limit the impact of the change ensured that we only do extra work in the case that we detect a script tag with a `type` attribute.
- The parsing changes include normal HTML parsing behaviors when we detect that a script tag has a `type` attribute with value `text/html`.
- Added unit and code generation tests to validate `text/html` script tag behavior.

#502

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNet.Razor.Chunks.Generators;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Tokenizer.Symbols;
 
@@ -9,6 +12,9 @@ namespace Microsoft.AspNet.Razor.Parser
 {
     public partial class HtmlMarkupParser
     {
+        private static readonly char[] ValidAfterTypeAttributeNameCharacters =
+            new[] { ' ', '\t', '\r', '\n', '\f', '=' };
+
         public override void ParseDocument()
         {
             if (Context == null)
@@ -50,7 +56,7 @@ namespace Microsoft.AspNet.Razor.Parser
                         return;
                     }
 
-                    // We should behave like a normal tag that has a parser escape, fall through to the normal 
+                    // We should behave like a normal tag that has a parser escape, fall through to the normal
                     // tag logic.
                 }
                 else if (NextIs(HtmlSymbolType.QuestionMark))
@@ -79,7 +85,9 @@ namespace Microsoft.AspNet.Razor.Parser
                     Optional(HtmlSymbolType.ForwardSlash);
                     Optional(HtmlSymbolType.CloseAngle);
 
-                    if (scriptTag)
+                    // If the script tag expects javascript content then we should do minimal parsing until we reach
+                    // the end script tag. Don't want to incorrectly parse a "var tag = '<input />';" as an HTML tag.
+                    if (scriptTag && !CurrentScriptTagExpectsHtml())
                     {
                         Output(SpanKind.Markup);
                         tagBlock.Dispose();
@@ -106,6 +114,55 @@ namespace Microsoft.AspNet.Razor.Parser
                 // End tag block
                 tagBlock.Dispose();
             }
+        }
+
+        private bool CurrentScriptTagExpectsHtml()
+        {
+            var blockBuilder = Context.CurrentBlock;
+
+            Debug.Assert(blockBuilder != null);
+
+            var typeAttribute = blockBuilder.Children
+                .OfType<Block>()
+                .Where(block =>
+                    block.ChunkGenerator is AttributeBlockChunkGenerator &&
+                    block.Children.Count() >= 2)
+                .FirstOrDefault(IsTypeAttribute);
+
+            if (typeAttribute != null)
+            {
+                var contentValues = typeAttribute.Children
+                    .OfType<Span>()
+                    .Where(childSpan => childSpan.ChunkGenerator is LiteralAttributeChunkGenerator)
+                    .Select(childSpan => childSpan.Content);
+
+                var scriptType = string.Concat(contentValues).Trim();
+
+                // Does not allow charset parameter (or any other parameters).
+                return string.Equals(scriptType, "text/html", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+
+        private static bool IsTypeAttribute(Block block)
+        {
+            var span = block.Children.First() as Span;
+
+            if (span == null)
+            {
+                return false;
+            }
+
+            var trimmedStartContent = span.Content.TrimStart();
+            if (trimmedStartContent.StartsWith("type", StringComparison.OrdinalIgnoreCase) &&
+                (trimmedStartContent.Length == 4 ||
+                ValidAfterTypeAttributeNameCharacters.Contains(trimmedStartContent[4])))
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
@@ -18,6 +18,273 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
 {
     public class TagHelperParseTreeRewriterTest : TagHelperRewritingTestBase
     {
+        public static TheoryData InvalidHtmlScriptBlockData
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var blockFactory = new BlockFactory(factory);
+
+                return new TheoryData<string, MarkupBlock>
+                {
+                    {
+                        "<script type><input /></script>",
+                        new MarkupBlock(
+                            new MarkupTagBlock(
+                                factory.Markup("<script"),
+                                new MarkupBlock(factory.Markup(" type")),
+                                factory.Markup(">")),
+                            factory.Markup("<input />"),
+                            blockFactory.MarkupTagBlock("</script>"))
+                    },
+                    {
+                        "<script types='text/html'><input /></script>",
+                        new MarkupBlock(
+                            new MarkupTagBlock(
+                                factory.Markup("<script"),
+                                new MarkupBlock(
+                                    new AttributeBlockChunkGenerator(
+                                        name: "types",
+                                        prefix: new LocationTagged<string>(" types='", 7, 0, 7),
+                                        suffix: new LocationTagged<string>("'", 24, 0, 24)),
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+                {
+                    var contentValues = typeAttribute.Children
+                        .OfType<Span>()
+                        .Where(childSpan => childSpan.ChunkGenerator is LiteralAttributeChunkGenerator)
+                        .Select(childSpan => ((LiteralAttributeChunkGenerator)childSpan.ChunkGenerator))
+                        .Select(valueGenerator => valueGenerator.Prefix.Value + valueGenerator.Value.Value);
+
+                    var scriptType = string.Concat(contentValues).Trim();
+
+                    return !string.Equals(scriptType, "text/html", StringComparison.OrdinalIgnoreCase);
```
dougbu: Add a comment that the `charset` parameter is disallowed and `text/html` has no other parameters. Otherwise this looks unsafe.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+                        .TrimStart()
+                        .StartsWith("type", StringComparison.OrdinalIgnoreCase) == true);
+
+                if (typeAttribute != null)
+                {
+                    var contentValues = typeAttribute.Children
+                        .OfType<Span>()
+                        .Where(childSpan => childSpan.ChunkGenerator is LiteralAttributeChunkGenerator)
+                        .Select(childSpan => ((LiteralAttributeChunkGenerator)childSpan.ChunkGenerator))
+                        .Select(valueGenerator => valueGenerator.Prefix.Value + valueGenerator.Value.Value);
```
dougbu: Why include the prefix -- won't it always contain whitespace?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+
+                    var scriptType = string.Concat(contentValues).Trim();
```
dougbu: This and the previous Linq expression look very roundabout. In what case would "text/html" be split between multiple `Span`s? If none, suggest instead using `FirstOrDefault()` to find the non-`IsNullOrEmpty()` `Span` of interest. Ignore user errors such as specifying `... ; charset=utf-8`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+
+            Debug.Assert(blockBuilder != null);
+
+            var potentialAttributes = blockBuilder.Children
+                .OfType<Block>()
+                .Where(block =>
+                    block.ChunkGenerator is AttributeBlockChunkGenerator &&
+                    block.Children.Count() >= 2);
+
+            if (potentialAttributes.Any())
```
dougbu: Write one Linq expression (remove the `potentialAttributes` intermediate) and don't perform this additional evaluation. `FirstOrDefault()` does the `Any()` bit already.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
                 tagBlock.Dispose();
             }
         }
+
+        private bool CurrentTagBlockExpectsJavascriptContent()
```
dougbu: Method name is incorrect. Returns a value indicating the current `<script>` tag does not expect HTML. But it could expect an infinite number of other content types, not just JavaScript. Suggest reversing the various conditions where this is used, return `false` by default, and name the method `CurrentScriptTagExpectsHtml()`.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+                    block.ChunkGenerator is AttributeBlockChunkGenerator &&
+                    block.Children.Count() >= 2);
+
+            if (potentialAttributes.Any())
+            {
+                var typeAttribute = potentialAttributes
+                    .FirstOrDefault(attributeBlock => (attributeBlock.Children.First() as Span)
+                        ?.Content
+                        .TrimStart()
+                        .StartsWith("type", StringComparison.OrdinalIgnoreCase) == true);
```
dougbu: Does not appear safe, will match a `types` attribute for example.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+
+                if (typeAttribute != null)
+                {
+                    var contentValues = typeAttribute.Children
+                        .OfType<Span>()
+                        .Where(childSpan => childSpan.ChunkGenerator is LiteralAttributeChunkGenerator)
```
dougbu: Why not just use the `Span`'s `Content`?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,42 @@ private void ScanTagInDocumentContext()
...
+            {
+                var typeAttribute = potentialAttributes
+                    .FirstOrDefault(attributeBlock => (attributeBlock.Children.First() as Span)
+                        ?.Content
+                        .TrimStart()
+                        .StartsWith("type", StringComparison.OrdinalIgnoreCase) == true);
+
+                if (typeAttribute != null)
+                {
+                    var contentValues = typeAttribute.Children
```
dougbu: Shouldn't this instead use `Flatten()` to cover all constituent `Span`s?

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: To protect against cases such as: `<script type="text /html">` Or enable cases such as: `<script type="  text/html">`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: It's not about being split between different spans but more to ensure that 

`<script type="text/html andmore">` doesn't return positive. (Will be two different spans)

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: Guess we could

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: No need. At this point we can guaruntee that all content inside the attribute consist of spans. We just want to ensure that the one we inspect happen to have a `LiteralAttributeChunkGenerator` on them.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +112,59 @@ private void ScanTagInDocumentContext()
...
+        {
+            var blockBuilder = Context.CurrentBlock;
+
+            Debug.Assert(blockBuilder != null);
+
+            var typeAttribute = blockBuilder.Children
+                .OfType<Block>()
+                .Where(block =>
+                    block.ChunkGenerator is AttributeBlockChunkGenerator &&
+                    block.Children.Count() >= 2)
```
ajaybhargavb: Add a test to verify that this ignores minimized type attributes.

ajaybhargavb:
:watch:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: Done

ajaybhargavb:
:shipit:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -18,6 +18,273 @@ namespace Microsoft.AspNet.Razor.Test.TagHelpers
...
+        {
+            get
+            {
+                var factory = CreateDefaultSpanFactory();
+                var blockFactory = new BlockFactory(factory);
+
+                return new TheoryData<string, MarkupBlock>
+                {
+                    {
+                        "<script type><input /></script>",
```
dougbu: I'd be very surprised if any browser accepted this as valid. Neither "type" or the empty string is a [valid MIME type](http://www.w3.org/TR/html5/infrastructure.html#valid-mime-type). And minimized attributes only works for HTML [Boolean attributes](http://www.w3.org/TR/html5/infrastructure.html#boolean-attribute).

dougbu:
Sorry for scribbling on the middle commit.

dougbu:
:watch:

test/Microsoft.AspNet.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: Supposed to be invalid hence the `InvalidHtmlScriptBlockData`. This is just making sure Razor doesn't think it's valid :smile:

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -107,5 +115,55 @@ private void ScanTagInDocumentContext()
...
+
+            var trimmedStartContent = span.Content.TrimStart();
+            if (trimmedStartContent.StartsWith("type", StringComparison.OrdinalIgnoreCase) &&
+                (trimmedStartContent.Length == 4 ||
+                ValidAfterTypeAttributeNameCharacters.Contains(trimmedStartContent[4])))
+            {
+                return true;
+            }
+
+
```
dougbu: nuke extra line

dougbu:
:shipit:

Issue #137, opened 2014-09-15
Support symbol data bound attributes in HTML


NTaylorMullen:
Examples that we may need to support:

```HTML
<ul ng-repeat<="items">...</ul>
<button click+="doSomething()">Click Me</button>
<input value&="firstName">
<ul [ng-repeat|item]="items"></ul>
<button (click)="doSOmething()">Click Me</button>
```

These are just some, in general we should implement an attribute parsing approach that revolves around the '=' sign and less about the content around it.

Ultimately this showcases an issue with Razor where it expects attributes to be *exactly* in the form: `attributeName="attributeValue"` where the double quotes are optional.  Adding spaces or anything else around the = causes attribute parsing issues.  That specific issue is listed here: https://github.com/aspnet/Razor/issues/123

Eilon:
Need to update this issue with the most recent Angular 2 syntax.

We would most likely treat the entire LHS as the "name" of an attribute in tag helper cases, e.g. `(click)` would be the attribute name.

In the more general Razor case the key goal is to just not have parser errors on these and to reliably echo the output into the HTML stream.

NTaylorMullen:
After some research and testing. Here's the syntax angular 2 seems to enable:

```HTML
<ul [item]="items"></ul>
<ul [(item)]="items"></ul>
<button (click)="doSOmething()">Click Me</button>
<button (^click)="doSOmething()">Click Me</button>
<template *something="value"></template>
<div #local></div>
<div #local="value"></div>
```

Out of the supported syntaxes, the ones that currently aren't supported by Razor:

```HTML
<ul [item]="items"></ul>
<ul [(item)]="items"></ul>
<template *something="value"></template>
```

When working on this issue I'm going to investigate what it'll take to just get Razor allowing the full HTML5 spec of attribute names:
> Attribute names must consist of one or more characters other than the space characters, U+0000 NULL, """, "'", ">", "/", "=", the control characters, and any characters that are not defined by U

NTaylorMullen:
7239eb015c708e3e03eecd3c6785c595b8b8d938

PR #530, opened 2015-09-17
Add support for data bound symbols as attribute names.

NTaylorMullen:
- Took the HTML5 spec approach of disallowing specific characters and accepting all others.
- Added several parser and code generation tests to cover both `TagHelper` and non-`TagHelper` variations of symbol bound attribute names.

#137

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -464,7 +464,7 @@ namespace Microsoft.AspNet.Razor.Parser
             // Read the 'name' (i.e. read until the '=' or whitespace/newline)
             var name = Enumerable.Empty<HtmlSymbol>();
             var whitespaceAfterAttributeName = Enumerable.Empty<HtmlSymbol>();
-            if (At(HtmlSymbolType.Text))
+            if (IsValidAttributeNameSymbol(CurrentSymbol))
             {
                 name = ReadWhile(sym =>
                                  sym.Type != HtmlSymbolType.WhiteSpace &&
@@ -1135,6 +1135,29 @@ namespace Microsoft.AspNet.Razor.Parser
                 AddMarkerSymbolIfNecessary();
             }
             Output(SpanKind.Markup);
+        }
+
+        internal static bool IsValidAttributeNameSymbol(HtmlSymbol symbol)
+        {
+            if (symbol == null)
+            {
+                return false;
+            }
+
+            // These restrictions cover most of the spec defined: http://www.w3.org/TR/html5/syntax.html#attributes-0
+            // However, it's not all of it. For instance we don't special case control characters or allow OpenAngle.
+            // It also doesn't try to exclude Razor specific features such as the @ transition. This is based on the
+            // expectation that the parser handles such scenarios prior to falling through to name resolution.
+            var symbolType = symbol.Type;
+            return symbolType != HtmlSymbolType.WhiteSpace &&
+                symbolType != HtmlSymbolType.NewLine &&
+                symbolType != HtmlSymbolType.CloseAngle &&
+                symbolType != HtmlSymbolType.OpenAngle &&
+                symbolType != HtmlSymbolType.ForwardSlash &&
+                symbolType != HtmlSymbolType.DoubleQuote &&
+                symbolType != HtmlSymbolType.SingleQuote &&
+                symbolType != HtmlSymbolType.Equals &&
+                symbolType != HtmlSymbolType.Unknown;
         }
     }
 }
--- a/test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
@@ -14,6 +14,174 @@ namespace Microsoft.AspNet.Razor.Test.Parser.Html
 {
     public class HtmlAttributeTest : CsHtmlMarkupParserTestBase
     {
+        public static TheoryData SymbolBoundAttributeNames
+        {
+            get
+            {
+                return new TheoryData<string>
+                {
+                    "[item]",
+                    "[(item,",
+                    "(click)",
+                    "(^click)",
+                    "*something",
+                    "#local",
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(SymbolBoundAttributeNames))]
+        public void SymbolBoundAttributes_BeforeEqualWhitespace(string attributeName)
+        {
+            // Arrange
+            var attributeNameLength = attributeName.Length;
+            var newlineLength = Environment.NewLine.Length;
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1128,5 +1128,24 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
         }
+
+        internal static bool IsValidAttributeNameSymbol(HtmlSymbol symbol)
+        {
+            if (symbol == null)
+            {
+                return false;
+            }
+
+            var symbolType = symbol.Type;
```
dougbu: Long list and perhaps the `... == ... || ...` list would be (a bit) longer. Nonetheless suggest switching to the positive list for those of us who haven't memorized `HtmlSymbolType`.

ajaybhargavb:
Maybe add one or two tests with newlines and whitespace before and after the attribute name.

ajaybhargavb:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Nah, spec defines it via the negative.

NTaylorMullen:
Updated.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
```diff
@@ -14,6 +14,112 @@ namespace Microsoft.AspNet.Razor.Test.Parser.Html
...
+                lineIndex: 2,
+                characterIndex: 5 + attributeNameLength);
+            var valueLocation2 = new SourceLocation(
+                absoluteIndex: 14 + attributeNameLength * 2 + newlineLength * 2,
+                lineIndex: 2,
+                characterIndex: 2 + attributeNameLength);
+
+            // Act & Assert
+            ParseBlockTest(
+                $"<a {Environment.NewLine}  {attributeName}='Foo'\t{Environment.NewLine}{attributeName}='Bar' />",
```
ajaybhargavb: Suggest adding one more variation with Environment.NewLine before =.

ajaybhargavb:
:shipit:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1136,5 +1136,24 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
+            return symbolType != HtmlSymbolType.WhiteSpace &&
+                symbolType != HtmlSymbolType.NewLine &&
+                symbolType != HtmlSymbolType.CloseAngle &&
+                symbolType != HtmlSymbolType.OpenAngle &&
```
dougbu: `OpenAngle` is legal according to http://www.w3.org/TR/html5/syntax.html#attributes-0

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
dougbu: :ok: how 'bout a link to http://www.w3.org/TR/html5/syntax.html#attributes-0 ?

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1136,5 +1136,24 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
+                symbolType != HtmlSymbolType.ForwardSlash &&
+                symbolType != HtmlSymbolType.DoubleQuote &&
+                symbolType != HtmlSymbolType.SingleQuote &&
+                symbolType != HtmlSymbolType.Equals &&
+                symbolType != HtmlSymbolType.Unknown;
```
dougbu: `HtmlSymbolType` doesn't line up with HTML character classification. This expression allows `HtmlSymbolType.Transition`, `RazorComment`, `RazorCommentStar`, and `RazorCommentTransition`. None of those seem correct for attribute names in the Razor context.

Further `HtmlSymbolType.Text` (currently allowed) indicates _any_ non-whitespace non-symbol series of characters. Need to exclude control characters and unknown Unicode characters when that's the `symbolType`.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
```diff
@@ -14,6 +14,112 @@ namespace Microsoft.AspNet.Razor.Test.Parser.Html
...
+            get
+            {
+                return new TheoryData<string>
+                {
+                    "[item]",
+                    "[(item,",
+                    "(click)",
+                    "(^click)",
+                    "*something",
+                    "#local",
```
dougbu: Need to test more than just symbols Angular would allow. Non-Latin1 characters at least.

I also don't see tests verifying rejection of anything but whitespace and `=`.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: WIll add

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Yup, Razor don't like it though :smile:. This is an "almost" spec worthy fix. 

This PR solves all of the use-cases brought forward in issue #137 and takes an extra step in making it spec equivalent. I just filed an issue to cover the open angle derivation: https://github.com/aspnet/Razor/issues/538

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Block.cs
NTaylorMullen: Talked in person.

test/Microsoft.AspNet.Razor.Test/Parser/Html/HtmlAttributeTest.cs
NTaylorMullen: Talked in person.

NTaylorMullen:
Updated

dougbu:
Would still like to see a couple of negative tests. Show attribute name parsing stopping at excluded characters. E.g. confirm `name>` or `name'` are parsed as the attribute name `name` followed by a separate (garbage) symbol. Otherwise this PR LGTM.

dougbu:
Talked offline. The tests I wanted already exist and didn't need to be changed in this PR.

dougbu:
:shipit:

PR #531, opened 2015-09-18
Added temporary parse error for helper directive

ajaybhargavb:
#472 
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/SyntaxConstants.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/SyntaxConstants.cs
@@ -27,6 +27,9 @@ namespace Microsoft.AspNet.Razor.Parser
             public static readonly string ElseIfKeyword = "else if";
             public static readonly string NamespaceKeyword = "namespace";
             public static readonly string ClassKeyword = "class";
+
+            // Not supported. Only used for error cases.
+            public static readonly string HelperKeyword = "helper";
         }
     }
 }
--- a/src/Microsoft.AspNet.Razor/RazorResources.resx
+++ b/src/Microsoft.AspNet.Razor/RazorResources.resx
@@ -425,4 +425,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="TagHelperParseTreeRewriter_InvalidNestedTag" xml:space="preserve">
     <value>The &lt;{0}&gt; tag is not allowed by parent &lt;{1}&gt; tag helper. Only child tag helper(s) targeting tag name(s) '{2}' are allowed.</value>
   </data>
+  <data name="ParseError_HelperDirectiveNotAvailable" xml:space="preserve">
+    <value>The {0} directive is not supported.</value>
+  </data>
 </root>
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -425,4 +425,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="TagHelperParseTreeRewriter_InvalidNestedTag" xml:space="preserve">
     <value>The &amp;lt;{0}&amp;gt; tag is not allowed by parent &amp;lt;{1}&amp;gt; tag helper. Only child tag helper(s) targeting tag name(s) '{2}' are allowed.</value>
   </data>
+  <data name="ParseError_HelperDirectiveNotAvailable" xml:space="preserve">
+    <value>The @helper directive is not available in Razor 4.</value>
```
ajaybhargavb: Suggestions for error message welcome :smile:

src/Microsoft.AspNet.Razor/Parser/SyntaxConstants.cs
```diff
@@ -27,6 +27,7 @@ public static class CSharp
             public static readonly string ElseIfKeyword = "else if";
             public static readonly string NamespaceKeyword = "namespace";
             public static readonly string ClassKeyword = "class";
+            public static readonly string HelperKeyword = "helper";
```
NTaylorMullen: Separate this from the rest and leave a comment above it stating that it's only for error use-cases.

src/Microsoft.AspNet.Razor/RazorResources.resx
NTaylorMullen: `The {0} directive is no longer supported.`

Don't want to specify the version for when we rev, and also don't want to inline `@helper` because it may be localized.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpErrorTest.cs
```diff
@@ -28,6 +28,21 @@ public void ParseBlockHandlesQuotesAfterTransition()
         }
 
         [Fact]
+        public void ParseBlockWithHelperDirectiveProducesError()
+        {
+            ParseBlockTest("@helper { }",
+                           new ExpressionBlock(
```
NTaylorMullen: - `@helper someHelper {  }`
- nit: Use the new spacing format (aka 1 indent in).

NTaylorMullen:
:watch:

ajaybhargavb:
Updated.

src/Microsoft.AspNet.Razor/RazorResources.resx
```diff
@@ -425,4 +425,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="TagHelperParseTreeRewriter_InvalidNestedTag" xml:space="preserve">
     <value>The &amp;lt;{0}&amp;gt; tag is not allowed by parent &amp;lt;{1}&amp;gt; tag helper. Only child tag helper(s) targeting tag name(s) '{2}' are allowed.</value>
   </data>
+  <data name="ParseError_HelperDirectiveNotAvailable" xml:space="preserve">
+    <value>The {0} directive is no longer supported.</value>
```
Eilon: `is not supported`

NTaylorMullen:
:shipit: once @Eilon's comment is addressed.

Issue #532, opened 2015-09-18
@tagHelperPrefix does not work with [RestrictChildren(...)]


NTaylorMullen:
If you have a `TagHelper` with restricted children of `input` and `span` and your page has `@tagHelperPrefix "th:"`. You will not be able to write your `input` and `span` `TagHelper`s without receiving a Razor parse error.

NTaylorMullen:
Assigned to myself and for this milestone to unblock tooling.

NTaylorMullen:
36450c978b21048aa078ba4d7e78f5f8bd02ad35

PR #533, opened 2015-09-18
Add support for `@tagHelperPrefix` and `[RestrictChildren]`.

NTaylorMullen:
- Added a test to validate prefixed `TagHelper`s can have children with `TagHelper` prefixes.

#532


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -567,9 +567,12 @@ public TagHelperBlockTracker(TagHelperBlockBuilder builder)
 
                 if (Builder.Descriptors.Any(descriptor => descriptor.AllowedChildren != null))
                 {
+                    var tagHelperPrefix = Builder.Descriptors.First().Prefix;
```
pranavkm: Throw a Debug.Assert?

pranavkm:
:shipit:

PR #535, opened 2015-09-19
Update tests to run correctly on stable mono.

NTaylorMullen:
/cc @pranavkm


NTaylorMullen:
just FYI for @pranavkm

PR #536, opened 2015-09-19
ITypeInfo.FullName should return the same value as TypeInfo.FullName

### LINKED ISSUE #523: ITypeInfo.FullName should return the same value as TypeInfo.FullName
Right now we sanitize the FullName to remove assembly name for generic types for e.g.
`typeof(List<string>).FullName` - `System.Collections.Generic.List`1[[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]`

versus
`new RuntimeTypeInfo(typeof(List<string>).GetTypeInfo()).FullName` - `System.Collections.List`1[[System.String]]`

The primary motive for sanitizing these type names was to workaround an issue in precompilation where CoreCLR types are compiled against types in System.Runtime (consequently FullName has `[[System.String, System.Runtime...]]`) but get type forwarded to mscorlib at runtime. 

To resolve this issue, we should declare make `ITypeInfo: IEquatable<ITypeInfo>` which can do the right thing in terms of comparisons rather than relying on comparability of `ITypeInfo.FullName`

pranavkm:
Fixes #523


pranavkm:
cc @NTaylorMullen

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -8,11 +9,18 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
     /// Contains type metadata.
     /// </summary>
-    public interface ITypeInfo : IMemberInfo
+    public interface ITypeInfo : IMemberInfo, IEquatable<ITypeInfo>
     {
         /// <summary>
         /// Fully qualified name of the type.
         /// </summary>
+        /// <remarks>
+        /// On CoreCLR, we compile against the System.String contract in System.Runtime
```
NTaylorMullen: Makes it seem like this FUllName is only for strings. should clarify the string scenario as an example

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
```diff
@@ -116,7 +117,6 @@ public string TypeName
                 }
 
                 _typeName = value;
-                IsStringProperty = string.Equals(_typeName, typeof(string).FullName, StringComparison.Ordinal);
```
NTaylorMullen: Why move this to the ctor? By making the `IsStringProperty` public set it'll now be serialized when sent over the wire. (not needed I think).

test/Microsoft.AspNet.Razor.Runtime.Precompilation.Test/PrecompilationTagHelperDescriptorFactoryTest.cs
```diff
@@ -27,5 +27,54 @@ public override ITypeInfo GetTypeInfo(Type tagHelperType)
             return Assert.Single(typeResolver.GetExportedTypes(CompilationUtility.GeneratedAssemblyName),
                 generatedType => string.Equals(generatedType.FullName, tagHelperType.FullName, StringComparison.Ordinal));
         }
+
+        [Theory]
+        [MemberData(nameof(TagHelperWithPrefixData))]
+        public void CreateDescriptors_WithPrefixes_ReturnsExpectedAttributeDescriptors(
```
NTaylorMullen: Oo we weren't able to do this before?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -13,44 +13,18 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
-            };
-
         [Theory]
-        [MemberData(nameof(RuntimeTypeInfo_ReturnsMetadataOfAdaptingTypeData))]
-        public void RuntimeTypeInfo_ReturnsMetadataOfAdaptingType(Type type, string expectedFullName)
+        [InlineData(typeof(int))]
+        [InlineData(typeof(string))]
+        [InlineData(typeof(Tuple<>))]
+        [InlineData(typeof(Tuple<>))]
+        [InlineData(typeof(IDictionary<string, string>))]
```
NTaylorMullen: What happened to `typeof(IDictionary<string, IDictionary<string, CustomType>>`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
```diff
@@ -580,6 +554,11 @@ public IEnumerable<IPropertyInfo> Properties
                 }
             }
 
+            public bool Equals(ITypeInfo other)
```
NTaylorMullen: ?

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
pranavkm: It's set in the test-only ctor. We let the factory set the other value.

test/Microsoft.AspNet.Razor.Runtime.Precompilation.Test/PrecompilationTagHelperDescriptorFactoryTest.cs
pranavkm: We were. I had to copy the test over here so I could customize the behavior (the CoreCLR sanitization) for precompilation.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
pranavkm: Interface change.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/RuntimeTypeInfoTest.cs
pranavkm: Most of those cases aren't super interesting per se now - they were more useful when we were sanitizing the `FullName`s

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
NTaylorMullen: Why? I think this change would actually break back compatibility within the `TagHelperDescriptor` resolution that we use at design time. For instance:

Old `Razor.Tooling` (`isStringProperty` is not serialized across the wire) + new VisualStudio = VS trying to deserialize the payload into the new `TagHelperAttributeDescriptor` format would result in `IsStringProperty` not being set (unless VS hardcoded that logic on their end, badd).

pranavkm: But we decided relying on comparing FullNames isn't necessarily correct. Why would we continue doing that?

NTaylorMullen: Ah I see your point. This is tough. We could potentially have the `IsStringProperty` here call through to `SanitizeFullName` couldn't we?

pranavkm: We could leave it around for now and fix it in RC1 if it stops tooling from regressing.

NTaylorMullen: > We could leave it around for now  

To clarify, as in keep the `IsStingProperty` private setter & set via setting of `TypeName` right?

pranavkm:
Updated

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ITypeInfo.cs
```diff
@@ -8,11 +9,18 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
...
     /// Contains type metadata.
     /// </summary>
-    public interface ITypeInfo : IMemberInfo
+    public interface ITypeInfo : IMemberInfo, IEquatable<ITypeInfo>
     {
         /// <summary>
         /// Fully qualified name of the type.
         /// </summary>
+        /// <remarks>
+        /// On CoreCLR, some BCL types get type forwarded to the full desktop framework implementations at
```
NTaylorMullen: Beautiful :100:

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperAttributeDescriptor.cs
NTaylorMullen: This was done :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -671,6 +673,7 @@ private static bool ShouldSkipDescriptorCreation(bool designTime, IMemberInfo me
             string attributeName,
             string typeName,
             bool isIndexer,
+            bool isStringProperty,
```
NTaylorMullen: Do we still need to take in the `isStringProperty` here? Doesn't it get calculated via `typeName` correctly for these tests?

test/Microsoft.AspNet.Razor.Runtime.Precompilation.Test/project.json
```diff
@@ -23,6 +23,9 @@
     },
     "frameworks": {
         "dnx451": {
+            "frameworkAssemblies": {
+                "System.Text.Encoding": "4.0.0.0"
```
NTaylorMullen: `"System.Text.Encoding": ""`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -963,7 +963,8 @@ public void CreateDescriptors_ResolvesMultipleTagHelperDescriptorsFromSingleType
                         {
                             Name = "valid-attribute",
                             PropertyName = nameof(MultiTagTagHelper.ValidAttribute),
-                            TypeName = typeof(string).FullName
+                            TypeName = typeof(string).FullName,
+                            IsStringProperty = true
```
NTaylorMullen: no need here and elsewhere, wont leave any more comments on it being set.

NTaylorMullen:
Looks good, once all the various extra `IsStringProperty` test bits are removed :shipit:

test/Microsoft.AspNet.Razor.Runtime.Precompilation.Test/project.json
pranavkm: For some reason I couldn't build the project locally - it said it needed this reference.

NTaylorMullen: It's due to the Testing change @kichalla made. Does it not like the empty string as the version number?

dougbu: I am about to submit a commit removing the version numbers from all `frameworkAssemblies` entries in Testing. Hopefully that will allow @pranavkm to clean this up too.

Issue #526, opened 2015-09-16
Add extension methods on HtmlContentBuilder for String.Format style formatting


danroth27:

danroth27:
Talk to @rynowak and @Eilon to discuss the design.

rynowak:
a602b47e262b1733611018f1d357784dd7e10c5c

Issue #479, opened 2015-08-10
Consider removing `TagHelperContent.GetContent()`


sornaks:
`TagHelperContent.GetContent()` is hardcoded to use the default encoder, which is bad because it's a replaceable service. We recently added `TagHelperContent.GetContent(IHtmlEncoder)` - we should remove the no-param version.

Original Text:
~~Currently the content which is stored in `TagHelperContent.GetContent()` is unencoded. The only way to get encoded content is using `WriteTo(TextWriter, IHtmlEncoder)`. Consider having an overload like `GetContent(IHtmlEncoder)` which returns the encoded content.~~

~~Reference PR comment: https://github.com/aspnet/Razor/pull/473/files#r36673441~~

rynowak:
See discussion here: https://github.com/aspnet/Razor/pull/541/files#r40223087

Eilon:
Close?

NTaylorMullen:
@rynowak do we still feel that this is important? It's relatively simple to do.

rynowak:
It's not really that important, especially since we have the overload that lets you pass in the encoder. If you get blocked on this it's easy to work around.

NTaylorMullen:
kk, going to close as wontfix since it's an easy mechanism to use/workaround if it causes you trouble.

PR #541, opened 2015-09-22
Avoid GetContent and ToString in TagHelperContent tests

rynowak:
In preparation for changing the encoding behavior of `AppendFormat(...)`...

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
@@ -202,6 +202,13 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <returns>A <see cref="string"/> containing the content.</returns>
         public abstract string GetContent();
 
+        /// <summary>
+        /// Gets the content.
+        /// </summary>
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/>.</param>
+        /// <returns>A <see cref="string"/> containing the content.</returns>
+        public abstract string GetContent(IHtmlEncoder encoder);
+
         /// <inheritdoc />
         public abstract void WriteTo(TextWriter writer, IHtmlEncoder encoder);
     }
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
@@ -4,9 +4,7 @@
 using System;
 using System.Globalization;
 using System.IO;
-using System.Linq;
 using Microsoft.AspNet.Testing;
-using Microsoft.Framework.WebEncoders;
 using Microsoft.Framework.WebEncoders.Testing;
 using Xunit;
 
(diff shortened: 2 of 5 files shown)


test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -265,23 +283,25 @@ public void CanAppendFormat_WithDifferentCurrentCulture()
...
 
             // Assert
-            Assert.Equal("01 February 2015", tagHelperContent.GetContent());
+            Assert.Equal(
+                "HtmlEncode[[01 February 2015]]",
+                tagHelperContent.GetContent(new CommonTestEncoder()));
         }
 
-        [Fact]
+        [Fact(Skip = "Blocked by #256")]
```
NTaylorMullen: Hu? https://github.com/aspnet/Razor/pull/256

rynowak: Sorry, this should be #526 - Any other feedback?

NTaylorMullen: Overall looks good. Should make sure travis builds successfully after rebasing.

rynowak: That sounds like a :shipit: to me.

NTaylorMullen: As long as Travis passes si :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -202,6 +202,13 @@ public TagHelperContent SetContent(IHtmlContent htmlContent)
         /// <returns>A <see cref="string"/> containing the content.</returns>
         public abstract string GetContent();
 
+        /// <summary>
+        /// Gets the content.
+        /// </summary>
+        /// <param name="encoder">The <see cref="IHtmlEncoder"/>.</param>
+        /// <returns>A <see cref="string"/> containing the content.</returns>
+        public abstract string GetContent(IHtmlEncoder encoder);
```
Eilon: Though I can't imagine this to be a big extensibility point of tag helpers, isn't it icky to have *two* abstract methods that do essentially the same thing -- especially when one calls the other? I think a more traditional pattern would be for the parameterless method to be virtual and call the parameterful method with the default `HtmlEncoder`, no? (And then don't override the parameterless method in the derived type.)

NTaylorMullen: We only expose the abstract class to users when they do things like `output.Content`; therefore it needs to be rich enough to work with. 

Maybe a better alternative would be for this version to be the default and have a `GetContent()` extension method?

rynowak: This change was made as a bit of a drive-by.

The problem with what both are you are suggesting is that the HtmlEncoder is a service and is replaceable. We either need to pass it in to the constructor, or drop the parameterless overload.

So while I agree that having `GetContent()` be virtual non-abstract here would be a more expected design, it's not really correct either because we just shouldn't provide that API.

Eilon: Log a bug to address in RC? For beta8 this is certainly fine.

rynowak: Sure

rynowak: #479 updated to reflect the current state of things

Issue #426, opened 2015-06-22
Simplify '@addTagHelper' syntax for globbing types


Eilon:
Apparently there's a regex that's used to allow importing entire namespaces or partial type name matches via the `*` character.

This code needs to be simplified to not use regex. It should support s `*` only at the end of the type name.

ajaybhargavb:
3f34add49704b4f476d6d354d85f44daf77d06c7

PR #542, opened 2015-09-22
Removed support for regex type names in @addTagHelper and @removeTagH…

ajaybhargavb:
…elper

#426 
Updated tests to not use regex patterns

@NTaylorMullen or @dougbu

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
@@ -236,16 +236,20 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 return false;
             }
 
-            // We need to escape the TypePattern so we can choose to only allow specific regex.
-            var escaped = Regex.Escape(lookupInfo.TypePattern);
+            if (lookupInfo.TypePattern.EndsWith("*", StringComparison.Ordinal))
+            {
+                if (lookupInfo.TypePattern.Length == 1)
+                {
+                    // TypePattern is "*".
+                    return true;
+                }
 
-            // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire
-            // string. We also replace any '*' or '?' characters with regex to match appropriate content.
-            // '*' matches 0 or more characters lazily and '?' matches 1 character.
-            var pattern = "^" + escaped.Replace(@"\?", ".").Replace(@"\*", ".*?") + "$";
-            var regex = new Regex(pattern, RegexOptions.Singleline, Constants.RegexMatchTimeout);
+                var lookupTypeName = lookupInfo.TypePattern.Substring(0, lookupInfo.TypePattern.Length - 1);
 
-            return regex.IsMatch(descriptor.TypeName);
+                return descriptor.TypeName.StartsWith(lookupTypeName, StringComparison.Ordinal);
+            }
+
+            return string.Equals(descriptor.TypeName, lookupInfo.TypePattern, StringComparison.Ordinal);
         }
 
         private static LookupInfo GetLookupInfo(
@@ -255,9 +259,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             var lookupText = directiveDescriptor.DirectiveText;
             var lookupStrings = lookupText?.Split(new[] { ',' });
 
-            // Ensure that we have valid lookupStrings to work with. Valid formats are:
-            // "assemblyName"
-            // "typeName, assemblyName"
+            // Ensure that we have valid lookupStrings to work with. The valid format is "typeName, assemblyName"
             if (lookupStrings == null ||
                 lookupStrings.Any(string.IsNullOrWhiteSpace) ||
                 lookupStrings.Length != 2)
--- a/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
+++ b/test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
@@ -149,7 +149,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                             },
                             new TagHelperDirectiveDescriptor
                             {
-                                DirectiveText = "*Plain*, " + assemblyA,
+                                DirectiveText = "Microsoft.AspNet.Razor.Runtime.TagHelpers.Valid_P*, " + assemblyA,
                                 Location = directiveLocation1,
                                 DirectiveType = TagHelperDirectiveType.AddTagHelper
                             }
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -236,16 +236,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
...
-            // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire
-            // string. We also replace any '*' or '?' characters with regex to match appropriate content.
-            // '*' matches 0 or more characters lazily and '?' matches 1 character.
-            var pattern = "^" + escaped.Replace(@"\?", ".").Replace(@"\*", ".*?") + "$";
-            var regex = new Regex(pattern, RegexOptions.Singleline, Constants.RegexMatchTimeout);
+                return descriptor.TypeName.StartsWith(lookupInfo.TypePattern.Substring(0, lookupInfo.TypePattern.Length - 1));
+            }
 
-            return regex.IsMatch(descriptor.TypeName);
+            return string.Equals(descriptor.TypeName, lookupInfo.TypePattern);
```
NTaylorMullen: Comparers throughout.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -236,16 +236,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
...
+                {
+                    return true;
+                }
 
-            // We surround the escaped with ^ and $ in order ot ensure a regex match matches the entire
-            // string. We also replace any '*' or '?' characters with regex to match appropriate content.
-            // '*' matches 0 or more characters lazily and '?' matches 1 character.
-            var pattern = "^" + escaped.Replace(@"\?", ".").Replace(@"\*", ".*?") + "$";
-            var regex = new Regex(pattern, RegexOptions.Singleline, Constants.RegexMatchTimeout);
+                return descriptor.TypeName.StartsWith(lookupInfo.TypePattern.Substring(0, lookupInfo.TypePattern.Length - 1));
```
NTaylorMullen: Break this apart so it's easier to debug. Aka:

```C#
var foo = lookupInfo.TypePattern.Substring(0, lookupInfo.TypePattern.Length - 1);

return descriptor.TypeName.StartsWith(foo, StringComparer.Ordinal);
```

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -255,9 +256,7 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
             var lookupText = directiveDescriptor.DirectiveText;
             var lookupStrings = lookupText?.Split(new[] { ',' });
 
-            // Ensure that we have valid lookupStrings to work with. Valid formats are:
-            // "assemblyName"
-            // "typeName, assemblyName"
+            // Ensure that we have valid lookupStrings to work with. The valid format is "typeName, assemblyName"
```
NTaylorMullen: :+1:

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
```diff
@@ -149,7 +149,7 @@ public static TheoryData ResolveDirectiveDescriptorsInvalidTagHelperPrefixData
                             },
                             new TagHelperDirectiveDescriptor
                             {
-                                DirectiveText = "*Plain*, " + assemblyA,
+                                DirectiveText = "Microsoft.AspNet.Razor.Runtime.TagHelpers.Valid_P*, " + assemblyA,
```
NTaylorMullen: Do these set of tests (`*___*`) do anything unique compared to other variations here. Aka do we have other tests that appropriately cover `___*`?

NTaylorMullen:
:watc

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -236,16 +236,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
                 return false;
             }
 
-            // We need to escape the TypePattern so we can choose to only allow specific regex.
-            var escaped = Regex.Escape(lookupInfo.TypePattern);
+            if (lookupInfo.TypePattern.EndsWith("*"))
```
pranavkm: `StringComparison.Ordinal`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -236,16 +236,17 @@ private static bool MatchesLookupInfo(TagHelperDescriptor descriptor, LookupInfo
...
+            {
+                if (lookupInfo.TypePattern.Length == 1)
```
pranavkm: A comment for what this actually looks like?

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
ajaybhargavb: There are other tests in `Resolve_ReturnsPrefixedDescriptorsBasedOnDirectiveDescriptors` that cover `___*` scenario. There is no need to cover `*___*` because it is the same as testing `___*`.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorResolver.cs
pranavkm: Would inverting the check work: `lookupInfo.TypePattern.IndexOf(descriptor.TypeName, 0, lookupInfo.TypePattern - 1) == 0`

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
NTaylorMullen: Of course, I'm more or less just asking if we should delete these cases entirely instead of repurposing to further test `___*`.

ajaybhargavb: Ah. I already deleted some of those. Will check and remove if more exists.

ajaybhargavb:
Updated.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/TagHelperDescriptorResolverTest.cs
ajaybhargavb: Deleted a few. Some of those test the prefix variations. So leaving them there.

NTaylorMullen:
:shipit:

Issue #474, opened 2015-08-05
Different TagHelpers with the same element name, depending on scope


AlekseyMartynov:
Now we have `context.Items` (#238) and can compose hierarchies of tag helpers like:

    <parent>
        <child attr="value" />
    </parent>

It would be nice to be able to map an element name to different tag helpers, depending on a parent tag:

    <some-widget>
        <datasource attrs="for" some="widget" />
    </some>

    <other-widget>
        <datasource something="different" />
    </other-widget>

Probably possible in the context of #255: restrict by type name, not by element name.

dougbu:
@AlekseyMartynov do you have a scenario in mind that would benefit from this type of extension beyond what's described in #255?

NTaylorMullen:
@AlekseyMartynov @DamianEdwards, @madskristensen and I discussed an extension like this when talking about #255.  It's on our road map. At first I was thinking it'd be done as part of #255 but it honestly should be something separate; lets leave this issue open and use it to complete the feature.

Added feature and ready label, will wait for it to be picked up by triage to be assigned a milestone.

AlekseyMartynov:
@dougbu it's relevant for any tag helper with rich configuration, when you want to use nested tags. Consider:

Case 1 (repeating branded prefix is untidy):

    <brandedTag>
        <brandedConfig ... />
    </brandedTag>

Case 2 (using simple names globally is conflict-prone):
    
    <brandedTag>
        <config ... />
    </brandedTag>

AlekseyMartynov:
@NTaylorMullen sounds good!

Eilon:
We decided to add another property to the `HtmlTargetElement` attribute: `string ParentTag { get; set; }`

This would restrict a child tag to being that tag helper iff its containing parent is a tag helper with the specified tag name.

NTaylorMullen:
Hmmmm, so I'm working on this issue and it got me thinking about the required parent interaction. If we have:

```HTML
<tagHelperA>
  <p>
    <requiresParentA>
    </requiresParentA>
  </p>
</tagHelperA>
```

And `requiresParentA` says it has a required parent of `tagHelperA`. Does the nested `requiresParentA` tag fulfill its required parent aka have the `RequiresParentATagHelper` execute on it? 

I'm thinking as long as there's a parent (no matter how far up) that fulfills the parent tag then the `TagHelper` should apply. Thoughts?

Eilon:
[...]

PR #545, opened 2015-09-23
Add `ParentTag` to `[HtmlTargetElement]`.

NTaylorMullen:
- `ParentTag` allows `TagHelper`s to restrict where they apply based on their immediate parent tag.
- Changed the `TagHelperParseTreeRewriter` to understand non-`TagHelper` HTML elements to properly determine a parent tag when applying `TagHelperDescriptor.RequiredParent`. This change will also enable `[RestrictChildren]` to apply to more than just `TagHelper`s.
- Added tests to validate that partial and well formed tags properly discover `TagHelper`s. Also added tests that validate that descriptors are properly created based on `TagHelper` types.

#474

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlTargetElementAttribute.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlTargetElementAttribute.cs
@@ -79,5 +79,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// </para>
         /// </remarks>
         public TagStructure TagStructure { get; set; }
+
+        /// <summary>
+        /// The required HTML element name of the direct parent.
+        /// </summary>
+        /// <remarks>A <c>null</c> value indicates any HTML element name is appropriate.</remarks>
+        public string ParentTag { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -142,6 +142,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                         requiredAttributes: Enumerable.Empty<string>(),
                         allowedChildren: allowedChildren,
                         tagStructure: default(TagStructure),
+                        parentTag: null,
                         designTimeDescriptor: typeDesignTimeDescriptor)
                 };
             }
@@ -231,6 +232,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 attributeDescriptors,
                 requiredAttributes,
                 allowedChildren,
+                targetElementAttribute.ParentTag,
                 targetElementAttribute.TagStructure,
                 designTimeDescriptor);
         }
@@ -242,6 +244,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             IEnumerable<string> requiredAttributes,
             IEnumerable<string> allowedChildren,
+            string parentTag,
             TagStructure tagStructure,
             TagHelperDesignTimeDescriptor designTimeDescriptor)
         {
@@ -253,6 +256,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 Attributes = attributeDescriptors,
                 RequiredAttributes = requiredAttributes,
                 AllowedChildren = allowedChildren,
+                RequiredParent = parentTag,
                 TagStructure = tagStructure,
                 DesignTimeDescriptor = designTimeDescriptor
             };
@@ -286,7 +290,27 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 }
             }
 
-            return validTagName && validAttributeNames;
+            var validParentTagName = ValidateParentTagName(attribute.ParentTag, errorSink);
+
+            return validTagName && validAttributeNames && validParentTagName;
+        }
+
+        /// <summary>
+        /// Internal for unit testing.
+        /// </summary>
+        internal static bool ValidateParentTagName(string parentTag, ErrorSink errorSink)
+        {
+            return parentTag == null ||
+                TryValidateName(
+                    parentTag,
+                    Resources.FormatHtmlTargetElementAttribute_NameCannotBeNullOrWhitespace(
+                        Resources.TagHelperDescriptorFactory_ParentTag),
+                    characterErrorBuilder: (invalidCharacter) =>
+                        Resources.FormatHtmlTargetElementAttribute_InvalidName(
+                            Resources.TagHelperDescriptorFactory_ParentTag.ToLower(),
+                            parentTag,
+                            invalidCharacter),
+                    errorSink: errorSink);
         }
 
         private static bool ValidateName(
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Globalization;
 using System.Linq;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 using Microsoft.AspNet.Razor.Runtime.TagHelpers;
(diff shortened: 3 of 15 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlTargetElementAttribute.cs
```diff
@@ -79,5 +79,11 @@ public HtmlTargetElementAttribute(string tag)
         /// </para>
         /// </remarks>
         public TagStructure TagStructure { get; set; }
+
+        /// <summary>
+        /// The required HTML element name of the direct parent.
+        /// </summary>
+        /// <remarks>A <c>null</c> value indicates any HTML element name is appropriate.</remarks>
```
pranavkm: Do the null or whitespace checks apply to this thing? If so, should this comment reflect that?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -286,7 +290,27 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
                 }
             }
 
-            return validTagName && validAttributeNames;
+            var validParentTagName = ValidateParentTagName(attribute.ParentTag, errorSink);
+
+            return validTagName && validAttributeNames && validParentTagName;
```
pranavkm: Is this intentionally not short-circuited so invalid parent tag errors get added to error sink?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -15,16 +14,38 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
 {
     public class TagHelperParseTreeRewriter : ISyntaxTreeRewriter
     {
+        // From http://dev.w3.org/html5/spec/Overview.html#elements-0
+        private static readonly ISet<string> VoidElements = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
pranavkm: `HashSet<string> VoidElements`. No reason to make all its be methods be virtual calls.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -450,11 +504,23 @@ private void BuildCurrentlyTrackedBlock()
...
+            while (tagHelperTracker == null);
+
             // Track the original end tag so the editor knows where each piece of the TagHelperBlock lies
             // for formatting.
-            _trackerStack.Pop().Builder.SourceEndTag = endTag;
+            tagHelperTracker.Builder.SourceEndTag = endTag;
 
-            _currentTagHelperTracker = _trackerStack.Count > 0 ? _trackerStack.Peek() : null;
+            _currentTagHelperTracker =
+                _trackerStack.FirstOrDefault(tagBlockTracker => tagBlockTracker.IsTagHelper) as TagHelperBlockTracker;
```
pranavkm: Is this guaranteed to be a `TagHelperBlockTracker` if it `IsTagHelper` is true?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -492,7 +558,7 @@ private bool TryRecoverTagHelper(string tagName, Block endTag, RewritingContext
 
             foreach (var tracker in _trackerStack)
             {
-                if (tracker.Builder.TagName.Equals(tagName, StringComparison.OrdinalIgnoreCase))
+                if (tracker.TagName.Equals(tagName, StringComparison.OrdinalIgnoreCase) && tracker.IsTagHelper)
```
pranavkm: Consider inverting these checks because `IsTagHelper` is probably quicker

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
```diff
@@ -75,10 +79,20 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(string tagName, IEnumerab
             }
 
             var applicableDescriptors = ApplyRequiredAttributes(descriptors, attributeNames);
+            applicableDescriptors = ApplyRequiredParentTag(applicableDescriptors, parentTagName);
 
             return applicableDescriptors;
         }
 
+        private IEnumerable<TagHelperDescriptor> ApplyRequiredParentTag(
```
pranavkm: `ApplyRequiredParentTagFilter`?

pranavkm:
Does this need functional tests? Cases where you have a potential tag helper that is ignored because it does not have the right parent tag?

pranavkm:
:watch:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlTargetElementAttribute.cs
NTaylorMullen: Not sure what you mean. Like invalid characters? If so we don't mention such restrictions for `AllowedChildren` or `TagName`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: :+1: si

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
NTaylorMullen: Yes

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs

src/Microsoft.AspNet.Razor/TagHelpers/TagHelperDescriptorProvider.cs
NTaylorMullen: :+1:

NTaylorMullen:
@pranavkm I decided not to add code generation tests because we haven't touched any of the code generation pieces. Very similar to the `[RestrictChildren]` change (didn't add any code generation change).

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
pranavkm: In which would a cast work?

NTaylorMullen: The `as` is there in-case there is no `TagHelper` block, aka `FirstOrDefault` returns `null`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/HtmlTargetElementAttribute.cs
pranavkm: There's a couple of references to `string.IsNullOrWhitespace`. Do the checks apply to the value of this property? If so, the comment should say `A <c>null</c> or empty string ...`

NTaylorMullen: Oo, the `IsNullOrWhiteSpace` check is essentially just checking for whitespace in the characters. Prior to doing any character validation checks we validate if it's `null`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -286,7 +290,27 @@ internal static IEnumerable<string> GetCommaSeparatedValues(string text)
...
+        }
+
+        /// <summary>
+        /// Internal for unit testing.
```
pranavkm: IMO these comments are kinda pointless. With the exception of shared code, we exclusively use `internal` method for InternalsVisibleTo unit testing. We should only be calling out the opposite case (when `internal` is used for non-test cases)

NTaylorMullen: Do we do that anywhere else?

pranavkm:
:shipit:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
pranavkm: Casting should still work for null values.

Issue #543, opened 2015-09-23
Change [RestrictChildren] to work for non-TagHelper children.


NTaylorMullen:
Today if a user provides `[RestrictChildren]` any children specified MUST be a `TagHelper`. This decision was initially made so we didn't have to maintain HTML element scopes in the `TagHelperParserRewriter`. With the completion of #474 we will track non-`TagHelper` HTML elements making this much easier to accomplish.

NTaylorMullen:
d2a7a355cc9c1410112c85769588ad37d79a23f2

PR #547, opened 2015-09-24
Change `[RestrictChildren]` to allow non-`TagHelper` tags.

NTaylorMullen:
- Updated the `TagHelperParseTreeRewriter` loosen child restrictions to non-`TagHelper` HTML elements.
- Updated tests to showcase that non-`TagHelper` elements are allowed to be restricted.
- Added an additional test case to showcase sub-sub nesting of non-`TagHelper` restricted children.

#543

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
@@ -81,10 +81,10 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                         }
                         else
                         {
-                            TrackTagBlock(childBlock);
-
                             // Non-TagHelper tag.
-                            ValidateParentTagHelperAllowsPlainTag(childBlock, context.ErrorSink);
+                            ValidateParentAllowsPlainTag(childBlock, context.ErrorSink);
+
+                            TrackTagBlock(childBlock);
                         }
 
                         // If we get to here it means that we're a normal html tag.  No need to iterate any deeper into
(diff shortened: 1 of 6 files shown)


NTaylorMullen:
/cc @pranavkm @dougbu @ajaybhargavb 

**Note:** Would recommend appending `?w=0` when reviewing.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -335,10 +335,22 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
             return attributeNames;
         }
 
-        private void ValidateParentTagHelperAllowsContent(Span child, ErrorSink errorSink)
+        private bool ChildRestrictionsPresent()
```
pranavkm: `HasAllowedChildren`?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -358,31 +371,32 @@ private void ValidateParentTagHelperAllowsContent(Span child, ErrorSink errorSin
...
+            if (string.IsNullOrEmpty(tagName))
             {
-                var tagName = GetTagName(tagBlock);
+                Debug.Assert(tagBlock.Children.First() is Span);
 
-                // Treat partial tags such as '</' which have no tag names as content based errors.
-                if (string.IsNullOrEmpty(tagName))
-                {
-                    Debug.Assert(tagBlock.Children.First() is Span);
+                ValidateParentAllowsContent(tagBlock.Children.First() as Span, errorSink);
```
pranavkm: `ValidateParentAllowsContent((Span)tagBlock.Children.First(), errorSink);`

pranavkm:
:shipit:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs

PR #550, opened 2015-09-28
Fix #526 - use extension methods for AppendFormat

### LINKED ISSUE #526: Add extension methods on HtmlContentBuilder for String.Format style formatting

rynowak:
These changes make TagHelperContent with the IHtmlBuilder pattern, but
retain the signatures returning TagHelperContent.

Removed a bunch of tests that are no redundant with tests for the
extension methods in HttpAbstractions. Kept a few explemars in place to
ensure that the basic plumbing functions as desired.


NTaylorMullen:
:shipit:. We might want to consider eventually moving `TagHelperContent` HTML building methods to only return `IHtmlContentBuilder`. If we have a rich enough HTML content building surface chaining only `IHtmlContentBuilder` may suffice.

rynowak:
@NTaylorMullen - the reason I kept it as-is is that `IHtmlContentBuilder` and friends are in a different namespace

NTaylorMullen:
@rynowak oh duh, ya that was the reason. Ugh.

rynowak:
The other change we could make would be to define the extensions on `<T> where T : IHtmlContentBuilder`, but that's ugly.

NTaylorMullen:
Hmm, that honestly doesn't sound too bad compared to what we're doing here (duplicating everything)

PR #554, opened 2015-09-30
Refactor WriteAttribute \ AddHtmlAttribute

### LINKED ISSUE #177: Refactor WriteAttribute API surface.
Today we generate a single `WriteAttribute` method which contains values that are only used for instrumentation. The suggestion here is to instead have two methods

* `WriteAttribute` which does not carry any instrumentation information. I'll update this method's signature once I investigate.
* `WriteInstrumentedAttribute` which carries instrumentation information. This method will look identical to the current `WriteAttribute`. 

The primary motive for this method is to have minimal runtime effects when instrumentation (BrowserLink) is not enabled.

pranavkm:
Fixes #177

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -108,7 +108,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         {
             WriteStringLiteral(value.Value);
             WriteParameterSeparator();
-            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.InvariantCulture));
+            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.CurrentCulture));
 
             return this;
         }
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -106,11 +106,9 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
 
         public CSharpCodeWriter WriteLocationTaggedString(LocationTagged<string> value)
         {
-            WriteStartMethodInvocation("Tuple.Create");
             WriteStringLiteral(value.Value);
             WriteParameterSeparator();
-            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.CurrentCulture));
-            WriteEndMethodInvocation(false);
+            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.InvariantCulture));
 
             return this;
         }
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Microsoft.AspNet.Razor.Chunks;
@@ -26,6 +27,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         private readonly CodeGeneratorContext _context;
         private readonly IChunkVisitor _bodyVisitor;
         private readonly IChunkVisitor _literalBodyVisitor;
+        private readonly TagHelperAttributeCodeVisitor _attributeCodeVisitor;
         private readonly GeneratedTagHelperContext _tagHelperContext;
         private readonly bool _designTimeMode;
 
@@ -63,6 +65,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
             _designTimeMode = context.Host.DesignTimeMode;
 
             _literalBodyVisitor = new CSharpLiteralCodeVisitor(this, writer, context);
+            _attributeCodeVisitor = new TagHelperAttributeCodeVisitor(writer, context);
             AttributeValueCodeRenderer = new TagHelperAttributeValueCodeRenderer();
         }
 
@@ -467,15 +470,32 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     // Dynamic attribute value should be run through the conditional attribute removal system. It's
                     // unbound and contains C#.
 
+                    // TagHelper attribute rendering is buffered by default. We do not want to write to the current
+                    // writer.
+                    var currentTargetWriter = _context.TargetWriterName;
+                    var currentWriteAttributeMethodName = _context.Host.GeneratedClassContext.WriteAttributeValueMethodName;
+                    _context.TargetWriterName = null;
+
+                    Debug.Assert(attributeValueChunk is ParentChunk);
+                    var children = ((ParentChunk)attributeValueChunk).Children;
+                    var attributeCount = children.Count(c => c is DynamicCodeAttributeChunk || c is LiteralCodeAttributeChunk);
+
                     _writer
-                        .WriteStartMethodInvocation(_tagHelperContext.AddHtmlAttributeValuesMethodName)
+                        .WriteStartMethodInvocation(_tagHelperContext.BeginAddHtmlAttributeValuesMethodName)
+                        .Write(ExecutionContextVariableName)
+                        .WriteParameterSeparator()
                         .WriteStringLiteral(attributeName)
                         .WriteParameterSeparator()
-                        .Write(ExecutionContextVariableName);
+                        .Write(attributeCount.ToString(CultureInfo.InvariantCulture))
+                        .WriteEndMethodInvocation();
 
-                    _bodyVisitor.Accept(attributeValueChunk);
+                    _attributeCodeVisitor.Accept(attributeValueChunk);
 
-                    _writer.WriteEndMethodInvocation();
+                    _writer.WriteMethodInvocation(
+                        _tagHelperContext.EndAddHtmlAttributeValuesMethodName,
+                        ExecutionContextVariableName);
+
+                    _context.TargetWriterName = currentTargetWriter;
                 }
                 else
                 {
@@ -721,6 +741,19 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     return Context.Host.GeneratedClassContext.WriteLiteralToMethodName;
                 }
             }
+        }
+
+        private class TagHelperAttributeCodeVisitor : CSharpCodeVisitor
+        {
+            public TagHelperAttributeCodeVisitor(
+                CSharpCodeWriter writer,
+                CodeGeneratorContext context)
+                : base(writer, context)
+            {
+            }
+
+            protected override string WriteAttributeValueMethodName =>
+                Context.Host.GeneratedClassContext.GeneratedTagHelperContext.AddHtmlAttributeValueMethodName;
         }
     }
 }
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
@@ -207,9 +207,9 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
 
                 Writer
                     .WriteParameterSeparator()
-                    .Write(chunk.Start.AbsoluteIndex.ToString(CultureInfo.InvariantCulture))
+                    .Write(chunk.Start.AbsoluteIndex.ToString(CultureInfo.CurrentCulture))
                     .WriteParameterSeparator()
-                    .Write(chunk.Association.Length.ToString(CultureInfo.InvariantCulture))
+                    .Write(chunk.Association.Length.ToString(CultureInfo.CurrentCulture))
                     .WriteParameterSeparator()
                     .WriteBooleanLiteral(value: false)
                     .WriteEndMethodInvocation();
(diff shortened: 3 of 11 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -106,11 +106,9 @@ public CSharpCodeWriter WriteStartNewObject(string typeName)
 
         public CSharpCodeWriter WriteLocationTaggedString(LocationTagged<string> value)
         {
-            WriteStartMethodInvocation("Tuple.Create");
             WriteStringLiteral(value.Value);
             WriteParameterSeparator();
-            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.CurrentCulture));
-            WriteEndMethodInvocation(false);
+            Write(value.Location.AbsoluteIndex.ToString(CultureInfo.InvariantCulture));
```
dougbu: Why change the culture here?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -467,15 +470,32 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+
                     _writer
-                        .WriteStartMethodInvocation(_tagHelperContext.AddHtmlAttributeValuesMethodName)
+                        .WriteStartMethodInvocation(_tagHelperContext.BeginAddHtmlAttributeValuesMethodName)
+                        .Write(ExecutionContextVariableName)
+                        .WriteParameterSeparator()
                         .WriteStringLiteral(attributeName)
                         .WriteParameterSeparator()
-                        .Write(ExecutionContextVariableName);
+                        .Write(attributeCount.ToString(CultureInfo.InvariantCulture))
```
dougbu: Doubt this should be invariant either.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
pranavkm: No reason for generated code to rely on user culture.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -722,5 +742,18 @@ protected override string WriteToMethodName
...
+        {
+            public TagHelperAttributeCodeVisitor(
+                CSharpCodeWriter writer,
+                CodeGeneratorContext context)
+                : base(writer, context)
+            {
+            }
+
+            protected override string WriteAttributeValueMethodName =>
+                Context.Host.GeneratedClassContext.GeneratedTagHelperContext.AddHtmlAttributeValueMethodName;
```
dougbu: Why not a get-only auto-property set in the constructor?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -467,15 +470,32 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                    _context.TargetWriterName = null;
+
+                    Debug.Assert(attributeValueChunk is ParentChunk);
+                    var children = ((ParentChunk)attributeValueChunk).Children;
+                    var attributeCount = children.Count(c => c is DynamicCodeAttributeChunk || c is LiteralCodeAttributeChunk);
+
                     _writer
-                        .WriteStartMethodInvocation(_tagHelperContext.AddHtmlAttributeValuesMethodName)
+                        .WriteStartMethodInvocation(_tagHelperContext.BeginAddHtmlAttributeValuesMethodName)
+                        .Write(ExecutionContextVariableName)
```
NTaylorMullen: Curious: We typically have context parameters first followed by more specific variables?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -467,15 +470,32 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
 
+                    // TagHelper attribute rendering is buffered by default. We do not want to write to the current
+                    // writer.
+                    var currentTargetWriter = _context.TargetWriterName;
+                    var currentWriteAttributeMethodName = _context.Host.GeneratedClassContext.WriteAttributeValueMethodName;
+                    _context.TargetWriterName = null;
+
+                    Debug.Assert(attributeValueChunk is ParentChunk);
+                    var children = ((ParentChunk)attributeValueChunk).Children;
+                    var attributeCount = children.Count(c => c is DynamicCodeAttributeChunk || c is LiteralCodeAttributeChunk);
```
NTaylorMullen: Is this just to protect against Razor comments?

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -174,50 +181,60 @@ protected override void Visit(DynamicCodeAttributeChunk chunk)
...
-                       .Write(chunk.Start.AbsoluteIndex.ToString(CultureInfo.CurrentCulture))
-                       .WriteEndMethodInvocation(endLine: false)
-                       .WriteParameterSeparator()
-                       .WriteBooleanLiteral(value: false)
-                       .WriteEndMethodInvocation(endLine: false);
+                Writer
+                    .WriteParameterSeparator()
+                    .Write(chunk.Start.AbsoluteIndex.ToString(CultureInfo.InvariantCulture))
+                    .WriteParameterSeparator()
+                    .Write(chunk.Association.Length.ToString(CultureInfo.InvariantCulture))
```
NTaylorMullen: Shouldn't need the length at runtime.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -174,50 +181,60 @@ protected override void Visit(DynamicCodeAttributeChunk chunk)
...
-                       .WriteEndMethodInvocation(endLine: false)
-                       .WriteParameterSeparator()
-                       .WriteBooleanLiteral(false)
-                       .WriteEndMethodInvocation(false);
+                Writer
+                    .WriteEndMethodInvocation(false)
+                    .WriteParameterSeparator()
+                    .Write(chunk.Start.AbsoluteIndex.ToString(CultureInfo.InvariantCulture))
+                    .WriteParameterSeparator()
+                    .Write(chunk.Association.Length.ToString(CultureInfo.InvariantCulture))
```
NTaylorMullen: Ditto

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -255,19 +280,24 @@ protected override void Visit(LiteralCodeAttributeChunk chunk)
...
-                       .Write(chunk.ValueLocation.AbsoluteIndex.ToString(CultureInfo.CurrentCulture))
-                       .WriteEndMethodInvocation(false)
-                       .WriteParameterSeparator()
-                       .WriteBooleanLiteral(false)
-                       .WriteEndMethodInvocation(false);
+                Writer
+                    .WriteParameterSeparator()
+                    .Write(chunk.ValueLocation.AbsoluteIndex.ToString(CultureInfo.InvariantCulture))
+                    .WriteParameterSeparator()
+                    .Write(chunk.Association.Length.ToString(CultureInfo.InvariantCulture))
```
NTaylorMullen: Dittttooo

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -255,19 +280,24 @@ protected override void Visit(LiteralCodeAttributeChunk chunk)
...
             else
             {
-                Writer.WriteLocationTaggedString(chunk.Value)
-                       .WriteParameterSeparator()
-                       .WriteBooleanLiteral(true)
-                       .WriteEndMethodInvocation(false);
+                Writer
+                    .WriteLocationTaggedString(chunk.Value)
+                    .WriteParameterSeparator()
+                    .Write(chunk.Association.Length.ToString(CultureInfo.InvariantCulture))
```
NTaylorMullen: uno mas

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -467,15 +470,32 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
                     // Dynamic attribute value should be run through the conditional attribute removal system. It's
                     // unbound and contains C#.
 
+                    // TagHelper attribute rendering is buffered by default. We do not want to write to the current
```
NTaylorMullen: I'm pretty sure we don't need to nuke the target writer here. I missing something?

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
dougbu: Really starting to think you're mixing an unrelated change into this PR. If there's a culture-related issue, file it separately.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: As mentioned later, this doesn't fit in this PR.

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
pranavkm: Follows the pattern from type above.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
pranavkm: Yup. These looked like the only types that produced `WriteAttributeValues`

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
pranavkm: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.PageExecutionInstrumentation.Interfaces/IPageExecutionContext.cs#L18

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
pranavkm: See above.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
pranavkm: See above.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
pranavkm: See above

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
pranavkm: We're specifically overriding the `WriteAttribute` method in `TagHelperAttributeCodeVisitor`. Don't want it to generate `WriteAttributeValueTo` methods since it's all buffered to the `StringCollectionTextWriter`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
pranavkm: Leaving as is. It's too small a change and has no user impact.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
pranavkm: Yeah. In general most of ours APIs tend to have least variable -> most variable going left to right. In this case the context changes the least (for a single attribute). Plus it also makes the start and end look a bit symmetrical

```C#
BeginAddHtmlAttributeValues(context, "attrname", 1);
EndAddHtmlAttributeValues(context);
```

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Either this change has no user impact and isn't worth making or there's a scenario where the user's culture could result in compilation failures and we should be tracking it as an issue, making sure it the problem doesn't exist elsewhere, testing, ...

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Gots it :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -467,15 +470,32 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
...
+                    // writer.
+                    var currentTargetWriter = _context.TargetWriterName;
```
NTaylorMullen: Move this down to where it has effect where we do the `_attributeCodeVisitor.Accept(attributeValueChunk);`

NTaylorMullen:
To fulfill the purpose of the issue we should be conditionally generating the instrumented bits (value length, value offset etc.) based on instrumentation.

:watch:

pranavkm:
@NTaylorMullen The primary issue with conditionally generating these values is that we'll have to factor it in the page cache invalidation which is more work than it needs to be. Right now, we always generate it instrumented with the assumption that having no-op instrumentation methods and parameters is very cheap.

I've updated the work item to reflect this design choice.

NTaylorMullen:
:shipit:

Issue #555, opened 2015-09-30
Editing a tag helper in a view throws exception causing VS to crash


kichalla:
Example:
After typing in `<a asp-action="` especially if its the **last** tag in the .cshtml file there is an exception being thrown which causes VS to crash.

This was with Canary build.

@NTaylorMullen @ajaybhargavb have already investigated the cause and this issue is to track it

NTaylorMullen:
The tag has to be the last tag in the file correct?

kichalla:
Yup, I mentioned it above :-)

NTaylorMullen:
You said *especially* so wanted to verify that it was a requirement :smile:

ajaybhargavb:
bc103fcaa404c3fabd902259e170f18f2f183738

PR #556, opened 2015-09-30
Fix: Index out of bounds for incomplete attribute value

ajaybhargavb:
#555 
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
@@ -215,34 +215,29 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
 
                     // The coming symbols will either be a quote or value (in the case that the value is unquoted).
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -215,34 +215,28 @@ public static class TagHelperBlockRewriter
...
+                        symbolStartLocation = htmlSymbols[i].Start;
 
                         // If there's a start quote then there must be an end quote to be valid, skip it.
                         symbolOffset = 1;
-
-                        i = valueStartIndex - 1;
                     }
                     else
                     {
+                        // No quote found. Go back.
```
NTaylorMullen: Clarify comment. This `i--` ensures we don't skip past the value after the `=`. Such as: `<input checked=true>` would result in us skipping the `true` symbol if we did not have this `i--`.

NTaylorMullen:
:shipit:

Eilon:
Please send mail ASAP for approval to `release` (and then rebase this on `release`, then merge back to `dev`).

ajaybhargavb:
bc103fcaa404c3fabd902259e170f18f2f183738

Issue #553, opened 2015-09-30
`TreeStructureChanged` marked as `true` when whitespace is added to a page with `TagHelper`s


NTaylorMullen:
The Razor editor uses the [TreeStructureChanged](https://github.com/aspnet/Razor/blob/0a1edbe3f96bcf3807ba8483d47a6ecc00ee0562/src/Microsoft.AspNet.Razor/DocumentParseCompleteEventArgs.cs#L18) property to determine if a full-parse should occur. Today, if there's a `TagHelper` on a page hitting *Enter* results in its value being set to `true`. This winds up causing unneeded parsing of the document.

Eilon:
So this is a perf issue?

NTaylorMullen:
On the editor side yes. I didn't want to mark it as perf because for Razor it's more of a bug that alleviates the editors perf issue :smile:

NTaylorMullen:
deaf2dc828b549f425a54d484d94afc9f8770635

PR #572, opened 2015-10-15
Enable `TagHelper`s to properly flatten to enable partial parsing.

NTaylorMullen:
- `TagHelper`s used to not flatten correctly resulting in inconsistent start locations for `SyntaxTreeNode`s following/within `TagHelper`s.
- This change indirectly corrected bad indentation that existed in generated C#.
- Added a test to validate `TreesAreDifferent` returns expected behavior when modifying content inside of `TagHelper`.
- Added `TagHelperBlock` `Flatten` test.

#553


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperBlockBuilder.cs
```diff
@@ -134,6 +134,6 @@ public override void Reset()
         /// <summary>
         /// The starting <see cref="SourceLocation"/> of the tag helper.
         /// </summary>
-        public SourceLocation Start { get; private set; }
+        public SourceLocation Start { get; set; }
```
ajaybhargavb: Is this change needed only for testing? If so shouldn't this just be `internal`?

NTaylorMullen: In this PR we need it accessible for testing yes. However, the builder was a bit odd to begin with only having a `private set`. All of our other builds have public get/set methods, thought I'd just normalize this property since I had to touch it anyways.

ajaybhargavb: Got it.

ajaybhargavb:
:shipit:

dougbu:
:shipit:

Issue #552, opened 2015-09-29
Expose `FilePath` on `MappingLocation` to let Razor editor know mapping origination.


NTaylorMullen:
Today the Razor editor has to jump through hoops to determine if a projected [mapping location](https://github.com/aspnet/Razor/blob/0a1edbe3f96bcf3807ba8483d47a6ecc00ee0562/src/Microsoft.AspNet.Razor/CodeGenerators/MappingLocation.cs) originated from the current file or from a `_ViewImports.cshtml`. Due to their roundabout way of determining a mappings original location it can occasionally be wrong. We should expose what file created the mapping location so tooling can respond accordingly.

/cc @pranavkm @ToddGrun

Eilon:
Do we need to have a matching tooling bug to consume this new property?

NTaylorMullen:
@ToddGrun already has one :smile:

Eilon:
Cool.

NTaylorMullen:
be324b220a9d3adb98b7dec2c1eaf5a0924c3a8f

PR #573, opened 2015-10-16
Expose `FilePath` on `MappingLocation`.

NTaylorMullen:
- We had the `FilePath` information available to us but we weren't exposing it on the `MappingLocation` context object.
- By exposing the `FilePath` the Razor editor can 100% know where a `MappingLocation` is resolved from. If `null` the location is deemed to be the current source/generated file.

#552

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/MappingLocation.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/MappingLocation.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
bump

src/Microsoft.AspNet.Razor/CodeGenerators/MappingLocation.cs
```diff
@@ -45,6 +50,7 @@ public override bool Equals(object obj)
         public override int GetHashCode()
         {
             var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(FilePath);
```
pranavkm: `hashCodeCombiner.Add(FilePath, StringComparer.Ordinal)`

pranavkm:
:shipit:

Issue #569, opened 2015-10-10
Compilation error within tag attribute shows generated code instead of markup


Eilon:
Create this Razor page (nothing else in the app really matters):

```html
<a href="@doesNotExist">Hi!</a>
```

You get this error:

```
An error occurred during the compilation of a resource required to process this request. Please review the following specific error details and modify your source code appropriately.

Generated Code

The name 'doesNotExist123' does not exist in the current context

84.            Tuple.Create(Tuple.Create("", 11), Tuple.Create<System.Object, System.Int32>(doesNotExist123, 11), false));
```

Note that the error shows the *generated* code instead of the Razor markup.

This is using Razor 4.0.0-beta8-15751 (should be the newest beta8 release build).

Eilon:
BTW other errors show up fine. E.g. a random `@foo` in the page shows the markup and not the generated code.

NTaylorMullen:
2d4092ca32f38db18b700d1df0d2c4cb9a7ad9f2

PR #574, opened 2015-10-16
Add line pragmas to dynamic attributes.

NTaylorMullen:
- This enables debugging and proper error reporting for dynamic attributes. Without the pragmas errors would showcase generated Razor C# instead of their corresponding .cshtml files.
- Did not have to do any `TagHelper` specific changes since they utilize the same core attribute rendering logic.
- Updated generated C# files for tests.

#569


pranavkm:
:shipit:

Issue #578, opened 2015-10-19
API namespace cleanup for tag helpers


Eilon:
From a meeting with @dougbu @rynowak @DamianEdwards

Move what’s in the `Microsoft.AspNet.Razor.TagHelpers` namespace to something like 
`Microsoft.AspNet.Razor.Compilation.TagHelpers`.

And move what's in the `Microsoft.AspNet.Razor.Runtime.TagHelpers` namespace to `Microsoft.AspNet.Razor.TagHelpers`. An alternative that we considered was to rename it to `Microsoft.AspNet.Html.TagHelpers`, because it's not really Razor-specific, but meh.

NTaylorMullen:
bd2a98e5ecb7afe7ca621b01af68c03e43e29669
b1ad14fd46c50d6613a5d9dc691cefbe21516e0e
https://github.com/aspnet/Mvc/commit/0e254706607e7cccef11dc50d6b69e746c5827c3

dougbu:
aspnet/Mvc@0e25470 leaves references to `Microsoft.AspNet.Razor.Runtime.TagHelpers` in the `LinkTagHelper` and `ScriptTagHelper`. Either unused `using`s were not removed or this issue should be reopened and the dangling type moved to `Microsoft.AspNet.Razor.TagHelpers`.

Who reviewed that commit BTW? I don't remember a PR for it.

NTaylorMullen:
It was a reaction. Shouldn't need a review. As for them having `Runtime.TagHelper`s. That's due to them creating new `DefaultTagHelperContent`'s

dougbu:
If they need to do that, others will as well. `DefaultTagHelperContent` is not in the correct namespace yet.

NTaylorMullen:
Moved `DefaultTagHelperContent` to `Razor.TagHelpers`

PR #579, opened 2015-10-20
TagHelper API Cleanup

NTaylorMullen:
- Changed `Microsoft.AspNet.Razor.TagHelpers` to `Microsoft.AspNet.Razor.Compilation.TagHelpers`.
- Changed `Microsoft.AspNet.Razor.Runtime.TagHelpers` to `Microsoft.AspNet.Razor.TagHelpers`.
- Updated folder structure to reflect new namespaces.
- Updated generated code files to reflect new runtime type namespaces.

#578


NTaylorMullen:
/cc @Eilon @rynowak @dougbu @DamianEdwards 

If all looks good i'll do the reaction to MVC/RazorTooling etc.

Eilon:
Looks :shipit: to me, but would like at least one of @dougbu and @rynowak to confirm this matches their expectations.

dougbu:
Looks like doc comments need to change:
```
Invalid documentation syntax in src\Microsoft.AspNet.Razor\bin\Debug\dnx451\Microsoft.AspNet.Razor.xml
```

src/Microsoft.AspNet.Razor/Chunks/AddTagHelperChunk.cs
```diff
@@ -4,12 +4,12 @@
 namespace Microsoft.AspNet.Razor.Chunks
 {
     /// <summary>
-    /// A <see cref="Chunk"/> used to look up <see cref="TagHelpers.TagHelperDescriptor"/>s.
+    /// A <see cref="Chunk"/> used to look up <see cref="Compilation.TagHelperDescriptor"/>s.
```
dougbu: `Compilation.TagHelpers.TagHelperDescriptor`

dougbu:
Why are a few classes e.g. `RuntimePropertyInfo` and `XMLDocumentationProvider` (hmm, shouldn't that be `XmlDocumentationProvider`?) still in the `Microsoft.AspNet.Razor.Runtime` namespace? Should be the case that nothing except resources are left there.

dougbu:
Suggest we follow up by moving infrastructure classes that tag helper authors don't need into `...TagHelpers.Internal`. This would include e.g. `DefaultTagHelperContent`, `TagHelperRunner` and `TagHelperDescriptorResolver`. Tag helper authors should only need the attributes, `TagHelper`, and what they reference.

dougbu:
;watch:

NTaylorMullen:
@dougbu The `RuntimePropertyInfo` type classes aren't `TagHelper` specific. Thoughts on moving them to `Microsoft.AspNet.Razor.Runtime.Internal`?

dougbu:
> The `RuntimePropertyInfo` type classes aren't `TagHelper` specific.

I was looking at an inconsistency. Some of these `...Info` classes moved and a few did not.

In any case, while these classes have no inherent ties to `TagHelper`s, aren't they used only in that context?

> moving them to `Microsoft.AspNet.Razor.Runtime.Internal`?

I suggested that as a follow-up. It's a fine idea now as well but I'd prefer to cover everything that `TagHelper` authors never need (e.g. the factories) in one pass.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
```diff
@@ -93,11 +97,15 @@ protected override void Visit(TagHelperChunk chunk)
             {
                 _foundTagHelpers = true;
 
-                if (!ImportedUsings.Contains(TagHelpersRuntimeNamespace))
+                foreach (var tagHelperRuntimeNamespace in TagHelpersRuntimeNamespaces)
                 {
-                    // If we find TagHelpers then we need to add the TagHelper runtime namespace to our list of usings.
-                    Writer.WriteUsing(TagHelpersRuntimeNamespace);
-                    ImportedUsings.Add(TagHelpersRuntimeNamespace);
+                    if (!ImportedUsings.Contains(tagHelperRuntimeNamespace))
```
dougbu: Drive-by suggestion but could `ImportedUsings` be a `HashSet<string>`? This would then simplify to
``` c#
if (ImportedUsings.Add(tagHelperRuntimeNamespace))
{
    Writer.WriteUsing(tagHelperRuntimeNamespace);
}
```

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -1,14 +1,13 @@
...
-using System;
 using System.Globalization;
 using System.IO;
-using Microsoft.AspNet.Testing;
+using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.Extensions.WebEncoders.Testing;
 using Xunit;
 
-namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
+namespace Microsoft.AspNet.Razor.Runtime.TagHelpers.Internal
```
dougbu: Looks as if test classes are in the correct namespaces but not the correct directories.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
```diff
@@ -11,7 +11,11 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
 {
     public class CSharpUsingVisitor : CodeVisitor<CSharpCodeWriter>
     {
-        private const string TagHelpersRuntimeNamespace = "Microsoft.AspNet.Razor.Runtime.TagHelpers";
+        private static readonly string[] TagHelpersRuntimeNamespaces = new[]
+        {
+            "Microsoft.AspNet.Razor.TagHelpers",
+            "Microsoft.AspNet.Razor.Runtime.TagHelpers.Internal"
```
dougbu: Why does the generated code need the `Internal` namespace? Does this make the types show up in IntelliSense when the user is in C# mode? I expected more of the explicit type qualification used elsewhere in the generated code.

NTaylorMullen: It's identical to what we had before. We need the `Internal` namespace to enable access to the variety of `TagHelper` enabling types such as `TagHelperExecutionContext`, `TagHelperRunner` etc. We could at some point remove these namespaces and generate all the required fields using their fully qualified names.

test/Microsoft.AspNet.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
NTaylorMullen: Done, for some reason git diff  doesn't show unless you look at all all file changes.

NTaylorMullen:
Updated.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpUsingVisitor.cs
```diff
@@ -89,15 +93,23 @@ protected override void Visit(UsingChunk chunk)
 
         protected override void Visit(TagHelperChunk chunk)
         {
+            if (Context.Host.DesignTimeMode)
+            {
+                return;
+            }
```
dougbu: :clap:

dougbu:
Yah know, well I just thought, and I realize you're going to absolutely love this... Could now use `Microsoft.AspNet.Razor.Runtime.TagHelpers` where you've got `Microsoft.AspNet.Razor.Runtime.TagHelpers.Internal`. Ignore me if you're thinking of killing me.

dougbu:
:shipit: after fixing up an MVC fix-up PR.

Issue #562, opened 2015-10-02
Specifying `RestrictChildren` and empty `HtmlTargetElement` results in an error


pranavkm:
```
[RestrictChildren("div")]
[HtmlTargetElement]
public class MyTagHelper : TagHelper
{

}
```

```HTML
<p>
    <my>
        <div></div>
    </my>
</p>
```

Expected: This should render MyTagHelper correctly
Actual: `The <my> tag is not allowed by parent <p> tag helper. Only child tags with name(s) 'div' are allowed.`

NTaylorMullen:
By design. By specifying `[HtmlTargetElement]` you're saying that the `TagHelper` applies to every tag.

pranavkm:
We need to update the docs on that attribute to indicate it's a catch all tag helper.

danroth27:
The doc comments for `HtmlTargetElementAttribute` could be improved. The default constructor summary should specify what `*` means - you shouldn't have to read the remarks.

Also, the error message doesn't specify clearly which tag helper is causing the issue. What if you have multiple tag helpers targeting `<p>`?

dougbu:
1c7af0d

PR #584, opened 2015-10-23
Improve `[HtmlTargetElement]` doc comments

dougbu:
/cc @pranavkm 

I did not "improve" the error message for the #562 case because the tag helper type name is usually not helpful in general and is never helpful in an error page. Am open to further discussion on that point but we need to keep in mind the tag helper author isn't always the same as the person writing a `.cshtml` file or using an MVC site.


pranavkm:
Yeah - I guess tag helper authors might learn quickly once they understand it targets all elements.

:shipit:

dougbu:
1c7af0d

Issue #561, opened 2015-10-02
Consider not requiring quotes around @tagHelperPrefix values


pranavkm:
The usage for this is `@tagHelperPrefix "th"` where the value cannot have whitespaces. Unless I'm missing something obvious, there isn't a whole lot of use for the extra quotes.

NTaylorMullen:
If we decide to do this, we should do the same for `@addTagHelper` and `@removeTagHelper`

danroth27:
@DamianEdwards

Eilon:
Please post an announcement when done. Also need to consider if any tooling changes are needed (including templates).

NTaylorMullen:
Yup, will need changes across the board to react to this change. Will breakk evvverything :imp:

NTaylorMullen:
013f3a27afc6ecdc58a8c246a67cbc5b2c531ce3
0c953155001e49d92a15dbce3db1a9ea98d0697f
https://github.com/aspnet/Mvc/commit/d564a4a96b891bea1f321824b9ed7c7c77c3cf85
https://github.com/aspnet/Docs/commit/031406a2f1b7c956ef4644beecb7f5346154c61c
https://github.com/aspnet/Templates/commit/94946892c0ae93cd93acfaf1144a57a66bdb20ad

dougbu:
and aspnet/MusicStore@dddc6a2

Kagamine:
I think this should be filed on aspnet/Announcements

NTaylorMullen:
@Kagamine it is :smile: https://github.com/aspnet/Announcements/issues/103

PR #591, opened 2015-10-31
Remove quote requirement for `TagHelper` directives.

NTaylorMullen:
- `@addTagHelper`, `@removeTagHelper` and `@tagHelperPrefix` values are all written without quotes now.
- Updated the `ChunkGenerator`s to be in-line with other `ChunkGenerator`s in the system now that they don't have vastly different values than what their corresponding `Span`s have.
- Split `AddOrRemoveTagHelperChunkGenerator` into two smaller `ChunkGenerator`s to stay consistent with other generators.
- Created a unified `TagHelperDirectiveChunkGenerator` so we can unify parsing logic when dealing with `TagHelperDirective`s. A prime example of this is the work done in `TagHelperDirectiveSpanVisitor`.
- Removed error cases introduced by having quotes around directives.
- Updated code generation logic to not generate line pragmas (they should never have been generated to start with) and more specifically only map to the inner string that's generated. Without this change directive mappings would be offeset by a generated quote.

*Update tests to not require quotes for TH directives.*
- Updated .cshtml files to not provide quotes for the various `TagHelper` directives.
- Updated `TagHelper` directive parsing tests. This also involved removing several tests that expected errors which no longer occur since we don't have to ensure quotes are balanced.
- Updated line mappings which significantly changed due to us no longer generating line pragmas at design time for `TagHelper` directives.

**Note:** I did not document the public API surface for core Razor bits. We decided `///` documentation was not important for core Razor due to its lack of direct usage by users.

#561


src/Microsoft.AspNet.Razor/Chunks/Generators/TagHelperDirectiveChunkGenerator.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Razor.Parser.SyntaxTree;
+
+namespace Microsoft.AspNet.Razor.Chunks.Generators
+{
+    public abstract class TagHelperDirectiveChunkGenerator : SpanChunkGenerator
```
NTaylorMullen: Instead of having this common class we could duplicate the `GetDirectiveText` on each of the inheriting `ChunkGenerator`s but it'd still require an interface to avoid ugliness in `TagHelperDirectiveSpanVisitor`. I felt like this was the cleanest approach.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -82,50 +83,42 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(Block root)
 
         public override void VisitSpan(Span span)
```
NTaylorMullen: All of the changes in this method are a reaction to the new `TagHelper` directive `ChunkGenerator` formats of not holding onto the `LookupText`. Was able to significantly simplify the code in this method as a result. In the end does the same thing with the exception of `.LookupText` => `GetDirectiveText(span)`.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/AddTagHelperDirective.cs
```diff
@@ -10,13 +10,7 @@ private void @__RazorDesignTimeHelpers__()
         {
             #pragma warning disable 219
             string __tagHelperDirectiveSyntaxHelper = null;
-            __tagHelperDirectiveSyntaxHelper = 
-#line 1 "AddTagHelperDirective.cshtml"
```
NTaylorMullen: Line pragmas don't gain us any value during design time. They enabled users to hit breakpoints on `@addTagHelper` etc. directives which is very misleading because they're directives, not in-place methods or instructions. See description for more detail.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -82,50 +83,42 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(Block root)
...
+                directiveType = TagHelperDirectiveType.TagHelperPrefix;
+            }
+            else
+            {
+                // Not a chunk generator that we're interested in.
+                return;
             }
-        }
 
-        private static SourceLocation GetSubTextSourceLocation(Span span, string text)
```
NTaylorMullen: Inlined this method because it was no longer called multiple times.

NTaylorMullen:
/cc @dougbu @ajaybhargavb @pranavkm 

I'd recommend initially looking at the src based commit to avoid the masssses of test changes :smile:

src/Microsoft.AspNet.Razor/Chunks/Generators/TagHelperDirectiveChunkGenerator.cs
dougbu: Just inline the one line. Tiny methods like this add very little value and the surrounding class is dubious. Further `TagHelperPrefixDirectiveChunkGenerator` was completely separate before without everything falling apart.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -289,7 +281,8 @@ protected void BaseTypeDirective(string noTypeNameError, Func<string, SpanChunkG
             Output(SpanKind.Code, AcceptedCharacters.AnyExceptNewline);
         }
 
-        private void TagHelperDirective(string keyword, Func<string, ISpanChunkGenerator> buildChunkGenerator)
+        private void TagHelperDirective<TDirectiveChunkGenerator>(string keyword)
+            where TDirectiveChunkGenerator : TagHelperDirectiveChunkGenerator, new()
```
dougbu: No need for a generic method. Pass in the `SpanChunkGenerator` instance.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -82,50 +83,42 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(Block root)
...
-        {
-            var startOffset = span.Content.IndexOf(text);
+            var directiveText = ((TagHelperDirectiveChunkGenerator)span.ChunkGenerator).GetDirectiveText(span);
```
dougbu: Bizarre this code reaches past the `GenerateChunk()` methods to perform this one-liner. Where will `GenerateChunk()` actually be called?

dougbu: Also the `(TagHelperDirectiveChunkGenerator)` cast is fragile. This class assumes `AddTagHelperChunkGenerator` and so on share this common base. That's not information that should need to be shared.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
NTaylorMullen: > Bizarre this code reaches past the GenerateChunk() methods to perform this one-liner. Where will GenerateChunk() actually be called?
`GenerateChunk`s occur after the parse tree has been rewritten to have `TagHelper`s. This stage is all about finding the directives that are then responsible for adding `TagHelperDescriptor`s that are used to then re-write the syntax tree.


> Also the (TagHelperDirectiveChunkGenerator) cast is fragile. This class assumes AddTagHelperChunkGenerator and so on share this common base. That's not information that should need to be shared.
It definitely does. I could add an `if (.. is TagHelperDirectiveChunkGenerator) { return; }` up top if you don't like the cast.

src/Microsoft.AspNet.Razor/Chunks/Generators/TagHelperDirectiveChunkGenerator.cs
NTaylorMullen: If I inlined the method without an interface I'll have to check and then cast each of the `TagHelperDirectiveChunkGenerator`s to call the method. Definitely don't want to do that.

dougbu: The method doesn't add value especially since the `Span` is available. This code doesn't need these complications and an `interface` is definitely not required.

NTaylorMullen: Why duplicate the logic in each of the 3 chunk generators? Then ontop of that have to do the following:

```C#
TagHelperDirectiveType directiveType;
string directiveText;
if (span.ChunkGenerator is AddTagHelperChunkGenerator)
{
    directiveType = TagHelperDirectiveType.AddTagHelper;
    directiveText = ((AddTagHelperChunkGenerator)span.ChunkGenerator).GetDirectiveText(span);
}
else if (span.ChunkGenerator is RemoveTagHelperChunkGenerator)
{
    directiveType = TagHelperDirectiveType.RemoveTagHelper;
    directiveText = ((RemoveTagHelperChunkGenerator)span.ChunkGenerator).GetDirectiveText(span);
}
else if (span.ChunkGenerator is TagHelperPrefixDirectiveChunkGenerator)
{
    directiveType = TagHelperDirectiveType.TagHelperPrefix;
    directiveText = ((TagHelperPrefixDirectiveChunkGenerator)span.ChunkGenerator).GetDirectiveText(span);
}
else
{
    // Not a chunk generator that we're interested in.
    return;
}
```

dougbu: Get rid of the one line `GetDirectiveText()` method entirely.

NTaylorMullen: What'd be the alternative? Duplicate the logic in a non directly related class (`TagHelperDirectiveSpanVisitor`)?

dougbu: You're talking about one simple line of code using a `public` property of the `Span`. Why all the rigmarole?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
NTaylorMullen: Note:
> Also the (TagHelperDirectiveChunkGenerator) cast is fragile. This class assumes AddTagHelperChunkGenerator and so on share this common base. That's not information that should need to be shared.

This concern goes away when we remove the `TagHelperDirectiveChunkGenerator`.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNet.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// <param name="target">
+        /// The <see cref="Span"/> responsible for this <see cref="AddTagHelperChunkGenerator"/>.
+        /// </param>
+        /// <param name="context">A <see cref="ChunkGeneratorContext"/> instance that contains information about
+        /// the current chunk generation process.</param>
+        public override void GenerateChunk(Span target, ChunkGeneratorContext context)
+        {
+            var lookupText = target.Content.Trim();
+
+            context.ChunkTreeBuilder.AddAddTagHelperChunk(lookupText, target);
```
dougbu: Missed that the call here passes `target.Content.Trim(), target`. Should the method signature instead be `AddAddTagHelperChunk(Span target)`, moving the `Content.Trim()` bit into the different chunks? Or are the chunks generally dumber than their generators?

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -289,7 +283,7 @@ protected void BaseTypeDirective(string noTypeNameError, Func<string, SpanChunkG
             Output(SpanKind.Code, AcceptedCharacters.AnyExceptNewline);
         }
 
-        private void TagHelperDirective(string keyword, Func<string, ISpanChunkGenerator> buildChunkGenerator)
+        private void TagHelperDirective(string keyword, ISpanChunkGenerator chunkGenerator)
```
dougbu: :clap:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -82,50 +83,47 @@ public IEnumerable<TagHelperDescriptor> GetDescriptors(Block root)
...
+                DirectiveType = directiveType
+            };
 
-            return offsetTextLocation;
+            _directiveDescriptors.Add(directiveDescriptor);
+        }
+
+        private string GetDirectiveText(Span span)
+        {
+            return span.Content.Trim();
```
dougbu: This one-liner is called exactly once. Why is it split out?

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpDirectivesTest.cs
```diff
@@ -12,29 +12,16 @@ namespace Microsoft.AspNet.Razor.Test.Parser.CSharp
...
-                        .MetaCode(SyntaxConstants.CSharp.TagHelperPrefixKeyword + " ")
-                        .Accepts(AcceptedCharacters.None),
-                    Factory.Code("\"\"")
-                        .AsTagHelperPrefixDirective(string.Empty)));
-        }
-
-        [Fact]
         public void TagHelperPrefixDirective_Succeeds()
         {
-            ParseBlockTest("@tagHelperPrefix \"Foo\"",
```
dougbu: Have you preserved tests showing what happens when parsing a `.cshtml` file containing the old syntax? Where are those tests? And what will the end user see in that case, either in console logging or on the error page (assuming that's enabled)?

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/ComplexTagHelpers.cs
```diff
@@ -491,7 +491,7 @@ public ComplexTagHelpers()
 #line default
 #line hidden
             Instrumentation.EndContext();
-            Instrumentation.BeginContext(1376, 14, true);
+            Instrumentation.BeginContext(1374, 14, true);
```
dougbu: Shoot you were doing such a good job making the generated HTML monotonically larger. At least you're only saving a few characters :smiling_imp:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/AddTagHelperDirective.cshtml
```diff
@@ -1 +1 @@
-﻿@addTagHelper "**??**??*, nice"
\ No newline at end of file
+﻿@addTagHelper something, nice
```
dougbu: Why did this previously work? Didn't think we'd supported more than "something*" for a while.

dougbu:
:watch: mostly for a couple of open questions.

src/Microsoft.AspNet.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
NTaylorMullen: Chunks are generally dumber than their generators. :smile:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
NTaylorMullen: I fail, no other reason lmao.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpDirectivesTest.cs
NTaylorMullen: Nope, have not preserved any tests that have quotes. Don't typically maintain tests for old/changed features.

As for what the user will see if they still have quotes. You'll get a design time error because the code generation will look like: `_syntaxhelper = ""value""`

dougbu: :ok: my interest was primarily in the second question and that experience seems reasonable for those moving from prior versions.

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Source/AddTagHelperDirective.cshtml
NTaylorMullen: With how the tests are setup it doesn't matter about the value of the `@addTagHelper` as long as it can be parsed. If you want more detail we can talk offline :smile:

NTaylorMullen:
:up::date: And answered comments

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -325,31 +319,7 @@ private void TagHelperDirective(string keyword, Func<string, ISpanChunkGenerator
...
-                // The format for tag helper directives are: @directivename "SomeValue"
-                if (!startsWithQuote ||
-                    !rawValue.EndsWith("\"", StringComparison.OrdinalIgnoreCase))
-                {
-                    Context.OnError(
-                        startLocation,
-                        RazorResources.FormatParseError_DirectiveMustBeSurroundedByQuotes(keyword),
-                        rawValue.Length);
-                }
+                Span.ChunkGenerator = chunkGenerator;
```
dougbu: Any reason not to update the `ChunkGenerator` in the error case as well? I.e. could we move this below the `if`?

dougbu:
:shipit: once reaction work is done -- in MVC, MusicStore, and probably Entropy.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: :+1:

PR #592, opened 2015-11-02
Strong name everything.

cesarbs:
https://github.com/aspnet/dnx/issues/2615

cc @davidfowl @Eilon @muratg @moozzyk @pranavkm


Eilon:
Belated :shipit:

Issue #594, opened 2015-11-04
An inline if-statement in an attribute value crashes razor


twsouthwick:
The following code fails in beta8:

```html
<div class="alert alert-new-arrival @if (Model == null) { <text>hidden</text>}" id="alert-new-arrival" role="alert">
```

If this is replaced with:

```html
<div class="alert alert-new-arrival @(Model == null ? "hidden" : "") id="alert-new-arrival" role="alert">
````

it works. The error seen is coming from some generated code. The message is:

```
An error occurred during the compilation of a resource required to process this request. Please review the following specific error details and modify your source code appropriately.
•	Not all code paths return a value in lambda expression of type 'Func<TextWriter, Task>'
40.	        {
41.	            __tagHelperRunner = __tagHelperRunner ?? new Microsoft.AspNet.Razor.Runtime.TagHelpers.TagHelperRunner();
42.	            BeginContext(40, 4, true);
43.	            WriteLiteral("<div");
44.	            EndContext();
45.	            WriteAttribute("class", Tuple.Create(" class=\"", 44), Tuple.Create("\"", 118), Tuple.Create(Tuple.Create("", 52), Tuple.Create("alert", 52), true), Tuple.Create(Tuple.Create(" ", 57), Tuple.Create("alert-new-arrival", 58), true), 
46.	            Tuple.Create(Tuple.Create(" ", 75), Tuple.Create<System.Object, System.Int32>(new Microsoft.AspNet.Mvc.Razor.HelperResult((__razor_attribute_value_writer) => {
47.	#line 3 "/Views/Shared/Components/Announcement/Default.cshtml"
48.	                                     if (Model == null) {
49.	
50.	#line default
51.	#line hidden
52.	
53.	                BeginContext(104, 6, true);
54.	                WriteLiteralTo(__razor_attribute_value_writer, "hidden");
55.	                EndContext();
56.	#line 3 "/Views/Shared/Components/Announcement/Default.cshtml"
57.	                                                                            }
58.	
59.	#line default
60.	#line hidden
61.	
62.	            }

```

NTaylorMullen:
02d6b00d8f5d7fbc452c1c69a3330dddba9f3c91
https://github.com/aspnet/Mvc/commit/bebd3dc9a0122cc35a1ecc4e4305b767588c61b9

twsouthwick:
Thanks for the quick turn around!

PR #595, opened 2015-11-05
Generate `async` lambda for template attributes.

NTaylorMullen:
- Template attributes should be of type `Func<TextWriter, Task>`. We weren't generating an `async` lambda for attributes prior to this change resulting in a compilation failure when used at runtime and the inability to `@await` code (unless a user returns some sort of `Task`).
- Updated code generation files to reflect the new code generation behavior.

#594


ajaybhargavb:
:shipit: Will this need an MVC reaction?

NTaylorMullen:
@ajaybhargavb I thought it might so I tried it out and turns out there's no reaction needed. The case of having a block level (`@if ...`) Razor statement is rare enough that it's not used in MVC.

ajaybhargavb:
Ah. Maybe add a test that does that if its not too much effort for a rare case.

pranavkm:
Let's add a test to Mvc (would be ok to modify an existing Razor scenario) to make using `await` inside of it works correctly.

:shipit:

Issue #577, opened 2015-10-19
Lots of unit test failures in Microsoft.AspNet.Razor.Test on CoreCLR on Linux


cesarbs:
They all look similar:

    info: Exec
    info:   program: /usr/bin/env
    info:   commandline: bash -c ". \"/home/cesars/src/aspnet/universe/Razor/packages/KoreBuild/build/dnvm.sh\"; dnvm run default -runtime coreclr test"
    info:   workingdir: test/Microsoft.AspNet.Razor.Test
    xUnit.ne;t DNX Runner (64-bit DNXCore 5.0)
      Discovering: Microsoft.AspNet.Razor.Test
      Discovered:  Microsoft.AspNet.Razor.Test
      Starting:    Microsoft.AspNet.Razor.Test
          Assert.Equal() Failure
                                                   ↓ (pos 197)
          Expected: ···lock.cshtml"\r\n\r\n\r\n\r\n#line default\r\n#line hidden\r\n#line 5 "Fun···
          Actual:   ···lock.cshtml"\r\n\r\n\r\n\r\n\r\n#line default\r\n#line hidden\r\n#line 5 "F···
                                                   ↑ (pos 197)
          Stack Trace:
               at Microsoft.AspNet.Razor.Test.Generator.RazorCodeGeneratorTest`1.RunTestInternal(String name, String baselineName, Boolean generatePragmas, Boolean designTimeMode, IList`1 expectedDesignTimePragmas, TestSpan[] spans, Boolean withTabs, Func`2 hostConfig, Func`2 templateEngineConfig, Action`1 onResults)
               at Microsoft.AspNet.Razor.Test.Generator.RazorCodeGeneratorTest`1.RunTest(String name, String baselineName, Boolean generatePragmas, Boolean designTimeMode, IList`1 expectedDesignTimePragmas, TestSpan[] spans, TabTest tabTest, Func`2 hostConfig, Func`2 templateEngineConfig, Action`1 onResults)
               at Microsoft.AspNet.Razor.Test.Generator.RazorCodeGeneratorTest.CSharpChunkGeneratorCorrectlyGeneratesFunctionsBlocksAtDesignTime()
               at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
               at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
[...]

PR #609, opened 2015-11-17
Execute `dnu restore` with .NET Core and use Mono 4.0.5

dougbu:
- upgrade Linux to Trusty (Ubuntu 14.04) distribution, where .NET Core is supported
- Mono beta is now 4.2.1 which doesn't work with `nuget.exe`
  - see also aspnet/External#48

Note: cannot use `KOREBUILD_TEST_DNXCORE=true` due to #577

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,22 @@
 language: csharp
-sudo: false
-mono:
-  - beta
+sudo: required
+dist: trusty
+addons:
+  apt:
+    packages:
+    - gettext
+    - libcurl4-openssl-dev
+    - libicu-dev
+    - libssl-dev
+    - libunwind8
+    - zlib1g
+before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,9 +1,22 @@
 language: csharp
-sudo: false
-mono:
-  - beta
+sudo: required
+dist: trusty
+addons:
+  apt:
+    packages:
+    - gettext
```
davidfowl: wow

dougbu: Likes it does yah? Or are you suggesting http://docs.asp.net/en/latest/getting-started/installing-on-linux.html#install-the-net-execution-environment-dnx is out-of-date?

NTaylorMullen:
:shipit:

PR #611, opened 2015-11-18
Simplify ChunkTreeBuild

rynowak:
By making the tree more uniform and adding a root node, we simplify
the programming model inside the the builder.


src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
```diff
@@ -27,13 +30,13 @@ public void AddChunk(Chunk chunk, SyntaxTreeNode association, bool topLevel = fa
             chunk.Association = association;
 
             // If we're not in the middle of a parent chunk
-            if (_parentStack.Count == 0 || topLevel == true)
+            if (topLevel == true)
```
NTaylorMullen: Not sure why I had `topLevel == true`. Can just use this opportunity to simplify to `if (topLevel)` :smile:

NTaylorMullen:
:shipit:

rynowak:
I have one more idea I want to run by you tomorrow. Closing for now.

PR #612, opened 2015-11-18
Simplify ChunkTreeBuilder

rynowak:


src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
```diff
@@ -29,11 +32,11 @@ public void AddChunk(Chunk chunk, SyntaxTreeNode association, bool topLevel = fa
             // If we're not in the middle of a parent chunk
             if (_parentStack.Count == 0 || topLevel == true)
```
NTaylorMullen: Should fix this up. Looks like you undid the changes you previously made here.

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
rynowak: :+1:

rynowak: :+1:

PR #613, opened 2015-11-18
Fix aspnet/Mvc#3196 Razor Compilation Allocations

rynowak:
This change significantly reduces the amount of string and List<ISymbol>
allocations that occur during compilation by changing the way
LiteralChunks are combined.

This is a low impact fix that addresses the performance issue, the design
issues that caused it still exist.

The problem here lies in what Razor fundamentally does - it parses HTML/C#
into tokens, and then combines them back into 'chunks' a representation
friendly to code generation. When presenting with a large block of static
HTML, Razor parses it into individual HTML tokens, and then tries to join
them in back into a single chunk for rendering. Due to details of Razor's
representation of chunks/tokens, the process of combining literals is too
expensive.

Mainly, what's done here is to not try to combine instances of
LiteralChunk. The process of merging them is too expensive and requires
lots of interm List<ISymbol> and string allocations.

Instead we produce a new 'chunk' ParentLiteralChunk, which doesn't do so
much up-front computing. Various pieces of the code that deal with
LiteralChunk need to be updated to deal with ParentLiteralChunk also,
which is the bulk of the changes here.

Note that we still have the potential for LOH allocations to occur during
codegen, but it's likely to occur O(1) for each large block of HTML
instead of O(N) as it did in the old code.

### DIFF
--- a/src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
+++ b/src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNet.Razor.Chunks
         {
             _lastChunk = chunk;
 
-            chunk.Start = association?.Start ?? default(SourceLocation);
+            chunk.Start = association.Start;
             chunk.Association = association;
 
             // If we're not in the middle of a parent chunk
@@ -72,8 +72,12 @@ namespace Microsoft.AspNet.Razor.Chunks
         {
             ParentLiteralChunk parentLiteralChunk;
 
-            // We try to join literal chunks where possible, so that we have fewer 'writes' in the generated
-            // code.
+            // We try to join literal chunks where possible, so that we have fewer 'writes' in the generated code.
+            //
+            // Possible cases here:
+            //  - We just added a LiteralChunk and we need to add another - so merge them into ParentLiteralChunk.
+            //  - We have a ParentLiteralChunk - merge the new chunk into it.
+            //  - We just added something <else> - just add the LiteralChunk like normal.
             if (_lastChunk is LiteralChunk)
             {
                 parentLiteralChunk = new ParentLiteralChunk()
--- a/src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
+++ b/src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.AspNet.Razor.Parser.SyntaxTree;
 
 namespace Microsoft.AspNet.Razor.Chunks
@@ -26,7 +27,7 @@ namespace Microsoft.AspNet.Razor.Chunks
         {
             _lastChunk = chunk;
 
-            chunk.Start = association.Start;
+            chunk.Start = association?.Start ?? default(SourceLocation);
             chunk.Association = association;
 
             // If we're not in the middle of a parent chunk
@@ -69,22 +70,39 @@ namespace Microsoft.AspNet.Razor.Chunks
 
         public void AddLiteralChunk(string literal, SyntaxTreeNode association)
         {
-            // If the previous chunk was also a LiteralChunk, append the content of the current node to the previous one.
-            var literalChunk = _lastChunk as LiteralChunk;
-            if (literalChunk != null)
+            ParentLiteralChunk parentLiteralChunk;
+
+            // We try to join literal chunks where possible, so that we have fewer 'writes' in the generated
+            // code.
+            if (_lastChunk is LiteralChunk)
             {
-                // Literal chunks are always associated with Spans
-                var lastSpan = (Span)literalChunk.Association;
-                var currentSpan = (Span)association;
-
-                var builder = new SpanBuilder(lastSpan);
-                foreach (var symbol in currentSpan.Symbols)
+                parentLiteralChunk = new ParentLiteralChunk()
                 {
-                    builder.Accept(symbol);
-                }
+                    Start = _lastChunk.Start,
+                };
 
-                literalChunk.Association = builder.Build();
-                literalChunk.Text += literal;
+                parentLiteralChunk.Children.Add(_lastChunk);
+                parentLiteralChunk.Children.Add(new LiteralChunk
+                {
+                    Association = association,
+                    Start = association.Start,
+                    Text = literal,
+                });
+
+                Debug.Assert(Current.Children[Current.Children.Count - 1] == _lastChunk);
+                Current.Children.RemoveAt(Current.Children.Count - 1);
+                Current.Children.Add(parentLiteralChunk);
+                _lastChunk = parentLiteralChunk;
+            }
+            else if ((parentLiteralChunk = _lastChunk as ParentLiteralChunk) != null)
+            {
+                parentLiteralChunk.Children.Add(new LiteralChunk
+                {
+                    Association = association,
+                    Start = association.Start,
+                    Text = literal,
+                });
+                _lastChunk = parentLiteralChunk;
             }
             else
             {
--- a/src/Microsoft.AspNet.Razor/Chunks/ParentLiteralChunk.cs
+++ b/src/Microsoft.AspNet.Razor/Chunks/ParentLiteralChunk.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text;
+
+namespace Microsoft.AspNet.Razor.Chunks
+{
+    public class ParentLiteralChunk : ParentChunk
+    {
+        public string GetText()
+        {
+            var builder = new StringBuilder();
+            for (var i = 0; i < Children.Count; i++)
+            {
+                builder.Append(((LiteralChunk)Children[i]).Text);
+            }
+
+            return builder.ToString();
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -505,19 +505,18 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
 
         public CSharpCodeWriter WriteStartInstrumentationContext(
             ChunkGeneratorContext context,
-            int start,
+            int absoluteIndex,
             int length,
             bool isLiteral)
         {
             WriteStartMethodInvocation(context.Host.GeneratedClassContext.BeginContextMethodName);
-            Write(start.ToString(CultureInfo.InvariantCulture));
+            Write(absoluteIndex.ToString(CultureInfo.InvariantCulture));
             WriteParameterSeparator();
             Write(length.ToString(CultureInfo.InvariantCulture));
             WriteParameterSeparator();
             Write(isLiteral ? "true" : "false");
             return WriteEndMethodInvocation();
         }
-
 
         public CSharpCodeWriter WriteEndInstrumentationContext(ChunkGeneratorContext context)
         {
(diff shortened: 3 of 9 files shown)


src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
```diff
@@ -26,7 +27,7 @@ public void AddChunk(Chunk chunk, SyntaxTreeNode association, bool topLevel = fa
         {
             _lastChunk = chunk;
 
-            chunk.Start = association.Start;
+            chunk.Start = association?.Start ?? default(SourceLocation);
```
rynowak: ParentLiteralChunk doesn't have an association.

src/Microsoft.AspNet.Razor/Chunks/ParentLiteralChunk.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public string GetText()
+        {
+            var builder = new StringBuilder();
+            for (var i = 0; i < Children.Count; i++)
+            {
+                builder.Append(((LiteralChunk)Children[i]).Text);
+            }
+
+            return builder.ToString();
```
rynowak: Get your LOH allocations here!

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -496,15 +496,29 @@ private void WriteCStyleStringLiteral(string literal)
...
+                syntaxNode.Start.AbsoluteIndex,
+                syntaxNode.Length,
+                isLiteral);
+        }
+
+        public CSharpCodeWriter WriteStartInstrumentationContext(
+            ChunkGeneratorContext context,
+            int start,
+            int length,
+            bool isLiteral)
```
rynowak: Needed because `ParentLiteralChunk` doesn't have an association.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.AspNet.Razor.CodeGenerators.Visitors;
 using Microsoft.AspNet.Razor.TagHelpers;
 using Microsoft.AspNet.Razor.Compilation.TagHelpers;
+using System.Text;
```
rynowak: Will remove

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
```diff
@@ -69,22 +70,39 @@ public void AddRemoveTagHelperChunk(string lookupText, SyntaxTreeNode associatio
 
         public void AddLiteralChunk(string literal, SyntaxTreeNode association)
         {
-            // If the previous chunk was also a LiteralChunk, append the content of the current node to the previous one.
-            var literalChunk = _lastChunk as LiteralChunk;
-            if (literalChunk != null)
+            ParentLiteralChunk parentLiteralChunk;
```
NTaylorMullen: For a thought exercise I rewrote the following to normalize the Adding of literal chunks. Torn as to which one is more clear. Therefore this is probably fine as is. Here's it rewritten in-case you feel differently:

```C#
var parentLiteralChunk = _lastChunk as ParentLiteralChunk;

// We try to join literal chunks where possible, so that we have fewer 'writes' in the generated code.
if (parentLiteralChunk == null && _lastChunk is LiteralChunk)
{
    Debug.Assert(Current.Children[Current.Children.Count - 1] == _lastChunk);
    Current.Children.RemoveAt(Current.Children.Count - 1);

    parentLiteralChunk = new ParentLiteralChunk
    {
        Start = _lastChunk.Start,
    };
    parentLiteralChunk.Children.Add(_lastChunk);
}

if (parentLiteralChunk != null)
{
    parentLiteralChunk.Children.Add(new LiteralChunk
    {
        Association = association,
        Start = association.Start,
        Text = literal,
    });

    _lastChunk = parentLiteralChunk;
}
else
{
    AddChunk(new LiteralChunk
    {
        Text = literal,
    }, association);
}
```

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
```diff
@@ -69,22 +70,39 @@ public void AddRemoveTagHelperChunk(string lookupText, SyntaxTreeNode associatio
...
+
+            // We try to join literal chunks where possible, so that we have fewer 'writes' in the generated
+            // code.
```
NTaylorMullen: can fit on line above.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -496,15 +496,29 @@ private void WriteCStyleStringLiteral(string literal)
...
+            return WriteStartInstrumentationContext(
+                context,
+                syntaxNode.Start.AbsoluteIndex,
+                syntaxNode.Length,
+                isLiteral);
+        }
+
+        public CSharpCodeWriter WriteStartInstrumentationContext(
+            ChunkGeneratorContext context,
+            int start,
```
NTaylorMullen: rename to `absoluteIndex`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -496,15 +496,29 @@ private void WriteCStyleStringLiteral(string literal)
...
+            Write(start.ToString(CultureInfo.InvariantCulture));
             WriteParameterSeparator();
-            Write(syntaxNode.Length.ToString(CultureInfo.InvariantCulture));
+            Write(length.ToString(CultureInfo.InvariantCulture));
             WriteParameterSeparator();
             Write(isLiteral ? "true" : "false");
             return WriteEndMethodInvocation();
         }
 
+
```
NTaylorMullen: Nuke

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -116,6 +117,45 @@ protected override void Visit(TemplateChunk chunk)
             Writer.WriteEndMethodInvocation(false).WriteLine();
         }
 
+        protected override void Visit(ParentLiteralChunk chunk)
+        {
+            if (Context.Host.DesignTimeMode)
```
NTaylorMullen: Add a debug.assert above this that `chunk.Children.Count > 1`

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -116,6 +117,45 @@ protected override void Visit(TemplateChunk chunk)
...
+            {
+                // Skip generating the chunk if we're in design time or if the chunk is empty.
+                return;
+            }
+            
+            if (Context.Host.EnableInstrumentation)
+            {
+                var start = chunk.Children[0].Association.Start.AbsoluteIndex;
```
NTaylorMullen: Did you find that the `chunk.Start` was different from the `chunk.Children[0].Association.Start`? You set it [here](https://github.com/aspnet/Razor/pull/613/files?diff=split#diff-60b1e905fa1bc92ee8f36cc9963e8b83R81).

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -116,6 +117,45 @@ protected override void Visit(TemplateChunk chunk)
...
+
+                Writer.WriteStartInstrumentationContext(Context, start, length, isLiteral: true);
+            }
+
+            if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
+            {
+                RenderPreWriteStart();
+            }
+
+            Writer.WriteStringLiteral(chunk.GetText());
```
NTaylorMullen: For easy debugging lets add a variable for `chunk.GetText()`.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -116,6 +117,45 @@ protected override void Visit(TemplateChunk chunk)
...
+            }
+            
+            if (Context.Host.EnableInstrumentation)
+            {
+                var start = chunk.Children[0].Association.Start.AbsoluteIndex;
+
+                var length = 0;
+                for (var i = 0; i < chunk.Children.Count; i++)
+                {
+                    length += chunk.Children[i].Association.Length;
```
NTaylorMullen: Can't we derive this from `chunk.GetText().Length`? You call `.GetText()` below anyways.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperAttributeValueVisitor.cs
```diff
@@ -102,6 +102,11 @@ protected override void Visit(LiteralChunk chunk)
             RenderCode(chunk.Text, (Span)chunk.Association);
         }
 
+        protected override void Visit(ParentLiteralChunk chunk)
+        {
+            RenderCode(chunk.GetText(), (Span)chunk.Children[0].Association);
```
NTaylorMullen: Could refactor `RenderCode` to just take in the start and then you can do `chunk.Start` (assuming it's equivalent to the association's start as mentioned in my comment [here](https://github.com/aspnet/Razor/pull/613/files?diff=split#r45247324)

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
rynowak: This code just never got updated. Will fix.

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
rynowak: Actually this isn't needed because we don't call `AddChunk` with the parent.

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
rynowak: I don't like it better. :-1: I wrote the code the way I did to make it clear that there are three distinct states, and make it easy to grasp what to do in each case. I added some comments here that hopefully makes it even better.

src/Microsoft.AspNet.Razor/Chunks/ChunkTreeBuilder.cs
rynowak: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
rynowak: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
rynowak: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
rynowak: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpTagHelperAttributeValueVisitor.cs
rynowak: Yeah, this is a good change all around. Made all of the usages simpler.

rynowak:
**Before**
image

I actually had to kill the process :laughing: I have no idea how much it would have eventually allocated.

**After**
image

This still needs a lot of love. Compilation time dropped from 5+ mins down to 4 seconds.

NTaylorMullen:
:shipit:

Issue #358, opened 2015-04-14
Analyze performance impact of TagHelperContent not being lazy on TagHelperOutput properties.


NTaylorMullen:
The initial implementations of the "new" `TagHelperOutput` data structure (one with buffers instead of strings) was originally lazy with the buffers construction but was then removed due to the added complexity via code review. We should analyze its perf impact to determine if being lazy about construction of `PreElement`, `PreContent`, `Content`, `PostContent` and `PostElement`'s `TagHelperContent` objects is beneficiary.

Eilon:
Please see https://github.com/aspnet/Razor/issues/317 for additional info.

Eilon:
Please have a chat with @rynowak before starting the perf investigation.

rynowak:
> beneficiary

Yes. Let's discuss

Eilon:
Hah!

rynowak:
Let's discuss
![image](https://cloud.githubusercontent.com/assets/1430011/11051973/17d183e4-8709-11e5-9a9b-4f04e239baad.png)

This is definitely worth addressing

NTaylorMullen:
Upon investigating a proper fix for this issue I decided to investigate what it'd take to have `TagHelperOutput` be an `IHtmlContent` (since it's responsible for lazily initializing buffers). After a lot of playing around with the idea and talks with @DamianEdwards it definitely fits. 

This will also involve removing the `WriteTagHelperAsync` method from Mvc.

NTaylorMullen:
bdf869c3d586e1ac4b94882aa91a719058ec9aa3
e8327eeec76afcb4156e95763e580df3693f4d8e
6b43ebc88b4645ab62e0f0f87a687d477f7f7c39
https://github.com/aspnet/Mvc/commit/2dc13b523d9b346fc99f998902f8d69a44222948

PR #615, opened 2015-11-18
Make `TagHelperOutput` an `IHtmlContent` to lazily initialize `TagHelperOutput`s `Content` properties.

NTaylorMullen:
- This allows users to write `TagHelperOutput` directly to an `IHtmlContent` accepting `TextWriter`.
- This also enables us to inspect backing fields for all of the various contents to lazily initialize them.
- Added backing fields to each of the content properties and added null propagation checks throughout the class.
- Update code generation tests to account for new `TagHelperOutput`.
- Add `TagHelperOutput` unit tests.

Overall we got a 73% improvement in `DefaultTagHelperContent` allocations for the application that was run.

**Before:**
image

**After:**
image

#358


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
...
+
+                    if (attribute.Minimized)
+                    {
+                        continue;
+                    }
+
+                    writer.Write("=\"");
+                    var value = attribute.Value;
+                    var htmlContent = value as IHtmlContent;
+                    if (htmlContent != null)
```
NTaylorMullen: Inlined the attribute rendering here from MVC's `WriteTagHelperAsync`/`WriteTo` overload. MVC PR incoming.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
         {
             return _getChildContentAsync(useCachedResult);
         }
+
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
```
NTaylorMullen: This entire method is copied from MVC `WriteTagHelperToAsync` with the exception of [this](https://github.com/aspnet/Razor/pull/615/files#r45275825).

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -61,32 +71,91 @@ internal TagHelperOutput(string tagName)
...
+                if (_content == null)
+                {
+                    _content = new DefaultTagHelperContent();
+                }
+
+                return _content;
+            }
+            set
+            {
+                _content = value ?? _content;
```
rynowak: Should it be an exception to set this to `null`? Generally doing something like this gives me the heebie-jeebies.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,88 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
         {
             return _getChildContentAsync(useCachedResult);
         }
+
+        /// <inheritdoc />
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
```
NTaylorMullen: This entire method is copied from MVC `WriteTagHelperToAsync` with the exception of [this](https://github.com/aspnet/Razor/pull/615/files#r45275825).

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -533,26 +533,50 @@ private void RenderEndTagHelpersScope()
                                                   _tagHelperContext.ScopeManagerEndMethodName);
         }
 
-        private void RenderWriteTagHelperMethodCall(TagHelperChunk chunk)
+        private void RenderTagHelperOutput(TagHelperChunk chunk)
         {
+            var tagHelperOutputAccessor =
```
NTaylorMullen: Could inline the triple write statements in each location. This was just re-used a lot so wanted to reduce duplication.

test/Microsoft.AspNet.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -0,0 +1,698 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
NTaylorMullen: This was taken directly from MVC since rendering is now handled by `TagHelperOutput` instead of a method on RazorPage.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,88 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
...
+                            var stringValue = stringWriter.ToString();
+                            stringValue = stringValue.Replace("\"", "&quot;");
+
+                            writer.Write(stringValue);
+                        }
+                    }
+                    else if (value != null)
+                    {
+                        var encodedValue = encoder.Encode(value.ToString());
+                        writer.Write(encodedValue);
```
rynowak: this can be `encoder.Encode(writer, value.ToString())`

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: Ya could definitely do that. :+1:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,88 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
...
+                }
+
+                writer.Write('>');
+            }
+
+            if (isTagNameNullOrWhitespace || TagMode == TagMode.StartTagAndEndTag)
+            {
+                _preContent?.WriteTo(writer, encoder);
+
+                _content?.WriteTo(writer, encoder);
```
rynowak: Do we need to check `_suppressContent` here?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +221,88 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
...
+
+                _postContent?.WriteTo(writer, encoder);
+            }
+
+            if (!isTagNameNullOrWhitespace && TagMode == TagMode.StartTagAndEndTag)
+            {
+                writer.Write("</");
+                writer.Write(TagName);
+                writer.Write(">");
```
rynowak: `string.Format`? 








:trollface:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -533,26 +533,50 @@ private void RenderEndTagHelpersScope()
...
+                using (_writer.BuildScope())
+                {
+                    _writer
+                        .Write(tagHelperOutputAccessor)
+                        .Write(".")
+                        .WriteStartAssignment(_tagHelperContext.TagHelperOutputContentPropertyName)
+                        .Write("await ")
+                        .WriteInstanceMethodInvocation(
+                            tagHelperOutputAccessor,
+                            _tagHelperContext.TagHelperOutputGetChildContentAsyncMethodName);
```
rynowak: Doing this with codegen is a pretty smart way to improve it.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -688,6 +712,23 @@ private void RenderBufferedAttributeValueAccessor(CSharpCodeWriter writer)
             }
         }
 
+        private static bool ContainsChildContent(IList<Chunk> children)
+        {
+            foreach (var child in children)
+            {
+                if (!(child is ParentChunk) || child is TagHelperChunk)
```
rynowak: This is why I was suggesting that we make `ParentChunk` abstract, and specialize other nodes, so that you can generically traverse it :+1: 

Not relevant for this PR, just making a point.

test/Microsoft.AspNet.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -143,34 +143,6 @@ public void ParentItems_SetsItemsProperty()
             Assert.Empty(content3.GetContent(new HtmlTestEncoder()));
         }
 
-        [Fact]
-        public async Task ExecuteChildContentAsync_IsNotMemoized()
```
rynowak: So is this gone from the user programming surface?

NTaylorMullen: It's gone, yes. However, I wouldn't say it was ever user surface. It only existed on the `TagHelperExecutionContext` for the `TagHelperRunner` system to consume.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: Nope, because if you suppress the content (`.Content.Clear()`) and then go ahead and append to it; any appended content will override the suppressed content and will result in content being rendered.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: So if you suppress, and then write to it ???

NTaylorMullen: `TagHelperOutput` is just a poco object where `SupressOutput` is just a helper method (documented that way too).

NTaylorMullen:
[MVC Reaction PR](https://github.com/aspnet/Mvc/pull/3582)

rynowak:
@NTaylorMullen - what's the overall impact? I'd expect this to improve a bunch of stuff beyond just `THC`

NTaylorMullen:
Including the current Kestrel bug (I haven't pulled latest) here's the difference.

**Before:**
image

**After:**
image

Nearly 9% decrease in allocations overall :smile:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -151,5 +226,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
...
+                    writer.Write("=\"");
+                    var value = attribute.Value;
+                    var htmlContent = value as IHtmlContent;
+                    if (htmlContent != null)
+                    {
+                        // There's no way of tracking the attribute value quotations in the Razor source. Therefore, we
+                        // must escape any IHtmlContent double quote values in the case that a user wrote:
+                        // <p name='A " is valid in single quotes'></p>
+                        //
+                        // This is really not optimal for perf, but it's the best we can do for right now.
```
dougbu: Suggest removing this comment; it reads like a TODO. The code is obviously not optimal but don't need to worry about it 'til a profile takes us here. For one thing attribute values may normally contain few enough (e.g. one) segments that the sub-optimal code just doesn't matter.

Note as well that we _can_ do better with a `TextWriter` that wraps `writer` and replaces the quote in the appropriate methods. May never need to build that.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -688,6 +712,23 @@ private void RenderBufferedAttributeValueAccessor(CSharpCodeWriter writer)
             }
         }
 
+        private static bool ContainsChildContent(IList<Chunk> children)
+        {
+            foreach (var child in children)
```
dougbu: Why is this a loop? Never looks at a second chunk, just recurses deeper.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -688,6 +712,23 @@ private void RenderBufferedAttributeValueAccessor(CSharpCodeWriter writer)
...
+            {
+                if (!(child is ParentChunk) || child is TagHelperChunk)
+                {
+                    return true;
+                }
+                else // Non-TagHelper ParentChunk
```
dougbu: The `!` above says this `else` is entered only when `child` _is_ a `ParentChunk` or `TagHelperChunk`.

NTaylorMullen: The else does not get entered for `TagHelperChunk`s

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Ahh, I messed up line 725 :+1: . That should only return IF it's value is `true`. Thankfully today that edge case wont ever be hit due to how the parser parses but should correct the method nevertheless.

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -186,24 +187,29 @@ public GeneratedTagHelperContext()
...
-        /// The name of the method used to write <see cref="ExecutionContextTypeName"/> to a specified
-        /// <see cref="System.IO.TextWriter"/>.
+        /// The name of the method used to convert a <c>TagHelperContent</c> into a <see cref="string"/>.
         /// </summary>
-        public string WriteTagHelperToAsyncMethodName { get; set; }
+        public string TagHelperContentGetContentMethodName { get; set; }
 
         /// <summary>
-        /// The name of the property containing the <c>HtmlEncoder</c>.
+        /// The property accessor to indicate if the tag helper's content has been modified.
```
dougbu: Ick. This is still a property name and that wording is consistent with the rest of this class.

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -186,24 +187,29 @@ public GeneratedTagHelperContext()
...
         /// </summary>
-        public string HtmlEncoderPropertyName { get; set; }
+        public string TagHelperOutputIsContentModifiedPropertyName { get; set; }
 
         /// <summary>
-        /// The name of the method used to convert a <c>TagHelperContent</c> into a <see cref="string"/>.
+        /// The property accessor for the tag helper's output content.
```
dougbu: Another property name.

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -186,24 +187,29 @@ public GeneratedTagHelperContext()
...
         /// </summary>
-        public string TagHelperContentGetContentMethodName { get; set; }
+        public string TagHelperOutputContentPropertyName { get; set; }
+
+        /// <summary>
+        /// The name of the method on the property <see cref="ExecutionContextOutputPropertyName"/> used to retrieve
+        /// tag helper child content.
```
dougbu: Suggest updating [line 128](https://github.com/aspnet/Razor/blob/nimullen/thctho.358/src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs#L128) to mention the `TagHelperContext` type. Then you could use simpler wording such as on [line 123](https://github.com/aspnet/Razor/blob/nimullen/thctho.358/src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs#L123) e.g. `The name of the <c>TagHelperOutput</c> method used to get and set the tag helper child content.` (The `ExecutionContextOutputPropertyName` is `The The name of the <see cref="ExecutionContextTypeName"/> property used to access the <c>TagHelperOutput<c>.`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -533,26 +533,50 @@ private void RenderEndTagHelpersScope()
                                                   _tagHelperContext.ScopeManagerEndMethodName);
         }
 
-        private void RenderWriteTagHelperMethodCall(TagHelperChunk chunk)
+        private void RenderTagHelperOutput(TagHelperChunk chunk)
         {
+            var tagHelperOutputAccessor =
+                $"{ExecutionContextVariableName}.{_tagHelperContext.ExecutionContextOutputPropertyName}";
+
+            if (ContainsChildContent(chunk.Children))
```
dougbu: Why not unconditionally write this check? Or at least only skip this output of there are no `Children` at all. Very special-case-y at the moment.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: After looking at aspnet/Mvc#3582 doesn't look like the intricacies of this method are tested. Can see this method does the Right Thing:tm: with self-closing elements, those containing straight text, those containing tag helpers, and not much else.

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
NTaylorMullen: Hu? Not sure I follow. The two configuration values are for the `TagHelperExecutionContext`. One for `TagHelperExecutionContext.Output` (of type `TagHelperOutput` used to render the `TagHelper`s output) and the other for `TagHelperExecutionContext.Add(TagHelper tagHelper)`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: > Why not unconditionally write this check? Or at least only skip this output of there are no Children at all. Very special-case-y at the moment.

We can only do this at compile time because we know if there's whitespace etc. that exist within the child content.



> After looking at aspnet/Mvc#3582 doesn't look like the intricacies of this method are tested. Can see this method does the Right Thing with self-closing elements, those containing straight text, those containing tag helpers, and not much else.

We verify it, no worries. Several self-closing examples in the code generation tests. Heck we have every tag form in the tests :smile:

NTaylorMullen:
:up: :date:

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
dougbu: My general point is you're making the doc wording in this class inconsistent and harder to understand.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: > We can only do this at compile time because we know if there's whitespace etc. that exist within the child content.

I was saying "Why bother with this added complexity at all?" Still need to write output even if it's only whitespace.

> Several self-closing examples

I said self-closing was covered. It's Razor comments, complex statement blocks, templates, ... inside tag helpers that have poor coverage. No way to know if we're missing a case where `Children` contains something obscure that will no longer render. Again, why bother with these special cases?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -688,6 +712,21 @@ private void RenderBufferedAttributeValueAccessor(CSharpCodeWriter writer)
             }
         }
 
+        private static bool ContainsChildContent(IList<Chunk> children)
+        {
+            foreach (var child in children)
+            {
+                if (!(child is ParentChunk) ||
+                    child is TagHelperChunk ||
+                    ContainsChildContent(((ParentChunk)child).Children))
```
dougbu: If this is really required, easier to read as
``` c#
                if ((child is ParentChunk && ContainsChildContent(((ParentChunk)child).Children) ||
                    child is TagHelperChunk)
```

dougbu: Since `ParentChunk` is a `class`, also more readable (and better in general) to use `as` and a temporary variable than to do the cast twice.

src/Microsoft.AspNet.Razor/CodeGenerators/GeneratedTagHelperContext.cs
NTaylorMullen: I just added the docs to be semi-nice to users. We actually don't have any plans to doc core Razor stuff like this :smile:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: > I was saying "Why bother with this added complexity at all?" Still need to write output even if it's only whitespace.

Whitespace is still content, this if statement will be generated if there's whitespace. This statement is to prevent us creating a brand new `DefaultTagHelperContent` (which wont have anything in it) via `GetChildContentAsync` for self-closing tags that do nothing with their `output.Content`.

> I said self-closing was covered. It's Razor comments, complex statement blocks, templates, ... inside tag helpers that have poor coverage. No way to know if we're missing a case where Children contains something obscure that will no longer render. Again, why bother with these special cases?

We have suuuuuuper broad `TagHelper` tests across Razor and MVC repos to cover pretty much every scenario possible; I don't see the value in duplicating that here when this code generation addition affects every `TagHelper` scenario. 

Overall this `ContainsChildContent` check is very broad and returns `true` when in doubt, the only specific `Chunk` we check is `TagHelperChunk`. All other `ParentChunk`s are treated as pure containers (which they are in Razor) and we validate their children.

dougbu: DIsagree on both the need for this complexity and "suuuuuuper broad `TagHelper` tests across Razor and MVC". Let's chat.

dougbu: Discussed offline. There are some odd cases in Razor where useless / empty ParentChunk nodes are created and we need to avoid creating an empty `DefaultTagHelperContent` in those cases.

So, with a few more comments and the changes I suggested in `ContainsChildContent()`, I'm alright w/ this.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -2,16 +2,26 @@
...
-    public class TagHelperOutput
+    public class TagHelperOutput : IHtmlContent
     {
         private readonly Func<bool, Task<TagHelperContent>> _getChildContentAsync;
+        private TagHelperContent _preElement;
+        private TagHelperContent _preContent;
+        private TagHelperContent _content;
+        private TagHelperContent _postContent;
+        private TagHelperContent _postElement;
+        private bool _contentSupressed;
```
dougbu: Hmm, "content suppressed" is the state only immediately after a call to `SuppressOutput()`. Suggest `_wasSuppressOutputCalled` especially to avoid confusing @rynowak again.

dougbu:
:shipit: once aspnet/Mvc#3582 is done and dusty.

Issue #601, opened 2015-11-12
Remove use of OrderBy in TagHelperRunner


rynowak:
This is a hotpath, no Linq allowed

![image](https://cloud.githubusercontent.com/assets/1430011/11107345/e71bfd62-8891-11e5-99d9-2e2ffb454424.png)

NTaylorMullen:
43aabcb2d616e61f1623447da1d367b8b3fdcaa9

PR #617, opened 2015-11-19
Order `TagHelperExecutionContext.TagHelpers` without `.OrderBy`.

NTaylorMullen:
- By using linq we created several `OrderedEnumerable` allocations.

**Before:**
image

**After:**
image

**Result:** Roughly a 2.3% reduction in overall allocations.

#601

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -135,7 +135,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
         /// <summary>
         /// <see cref="ITagHelper"/>s that should be run.
         /// </summary>
-        public IEnumerable<ITagHelper> TagHelpers
+        public IList<ITagHelper> TagHelpers
         {
             get
             {
--- a/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Razor.TagHelpers;
@@ -31,11 +32,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 executionContext.AllAttributes,
                 executionContext.Items,
                 executionContext.UniqueId);
-            var orderedTagHelpers = executionContext.TagHelpers.OrderBy(tagHelper => tagHelper.Order);
 
-            foreach (var tagHelper in orderedTagHelpers)
+            OrderTagHelpers(executionContext.TagHelpers);
+
+            for (var i = 0; i < executionContext.TagHelpers.Count; i++)
             {
-                tagHelper.Init(tagHelperContext);
+                executionContext.TagHelpers[i].Init(tagHelperContext);
             }
 
             var tagHelperOutput = new TagHelperOutput(
@@ -46,12 +48,31 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 TagMode = executionContext.TagMode,
             };
 
-            foreach (var tagHelper in orderedTagHelpers)
+            for (var i = 0; i < executionContext.TagHelpers.Count; i++)
             {
-                await tagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
+                await executionContext.TagHelpers[i].ProcessAsync(tagHelperContext, tagHelperOutput);
             }
 
             return tagHelperOutput;
+        }
+
+        private static void OrderTagHelpers(IList<ITagHelper> tagHelpers)
+        {
+            // Using bubble-sort here due to its simplicity. It'd be an extreme corner case to ever have more than 3 or
+            // 4 tag helpers simultaneously.
+            ITagHelper temp = null;
+            for (var i = 0; i < tagHelpers.Count; i++)
+            {
+                for (var j = i + 1; j < tagHelpers.Count; j++)
+                {
+                    if (tagHelpers[j].Order < tagHelpers[i].Order)
+                    {
+                        temp = tagHelpers[i];
+                        tagHelpers[i] = tagHelpers[j];
+                        tagHelpers[j] = temp;
+                    }
+                }
+            }
         }
     }
 }


src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -46,12 +48,34 @@ public class TagHelperRunner
...
             {
                 await tagHelper.ProcessAsync(tagHelperContext, tagHelperOutput);
             }
 
             return tagHelperOutput;
         }
+
+        private static void OrderTagHelpers(IList<ITagHelper> tagHelpers)
+        {
+            if (tagHelpers.Count == 1)
```
NTaylorMullen: This is the 90% case where there's only 1 `TagHelper`.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -135,7 +135,7 @@ public bool ChildContentRetrieved
         /// <summary>
         /// <see cref="ITagHelper"/>s that should be run.
         /// </summary>
-        public IEnumerable<ITagHelper> TagHelpers
+        public IList<ITagHelper> TagHelpers
```
dougbu: `IReadOnlyList<ITagHelper>`

dougbu: OIC you're updating the collection in place. Still seems incorrect for the rest of the system. Why not sort upfront and expose the sorted enumeration here?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -46,12 +48,34 @@ public class TagHelperRunner
...
+                for (var j = i + 1; j < tagHelpers.Count; j++)
+                {
+                    if (tagHelpers[j].Order < tagHelpers[i].Order)
+                    {
+                        temp = tagHelpers[i];
+                        tagHelpers[i] = tagHelpers[j];
+                        tagHelpers[j] = temp;
+                    }
+                }
+            }
```
dougbu: Doubt this is efficient for mid-sized collections but agree it's good enough for the small collections we expect. Add a comment about not spending a lot of time choosing the algorithm.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Ah, so I thought about that. The issue with that approach is that `TagHelper` authors wouldn't be able to determine their `Order` dynamically based on their provided properties with how we code gen today.

Ex:
```C#
 __TestNamespace_InputTagHelper = CreateTagHelper<global::TestNamespace.InputTagHelper>();
__tagHelperExecutionContext.Add(__TestNamespace_InputTagHelper); // Would be ordered on Insert, in this case we don't have any properties set yet.
__TestNamespace_InputTagHelper2 = CreateTagHelper<global::TestNamespace.InputTagHelper2>();
__tagHelperExecutionContext.Add(__TestNamespace_InputTagHelper2);
__TestNamespace_InputTagHelper.Type = "checkbox";
```

It'd be a larger change to transition the `.Add` after the attribute initialization of the `TagHelper`s. Wanted to avoid the larger change. If you feel like it has a lot of value I can do it.

dougbu:
Razor also uses `OrderBy` in `TagHelperDescriptorComparer` and `CaseSensitiveTagHelperDescriptorComparer`. One of those are likely responsible for the remaining allocations. Suggest cleaning them up too if possible.

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Oh and I also felt that `Order` was more of a `TagHelperRunner`s responsibility.

NTaylorMullen:
> Razor also uses OrderBy in TagHelperDescriptorComparer and CaseSensitiveTagHelperDescriptorComparer. One of those are likely responsible for the remaining allocations. Suggest cleaning them up too if possible.

Talked offline about this. Those classes are compile time only and not a worry for their `OrderBy`s. The remaining `OrderedEnumerable`s are actually caused by [this](https://github.com/aspnet/Mvc/blob/d6bda0ec11d7d29782cb19b09073ea67710ce8a5/src/Microsoft.AspNet.Mvc.ViewFeatures/ViewFeatures/DefaultHtmlGenerator.cs#L760) method in MVC.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: The general principle should be to expose what's required and this doesn't do that -- the only reader needs a sorted read-only list and doesn't get one. But it's probably not that big a deal because the `TagHelperExecutionContext` isn't exposed to tag helper authors (or most other user code) directly.

rynowak: I'd also expect that the TagHelpers are ordered by whatever builds their descriptors.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -31,9 +32,10 @@ public class TagHelperRunner
                 executionContext.AllAttributes,
                 executionContext.Items,
                 executionContext.UniqueId);
-            var orderedTagHelpers = executionContext.TagHelpers.OrderBy(tagHelper => tagHelper.Order);
 
-            foreach (var tagHelper in orderedTagHelpers)
+            OrderTagHelpers(executionContext.TagHelpers);
```
rynowak: While you're here, these should be `for` loops if they happen at runtime with an `IList` or `IReadOnlyList`.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: @rynowak but would you expect this collection to be reordered when that happens?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -31,9 +32,10 @@ public class TagHelperRunner
...
+
+            foreach (var tagHelper in executionContext.TagHelpers)
```
rynowak: While you're here, these should be `for` loops if they happen at runtime with an `IList` or `IReadOnlyList`.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
rynowak: Not needed. You'd never execute the body of the inner loop.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: @rynowak You couldn't determine order when constructing descriptors. The `TagHelper` types need to be instantiated in order to access their `.Order`.

NTaylorMullen:
:up: :date:

rynowak:
:shipit: from me

dougbu:
:shipit:

Issue #599, opened 2015-11-12
Optimize allocations in TagHelperAttributeList and friends


rynowak:
This is a particular hotspot for allocations as these are created and used all over.

![image](https://cloud.githubusercontent.com/assets/1430011/11107265/02bc23b8-8891-11e5-8b5c-0a22971f3837.png)

NTaylorMullen:
This significantly overlaps with https://github.com/aspnet/Razor/issues/605 but is a stepping stone. Going to grab this one and start work.

NTaylorMullen:
31a68a0705b8274e6345705787ea11b6820885b9

PR #619, opened 2015-11-20
Optimize `TagHelperAttributeList` allocations.

NTaylorMullen:
- Today `TagHelperAttributeList` and corresponding infrastructure copy themselves too frequently. I've reduced the copying since we own `TagHelperExecutionContext`.
- Removed usage of linq inside of `TagHelperAttributeList` to reduce allocations.

**Before:**
image

**After:**
image

**Result:** Roughly a 10% reduction in overall allocations.

#599

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNet.Razor.TagHelpers
 {
@@ -50,13 +49,7 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         protected List<TAttribute> Attributes { get; }
 
         /// <inheritdoc />
-        public TAttribute this[int index]
-        {
-            get
-            {
-                return Attributes[index];
-            }
-        }
+        public TAttribute this[int index] => Attributes[index];
 
         /// <summary>
         /// Gets the first <typeparamref name="TAttribute"/> with <see cref="IReadOnlyTagHelperAttribute.Name"/>
@@ -78,18 +71,21 @@ namespace Microsoft.AspNet.Razor.TagHelpers
                     throw new ArgumentNullException(nameof(name));
                 }
 
-                return Attributes.FirstOrDefault(attribute => NameEquals(name, attribute));
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < Attributes.Count; i++)
+                {
+                    if (NameEquals(name, Attributes[i]))
+                    {
+                        return Attributes[i];
+                    }
+                }
+
+                return default(TAttribute);
             }
         }
 
         /// <inheritdoc />
-        public int Count
-        {
-            get
-            {
-                return Attributes.Count;
-            }
-        }
+        public int Count => Attributes.Count;
 
         /// <summary>
         /// Determines whether a <typeparamref name="TAttribute"/> matching <paramref name="item"/> exists in the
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -50,13 +48,7 @@ public ReadOnlyTagHelperAttributeList(IEnumerable<TAttribute> attributes)
...
 
         /// <inheritdoc />
-        public TAttribute this[int index]
-        {
-            get
-            {
-                return Attributes[index];
-            }
-        }
+        public TAttribute this[int index] => Attributes[index];
```
rynowak: Many PranavPoints
Such Pranav
Wow

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -78,18 +70,20 @@ public ReadOnlyTagHelperAttributeList(IEnumerable<TAttribute> attributes)
                     throw new ArgumentNullException(nameof(name));
                 }
 
-                return Attributes.FirstOrDefault(attribute => NameEquals(name, attribute));
+                for (var i = 0; i < Attributes.Count; i++)
```
rynowak: I generally add a `// Perf: Avoid allocating enumerator` in these kind of micro things

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -197,9 +199,17 @@ public bool TryGetAttributes(string name, out IEnumerable<TAttribute> attributes
...
+            {
+                if (NameEquals(name, Attributes[i]))
+                {
+                    matchedAttributes.Add(Attributes[i]);
+                }
+            }
+            attributes = matchedAttributes;
 
-            return attributes.Any();
+            return matchedAttributes.Count > 0;
```
rynowak: How commonly is this used? you could avoid allocating this list if nothing matches

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -48,7 +49,18 @@ public class TagHelperContext
...
         /// Every attribute associated with the current HTML element.
         /// </summary>
-        public ReadOnlyTagHelperAttributeList<IReadOnlyTagHelperAttribute> AllAttributes { get; }
+        public ReadOnlyTagHelperAttributeList<IReadOnlyTagHelperAttribute> AllAttributes
+        {
+            get
+            {
+                if (_allAttributes == null)
+                {
+                    _allAttributes = new ReadOnlyTagHelperAttributeList<IReadOnlyTagHelperAttribute>(_allAttributesData);
```
rynowak: How deep is this copy?

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
NTaylorMullen: Hmmmmmm good point :+1:. We don't use it at all but I'll optimize.

rynowak:
:shipit: in general. Left a few minor comments, but most of them are questions. Let me know if you need another iteration.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Just a wrapper for API.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
rynowak: The other consideration here if this was a super commonly used API is that you could return a struct here.

Issue #630, opened 2015-12-03
TagHelperDesignTimeDescriptorFactory performance


IlyaKhD:
Documentation file is read each time a `DocumentationDescriptor` is built (see the `CreateDocumentationDescriptor` method), though while resolving TagHelpers this file could be loaded only once for each assembly.

The process of TagHelpers resolving takes much more time for documented TagHelpers.

Check PerfTest console app in this [branch](https://github.com/IlyaKhD/RazorTooling/tree/khd/resolve-taghelpers.performance).

I've created two fake TagHelper NuGet packs (hosted on MyGet) containing 200 TagHelpers (each has 20 properties). The only difference is that properties in the second pack have summary.
You can run the following commands:

`> dnx resolve-taghelpers`
`> dnx resolve-taghelpers-withsummary`

On my environment it takes 3 sec and 21 sec accordingly.

Caching documentation files decreases execution time from 21 sec to 5sec (check this illustration: https://github.com/IlyaKhD/Razor/commit/db7d1a5b4860fe3e5ef19bc889cd3f37e0f812c0)

This affects greatly Design Time since this code is used in [RazorTooling](https://github.com/aspnet/RazorTooling).
The original thread is here: https://github.com/aspnet/RazorTooling/issues/35

NTaylorMullen:
Talked with @Eilon offline about taking this one for RC2. + @danroth27

NTaylorMullen:
dbe4ce06f2d4dcd3924204ce8bf0673686b5a2a9

PR #637, opened 2015-12-09
Cache XML file information when creating design time based `TagHelperDescriptor`s.

NTaylorMullen:
- Prior to this change we'd hit the disk every time we'd create a `TagHelperDesignTimeDescriptor` to locate an assemblies XML files.
- Did not want to tie the XML cache to a static member in-case a user updates their XML information between parses. Therefore, changed `TagHelperDescriptorFactory` and `TagHelperDesignTimeDescriptorFactory` to no longer be static.
- Put the same `TagHeleprDescriptorFactory` extensibility point as we had for its counterpart `TagHelperTypeResolver` to stay consistent. This involved making `CreateDescriptors` virtual and allowing it to be provided in the `TagHelperDescriptorResolver` constructor.

#630

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -15,7 +15,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// <summary>
     /// Factory for <see cref="TagHelperDescriptor"/>s from <see cref="ITypeInfo"/>s.
     /// </summary>
-    public static class TagHelperDescriptorFactory
+    public class TagHelperDescriptorFactory
     {
         private const string DataDashPrefix = "data-";
         private const string TagHelperNameEnding = "TagHelper";
@@ -33,6 +33,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
 
         private static readonly ITypeInfo StringTypeInfo = new RuntimeTypeInfo(typeof(string).GetTypeInfo());
 
+#if !DOTNET5_4
+        private readonly TagHelperDesignTimeDescriptorFactory _designTimeDescriptorFactory;
+#endif
+        private readonly bool _designTime;
+
         // TODO: Investigate if we should cache TagHelperDescriptors for types:
         // https://github.com/aspnet/Razor/issues/165
 
@@ -40,22 +45,37 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'', '*' });
 
         /// <summary>
+        /// Instantiates a new <see cref="TagHelperDescriptorFactory"/>.
+        /// </summary>
+        /// <param name="designTime">
+        /// Indicates if <see cref="TagHelperDescriptor"/>s should be created for design time.
+        /// </param>
+        public TagHelperDescriptorFactory(bool designTime)
+        {
+#if !DOTNET5_4
+            if (designTime)
+            {
+                _designTimeDescriptorFactory = new TagHelperDesignTimeDescriptorFactory();
+            }
+#endif
+
+            _designTime = designTime;
+        }
+
+        /// <summary>
         /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="typeInfo"/>.
         /// </summary>
         /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
         /// <param name="typeInfo">The <see cref="ITypeInfo"/> to create a <see cref="TagHelperDescriptor"/> from.
         /// </param>
-        /// <param name="designTime">Indicates if the returned <see cref="TagHelperDescriptor"/>s should include
-        /// design time specific information.</param>
         /// <param name="errorSink">The <see cref="ErrorSink"/> used to collect <see cref="RazorError"/>s encountered
         /// when creating <see cref="TagHelperDescriptor"/>s for the given <paramref name="typeInfo"/>.</param>
         /// <returns>
         /// A collection of <see cref="TagHelperDescriptor"/>s that describe the given <paramref name="typeInfo"/>.
         /// </returns>
-        public static IEnumerable<TagHelperDescriptor> CreateDescriptors(
+        public virtual IEnumerable<TagHelperDescriptor> CreateDescriptors(
             string assemblyName,
             ITypeInfo typeInfo,
-            bool designTime,
             ErrorSink errorSink)
         {
             if (typeInfo == null)
@@ -68,12 +88,12 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 throw new ArgumentNullException(nameof(errorSink));
             }
 
-            if (ShouldSkipDescriptorCreation(designTime, typeInfo))
+            if (ShouldSkipDescriptorCreation(typeInfo))
             {
                 return Enumerable.Empty<TagHelperDescriptor>();
             }
 
-            var attributeDescriptors = GetAttributeDescriptors(typeInfo, designTime, errorSink);
+            var attributeDescriptors = GetAttributeDescriptors(typeInfo, errorSink);
             var targetElementAttributes = GetValidHtmlTargetElementAttributes(typeInfo, errorSink);
             var allowedChildren = GetAllowedChildren(typeInfo, errorSink);
 
@@ -83,8 +103,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                     assemblyName,
                     attributeDescriptors,
                     targetElementAttributes,
-                    allowedChildren,
-                    designTime);
+                    allowedChildren);
 
             return tagHelperDescriptors.Distinct(TagHelperDescriptorComparer.Default);
         }
@@ -99,24 +118,23 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
                 attribute => ValidHtmlTargetElementAttributeNames(attribute, errorSink));
         }
 
-        private static IEnumerable<TagHelperDescriptor> BuildTagHelperDescriptors(
+        private IEnumerable<TagHelperDescriptor> BuildTagHelperDescriptors(
             ITypeInfo typeInfo,
             string assemblyName,
             IEnumerable<TagHelperAttributeDescriptor> attributeDescriptors,
             IEnumerable<HtmlTargetElementAttribute> targetElementAttributes,
-            IEnumerable<string> allowedChildren,
-            bool designTime)
+            IEnumerable<string> allowedChildren)
         {
             TagHelperDesignTimeDescriptor typeDesignTimeDescriptor = null;
 
 #if !DOTNET5_4
-            if (designTime)
+            if (_designTime)
             {
                 var runtimeTypeInfo = typeInfo as RuntimeTypeInfo;
                 if (runtimeTypeInfo != null)
                 {
                     typeDesignTimeDescriptor =
-                        TagHelperDesignTimeDescriptorFactory.CreateDescriptor(runtimeTypeInfo.TypeInfo.AsType());
+                        _designTimeDescriptorFactory.CreateDescriptor(runtimeTypeInfo.TypeInfo.AsType());
                 }
             }
 #endif
@@ -387,10 +405,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             return validName;
         }
 
-        private static IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(
-            ITypeInfo type,
-            bool designTime,
-            ErrorSink errorSink)
+        private IEnumerable<TagHelperAttributeDescriptor> GetAttributeDescriptors(ITypeInfo type, ErrorSink errorSink)
         {
             var attributeDescriptors = new List<TagHelperAttributeDescriptor>();
 
@@ -400,7 +415,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
             var accessibleProperties = type.Properties.Where(IsAccessibleProperty);
             foreach (var property in accessibleProperties)
             {
-                if (ShouldSkipDescriptorCreation(designTime, property))
+                if (ShouldSkipDescriptorCreation(property))
                 {
                     continue;
                 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -40,22 +45,38 @@ public static class TagHelperDescriptorFactory
...
+
+        /// <summary>
         /// Creates a <see cref="TagHelperDescriptor"/> from the given <paramref name="typeInfo"/>.
         /// </summary>
         /// <param name="assemblyName">The assembly name that contains <paramref name="type"/>.</param>
         /// <param name="typeInfo">The <see cref="ITypeInfo"/> to create a <see cref="TagHelperDescriptor"/> from.
         /// </param>
-        /// <param name="designTime">Indicates if the returned <see cref="TagHelperDescriptor"/>s should include
-        /// design time specific information.</param>
+
```
dougbu: Is this still a design PR? If not remove weird blankie.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -84,7 +105,7 @@ public static class TagHelperDescriptorFactory
                     attributeDescriptors,
                     targetElementAttributes,
                     allowedChildren,
-                    designTime);
+                    _designTime);
```
dougbu: Why pass a field to a non-`static` method?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -17,8 +18,11 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// Factory for providing <see cref="TagHelperDesignTimeDescriptor"/>s from <see cref="Type"/>s and
     /// <see cref="TagHelperAttributeDesignTimeDescriptor"/>s from <see cref="PropertyInfo"/>s.
     /// </summary>
-    public static class TagHelperDesignTimeDescriptorFactory
+    public class TagHelperDesignTimeDescriptorFactory
     {
+        private readonly Dictionary<int, XmlDocumentationProvider> _documentationProviderCache =
+            new Dictionary<int, XmlDocumentationProvider>();
```
dougbu: To confirm: No possible concurrency issues given how instances of this class are shared?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -90,31 +92,52 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
             return null;
         }
 
-        private static DocumentationDescriptor CreateDocumentationDescriptor(Assembly assembly, string id)
+        private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
```
dougbu: Why did you split this `private` method in two? Makes it harder to see what really changed.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.AspNet.Razor.Runtime.TagHelpers
     /// <summary>
     /// Factory for <see cref="TagHelperDescriptor"/>s from <see cref="ITypeInfo"/>s.
     /// </summary>
-    public static class TagHelperDescriptorFactory
+    public class TagHelperDescriptorFactory
```
dougbu: Understand `TagHelperDescriptorFactory` doesn't itself hit the disk. But it's not clear caching at the lower `TagHelperDesignTimeDescriptorFactory` level is the way to do. This class is rather allocation-heavy and could contain the cached information.

dougbu:
Looks like this approach could get very confused if XML files are parsed before a build. Is the entire Razor host thrown away and rebuilt after every build? Put another way, what is the exact lifetime of the `TagHelperDesignTimeDescriptorFactory`'s cache?

dougbu:
:watch:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Still design.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: :+1:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: Hmmm, if the same Razor parser is invoked on two separate threads simultaneously then possibly. It ultimately depends on how the user is setting the `TagHelperDescriptorResolver`, by default it's set once per `MvcRazorHost`. Going to swap to `ConcurrentDictionary`.

NTaylorMullen: Actually, not enough. Going to lock around the dictionary instead

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: It hurt to look at the method as a single blob. Once my eyes started to bleed I immediately took action and split it out.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
NTaylorMullen: Don't want to cache any higher than the `TagHelperDesignTimeDescriptorFactory` because higher level caching prevents change in `TagHelperDescriptor`s from being picked up. The documentation of design time descriptors on the other hand is less important and has a significant perf hit.

NTaylorMullen:
> Looks like this approach could get very confused if XML files are parsed before a build. Is the entire Razor host thrown away and rebuilt after every build? Put another way, what is the exact lifetime of the TagHelperDesignTimeDescriptorFactory's cache?

Not sure I follow. 1 RazorHost is created per `.cshtml` file and is re-used for the lifetime of the file. It then has a custom `TagHelperDescriptorResolver` that offloads `TagHelperDescriptor` resolution to a separate process (RazorTooling). 

This cache is meant to be super low-impact as far as functionality is concerned because it'll be invalidated the next time the RazorTooling process is invoked to resolve new `TagHelperDescriptor`s. The value comes when there's N number of `TagHelper`s in a single assembly and they all have their XML documentation in the same file. We'd previously re-read the XML file and hit the disk for every `TagHelper` and then again for each of the `TagHelper`s properties, this prevents that.

NTaylorMullen:
:up::date:

dougbu:
> Not sure I follow.

My comment was a broader version of my "No possible concurrency issues" question. So you've covered my concern.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
dougbu: :ok:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorResolver.cs
```diff
@@ -137,7 +138,7 @@ public IEnumerable<TagHelperDescriptor> Resolve(TagHelperDescriptorResolutionCon
 
             // Convert types to TagHelperDescriptors
             var descriptors = tagHelperTypes.SelectMany(
-                type => TagHelperDescriptorFactory.CreateDescriptors(assemblyName, type, _designTime, errorSink));
+                type => _descriptorFactory.CreateDescriptors(assemblyName, type, errorSink));
```
pranavkm: Super unrelated, but would this benefit from un-Linq-ifying?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -90,31 +93,55 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
...
-            {
-                return null;
+                        // Normalize the path to UNC path. This will remove things like file:// from start of uri.Path.
+                        assemblyLocation = Uri.UnescapeDataString(uri.Path);
+                    }
+
+                    // Couldn't resolve a valid assemblyLocation.
+                    if (string.IsNullOrEmpty(assemblyLocation))
+                    {
+                        return null;
```
pranavkm: Should we cache misses?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -90,31 +93,55 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
...
+                    }
+
+                    var xmlDocumentationFile = GetXmlDocumentationFile(assembly, assemblyLocation);
```
pranavkm: Is this operation expensive? If so, maybe a `ConcurrentDictionary` would be a better alternative to the lock (since you wouldn't block all reads while you're constructing the documentation)

pranavkm:
Design looks good

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorResolver.cs
NTaylorMullen: Eh, not really. It's not hit nearly enough to make an impact.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: Hmm :+1:

NTaylorMullen:
:up::date:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -90,31 +93,51 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
...
+        private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
         {
-            var assemblyLocation = assembly.Location;
+            var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(assembly);
+            hashCodeCombiner.Add(CultureInfo.CurrentCulture);
+            var cacheKey = hashCodeCombiner.CombinedHash;
 
-            if (string.IsNullOrEmpty(assemblyLocation) && !string.IsNullOrEmpty(assembly.CodeBase))
+            var documentationProvider = _documentationProviderCache.GetOrAdd(cacheKey, valueFactory: (key) =>
```
pranavkm:
```C#
_documentationProviderCache.GetOrAdd(cacheKey, _ => 
{

};
```

since you don't actually use `key`

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -90,31 +93,51 @@ public static TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
             return null;
         }
 
-        private static DocumentationDescriptor CreateDocumentationDescriptor(Assembly assembly, string id)
+        private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
         {
-            var assemblyLocation = assembly.Location;
+            var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(assembly);
+            hashCodeCombiner.Add(CultureInfo.CurrentCulture);
```
pranavkm: In design time scenarios, can `CurrentCulture` change half way? Could we get away with simply varying by the `Assembly`?

pranavkm:
Might be nice to get a test or two to figure out if it actually caches. In particular if https://github.com/aspnet/Razor/pull/637/files#r47926057 is true, it would be nice to see it varies by culture. You could let it be if it's particularly tricky to mock the documentation types.

pranavkm:
:watch:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: Culture can change half way, yes. There's a setting in VS that exposes itself via `CultureInfo.CurrentCulture`.

pranavkm: Cool. Then let's try and get a test for this scenario so we know it caches per Assembly \ Culture.

NTaylorMullen:
:up::date: - Chose not to test the scenario of the documentation file changing half way through and then ensuring the original was cached as to not retrieve the changes. To test that we'd need to refactor/expose many `Factory`/`XmlDocumentationProvider` pieces. Didn't feel like that was worth it. If you feel otherwise let me know.

pranavkm:
Nah, the culture tests should suffice. 

:shipit:

Issue #625, opened 2015-11-30
`@section mySection {` creates an invisible error in the editor.


NTaylorMullen:
Technically an error is logged 1 character past the starting `{` but the editor can't show the red suiqggly since it's passed the EOF.

We should probably investigate other errorswhere +1 is used to ensure we don't surpass the end of the file.

Eilon:
Is this a problem only at the very end of the file?

NTaylorMullen:
Yes

Eilon:
So while you're typing at the end of the file you don't see an error? That doesn't sound super terrible, right?

NTaylorMullen:
You don't see a red squiggly but you do see a (1) pop up in your error dialog. Not horrible, but definitely inconsistent. Also need to consider scenarios where the user legitimately forgot to close out the section and continued writing.

Eilon:
What happens if there's content in the un-closed section?

NTaylorMullen:
Same deal. The error was built assuming the section had content following it.

Eilon:
Likely an easy fix? If so, I'm fine doing it. If not, backlog.

NTaylorMullen:
Yup, easy fix. However, we should probably take a look at all the errors that do similar behavior and fix them aswell.

Eilon:
Make it so.

NTaylorMullen:
392871beb62683bc366a8add4697af657b61a859

/cc @ToddGrun

PR #640, opened 2015-12-10
Change missing section end brace error to work at EOF.

NTaylorMullen:
- Updated existing tests and added a new case to understand `@section {....` scenarios.
- Instead of trying to guess 1 character prior to EOF decided to log error on opening brace since we know its position. Updated error to account for change in location.

#625

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
@@ -114,6 +114,8 @@ namespace Microsoft.AspNet.Razor.Parser
                 Accept(whitespace);
             }
 
+            var startingBraceLocation = CurrentLocation;
+
             // Set up edit handler
             var editHandler = new AutoCompleteEditHandler(Language.TokenizeString, autoCompleteAtEndOfSpan: true);
 
@@ -130,8 +132,11 @@ namespace Microsoft.AspNet.Razor.Parser
             {
                 editHandler.AutoCompleteString = "}";
                 Context.OnError(
-                    CurrentLocation,
-                    RazorResources.FormatParseError_Expected_X(Language.GetSample(CSharpSymbolType.RightBrace)),
+                    startingBraceLocation,
+                    RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(
+                        SyntaxConstants.CSharp.SectionKeyword,
+                        Language.GetSample(CSharpSymbolType.RightBrace),
+                        Language.GetSample(CSharpSymbolType.LeftBrace)),
                     length: 1 /* } */);
             }
             else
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -130,8 +132,11 @@ protected virtual void SectionDirective()
             {
                 editHandler.AutoCompleteString = "}";
                 Context.OnError(
-                    CurrentLocation,
-                    RazorResources.FormatParseError_Expected_X(Language.GetSample(CSharpSymbolType.RightBrace)),
+                    startingBraceLocation,
+                    RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF(
+                        SyntaxConstants.CSharp.SectionKeyword,
+                        Language.GetSample(CSharpSymbolType.RightBrace),
```
pranavkm: `Language.GetSample` wtf

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpSectionTest.cs
```diff
@@ -170,8 +170,30 @@ public void ParseSectionBlockHandlesEOFAfterOpenBrace()
                                .AutoCompleteWith("}", atEndOfSpan: true),
                         new MarkupBlock())),
                 new RazorError(
-                    RazorResources.FormatParseError_Expected_X("}"),
-                    new SourceLocation(14, 0, 14),
+                    RazorResources.FormatParseError_Expected_EndOfBlock_Before_EOF("section", "}", "{"),
```
pranavkm: Do we reference the error message in all the tests in this type (rather than handwriting it?)

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
```diff
@@ -130,8 +132,11 @@ protected virtual void SectionDirective()
             {
                 editHandler.AutoCompleteString = "}";
                 Context.OnError(
```
pranavkm: Does this work if you have nested braces?
```
@section Test
{
    @if (1 == 1)
   {
       <p>Hello world</p>
    }<EOF>
```

pranavkm:
:watch: for questions.

test/Microsoft.AspNet.Razor.Test/Parser/CSharp/CSharpSectionTest.cs
NTaylorMullen: yup lol

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: Yup, works. Added a test for this.

src/Microsoft.AspNet.Razor/Parser/CSharpCodeParser.Directives.cs
NTaylorMullen: lol, somewhat legacy, but it's just converting from Enum => string essentially. Didn't change this when we removed VB from Razor because it's still needed for the conversion.

NTaylorMullen:
:up: :date:

pranavkm:
:shipit:

PR #650, opened 2015-12-31
Rename TagHelperContent.Append(IHtmlContent` overload to be `TagHelpe…

### LINKED ISSUE #634: Rename `TagHelperContent.Append(IHtmlContent)` overload to be `TagHelperContent.AppendHtml(IHtmlContent)`
@DamianEdwards and I discussed the current implementation of `TagHelperContent.Append(IHtmlContent htmlContent)` and felt that it makes more sense to append an `IHtmlContent` via an `AppendHtml` method rather than an overload to `Append`. In `TagHelperContent`s case we already have an `AppendHtml` helper method that wraps `string`s to ensure they aren't encoded. We should just overload its existing `AppendHtml` to take an `IHtmlContent` and remove the current `Append(IHtmlContent)` overload.

It's already decently confusing when determining how to append encoded/unencoded content. This would help make it clearer since `IHtmlContent` is HTML.

pranavkm:
…rContent.AppendHtml(IHtmlContent)`

Fixes #634

### DIFF
--- a/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNet.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -142,7 +142,7 @@ namespace Microsoft.AspNet.Razor.TagHelpers
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


pranavkm:
cc @ajaybhargavb

Issue #614, opened 2015-11-18
Too much string.Concat in generated code with large blocks of HTML


rynowak:
Take a gander at the generated code for a page with a lot of 'literal text'

https://github.com/rynowak/mvc-sandbox/blob/master/workloads/RazorCodeGenerator/LargePage.cshtml

```
namespace Test
{
    using System;
    using System.Linq;
    using System.Collections.Generic;
    using Microsoft.AspNet.Mvc;
    using Microsoft.AspNet.Mvc.Rendering;
    using System.Threading.Tasks;

    public class LargePage : Microsoft.AspNet.Mvc.Razor.RazorPage<dynamic>
    {
        #line hidden
        public LargePage()
        {
        }
        #line hidden
        [Microsoft.AspNet.Mvc.Razor.Internal.RazorInjectAttribute]
        public Microsoft.AspNet.Mvc.IUrlHelper Url { get; private set; }
        [Microsoft.AspNet.Mvc.Razor.Internal.RazorInjectAttribute]
        public Microsoft.AspNet.Mvc.IViewComponentHelper Component { get; private set; }
        [Microsoft.AspNet.Mvc.Razor.Internal.RazorInjectAttribute]
        public Microsoft.AspNet.Mvc.Rendering.IJsonHelper Json { get; private set; }
        [Microsoft.AspNet.Mvc.Razor.Internal.RazorInjectAttribute]
        public Microsoft.AspNet.Mvc.Rendering.IHtmlHelper<dynamic> Html { get; private set; }

        #line hidden

        #pragma warning disable 1998
        public override async Task ExecuteAsync()
        {
            BeginContext(0, 160056, true);
            WriteLiteral("<p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><" +
"p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><" +
"p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><" +
"p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><" +
"p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><p>Paragraph</p><" +
[...]

Issue #635, opened 2015-12-09
Parsing and tokenizing large views thrashes CPU, Memory & GC


SimonOrdo:
I'm running RC1-Final.  Pages are served fine, but with a large view, the server gets so bogged down, apparently trying to parse all the tags, that the server complains and I end up with HTTP errors at times.

In this instance, I'm serving a view with no razor "@" tags and no kind of inline code.  Aside from two "section" tags, it's purely HTML and Javascript.  The view defines several data tables and is 188kb in size.  Here is what the start and end of requests look like in the VS performance tools (red arrows are start of request and end of request):

![image](https://cloud.githubusercontent.com/assets/15930167/11696711/4f41f122-9e7b-11e5-9b15-ef84f4cb2e15.png)

Memory snapshot before the page request
![image](https://cloud.githubusercontent.com/assets/15930167/11696754/8bf3e184-9e7b-11e5-9b55-f4bddb74e624.png)

Memory snapshot during the page request
![image](https://cloud.githubusercontent.com/assets/15930167/11696774/a37d6c9e-9e7b-11e5-8acf-5ce8daa9de3e.png)

Memory snapshot after the page request
![image](https://cloud.githubusercontent.com/assets/15930167/11696790/b8619040-9e7b-11e5-9a5d-0efbefd20f83.png)

Any ideas how to mitigate this beyond "write smaller views with fewer tags to parse" ?

rynowak:
The proposed fixes here:

**1). Rename `ISymbol` to `IToken` - because these are tokens not symbols** :trollface:

**2). Steal most of the design of Roslyn's tokens/symbols**

Specifically, token's shouldn't create a substring, they should have a span and a pointer to the original source. 

Additionally, `SyntaxNodes` should have a list of tokens, they should have a span of tokens and a pointer to the token stream.

rynowak:
Updated with some more info

Eilon:
@rynowak how related is this to https://github.com/aspnet/Razor/issues/614?

rynowak:
> @rynowak how related is this to #614?

They are both razor bugs we need to fix :+1: These are both issues you encounter with trying to use large razor pages that are usually completely blocking.

Eilon:
Cool. Nicholas will take care of these!

dougbu:
> For building of spans, it's the wrong approach to `ToString()` and then write it out

[...]

PR #652, opened 2016-01-05
Change string rendering to be chunked.

NTaylorMullen:
- Roslyn currently has an issue where too large of strings result in out of memory exceptions at compile time. To combat this I broke down literal strings into 1k pieces each resulting in their own `WriteLiteral`/`WriteLiteralTo` calls.
- Added tests to validate large string rendering.

#614

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -498,23 +498,6 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                         Write(literal[i].ToString());
                         break;
                 }
-                if (i > 0 && i % 80 == 0)
-                {
-                    // If current character is a high surrogate and the following
-                    // character is a low surrogate, don't break them.
-                    // Otherwise when we write the string to a file, we might lose
-                    // the characters.
-                    if (Char.IsHighSurrogate(literal[i])
-                        && (i < literal.Length - 1)
-                        && Char.IsLowSurrogate(literal[i + 1]))
-                    {
-                        Write(literal[++i].ToString());
-                    }
-
-                    Write("\" +");
-                    Write(NewLine);
-                    Write("\"");
-                }
             }
             Write("\"");
         }
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
         private const string ItemParameterName = "item";
         private const string ValueWriterName = "__razor_attribute_value_writer";
         private const string TemplateWriterName = "__razor_template_writer";
+        private const int MaxStringLiteralLength = 1024;
 
         private CSharpPaddingBuilder _paddingBuilder;
         private CSharpTagHelperCodeRenderer _tagHelperCodeRenderer;
@@ -134,17 +135,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
                 Writer.WriteStartInstrumentationContext(Context, start, text.Length, isLiteral: true);
             }
 
-            if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
-            {
-                RenderPreWriteStart();
-            }
-
-            Writer.WriteStringLiteral(text);
-
-            if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
-            {
-                Writer.WriteEndMethodInvocation();
-            }
+            RenderStartWriteLiteral(text);
 
             if (Context.Host.EnableInstrumentation)
             {
@@ -165,17 +156,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
                 Writer.WriteStartInstrumentationContext(Context, chunk.Association, isLiteral: true);
             }
 
-            if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
-            {
-                RenderPreWriteStart();
-            }
-
-            Writer.WriteStringLiteral(chunk.Text);
-
-            if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
-            {
-                Writer.WriteEndMethodInvocation();
-            }
+            RenderStartWriteLiteral(chunk.Text);
 
             if (Context.Host.EnableInstrumentation)
             {
@@ -575,25 +556,47 @@ namespace Microsoft.AspNet.Razor.CodeGenerators.Visitors
                    Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput;
         }
 
-        private CSharpCodeWriter RenderPreWriteStart()
+        private void RenderStartWriteLiteral(string text)
         {
-            return RenderPreWriteStart(Writer, Context);
-        }
+            var charactersRendered = 0;
 
-        public static CSharpCodeWriter RenderPreWriteStart(CSharpCodeWriter writer, CodeGeneratorContext context)
-        {
-            if (!string.IsNullOrEmpty(context.TargetWriterName))
+            // Render the string in pieces to avoid Roslyn OOM exceptions at compile time:
+            // https://github.com/aspnet/External/issues/54
+            while (charactersRendered < text.Length)
             {
-                writer.WriteStartMethodInvocation(context.Host.GeneratedClassContext.WriteLiteralToMethodName)
-                      .Write(context.TargetWriterName)
-                      .WriteParameterSeparator();
-            }
-            else
-            {
-                writer.WriteStartMethodInvocation(context.Host.GeneratedClassContext.WriteLiteralMethodName);
-            }
+                if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
+                {
+                    if (!string.IsNullOrEmpty(Context.TargetWriterName))
+                    {
+                        Writer.WriteStartMethodInvocation(Context.Host.GeneratedClassContext.WriteLiteralToMethodName)
+                              .Write(Context.TargetWriterName)
+                              .WriteParameterSeparator();
+                    }
+                    else
+                    {
+                        Writer.WriteStartMethodInvocation(Context.Host.GeneratedClassContext.WriteLiteralMethodName);
+                    }
+                }
 
-            return writer;
+                string textToRender;
+                if (text.Length <= MaxStringLiteralLength)
+                {
+                    textToRender = text;
+                }
+                else
+                {
+                    var charactersToSubstring = Math.Min(MaxStringLiteralLength, text.Length - charactersRendered);
+                    textToRender = text.Substring(charactersRendered, charactersToSubstring);
+                }
+
+                Writer.WriteStringLiteral(textToRender);
+                charactersRendered += textToRender.Length;
+
+                if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
+                {
+                    Writer.WriteEndMethodInvocation();
+                }
+            }
         }
     }
 }
--- a/test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpRazorChunkGeneratorTest.cs
+++ b/test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpRazorChunkGeneratorTest.cs
@@ -52,6 +52,7 @@ namespace Microsoft.AspNet.Razor.Test.Generator
         }
 
         [Theory]
+        [InlineData("StringLiterals")]
         [InlineData("NestedCSharp")]
         [InlineData("NullConditionalExpressions")]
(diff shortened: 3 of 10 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -16,6 +15,7 @@ public class CSharpCodeVisitor : CodeVisitor<CSharpCodeWriter>
         private const string ItemParameterName = "item";
         private const string ValueWriterName = "__razor_attribute_value_writer";
         private const string TemplateWriterName = "__razor_template_writer";
+        private static readonly int MaxStringLiteralLength = 499; // Max 1k string size despite x86/x64
```
NTaylorMullen: Decided on 1k. Definitely open to other suggestions, this in all reality is just one of the many magic numbers that fit in the small object heap.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -470,23 +470,6 @@ private void WriteCStyleStringLiteral(string literal)
                         Write(literal[i].ToString());
                         break;
                 }
```
NTaylorMullen: If we want we could always render C# strings always as `@"hello world"` or `"\r\nhello\tworld"` but thought I'd reduce the amount of change by just nuking the below code that interferes with what the issue was geared towards. In the end this bit of code was from CodeDOM and I believe a lot of it was for readability.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -16,6 +15,7 @@ public class CSharpCodeVisitor : CodeVisitor<CSharpCodeWriter>
         private const string ItemParameterName = "item";
         private const string ValueWriterName = "__razor_attribute_value_writer";
         private const string TemplateWriterName = "__razor_template_writer";
+        private const int MaxStringLiteralLength = 1024;
```
NTaylorMullen: 1024 to correspond with the http contexts response stream [buffer size](https://github.com/aspnet/Mvc/blob/47a3aee2c152b44bb75075e1f7494ba2124c82fa/src/Microsoft.AspNet.Mvc.Core/HttpResponseStreamWriter.cs#L24)

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Suggest also removing `literal.Length >= 256 && literal.Length <= 1500` from [line 118](https://github.com/aspnet/Razor/blob/nimullen/stringconcats.614/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs#L118). A verbatim `string` is fine for anything lacking a `'\0'` character.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -581,6 +561,31 @@ private CSharpCodeWriter RenderPreWriteStart()
...
+                charactersRendered += textToRender.Length;
+
+                if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
+                {
+                    Writer.WriteEndMethodInvocation();
+                }
+            }
+        }
+
         public static CSharpCodeWriter RenderPreWriteStart(CSharpCodeWriter writer, CodeGeneratorContext context)
```
dougbu: Remove this `static` method and inline it into the unhelpful `RenderPreWriteStart()` method. Better if that's renamed to something meaningful e.g. `RenderStartWriteLiteralMethodInvocation()`. "pre write" is meaning-challenged.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -581,6 +561,31 @@ private CSharpCodeWriter RenderPreWriteStart()
             return RenderPreWriteStart(Writer, Context);
         }
 
+        private void RenderChunkedWriteLiteral(string text)
+        {
+            var charactersRendered = 0;
+
+            // Render the string in pieces to avoid Roslyn OOM exceptions at compile time:
+            // https://github.com/dotnet/roslyn/issues/7398
```
dougbu: Use an indirection through aspnet/External. That's where we track non-aspnet issues and how they're referenced in our code.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -581,6 +561,31 @@ private CSharpCodeWriter RenderPreWriteStart()
...
+            while (charactersRendered < text.Length)
+            {
+                if (Context.ExpressionRenderingMode == ExpressionRenderingMode.WriteToOutput)
+                {
+                    RenderPreWriteStart();
+                }
+
+                var charactersToSubstring = Math.Min(MaxStringLiteralLength, text.Length - charactersRendered);
+                var textToRender = text.Substring(charactersRendered, charactersToSubstring);
+                Writer.WriteStringLiteral(textToRender);
```
dougbu: Ick.
1. Don't `Substring()` if `text.Length <= MaxStringLiteralLength`.
2. Otherwise call `text.ToCharArray()` once before the loop and pass the array, start index, and length into the `Writer`.
3. Expose the necessary wrapper methods in `CodeWriter` (`WriteStringLiteral()`) and `CSharpCodeWriter` (`Write()`). Those APIs are anemic compared to the underlying `StringWriter` and lousy for this use case.

dougbu:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
rynowak: I agree with the Doug. :maple_leaf:

rynowak:
:watch:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
NTaylorMullen: Talked offline. Most of the perf changes for strings will be handled in [this](https://github.com/aspnet/Razor/issues/635) issue.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
NTaylorMullen: Chose to not remove the length restrictions to not muck with all the generated files in this PR. Also because the same code paths will be touched in [this](https://github.com/aspnet/Razor/issues/635) issue.

NTaylorMullen:
:up::date:

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -576,25 +556,38 @@ private bool ShouldGenerateInstrumentationForExpressions()
...
+                    }
+                    else
+                    {
+                        Writer.WriteStartMethodInvocation(Context.Host.GeneratedClassContext.WriteLiteralMethodName);
+                    }
+                }
 
-            return writer;
+                var charactersToSubstring = Math.Min(MaxStringLiteralLength, text.Length - charactersRendered);
+                var textToRender = text.Substring(charactersRendered, charactersToSubstring);
```
dougbu: There's a difference between deferring perf work and adding a useless `Substring()` to the mainline code path. Keep the `text.Length < MaxStringLiteralLength` case. That's the code path that'll be used 99% of the time this method is called.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -470,23 +470,6 @@ private void WriteCStyleStringLiteral(string literal)
...
-                        && (i < literal.Length - 1)
-                        && Char.IsLowSurrogate(literal[i + 1]))
-                    {
-                        Write(literal[++i].ToString());
-                    }
-
-                    Write("\" +");
-                    Write(NewLine);
-                    Write("\"");
-                }
```
dougbu: Note this change is unrelated to issue #614. Affects only `string`s (or the last `Substring`) of 256 `char`s or less, nothing that would hit the Roslyn issue.

dougbu: Probably fine. Just worth mentioning separately in the squished (or squashed) commit description.

test/Microsoft.AspNet.Razor.Test/CodeGenerators/CSharpRazorChunkGeneratorTest.cs
```diff
@@ -52,6 +52,7 @@ public void ConstructorAllowsEmptyRootNamespaceName()
         }
 
         [Theory]
+        [InlineData("StringLiterals")]
         [InlineData("NestedCSharp")]
         [InlineData("NullConditionalExpressions")]
         [InlineData("NestedCodeBlocks")]
```
dougbu: Sort alphabetically :trollface:

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/StringLiterals.cs
```diff
@@ -0,0 +1,238 @@
+#pragma checksum "StringLiterals.cshtml" "{ff1816ec-aa5e-4d10-87f7-6f4963833460}" "4bc2accb98c9c1f52bff719978da3f8449b7c9c6"
```
dougbu: This is one test of a brand new code path. Suggest adding a `CSharpCodeVisitorTest` or extending the `CSharpCodeGeneratorTest` class with a few unit tests at least for `Visit(ParentLiteralChunk)` and `Visit(LiteralChunk)`. Check near boundary conditions such as `string`s of length `0`, `1`, `255`, `256`, `257`, `1023`, `1024`, `1025`, ... Of course the middle few cases might be more interesting in the `CSharpCodeWriterTest` class.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Then again, no existing tests would change if you undid this. As-is, need a MVC reaction PR. Don't see one of those...

dougbu:
:watch:
I don't have a solid recommendation in the [discussion of "extra" change](https://github.com/aspnet/Razor/pull/652#discussion-diff-49003626) but know a MVC reaction PR is needed as-is. Also looking for more tests and a fix [here](https://github.com/aspnet/Razor/pull/652#discussion-diff-49002871).

test/Microsoft.AspNet.Razor.Test/TestFiles/CodeGenerator/Output/StringLiterals.cs
NTaylorMullen: This test covers most of those. Going to modify so 1 and 1023 are covered.

NTaylorMullen:
:up: :date:

Also did not need to do an MVC reaction PR. Not many code gen tests and none of them have literals longer than 80 chars.

src/Microsoft.AspNet.Razor/CodeGenerators/Visitors/CSharpCodeVisitor.cs
```diff
@@ -576,25 +556,47 @@ private bool ShouldGenerateInstrumentationForExpressions()
...
+                    }
+                    else
+                    {
+                        Writer.WriteStartMethodInvocation(Context.Host.GeneratedClassContext.WriteLiteralMethodName);
+                    }
+                }
 
-            return writer;
+                string textToRender;
+                if (text.Length <= MaxStringLiteralLength)
```
dougbu: Slightly odd to check this w/in a loop but :ok:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Guess you decided to stick w/ this change :8ball:

dougbu: Oops, misplaced.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Guess you decided to stick w/ this change...

dougbu:
:shipit:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
NTaylorMullen: Si :smile:

rynowak:
:shipit:

Issue #643, opened 2015-12-14
Option to bypass GetContent HTML encoding in TagHelper


khalidabuhakmeh:
The current implementation of `TagHelperContent.GetContent` automatically HTML encodes when called from within the `TagHelper` implementation. Using [Dave Paquette's](http://www.davepaquette.com/archive/2015/11/02/markdown-in-your-mvc-6-razor-pages.aspx) implementation of Markdown `Taghelper`, I am able to reproduce the unexpected behavior of *premature encoding*.

### Implementation

```csharp
[HtmlTargetElement("p", Attributes = "markdown")]
[HtmlTargetElement("markdown")]
[OutputElementHint("p")]
public class MarkdownTagHelper : TagHelper
{
    public async override Task ProcessAsync(TagHelperContext context, TagHelperOutput output)
    {
        if (output.TagName == "markdown")
        {
            output.TagName = "p";
        }
        var childContent = await context.GetChildContentAsync();

        // This call to get markdownContent is HTML encoded
        // prematurely and strips meaningful newline characters
        // necessary for the markdown output
        var markdownContent = childContent.GetContent();

        var markdown = new MarkdownSharp.Markdown();
        var htmlContent = markdown.Transform(markdownContent);
        output.Content.SetContentEncoded(htmlContent);
    }
}
```

### Input

```markdown
### Features\r\n\r\n* API namespace cleanup for tag helpers ([#578](https://github.com/aspnet/Razor/issues/578))\r\n* Refactor WriteAttribute API surface. ([#177](https://github.com/aspnet/Razor/issues/177))\r\n\r\n### Bugs Fixed\r\n\r\n* Multiple TagHelpers targeting same element cannot effectively communicate to children. ([#571](https://github.com/aspnet/Razor/issues/571))\r\n* Compilation error within tag attribute shows generated code instead of markup ([#569](https://github.com/aspnet/Razor/issues/569))\r\n* Specifying `RestrictChildren` and empty `HtmlTargetElement` results in an error ([#562](https://github.com/aspnet/Razor/issues/562))\r\n* `TreeStructureChanged` marked as `true` when whitespace is added to a page with `TagHelper`s ([#553](https://github.com/aspnet/Razor/issues/553))\r\n* Expose `FilePath` on `MappingLocation` to let Razor editor know mapping origination. ([#552](https://github.com/aspnet/Razor/issues/552))\r\n\r\n
```

### Expected

[...]

PR #653, opened 2016-01-06
Add `TagHelperOutput.GetChildContentAsync()` overloads including an `HtmlEncoder`

dougbu:
- #643
- change is viral and requires an update to `RazorPage.StartTagHelperWritingScope()`
- note `HtmlEncoder`s used elsewhere e.g. in other `RazorPage` instances are unaffected

WIP:
- new tests of the new `TagHelperOutput` methods yet
- test do not pass a non-`null` `HtmlEncoder` into `TagHelperScopeManager.Begin()`

Nit: remove unused `using`s in files I had open


dougbu:
/cc @NTaylorMullen 
FYI @DamianEdwards this looks fairly ugly.

DamianEdwards:
Which bit looks ugly? The implementation or the consumption?

dougbu:
> The implementation or the consumption?

Bit of both though mostly the additional cases that are unaffected (things I noticed while doing the implementation). Also have a look at the MVC reaction I'm about to open...

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -224,18 +228,24 @@ public void AddTagHelperAttribute(string name, object value)
...
 
         /// <summary>
-        /// Execute and retrieve the rendered child content asynchronously.
+        /// Execute children asynchronously with the given <paramref name="encoder"/> in scope and return their
+        /// rendered content.
         /// </summary>
+        /// <param name="useCachedResult">If <c>true</c> multiple calls to this method will not cause re-execution
+        /// of child content; cached content will be returned.</param>
+        /// <param name="encoder">
+        /// The <see cref="HtmlEncoder"/> to use. Does not override all HTML encodings which may occur.
```
DamianEdwards: This statement isn't particularly helpful without more detail is it? What other HTML encodings?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -224,18 +228,24 @@ public void AddTagHelperAttribute(string name, object value)
...
+        /// </param>
         /// <returns>A <see cref="Task"/> that on completion returns the rendered child content.</returns>
         /// <remarks>
         /// Child content is only executed once. Successive calls to this method or successive executions of the
         /// returned <see cref="Task{TagHelperContent}"/> return a cached result.
         /// </remarks>
-        public async Task<TagHelperContent> GetChildContentAsync(bool useCachedResult)
+        public async Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEncoder encoder)
```
DamianEdwards: I think we intended to have overloads, no? If you call the original overload if encodes using the page encoder. Call this one and it uses the one passed in.

NTaylorMullen: This is at the `TagHelperExecutionContext` level, the overloads you're expecting exist in `TagHelperOutput`

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -42,7 +43,10 @@ internal TagHelperExecutionContext(string tagName, TagMode tagMode)
         /// <see cref="ITagHelper"/>s</param>
         /// <param name="uniqueId">An identifier unique to the HTML element this context is for.</param>
         /// <param name="executeChildContentAsync">A delegate used to execute the child content asynchronously.</param>
-        /// <param name="startTagHelperWritingScope">A delegate used to start a writing scope in a Razor page.</param>
+        /// <param name="startTagHelperWritingScope">
+        /// A delegate used to start a writing scope in a Razor page and to override the page's current
```
NTaylorMullen: Hmm, when mentioning overriding the `HtmlEncoder` on the page it may be worth adding a remarks or additional detail stating that a `null` value results in no value being overridden.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
DamianEdwards: Ah OK. Those appear to be passing in `null` though. How do we differentiate between actually passing `null` to indicate "no encoding" and calling it without specifying an encoder, in which case the page's encoder should be used?

NTaylorMullen: Hmm, i thought we were going to use `null` to indicate we shouldn't override anything on the page. With this approach we can special case `null` in the `HtmlEncoder` `GetChildContentAsync` override to pass in a non-encoding encoder.

The other alternative would be to pass through additional parameters indicating if you truly want encoding. Neither way is that nice.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -206,24 +208,53 @@ public void SuppressOutput()
...
+        /// rendered content.
+        /// </summary>
+        /// <param name="encoder">
+        /// The <see cref="HtmlEncoder"/> to use. Does not override all HTML encodings which may occur.
+        /// </param>
+        /// <returns>A <see cref="Task"/> that on completion returns content rendered by children.</returns>
+        /// <remarks>This method is memoized.</remarks>
+        public Task<TagHelperContent> GetChildContentAsync(HtmlEncoder encoder)
+        {
+            return GetChildContentAsync(useCachedResult: true, encoder: encoder);
```
NTaylorMullen: If encoder is `null` we may need to do something based on [this](https://github.com/aspnet/Razor/pull/653/files#r48931958) conversation. Same with next overload.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Better alternative: Take in some sort of `HtmlEncoder` accessor for the current page in the `TagHelperScopeManager` and just pass that in the `GetChildContentAsync()` and `GetChildContentAsync(bool useCachedResult)` overloads. That would enable `null` to represent "do not encode".

DamianEdwards: Oh i see. If the outer method is called with null, we'd pass a no-op encoder to this level, yeah? It'll be null here if the outer method's default overload (no encoder param) is called.

NTaylorMullen: Correct. I prefer my [alternative](https://github.com/aspnet/Razor/pull/653/files#r48932143) though.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: > What other HTML encodings?

1. Another `RazorView`'s or `RazorPage`'s content
2. A view component's content
3. URLs `UrlResolutionTagHelper` and its subclasses touch
4. `@Html.BeginForm()`'s content
5. Anything written out using `FormContext.EndOfFormContent`
6. ...

Do you want the above list in these comments?

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
DamianEdwards: Yeah I thought your alternative is what we'd originally talked about in December (basically capture the page's encoder). Assuming you're talking about at this level, not on `TagHelperOutput`..

dougbu: You both want `RazorPage.HtmlEncoder` to be set to `null` and to special-case that where it's referenced, including in generated code? That plus the accessor gunk is an additional pile of code for this corner scenario over what's here..

NTaylorMullen: Ya thinking back, I can recall that. Only push back at the time was the pattern of a `null` encoder being passed around indicating "do not encode" would then add an inconsistency to the other encoding mechanics in MVC. Would need to be applied at a broader level most likely.

I feel like it'd be a pretty positive change overall though.

dougbu: > pretty positive change overall though

What's positive about introducing these inconsistencies?

NTaylorMullen: It'd be an inconsistency that we'd then change across the board to make passing `null` anywhere result in "do not encode".

NTaylorMullen:
:watch:

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: ##### Details galore
The following MVC classes store `HtmlEncoder` instances. This PR affects only the first use.

1. `RazorPage` uses an encoder for C# expressions in `string` tag helper attributes and when writing non-`IHtmlContent` C# expressions out.
2. `UrlResolutionTagHelper` uses an encoder to determine the original `string` value of a URL that includes C# expressions. The encoded value flows through.
 1. `LinkTagHelper` and `ScriptTagHelper` use the base `UrlResolutionTagHelper`'s encoder on the calculated URL when `asp-append-version="true"`.
3. `RazorViewEngine` uses an encoder when instantiating `RazorView` instances.
 1. `RazorView` uses that encoder when instantiating `RazorTextWriter` instances and when copying buffered content to an outer `TextWriter`.
 2. `RazorTextWriter` uses that encoder during and sometimes after `Flush[Async]()` operations.
4. `CacheTagHelper` uses an encoder to calculate the cached content. (Can't store the `IHtmlContent` for the body because that may come from a per-request pool.)
5. `DefaultViewComponentHelper` uses an encoder when instantiating `ViewComponentContext` instances.
 1. `ContentViewComponentResult` and `HtmlContentViewComponentResult` use the `ViewComponentContext` encoder when writing out to a non-`HtmlTextWriter` writer. This won't normally occur when the view component is invoked in a `.cshtml` file but can occur after a `Flush[Async]()` or if the view component is invoked directly from an action. (These class are used to handle specific view component return values.)
6. `ViewComponentResult` uses an encoder when writing the content out to the response stream. (This class is used when a view component is invoked directly from an action.)
7. `DefaultHtmlGenerator` uses an encoder in its support of `Html.Encode()`.
8. `HtmlHelper` uses an encoder when generating a `<form>` start tag.
9. `MvcForm` uses an encoder when writing out `FormContext.EndOfFormContent` to a non-`HtmlTextWriter` writer. However the writer should almost-always be a `HtmlTextWriter`.

DamianEdwards: Yes @dougbu, I want you to list all of that in the comment :trollface: 

Perhaps find a way to invert the wording such that it's clear what is being encoded, rather highlighting what isn't.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: > It'd be an inconsistency that we'd then change across the board to make passing null anywhere result in "do not encode".

Sorry I wasn't clear; was talking about the conversation up [here](https://github.com/aspnet/Razor/pull/653#discussion-diff-48931569).

In any case, there are two ways forward. @DamianEdwards and @NTaylorMullen, thoughts?

##### Use a `NullHtmlEncoder`
We or tag helper authors that are looking for no-op behaviour when calling `output.GetChildContentAsync` could create a relatively simple `NullHtmlEncoder`. We already have the [`HtmlTestEncoder`](https://github.com/aspnet/HtmlAbstractions/blob/dev/src/Microsoft.Extensions.WebEncoders/Testing/HtmlTestEncoder.cs) class in a production package (`Microsoft.Extensions.WebEncoders`) albeit in a special `...Testing` namespace. Implementation of `NullHtmlEncoder` would be similar.

This class (if we provide it) gives us an opportunity for further warnings about what will and won't be encoded, specifically around HTML-encoded values users type in directly. It also requires no changes to the approach shown in this PR and aspnet/MVC#3871 (except writing the class itself; again, if we provide it).

If we provide it, main questions on the class would be where to place it: Which namespace and which package? I'd suggest `System.Text.Encodings.Web` and the `Microsoft.AspNet.Razor.Runtime` package (i.e. same package as `TagHelperOutput`).

##### Special-case a `null` `HtmlEncoder` value
Requires rework of this PR to make an accessor for the default `HtmlEncoder` available in `TagHelperOutput`. Requirement would bubble up through `TagHelperRunner` and `TagHelperExecutionContext` into `TagHelperScopeManager`. From there would require generation changes for either the `TagHelperScopeManager` instantiation or all `TagHelperScopeManager.Begin()` calls. (First is a little more straightforward since the accessor never needs to change. But should probably be consistent with handling of `RazorPage.StartTagHelperWritingScope()` and `RazorPage.EndTagHelperWritingScope()`.)

Also requires changes to _every_ `IHtmlContent.WriteTo()` implementation across the HtmlAbstractions, Razor and MVC repos. Need to both remove `ArgumentNullException` cases and do something different when the passed encoder is `null`. This is a significant contract change that can't be enforced in 3rd party `IHtmlContent` implementations i.e. they'll break long after, when first used with a tag helper that calls `output.GetChildContentAsync(encoder: null)`.

Note I'm not whining this approach is lots of work (ok, not only whining). It's a lot for a small scenario but the contract changes are a larger concern. In addition `null` normally means "default" in our APIs, not "do this special other thing".

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: If you're :ok: with the exceptions, I'm :ok: with inverting the wording. Probably something like `The <see cref="HtmlEncoder"/> to use when the page handles non-<see cref="IHtmlContent"/> C# expressions.`

DamianEdwards: Why wouldn't the specified encoder be passed through to encountered `IHtmlContent` expressions?

dougbu: Because evaluating `IHtmlContent` is deferred until `WriteTo()` is called. In some cases however, this encoder _will_ be used e.g. when `GetContent()` is called on the buffer behind the `TextWriter` in a nested tag helper scope.

That raises another overlapping issue: Though changing this `HtmlEncoder` has little affect on other pages or views, it does affect all nested tag helpers. E.g.
``` html
  <outer comment="Calls GetChildContentAsync(NullHtmlEncoder) in its Process() method">
    <inner comment="Calls GetChildContentAsync(NullHtmlEncoder) in its Process() method">
      content
    </inner>
   </outer>
```
`InnerTagHelper` will _not_ see fully HTML-encoded body content.

dougbu: Corrected typo in last comment.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Forgot the 3rd approach ([mentioned](https://github.com/aspnet/Razor/issues/643#issuecomment-164317532) in #643).

##### Recommend `@Html.Raw()`
E.g. `<markdown>@Html.Raw(release.Body)</markdown>` or (as the original poster did) `<markdown>@(new HtmlString(release.Body))</markdown>`. This puts control in the `.cshtml` author's hands and tasks tag helper authors mainly with writing good doc comments for IntelliSense.

One upside is this maintains consistency w.r.t. to what is HTML encoded i.e. everything that isn't either in the `.cshtml` file itself or specially marked is encoded. The downside here is mainly discoverability, especially if the tag helper author does not write good comments or IntelliSense isn't working.

dougbu: FYI @DamianEdwards discussed the approach offline. Decided to go with __Use a `NullHtmlEncoder`__ because it doesn't just paper over #643 (as __Recommend `@Html.Raw()`__ does) and is somewhat minimal.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: FYI @DamianEdwards touched on the following bit when we talked offline.
> Though changing this `HtmlEncoder` has little affect on other pages or views, it does affect all nested tag helpers.

Main change we'll make is to disable memoization of the `GetChildContentAsync()` call by default if user calls the `GetChildContentAsync(HtmlEncoder)` overload.

I will go a bit further and pass `useCachedResult: encoder != null` to the full overload in that method. That way `GetChildContentAsync(encoder: null)` will behave :100:% identically to `GetChildContentAsync()`.

dougbu: Had another offline convo, this time w/ @NTaylorMullen. We went through a zillion and three corner cases. Ends up much more straightforward and easier to explain when memoization is per-encoder. Going with that and will update this PR soon to see how it really looks.

dougbu:
:up::date:

Still WIP since new stuff isn't tested. Working on adding tests in aspnet/Mvc#3871 to show something fairly end-to-end.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -708,7 +711,7 @@ private void RenderBufferedAttributeValueAccessor(CSharpCodeWriter writer)
                     StringValueBufferVariableName,
                     _tagHelperContext.TagHelperContentGetContentMethodName,
                     endLine: false,
-                    parameters: new string[] { _tagHelperContext.HtmlEncoderPropertyName });
+                    parameters: _tagHelperContext.HtmlEncoderPropertyName);
```
dougbu: Minor nit I fixed in passing. Array is allocated regardless but this way is easier on the eyes.

src/Microsoft.AspNet.Razor.Runtime/project.json
```diff
@@ -2,8 +2,9 @@
   "description": "Runtime components for rendering Razor pages.",
   "version": "4.0.0-*",
   "compilationOptions": {
-    "warningsAsErrors": true,
-    "keyFile": "../../tools/Key.snk"
+    "allowUnsafe": true,
```
dougbu: Unfortunately required to compile any `HtmlEncoder` subclass.

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
```diff
@@ -0,0 +1,118 @@
...
+
+using System.IO;
+
+namespace System.Text.Encodings.Web
+{
+    /// <summary>
+    /// A <see cref="HtmlEncoder"/> that does not encode. Should not be used when writing directly to a response
+    /// expected to contain valid HTML.
+    /// </summary>
+    public class NullHtmlEncoder : HtmlEncoder
```
NTaylorMullen: Might make more sense in Html abstractions.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -659,7 +659,9 @@ private void BuildBufferedWritingScope(Chunk htmlAttributeChunk, bool htmlEncode
                 // Scopes are a runtime feature.
                 if (!_designTimeMode)
                 {
-                    _writer.WriteMethodInvocation(_tagHelperContext.StartTagHelperWritingScopeMethodName);
+                    _writer.WriteMethodInvocation(
+                        _tagHelperContext.StartTagHelperWritingScopeMethodName,
+                        "encoder: null");
```
dougbu: Likely better just to pass an unnamed `null`. Any pushback on that change?

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        /// <inheritdoc />
+        public override unsafe int FindFirstCharacterToEncode(char* text, int textLength)
+        {
+            return -1;
+        }
+
+        /// <inheritdoc />
+        public override unsafe bool TryEncodeUnicodeScalar(
```
NTaylorMullen: So sad that this is abstract. I'd think that `WillEncode(int unicodeScalar)` returning false would result in this never being called.

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
dougbu: Why? It's specifically for this Razor / MVC feature.

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
dougbu: Following `HtmlTestEncoder` pattern.

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
NTaylorMullen: An encoder that doesn't encode would be more general use imo since we're now all using `IHtmlContent` whos only signature expects an encoder.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -224,23 +229,57 @@ public void AddTagHelperAttribute(string name, object value)
...
+        /// If <c>true</c>, multiple calls with the same <see cref="HtmlEncoder"/> will not cause children to
+        /// re-execute; returns cached content.
+        /// </param>
+        /// <param name="encoder">
+        /// The <see cref="HtmlEncoder"/> to use when the page handles
+        /// non-<see cref="Microsoft.AspNet.Html.IHtmlContent"/> C# expressions. If <c>null</c>, executes children with
+        /// the page's current <see cref="HtmlEncoder"/>.
+        /// </param>
         /// <returns>A <see cref="Task"/> that on completion returns the rendered child content.</returns>
-        /// <remarks>
```
NTaylorMullen: Somewhere in this doc comment we should mention that content is cached based on the encoder.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -206,24 +208,65 @@ public void SuppressOutput()
...
+        /// Executes children asynchronously with the given <paramref name="encoder"/> in scope and returns their
+        /// rendered content.
+        /// </summary>
+        /// <param name="encoder">
+        /// The <see cref="HtmlEncoder"/> to use when the page handles non-<see cref="IHtmlContent"/> C# expressions.
+        /// If <c>null</c>, executes children with the page's current <see cref="HtmlEncoder"/>.
+        /// </param>
+        /// <returns>A <see cref="Task"/> that on completion returns content rendered by children.</returns>
+        /// <remarks>
+        /// This method is memoized. Multiple calls with the same <see cref="HtmlEncoder"/> instance will not cause
```
NTaylorMullen: Should make some mention of this in the summary stating that content may be re-executed based on the encoder provided.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Ahhh, ya I wouldn't. It showcases a parameter that we're adding to fulfill an edge case. Don't want to make people think it's the right overload to use all the time.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: :+1:

NTaylorMullen:
@DamianEdwards one thing to note here. With this PR as it is this code sample would work as follows:

```HTML
@{
    var i = 0;
}
<myth>@($"<p>{i++}</p>")</myth>
```

Where 2 TagHelpers target myth: `MyTH1` (runs first) and `MyTH2`.

1. `MyTH1` calls `GetChildContentAsync` with a custom encoder that only encodes open angles it'd retrieve `&lt;p>0&lt;p>`.
1. `MyTH2` then calls `GetChildContentAsync` with the null encoder that doesn't encode anything, it'd retrieve `<p>1<p>`.
1. Neither one of them set `output.Content`, essentially they were both just inspecting the content.

The final output would be: `<myth>&lt;p&gt;2&lt;/p&gt;</myth>`.

Reason being: Each different encoder results in the body being re-executed with one final call to `output.GetChildContentAsync()` when the generated code detects that `output.Content` was not modified (and therefore uses the pages current `HtmlEncoder` when running the content).

This seems fine to me but wanted to raise the functionality.

src/Microsoft.AspNet.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: `multiple calls with the same <see cref="HtmlEncoder"/> will not cause children to re-execute` in `useCachedResult` description.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: That is made fairly explicit by `with the same <see cref="HtmlEncoder"/> instance`.

NTaylorMullen: Oh I agree, it just doesn't show up in IntelliSense. I'm simply suggesting moving that explanation to somewhere more visible to the user.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: There's no other overload of the `RazorPage.StartTagHelperWritingScope()` method. It's only called from generated code (though through a delegate in its only other generated use). Why would a user call that `RazorPage` method?

In any case, I was just talking about changing the generated parameter here from `encoder: null` to `null`.

NTaylorMullen: part of me would rather an overload without one. But I guess that's overkill for just this. Ya don't do named params here.

src/Microsoft.AspNet.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: Not sure that detail is interesting for normal use and it seems we agreed earlier. We made the decision to leave the old `This method is memoized.` in `<remarks>`.

src/Microsoft.AspNet.Razor.Runtime/Encodings/NullHtmlEncoder.cs
dougbu: As discussed, will move to `Microsoft.AspNet.Razor.TagHelpers` namespace and match `TagHelperOutput`.

dougbu:
Small :up::date:

dougbu:
:up::date: no longer in design (added tests)

NTaylorMullen:
Waiting for confirmation from @DamianEdwards on https://github.com/aspnet/Razor/pull/653#issuecomment-170088510

dougbu:
Ping @DamianEdwards. Blocked on your decision whether the implemented semantics work for you.

DamianEdwards:
:shipit:

test/Microsoft.AspNet.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -65,30 +80,73 @@ public void ParentItems_SetsItemsProperty()
...
-            Assert.Equal(expectedContent, content1.GetContent(new HtmlTestEncoder()));
-            Assert.Equal(expectedContent, content2.GetContent(new HtmlTestEncoder()));
+            Assert.Equal(1, executionCount);
         }
 
-        [Fact]
-        public async Task GetChildContentAsync_CanExecuteChildrenMoreThanOnce()
+        [Theory]
+        [MemberData(nameof(HtmlEncoderData))]
+        public async Task GetChildContentAsync_CachesValuePerEncoder(HtmlEncoder encoder)
```
NTaylorMullen: Should add one more variation of this where you call `GetChildContentAsync` twice for each encoder and ensure that `executionCount` remains 2.

NTaylorMullen:
:shipit: after https://github.com/aspnet/Razor/pull/653#discussion_r49897624

test/Microsoft.AspNet.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: How is that different from `GetChildContentAsync_CachesValue()` which calls one `HtmlEncoder` twice and confirms the `executionCount` is `1`?

dougbu: Will
- Add `null` to `HtmlEncoderData`.
- Add a test like this one but passing `null` instead of `firstEncoder`.

NTaylorMullen: :+1:

dougbu:
37850b5

PR #659, opened 2016-01-12
Removing Microsoft.AspNet.Razor.Runtime.Precompilation

pranavkm:


pranavkm:
cc @NTaylorMullen

davidfowl:
? dead?

pranavkm:
We'd have to write additional code to keep the tests working once we switch to DorkBuild. I spoke to @Eilon about it and he suggested removing the code for now and creating work items that tracked moving it to the right home for this code.

davidfowl:
That's fine. I just wanted to confirm. It might be time to move precomp into another tool.

pranavkm:
Split into 2 commits - the first one removes precompilation specific code. The second removes references to ITypeInfo. I'll combine them into a single commit prior to checking in.

dougbu:
So the idea is that whatever comes post-DorkBuild will _not_ involve `ITypeInfo` et cetera?

NTaylorMullen:
So, now that we're removing pre-compilation we definitely don't require `ITypeInfo` bits anymore. However, I do like the flexibility it gives us; reminds me a lot of XUnit extensibility and how they abstract everything to `ITypeInfo`s and their extensibility model has been well accepted. 

Needless to say I'm on the fence, would it be horribad to keep the `ITypeInfo` and friend abstractions in?

pranavkm:
We could always revive it if we need to later on. These aren't first class code paths so I don't think us changing it in the future to support new scenarios would be particularly breaking. cc @Eilon for thoughts on this.

Eilon:
Do the easiest thing now and deal with any future stuff in the future.

pranavkm:
@Eilon is ok with us reviving this code if we need to in the future. For now, it seems cleaner not to have an abstraction that isn't used.

NTaylorMullen:
Alright, sounds good.

NTaylorMullen:
:shipit:

pranavkm:
@dougbu, once we switch to the CLI, we don't have an an alternate source for `ITypeInfo` (since precompile modules don't exist in it as yet in the format DNX supported it)

PR #663, opened 2016-01-14
Eliminate extra List<RazorError> copies

rynowak:
Changing to IReadOnlyList since we always want to support indexing.
Replacing ToArray() with non-linq when needed, and with a static
EmptyArray when not needed.

Eliminates 50mb of list copies.


NTaylorMullen:
:shipit: if tests pass.

PR #664, opened 2016-01-15
Remove boxing of symbol types.

NTaylorMullen:
- The Equals operators were boxing the symbol types like crazy, added an abstract `SymbolTypeEquals` to avoid this.

**Before:**
image

**After:**
image

#635

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
@@ -105,7 +105,7 @@ namespace Microsoft.AspNet.Razor.Parser
                     // Whitespace here is invalid (according to the spec)
                     OptionalBangEscape();
                     Optional(HtmlSymbolType.Text);
-                    AcceptAll(HtmlSymbolType.WhiteSpace);
+                    Optional(HtmlSymbolType.WhiteSpace);
                     Optional(HtmlSymbolType.CloseAngle);
                 }
 
--- a/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
@@ -48,6 +48,8 @@ namespace Microsoft.AspNet.Razor.Parser
             get { return HtmlLanguageCharacteristics.Instance; }
         }
 
+        protected override bool SymbolTypeEquals(HtmlSymbolType x, HtmlSymbolType y) => x == y;
+
         public override void BuildSpan(SpanBuilder span, SourceLocation start, string content)
         {
             span.Kind = SpanKind.Markup;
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.cs
```diff
@@ -48,6 +48,8 @@ protected override ParserBase OtherParser
             get { return HtmlLanguageCharacteristics.Instance; }
         }
 
+        internal override bool SymbolTypeEquals(HtmlSymbolType x, HtmlSymbolType y) => x == y;
```
rynowak: `brotected`

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
```diff
@@ -105,7 +105,7 @@ private void ScanTagInDocumentContext()
                     // Whitespace here is invalid (according to the spec)
                     OptionalBangEscape();
                     Optional(HtmlSymbolType.Text);
-                    AcceptAll(HtmlSymbolType.WhiteSpace);
+                    Optional(HtmlSymbolType.WhiteSpace);
```
rynowak: Why is this a safe thing?

rynowak: Are there cases where there might be multiple whitespace tokens?

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.Helpers.cs
```diff
@@ -147,12 +149,12 @@ protected internal bool Balance(BalancingModes mode, TSymbolType left, TSymbolTy
...
         protected internal bool NextIs(TSymbolType type)
         {
-            return NextIs(sym => sym != null && Equals(type, sym.Type));
+            return NextIs(sym => sym != null && SymbolTypeEquals(type, sym.Type));
         }
 
         protected internal bool NextIs(params TSymbolType[] types)
         {
-            return NextIs(sym => sym != null && types.Any(t => Equals(t, sym.Type)));
+            return NextIs(sym => sym != null && types.Any(t => SymbolTypeEquals(t, sym.Type)));
```
rynowak: Is this called? We might want to clean this up to be more allocation friendly while here

rynowak:
:shipit:

src/Microsoft.AspNet.Razor/Parser/HtmlMarkupParser.Document.cs
NTaylorMullen: Talked inperson.

src/Microsoft.AspNet.Razor/Parser/TokenizerBackedParser.Helpers.cs
NTaylorMullen: Talked in person.

PR #665, opened 2016-01-15
Remove allocations from lookahead

rynowak:
This makes lookahead allocation-free, saving about 15MB in our current
benchmark.


NTaylorMullen:
:shipit:

PR #666, opened 2016-01-15
Optimize allocations of List<ISymbol> and related

rynowak:
**Before:**
image

**After:**
image

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
@@ -647,8 +647,9 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
             EnsureTagBlock(tagBlock);
 
             var childSpan = (Span)tagBlock.Children.First();
+
             // We grab the symbol that could be forward slash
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -648,7 +648,7 @@ private static bool IsEndTag(Block tagBlock)
 
             var childSpan = (Span)tagBlock.Children.First();
             // We grab the symbol that could be forward slash
-            var relevantSymbol = (HtmlSymbol)childSpan.Symbols.Take(2).Last();
+            var relevantSymbol = (HtmlSymbol)childSpan.Symbols[1];
```
NTaylorMullen: Doesn't do the same thing as it did before. `Take(2)` of a list of symbols of length 1 only returns a single length enumerable.

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
rynowak: Yeah, you're right. I'll fix it.

PR #667, opened 2016-01-15
Make the CodeWriter more efficient

rynowak:
**Before**
image
image

**After**
image

This is worth about 131mb

### DIFF
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using Microsoft.AspNet.Razor.Chunks.Generators;
@@ -14,6 +15,19 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
     public class CSharpCodeWriter : CodeWriter
     {
         private const string InstanceMethodFormat = "{0}.{1}";
+
+        private static readonly char[] CStyleStringLiteralEscapeChars = new char[]
+        {
+            '\r',
+            '\t',
+            '\"',
+            '\'',
+            '\\',
+            '\0',
+            '\n',
+            '\u2028',
+            '\u2029',
+        };
 
         public CSharpCodeWriter()
         {
@@ -195,8 +209,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                 file = location.FilePath;
             }
 
-            if (!string.IsNullOrEmpty(LastWrite) &&
-                !LastWrite.EndsWith(NewLine, StringComparison.Ordinal))
+            if (Builder.Length >= NewLine.Length && !IsAfterNewLine)
             {
                 WriteLine();
             }
@@ -445,17 +458,25 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         {
             Write("@\"");
 
-            foreach (char c in literal)
+            // We need to find the index of each '"' (double-quote) to escape it.
+            var start = 0;
+            int end;
+            while ((end = literal.IndexOf('\"', start)) > -1)
             {
-                if (c == '\"')
+                if (end > start)
                 {
-                    Write("\"\"");
+                    Write(literal, start, end - start);
                 }
-                else
-                {
-                    Write(c.ToString());
-                }
+
+                Write("\"\"");
+
+                start = end + 1;
             }
+
+            Debug.Assert(end == -1); // We've hit all of the double-quotes.
+
+            // Write the remainder after the last double-quote.
+            Write(literal, start, literal.Length - start);
 
             Write("\"");
         }
@@ -464,9 +485,18 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         {
             // From CSharpCodeGenerator.QuoteSnippetStringCStyle in CodeDOM
             Write("\"");
-            for (int i = 0; i < literal.Length; i++)
+
+            // We need to find the index of each escapable character to escape it.
+            var start = 0;
+            int end;
+            while ((end = literal.IndexOfAny(CStyleStringLiteralEscapeChars, start)) > -1)
             {
-                switch (literal[i])
+                if (end > start)
+                {
+                    Write(literal, start, end - start);
+                }
+
+                switch (literal[end])
                 {
                     case '\r':
                         Write("\\r");
@@ -492,13 +522,18 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     case '\u2028':
                     case '\u2029':
                         Write("\\u");
-                        Write(((int)literal[i]).ToString("X4", CultureInfo.InvariantCulture));
-                        break;
-                    default:
-                        Write(literal[i].ToString());
+                        Write(((int)literal[end]).ToString("X4", CultureInfo.InvariantCulture));
                         break;
                 }
+
+                start = end + 1;
             }
+
+            Debug.Assert(end == -1); // We've hit all of chars that need escaping.
+
+            // Write the remainder after the last escaped char.
+            Write(literal, start, literal.Length - start);
+
             Write("\"");
         }
 
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
@@ -463,10 +463,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
             int end;
             while ((end = literal.IndexOf('\"', start)) > -1)
             {
-                if (end > start)
-                {
-                    Write(literal, start, end - start);
-                }
+                Write(literal, start, end - start);
 
                 Write("\"\"");
 
@@ -491,10 +488,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
             int end;
             while ((end = literal.IndexOfAny(CStyleStringLiteralEscapeChars, start)) > -1)
             {
-                if (end > start)
-                {
-                    Write(literal, start, end - start);
-                }
+                Write(literal, start, end - start);
 
                 switch (literal[end])
                 {
@@ -523,6 +517,9 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
                     case '\u2029':
                         Write("\\u");
                         Write(((int)literal[end]).ToString("X4", CultureInfo.InvariantCulture));
+                        break;
+                    default:
+                        Debug.Assert(false, "Unknown escape character.");
                         break;
                 }
 
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
@@ -62,7 +62,7 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
         {
             if (_autoSpace && 
                 _writer.Builder.Length > 0 && 
-                !Char.IsWhiteSpace(_writer.Builder[_writer.Builder.Length - 1]))
+                !char.IsWhiteSpace(_writer.Builder[_writer.Builder.Length - 1]))
             {
                 _writer.Write(spaceCharacter);
             }
--- a/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
+++ b/src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
@@ -60,7 +60,9 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
 
         private void TryAutoSpace(string spaceCharacter)
         {
-            if (_autoSpace && _writer.LastWrite.Length > 0 && !Char.IsWhiteSpace(_writer.LastWrite.Last()))
+            if (_autoSpace && 
+                _writer.Builder.Length > 0 && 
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -445,18 +458,26 @@ private void WriteVerbatimStringLiteral(string literal)
...
+            while ((end = literal.IndexOf('\"', start)) > -1)
             {
-                if (c == '\"')
+                if (end > start)
                 {
-                    Write("\"\"");
-                }
-                else
-                {
-                    Write(c.ToString());
```
rynowak: This was allocating a string per-character

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -464,9 +485,18 @@ private void WriteCStyleStringLiteral(string literal)
         {
             // From CSharpCodeGenerator.QuoteSnippetStringCStyle in CodeDOM
             Write("\"");
-            for (int i = 0; i < literal.Length; i++)
+
+            // We need to find the index of each '"' (double-quote) to escape it.
```
rynowak: Will update comment.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -492,13 +522,18 @@ private void WriteCStyleStringLiteral(string literal)
...
+                        Write(((int)literal[end]).ToString("X4", CultureInfo.InvariantCulture));
                         break;
                 }
+
+                start = end + 1;
             }
+
+            Debug.Assert(end == -1); // We've hit all of the double-quotes.
+
+            // Write the remainder after the last double-quote.
```
rynowak: Will update comments

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -445,18 +458,26 @@ private void WriteVerbatimStringLiteral(string literal)
...
             Write("@\"");
 
-            foreach (char c in literal)
+            // We need to find the index of each '"' (double-quote) to escape it.
+            var start = 0;
+            int end;
+            while ((end = literal.IndexOf('\"', start)) > -1)
             {
-                if (c == '\"')
+                if (end > start)
```
NTaylorMullen: Wont end always > start?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
```diff
@@ -60,7 +60,9 @@ private void WriteEndScope()
 
         private void TryAutoSpace(string spaceCharacter)
         {
-            if (_autoSpace && _writer.LastWrite.Length > 0 && !Char.IsWhiteSpace(_writer.LastWrite.Last()))
+            if (_autoSpace && 
+                _writer.Builder.Length > 0 && 
+                !Char.IsWhiteSpace(_writer.Builder[_writer.Builder.Length - 1]))
```
NTaylorMullen: `char`

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
```diff
@@ -19,23 +18,13 @@ public class CodeWriter : IDisposable
...
-            get
-            {
-                return _writer.NewLine;
-            }
-            set
-            {
-                _writer.NewLine = value;
-            }
-        }
+        public bool IsAfterNewLine => _isAfterNewLine;
```
NTaylorMullen: Why have the backing field?

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
```diff
@@ -82,35 +70,42 @@ public CodeWriter Indent(int size)
 
         public CodeWriter Write(string data)
         {
-            Indent(CurrentIndent);
-
-            _writer.Write(data);
-            Flush();
+            if (data == null)
```
NTaylorMullen: Woaa, we do this?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -464,9 +485,18 @@ private void WriteCStyleStringLiteral(string literal)
...
+            int end;
+            while ((end = literal.IndexOfAny(CStyleStringLiteralEscapeChars, start)) > -1)
             {
-                switch (literal[i])
+                if (end > start)
+                {
+                    Write(literal, start, end - start);
+                }
+
+                switch (literal[end])
```
pranavkm: Could we cache the values to write?
```C#
private static readonly CStyleStringLiteralEscapeCharsLookup = new Dictionary<char, string>
{
  { '\r', "\\r" },
};
private static readonly CStyleStringLiteralEscapeCharsKeys = CStyleStringLiteralEscapeCharsLookup.Keys.ToArray();
```

rynowak: These are cached. They are constant strings

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
rynowak: Yep. Lots of stuff blows up without this.

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
rynowak: Sure, I can change this.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWritingScope.cs
rynowak: :+1:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
rynowak: yeah this isn't really needed, because in the case where `end == start` then `count == 0`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -15,6 +16,19 @@ public class CSharpCodeWriter : CodeWriter
...
+            '\r',
+            '\t',
+            '\"',
+            '\'',
+            '\\',
+            '\0',
+            '\n',
+            '\u2028',
+            '\u2029',
+        };
```
dougbu: Don't we have this information somewhere else in Razor?

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
```diff
@@ -207,7 +192,6 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                _writer.Dispose();
```
NTaylorMullen: `Builder.Clear()`

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
NTaylorMullen: :cry: :panda_face:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: @NTaylorMullen not when `literal[start] == '\"'`

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: This addresses [my comment](https://github.com/aspnet/Razor/issues/635#issuecomment-168457369) on #635.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
rynowak: No idea

rynowak:
updated

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -445,18 +458,26 @@ private void WriteVerbatimStringLiteral(string literal)
         {
             Write("@\"");
 
-            foreach (char c in literal)
+            // We need to find the index of each '"' (double-quote) to escape it.
+            var start = 0;
+            int end;
+            while ((end = literal.IndexOf('\"', start)) > -1)
```
dougbu: With this new implementation, `WriteVerbatimStringLiteral()` should be much more efficient than `WriteCStyleStringLiteral()`, writing fewer / shorter `string`s when the `Writer` is a `StringWriter`. Given that, suggest following through on an earlier [PR #652 discussion](https://github.com/aspnet/Razor/pull/652#discussion-diff-48799035) i.e. remove the `Length` checks in the condition on [line 132](https://github.com/aspnet/Razor/pull/667/files#diff-d020ecaca6e34f77668b2ffe8de6d65aR132). That way ``WriteCStyleStringLiteral()` will only be called in the unusual condition that the literal includes a null `char`.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -492,13 +522,18 @@ private void WriteCStyleStringLiteral(string literal)
                     case '\u2028':
                     case '\u2029':
                         Write("\\u");
-                        Write(((int)literal[i]).ToString("X4", CultureInfo.InvariantCulture));
-                        break;
-                    default:
```
dougbu: Suggest a `Debug.Assert(false)` in the `default:` case.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Removed part of previous comment since `CodeWriter` always writes to a `StringBuffer` (previously always wrote to a `StringBuilder`).

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
```diff
@@ -10,32 +9,21 @@ namespace Microsoft.AspNet.Razor.CodeGenerators
...
-                return _writer.NewLine;
-            }
-            set
-            {
-                _writer.NewLine = value;
-            }
-        }
+        public bool IsAfterNewLine { get; private set; }
+
+        public string NewLine { get; set; } = Environment.NewLine;
```
dougbu: Realize this is output and not input. But are newlines handled correctly when those in the `.cshtml` file don't match `Environment.NewLine` e.g. a Windows-ended file read on a Linux system?

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
rynowak: Sure

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
rynowak: Don't know and didn't change any behavior.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Discussed offline and agreed we should make a change here only if tests show it improves performance. That would  likely happen if the final `string`s are shorter i.e. doubled double-quotes happen less than backslashed characters.

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
NTaylorMullen: Yes, the test infrastructure changes this accordingly.

src/Microsoft.AspNet.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: NVM. Only other occurences of `'\u2028'` are later in this file and in `ParserHelpers.IsNewLine()`.

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
dougbu: :ok:

dougbu:
:watch: just for whether https://github.com/aspnet/Razor/pull/667#discussion-diff-49887256 makes enough difference to be worthwhile.

src/Microsoft.AspNet.Razor/CodeGenerators/CodeWriter.cs
dougbu: > Yes, the test infrastructure changes this accordingly.

If the test infrastructure ensures that mixes real people will hit don't occur when testing, that's bad. If you mean the test infrastructure mixes line endings to confirm they're correctly handled, that's good. So which did you mean?

rynowak:
Removing the special casing for verbatim was about 5kb better out of 1.545gb. Not worth doing right without a better reason to do it.

dougbu:
:shipit:

PR #668, opened 2016-01-15
Remove a String.Split

rynowak:
image

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
@@ -324,12 +324,25 @@ namespace Microsoft.AspNet.Razor.Parser.TagHelpers.Internal
                     childSpan = child as Span;
                 }
 
-                var attributeName = childSpan
-                    .Content
-                    .Split(separator: new[] { '=' }, count: 2)[0]
-                    .TrimStart();
+                var start = 0;
+                var end = 0;
+                for (; start < childSpan.Content.Length; start++)
+                {
+                    if (!char.IsWhiteSpace(childSpan.Content[start]))
+                    {
+                        break;
+                    }
+                }
 
-                attributeNames.Add(attributeName);
+                for (end = start; end < childSpan.Content.Length; end++)
+                {
+                    if (childSpan.Content[end] == '=')
+                    {
+                        break;
+                    }
+                }
+
+                attributeNames.Add(childSpan.Content.Substring(start, end - start));
             }
 
             return attributeNames;


src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -324,12 +324,25 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
...
+                for (end = start; end < childSpan.Content.Length; end++)
+                {
+                    if (childSpan.Content[end] == '=')
+                    {
+                        break;
+                    }
+                }
 
-                attributeNames.Add(attributeName);
+                attributeNames.Add(childSpan.Content.Substring(start, end - start));
```
rynowak: We could also make a bigger change here to look at the tokens individually instead of calling `.Content` and substringing. Ideally this info would be part of the syntax tree already, but that's a much bigger change.

NTaylorMullen:
:shipit:

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -324,12 +324,25 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
                     childSpan = child as Span;
                 }
 
-                var attributeName = childSpan
-                    .Content
-                    .Split(separator: new[] { '=' }, count: 2)[0]
-                    .TrimStart();
+                var start = 0;
+                var end = 0;
```
pranavkm: Declare this after this loop:

```
var end = start;
for (; end < childSpan.Content.Length;..)
```

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -324,12 +324,25 @@ private IEnumerable<string> GetAttributeNames(Block tagBlock)
...
+                for (; start < childSpan.Content.Length; start++)
```
pranavkm: Does accessing `childSpan.Content` cause revaluation of a buffer? Maybe read it into a variable.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
rynowak: Sure

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
rynowak: No, it's cached

PR #669, opened 2016-01-15
Prevent unneeded allocations of `Block` children enumerators.

NTaylorMullen:
**Before:**
image

**After:**
image

#635

### DIFF
--- a/src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
+++ b/src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
@@ -19,7 +19,7 @@ namespace Microsoft.AspNet.Razor.Parser.SyntaxTree
             source.Reset();
         }
 
-        protected Block(BlockType? type, IEnumerable<SyntaxTreeNode> contents, IParentChunkGenerator generator)
+        protected Block(BlockType? type, IReadOnlyList<SyntaxTreeNode> contents, IParentChunkGenerator generator)
         {
             if (type == null)
             {
@@ -30,14 +30,15 @@ namespace Microsoft.AspNet.Razor.Parser.SyntaxTree
             Children = contents;
             ChunkGenerator = generator;
 
-            foreach (SyntaxTreeNode node in Children)
+            // Perf: Avoid allocating an enumerator.
+            for (var i = 0; i < Children.Count; i++)
             {
-                node.Parent = this;
+                Children[i].Parent = this;
             }
         }
 
         // A Test constructor
-        internal Block(BlockType type, IEnumerable<SyntaxTreeNode> contents, IParentChunkGenerator generator)
+        internal Block(BlockType type, IReadOnlyList<SyntaxTreeNode> contents, IParentChunkGenerator generator)
         {
             Type = type;
             ChunkGenerator = generator;
@@ -46,7 +47,7 @@ namespace Microsoft.AspNet.Razor.Parser.SyntaxTree
 
         public BlockType Type { get; }
 
-        public IEnumerable<SyntaxTreeNode> Children { get; }
+        public IReadOnlyList<SyntaxTreeNode> Children { get; }
 
         public IParentChunkGenerator ChunkGenerator { get; }
 
@@ -133,9 +134,10 @@ namespace Microsoft.AspNet.Razor.Parser.SyntaxTree
 
         public virtual IEnumerable<Span> Flatten()
         {
-            // Create an enumerable that flattens the tree for use by syntax highlighters, etc.
-            foreach (SyntaxTreeNode element in Children)
+            // Perf: Avoid allocating an enumerator.
+            for (var i = 0; i < Children.Count; i++)
             {
+                var element = Children[i];
                 var span = element as Span;
                 if (span != null)
                 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
```diff
@@ -19,7 +19,7 @@ public Block(BlockBuilder source)
             source.Reset();
         }
 
-        protected Block(BlockType? type, IEnumerable<SyntaxTreeNode> contents, IParentChunkGenerator generator)
+        protected Block(BlockType? type, List<SyntaxTreeNode> contents, IParentChunkGenerator generator)
```
rynowak: Should this be `IList<>` since it's on a publicish API?

rynowak: Do we want to make these consistent and use `IReadOnlyList<>` here? (like the other constructor)

NTaylorMullen: `IList` doesn't implement `IReadOnlyList`.

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/BlockBuilder.cs
```diff
@@ -21,7 +21,7 @@ public BlockBuilder(Block original)
         }
 
         public BlockType? Type { get; set; }
-        public IList<SyntaxTreeNode> Children { get; private set; }
+        public List<SyntaxTreeNode> Children { get; private set; }
```
rynowak: why?

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -302,8 +302,13 @@ private bool TryRewriteTagHelper(Block tagBlock, RewritingContext context)
         private IEnumerable<string> GetAttributeNames(Block tagBlock)
         {
             // Need to calculate how many children we should take that represent the attributes.
-            var childrenOffset = IsPartialTag(tagBlock) ? 1 : 2;
-            var attributeChildren = tagBlock.Children.Skip(1).Take(tagBlock.Children.Count() - childrenOffset);
+            var childrenOffset = IsPartialTag(tagBlock) ? 0 : 1;
+            var childrenCount = tagBlock.Children.Count - childrenOffset;
```
rynowak: `childCount`?

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/Block.cs
NTaylorMullen: Oh shoot, this is `Block`. Yes `IReadOnlyList` works. Ignore my previous comment.

src/Microsoft.AspNet.Razor/Parser/TagHelpers/TagHelperParseTreeRewriter.cs
```diff
@@ -302,8 +302,13 @@ private bool TryRewriteTagHelper(Block tagBlock, RewritingContext context)
...
+            var attributeChildren = new List<SyntaxTreeNode>();
```
rynowak: Intialize with the capacity since you already know it

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/BlockBuilder.cs
NTaylorMullen: `List` : `IReadOnlyList`, `IList` does not.

rynowak: #BackingField

rynowak:
:shipit: after fixing minor comments

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/BlockBuilder.cs
NTaylorMullen: Needs to be exposed so `Block.cs` can consume it.

dougbu: But `Block` doesn't need to change the collection does it? `IReadOnlyList` should work here w/ a `List` backing field.

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
```diff
@@ -29,24 +29,19 @@ public StatementBlock(params SyntaxTreeNode[] children)
             : this(ParentChunkGenerator.Null, children)
         {
         }
-
-        public StatementBlock(IEnumerable<SyntaxTreeNode> children)
-            : this(ParentChunkGenerator.Null, children)
```
dougbu: Just curious: Were these constructor overloads ever used?

dougbu:
:shipit: w/ @rynowak's carefully-considered suggestions fixed in a carefully-considered fashion.

test/Microsoft.AspNet.Razor.Test/Framework/BlockTypes.cs
NTaylorMullen: Nope

src/Microsoft.AspNet.Razor/Parser/SyntaxTree/BlockBuilder.cs
NTaylorMullen: `Block` doesn't need to change it but this is a builder type. There'd be no way to add `Children` to `BlockBuilder`s unless I added an additional method to add children (not worth when its true type has the capability already).

PR #670, opened 2016-01-15
State machine

rynowak:
image

About 250mb of state machine overhead removed and replaced with a switch + enum


src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
```diff
@@ -31,6 +35,12 @@ protected Tokenizer(ITextDocument source)
             StartSymbol();
         }
 
+        protected abstract int StartState { get; }
+
+        protected int? CurrentState { get; set; }
```
NTaylorMullen: Do we see allocations for this bit a lot? Could have a separate flag to showcase the tokenizer is done rather than rely on this being `null`.

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
```diff
@@ -179,9 +277,10 @@ protected StateResult AfterRazorCommentTransition()
                 // We've been moved since last time we were asked for a symbol... reset the state
                 return Transition(StartState);
             }
+
             AssertCurrent('*');
             TakeCurrent();
-            return Transition(EndSymbol(RazorCommentStarType), RazorCommentBody);
+            return Transition(1002, EndSymbol(RazorCommentStarType));
```
NTaylorMullen: Curious: Why not just use the enum name instead of its magic `int` value?

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
```diff
@@ -189,42 +288,43 @@ protected StateResult RazorCommentBody()
             TakeUntil(c => c == '*');
```
NTaylorMullen: If this lambda is expensive we could inline the logic.

NTaylorMullen:
:watch: for questions.

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
rynowak: But why though? Why is that better? This is pretty simple

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
rynowak: I already fixed this by creating another enum. you'll see in the next iteration..

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
rynowak: We'll get there or not :laughing:

rynowak:
Updated

src/Microsoft.AspNet.Razor/Tokenizer/Tokenizer.cs
pranavkm: `Nullable` is a value type.

NTaylorMullen:
:shipit:

PR #673, opened 2016-01-22
Flatten TagHelperOutput when writing

rynowak:
This change 'flattens' a TagHelperOutput when writing it out to an
HtmlTextWriter. This is a beneficial perf change because more often than
not the thing being written to is a ViewBuffer in Razor, which uses pooled
memory. This allows us to 'join' islands of pooled ViewBuffer memory back
into the main buffer instead of keeping them wrapped up in a
TagHelperOutput or TagHelperContent.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -277,164 +277,77 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 throw new ArgumentNullException(nameof(writer));
             }
 
-            var htmlTextWriter = writer as HtmlTextWriter;
-            if (htmlTextWriter != null)
-            {
-                WriteTo(htmlTextWriter, encoder);
-                return;
-            }
-
             _preElement?.WriteTo(writer, encoder);
 
             var isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(TagName);
 
             if (!isTagNameNullOrWhitespace)
             {
-                WriteBeginTag(writer, encoder, TagName, Attributes, TagMode);
-            }
+                writer.Write("<");
+                writer.Write(TagName);
 
-            if (isTagNameNullOrWhitespace || TagMode == TagMode.StartTagAndEndTag)
-            {
-                _preContent?.WriteTo(writer, encoder);
-
-                _content?.WriteTo(writer, encoder);
-
-                _postContent?.WriteTo(writer, encoder);
-            }
-
-            if (!isTagNameNullOrWhitespace && TagMode == TagMode.StartTagAndEndTag)
-            {
-                WriteEndTag(writer, encoder, TagName);
-            }
-
-            _postElement?.WriteTo(writer, encoder);
-        }
-
-        private void WriteTo(HtmlTextWriter writer, HtmlEncoder encoder)
-        {
-            _preElement?.WriteTo(writer, encoder);
-
-            var isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(TagName);
-
-            if (!isTagNameNullOrWhitespace)
-            {
-                writer.Write(new BeginTag(TagName, Attributes, TagMode));
-            }
-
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
Talked offline: We can't delay render begin/end tags etc. due to attribute values potential to be `IHtmlContent`'s containing `@(i++)`. @rynowak has an idea on how to fix this.

Waiting till the concern is addressed to look over the PR further.

:watch:

dougbu:
> due to attribute values potential to be `IHtmlContent`'s containing `@(i++)`

@NTaylorMullen what `IHtmlContent` implementations contain user expressions? We have a few that defer encoding, avoid `Substring()` et cetera but I can't think of any containing `Func`s.

NTaylorMullen:
@dougbu `IHtmlContent` is just an interface, any implementation would cause this behavior. Even more worrisome though is if a `TagHelper` author holds onto their `TagHelperOutput`s Attributes and then modifies them. Delayed rendering would destroy both scenarios.

dougbu:
> Delayed rendering would destroy both scenarios.

We do delayed rendering all over the place. You're talking about two corner cases that, while possible, should really be warned about (if we had the appropriate Roslyn analyzers). For example "holds onto their `TagHelperOutput`s `Attributes` and then modifies them" involves tag helper authors (1) creating an `IHtmlContent` implementation containing `Func`s, (2) putting their own expressions into those `Func`s (note we don't give tag helper author's access to the user's expressions), and (3) also setting up their `IHtmlContent` for deferred execution.

rynowak:
Discussed with @NTaylorMullen - we agree that what this code is doing today isn't really a change in semantics. 

We're still going to try and make the behavior more straightforward, and avoid holding a reference to the `TagHelperAttributesList` after the call to `WriteTo` completes. We think we can do this without causing any new perf problems.

I'll send an update to this PR shortly

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -283,14 +283,14 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
 
             if (!isTagNameNullOrWhitespace)
             {
-                writer.Write('<');
+                writer.Write("<");
```
rynowak: Changing these to string to avoid boxing in the common case (`HtmlTextWriter` that stores string or `IHtmlContent`).

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -309,10 +309,9 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
                         using (var stringWriter = new StringWriter())
                         {
                             htmlContent.WriteTo(stringWriter, encoder);
+                            stringWriter.GetStringBuilder().Replace("\"", "&quot;");
```
rynowak: Doing this replace on the `StringBuilder`/`char[]` before creating a string is better.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -990,6 +992,34 @@ public void WriteTo_WritesFormattedTagHelper(TagHelperOutput output, string expe
             Assert.Equal(expected, writer.ToString(), StringComparer.Ordinal);
         }
 
+        // This tests a separate code path that's used by THO when the writer is an HtmlTextWriter.
+        // The output should be the same, but we do some specific perf optimizations on this path.
+        [Theory]
+        [MemberData(nameof(WriteTagHelper_InputData))]
+        public void WriteTo_WritesFormattedTagHelper_HtmlTextWriter(TagHelperOutput output, string expected)
```
rynowak: I'll update this comment. This feels like a good integration test to have even though it's not a separate code path anymore.

rynowak:
@dougbu @NTaylorMullen - updated

This now forces eager evaluation of attributes

NTaylorMullen:
:shipit:

dougbu:
:shipit:

Issue #604, opened 2015-11-12
Consider a design when TagHelperAttribute is a struct


rynowak:
It may be possible to mitigate some of this with caching, but here's the data as-of now.

![image](https://cloud.githubusercontent.com/assets/1430011/11107457/e4481912-8892-11e5-9067-e48491131751.png)

The HtmlHelpers equivalent to this allocates about 82mb **total**. 2mb of just attributes seems like enough to consider a design change.

NTaylorMullen:
fc99082ea8d97b6b5ffbde3c92a1f03377c9c2fd
b7b3273fa49d31f004a4b739be142db8d068e357
https://github.com/aspnet/Mvc/commit/dca15c0a60ae8282c4b3b5cd4f1ef33e09816834

PR #682, opened 2016-01-30
Make `TagHelperAttribute` immutable.

NTaylorMullen:
- Remove a lot of the ugly generic work done in `TagHelperAttributeList`/`ReadOnlyTagHelperAttributeList`.
- Removed error cases where we'd check on addition of `TagHelperAttribute`s that their `name` was not `null`. This was due to the pre-existing `string` indexer for `TagHelperAttributeList`. It no longer sets `TagHelperAttribute`s directly, instead it's `[key] = value` (instead of `[key] = new TagHelperAttribute(...))`.
- Updated tests to account for new immutability/string indexer format.
- Removed copy ctor from `TagHelperAttribute` since its immutability kills the value.
- This does not impact performance.

#604


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -183,18 +183,9 @@ public void AddMinimizedHtmlAttribute(string name)
...
-                    Minimized = true
-                });
-            AllAttributes.Add(
-                new TagHelperAttribute
-                {
-                    Name = name,
-                    Minimized = true
-                });
+            var attribute = new TagHelperAttribute(name);
+            HTMLAttributes.Add(attribute);
```
rynowak: Naming of this doesn't make .Net guidelines for abbreviations btw. It's supposed to be HtmlAttributes

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -8,67 +8,66 @@
...
     {
-        private static readonly IReadOnlyList<TAttribute> EmptyList =
+        private static readonly IReadOnlyList<TagHelperAttribute> EmptyList =
 #if NET451
-            new TAttribute[0];
+            new TagHelperAttribute[0];
 #else
-            Array.Empty<TAttribute>();
+            Array.Empty<TagHelperAttribute>();
 #endif
```
rynowak: IMO it's better without the ifdefs unless we know we're going to 4.6 fo shizzle

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -8,67 +8,66 @@
...
+    /// A read-only collection of <typeparamref name="TagHelperAttribute"/>s.
     /// </summary>
-    /// <typeparam name="TAttribute">
-    /// The type of <see cref="IReadOnlyTagHelperAttribute"/>s in the collection.
+    /// <typeparam name="TagHelperAttribute">
+    /// The type of <see cref="TagHelperAttribute"/>s in the collection.
     /// </typeparam>
-    public class ReadOnlyTagHelperAttributeList<TAttribute> : IReadOnlyList<TAttribute>
-        where TAttribute : IReadOnlyTagHelperAttribute
+    public class ReadOnlyTagHelperAttributeList : IReadOnlyList<TagHelperAttribute>
```
rynowak: I may have mentioned this before, but consider inheriting `ReadOnlyCollection<T>` to get rid of some of this boilerplate

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
-        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>
-        /// and <paramref name="value"/>.
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>,
+        /// <paramref name="value"/> and <paramref name="minimized"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         /// <param name="value">The <see cref="Value"/> of the attribute.</param>
-        public TagHelperAttribute(string name, object value)
+        /// <param name="minimized">The <see cref="Minimized"/> value of the attribute.</param>
+        public TagHelperAttribute(string name, object value, bool minimized)
```
rynowak: This doc doesn't explain what minimized means

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -289,7 +289,7 @@ public void HtmlAttributes_IgnoresCase(string originalName, string updatedName)
         {
             // Arrange
             var executionContext = new TagHelperExecutionContext("p", TagMode.StartTagAndEndTag);
-            executionContext.HTMLAttributes[originalName] = "hello";
+            executionContext.AddHtmlAttribute(originalName, "hello");
```
rynowak: Notice you've got inconsistent casing between this method and the property.

rynowak: Should there be an Add(string, object) method on the attributes collection instead?

rynowak:
:shipit: from me. Design overall looks like a positive change. I'll leave the details for someone more familiar to pore over.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -8,67 +8,66 @@
...
-        /// The <see cref="IReadOnlyTagHelperAttribute.Name"/> of the <typeparamref name="TAttribute"/> to get.
+        /// The <see cref="TagHelperAttribute.Name"/> of the <typeparamref name="TagHelperAttribute"/> to get.
         /// </param>
-        /// <returns>The first <typeparamref name="TAttribute"/> with <see cref="IReadOnlyTagHelperAttribute.Name"/>
+        /// <returns>The first <typeparamref name="TagHelperAttribute"/> with <see cref="TagHelperAttribute.Name"/>
         /// matching <paramref name="name"/>.
         /// </returns>
         /// <remarks><paramref name="name"/> is compared case-insensitively.</remarks>
-        public TAttribute this[string name]
+        public object this[string name]
```
dougbu: Could this be `TagHelperAttribute` instead? The shortcut to the attribute's `Value` leads to some odd usage and doesn't improve readability. Besides that, having the two indexers return different types is counter-intuitive.

dougbu: Note as well that doc comments do not match what is implemented.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -166,41 +165,49 @@ public int IndexOf(TAttribute item)
...
-            attribute = this[name];
+            attribute = null;
+            for (var i = 0; i < Attributes.Count; i++)
+            {
+                if (NameEquals(name, Attributes[i]))
+                {
+                    attribute = Attributes[i];
+                    break;
+                }
+            }
```
dougbu: Undo this change once https://github.com/aspnet/Razor/pull/682#discussion-diff-51338876 is fixed.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
 
         /// <summary>
-        /// Instantiates a new instance of <see cref="TagHelperAttribute"/>.
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>.
+        /// <see cref="Minimized"/> is set to <c>true</c> and <see cref="Value"/> to <c>null</c>.
         /// </summary>
-        public TagHelperAttribute()
+        /// <param name="name">The <see cref="Name"/> of the attribute.</param>
+        public TagHelperAttribute(string name)
+            : this(name, value: null, minimized: true)
```
dougbu: `minimized: false` since minimization must be an explicit opt-in choice.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
+                throw new ArgumentNullException(nameof(name));
+            }
+
             Name = name;
             Value = value;
+            Minimized = minimized;
         }
 
         /// <summary>
         /// Gets or sets the name of the attribute.
```
dougbu: `Gets or sets` -> `Gets` for all properties.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
+        public string Name { get; }
 
         /// <summary>
         /// Gets or sets the value of the attribute.
         /// </summary>
-        public object Value { get; set; }
+        public object Value { get; }
 
         /// <summary>
         /// Gets or sets an indication whether the attribute is minimized or not.
```
dougbu: After reading @rynowak's [comment](https://github.com/aspnet/Razor/pull/682/files#r51336706), see this comment goes down the rabbit hole without saying much.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -77,21 +67,8 @@ public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
...
-        /// // Will "value" be converted to a TagHelperAttribute with a null Name
-        /// attributes["name"] = "value";
-        ///
-        /// // TagHelperAttribute.Name must match the specified name.
-        /// attributes["name"] = new TagHelperAttribute("name", "value");
-        /// </code>
-        /// </example>
-        public new TagHelperAttribute this[string name]
+        /// <remarks><paramref name="name"/> is compared case-insensitively.</remarks>
+        public new object this[string name]
```
dougbu: As w/ https://github.com/aspnet/Razor/pull/682#discussion-diff-51338876, don't change the API here either.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -114,23 +91,6 @@ public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
...
-                else if (!NameEquals(name, value))
-                {
-                    throw new ArgumentException(
-                        Resources.FormatTagHelperAttributeList_CannotAddAttribute(
-                            nameof(TagHelperAttribute),
-                            nameof(TagHelperAttribute.Name),
-                            value.Name,
-                            name),
-                        nameof(name));
-                }
```
dougbu: After restoring previous API, will need the `!NameEquals(name, value)` check.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
-        /// </summary>
-        /// <param name="value">The <see cref="Value"/> of the created <see cref="TagHelperAttribute"/>.</param>
-        /// <remarks>Created <see cref="TagHelperAttribute"/>s <see cref="Name"/> is set to <c>null</c>.</remarks>
-        public static implicit operator TagHelperAttribute(string value)
-        {
-            return new TagHelperAttribute
-            {
-                Value = value
-            };
-        }
```
dougbu: Has this API change been approved? It means `attributes["name"] = "value"` no longer works but I'm not positive that was used a great deal. Do we use that in MVC tag helpers at all?

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: @rynowak FYI can see both sides of this point:
- `AddHtmlAttribute()` exists and does the same work as the proposed `Add(string, object)` method.
- `Add()` is a more obvious replacement for `attributes["name"] = value` and should be more discoverable than the `AddHtmlAttribute()` "shortcut".

dougbu: I lean toward adding `Add(string, object)`.

dougbu: __Wait__ `Add(string name, object value)` exists. @NTaylorMullen please use that as it's the more obvious replacement for the implicit `string` -> `TagHelperAttribute` conversion.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -305,10 +305,10 @@ public void AllAttributes_IgnoresCase(string originalName, string updatedName)
         {
             // Arrange
             var executionContext = new TagHelperExecutionContext("p", tagMode: TagMode.StartTagAndEndTag);
-            executionContext.AllAttributes.Add(originalName, value: false);
+            executionContext.AddTagHelperAttribute(originalName, value: false);
```
dougbu: This shouldn't need to change. If change was required, why?

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -305,10 +305,10 @@ public void AllAttributes_IgnoresCase(string originalName, string updatedName)
...
 
             // Act
-            executionContext.AllAttributes[updatedName].Value = true;
+            executionContext.AllAttributes[updatedName] = true;
```
dougbu: We lose old in-place update but replacement should be explicit i.e. executionContext.AllAttributes[updatedName] = new TagHelperAttribute(updatedName, true)`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
         }
 
         /// <summary>
-        /// Instantiates a new instance of <see cref="TagHelperAttribute" /> with values provided by the given
-        /// <paramref name="attribute"/>.
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>
+        /// and <paramref name="value"/>. <see cref="Minimized"/> is set to <c>false</c>.
         /// </summary>
-        /// <param name="attribute">A <see cref="IReadOnlyTagHelperAttribute"/> whose values should be copied.</param>
-        public TagHelperAttribute(IReadOnlyTagHelperAttribute attribute)
```
dougbu: A `TagHelperAttribute(IReadOnlyTagHelperAttribute attribute, object value)` overload would be useful though it should probably `throw` something if `attribute.Minimized`.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
NTaylorMullen: Gave it a shot and cleans this all up even more :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: `minimized` is HTML terminology much like `name` and `value` are.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
NTaylorMullen: This test slightly overlaps the changes dont [here](https://github.com/aspnet/Razor/pull/683). I split the changes after doing both and created two PRs. In this situation I call AddHtmlAttribute in order to instantiate the underlying `executionContext.HtmlAttributes` list. Otherwise it `null` refs.

dougbu: As mentioned in the other PR, collection properties that can return `null` are the worst.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: We can debate this one but if you're only providing a `name` I'd think it'd be more obvious to assume `minimzed` = `true`. If you want the less common `name=""` syntax you'd do `new TagHelperAttribute("name", string.Empty)` 

/cc @Eilon

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: This API assumed that the `TagHelperAttribute`s name was settable/nullable. Now that it's not this goes away.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/682#discussion_r51503651

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: Seems kind of odd to have

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: I agree we can't have this `operator` with the other changes you're making. But has the public API change been agreed to?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
NTaylorMullen: Talked with @DamianEdwards offline and the decision was to:

- Remove Setter
- Change getter to be `TagHelperAttribute`
- Add separate `Set` method(s).

dougbu: > Remove Setter

Nice. `new` dies and the API becomes more explicit overall. Slight downside is the churn in consuming code.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
Eilon: Yeah I think this is fine. If there's no value, it's typically a minimized attribute - why else would you call this ctor.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -51,112 +50,76 @@ public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
...
-        /// // TagHelperAttribute.Name must match the specified name.
-        /// attributes["name"] = new TagHelperAttribute("name", "value");
-        /// </code>
-        /// </example>
-        public new TagHelperAttribute this[string name]
+        /// <param name="value">
+        /// The <see cref="TagHelperAttribute.Value"/> to set.
+        /// </param>
+        /// <remarks><paramref name="name"/> is compared case-insensitively.</remarks>
+        public void SetAttribute(string name, object value)
```
NTaylorMullen: Followed same format of `.Add` where there's one overload for name, value and another for an entire `TagHelperAttribute`.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: I suspect the confusion here is between the C# property named `Minimized` and the "of the attribute" part of this sentence. These `<param>`s may be more clear if they all said something like `The <see cref="Minimized"/> value of the new instance.` Leave `... of the attribute` to the property `<summary>`s. (Perhaps even `... of the HTML attribute.`)

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>,
+        /// <paramref name="value"/> and <paramref name="minimized"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         /// <param name="value">The <see cref="Value"/> of the attribute.</param>
-        public TagHelperAttribute(string name, object value)
+        /// <param name="minimized">The <see cref="Minimized"/> value of the attribute.</param>
+        public TagHelperAttribute(string name, object value, bool minimized)
         {
+            if (name == null)
```
dougbu: Why is `string.Empty` a legal `Name`?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
         /// </summary>
-        /// <param name="attribute">A <see cref="IReadOnlyTagHelperAttribute"/> whose values should be copied.</param>
-        public TagHelperAttribute(IReadOnlyTagHelperAttribute attribute)
-            : this(attribute?.Name, attribute?.Value)
+        /// <param name="name">The <see cref="Name"/> of the attribute.</param>
+        /// <param name="value">The <see cref="Value"/> of the attribute.</param>
+        public TagHelperAttribute(string name, object value)
+            : this(name, value, minimized: false)
         {
-            if (attribute == null)
```
dougbu: Previous code threw when attempting to create a `TagHelperAttribute` with `value: null` and `minimized: false`. This either means the same as `value: string.Empty` with `minimized: false` or as `value: null` with `minimized: true`. I suspect the first but why is `value: null` allowed here at all?

Same for the full constructor overload -- should disallow `value: null` when passed `mininized: false`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
         /// <summary>
-        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>
-        /// and <paramref name="value"/>.
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>,
+        /// <paramref name="value"/> and <paramref name="minimized"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         /// <param name="value">The <see cref="Value"/> of the attribute.</param>
-        public TagHelperAttribute(string name, object value)
+        /// <param name="minimized">The <see cref="Minimized"/> value of the attribute.</param>
```
dougbu: Add something stating that `value` is completely ignored when `minimized: true`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -51,112 +50,76 @@ public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
...
+            if (name == null)
             {
-                if (name == null)
-                {
-                    throw new ArgumentNullException(nameof(name));
-                }
-
-                return base[name];
+                throw new ArgumentNullException(nameof(name));
             }
```
dougbu: No need for a check that's done in the `TagHelperAttribute` constructor. We're fine here because the parameter names match.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: Let's separate the two PRs. Stick w/ the obvious test setup: `executionContext.SetAttributes(originalName, "hello")`. No need to deal w/ `AllAttributes` here.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: `executionContext.AllAttributes.SetAttribute(originalName, value: false)`.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperRunnerTest.cs
```diff
@@ -216,10 +216,9 @@ public override void Process(TagHelperContext context, TagHelperOutput output)
 
                 output.TagName = "foo";
 
-                TagHelperAttribute classAttribute;
-                if (output.Attributes.TryGetAttribute("class", out classAttribute))
+                if (output.Attributes.ContainsName("class"))
                 {
-                    classAttribute.Value = "somethingelse";
+                    output.Attributes.SetAttribute("class", "somethingelse");
```
dougbu: Hmm, we might need a `ReadOnlyTagHelperAttributeList.IndexOf(string)` method.

The old code only changed the first attribute and did not remove additional matches. More like:
``` c#
    var index = output.Attributes.IndexOf("class");
    if (index >= 0)
    {
        output.Attributes[index] = new TagHelperAttribute("class", "something else");
    }
```

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
```diff
@@ -128,15 +128,16 @@ public static TheoryData StringIndexerData
...
+            TagHelperAttribute expectedAttribute)
         {
             // Arrange
-            var attributes = new ReadOnlyTagHelperAttributeList<IReadOnlyTagHelperAttribute>(initialAttributes);
+            var attributes = new ReadOnlyTagHelperAttributeList(initialAttributes);
+            TagHelperAttribute attribute;
 
             // Act
-            var attribute = attributes[nameToLookup];
+            attributes.TryGetAttribute(nameToLookup, out attribute);
```
dougbu: This is a new test and should be done in addition to the check of the indexer.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
```diff
@@ -668,33 +669,34 @@ public void ModifyingUnderlyingAttributes_AffectsExposedAttributes()
...
         {
             // Arrange
-            var attributes = new TestableReadOnlyTagHelperAttributes(Enumerable.Empty<IReadOnlyTagHelperAttribute>());
+            var attributes = new TestableReadOnlyTagHelperAttributes(Enumerable.Empty<TagHelperAttribute>());
             attributes.PublicAttributes.AddRange(initialAttributes);
+            TagHelperAttribute attribute;
 
             // Act
-            var attribute = attributes[nameToLookup];
+            attributes.TryGetAttribute(nameToLookup, out attribute);
```
dougbu: This is a new test and should be done in addition to the check of the indexer.

dougbu: See a few more changes away from the indexer later in this PR. Revert all of that because neither the indexer nor `TryGetAttribute()` are new or significantly changed. If `TryGetAttribute()` isn't tested sufficiently, add tests for it.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
```diff
@@ -668,33 +669,34 @@ public void ModifyingUnderlyingAttributes_AffectsExposedAttributes()
...
+        private class TestableReadOnlyTagHelperAttributes : ReadOnlyTagHelperAttributeList
         {
-            public TestableReadOnlyTagHelperAttributes(IEnumerable<IReadOnlyTagHelperAttribute> attributes)
+            public TestableReadOnlyTagHelperAttributes(IEnumerable<TagHelperAttribute> attributes)
                 : base(attributes)
             {
             }
 
-            public List<IReadOnlyTagHelperAttribute> PublicAttributes
+            public List<TagHelperAttribute> PublicAttributes
```
dougbu: Why `List<T>`? Wouldn't need the downcast if this was exposed as `IList<T>`.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperContextTest.cs
```diff
@@ -23,7 +23,7 @@ public void Constructor_SetsProperties_AsExpected()
 
             // Act
             var context = new TagHelperContext(
-                allAttributes: Enumerable.Empty<IReadOnlyTagHelperAttribute>(),
+                allAttributes: new ReadOnlyTagHelperAttributeList(Enumerable.Empty<TagHelperAttribute>()),
```
dougbu: Huh? `Enumerable.Empty<TagHelperAttribute>()` is a fine `IEnumerable<TagHelperAttribute` and the constructor will do the rest.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
dougbu: Ignore previous comment here.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: :ok:

dougbu:
:watch: especially for resolution of https://github.com/aspnet/Razor/pull/682#discussion-diff-51681956 and https://github.com/aspnet/Razor/pull/682#discussion-diff-51682582.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -8,77 +8,70 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
     /// <summary>
     /// An HTML tag helper attribute.
     /// </summary>
-    public class TagHelperAttribute : IReadOnlyTagHelperAttribute
+    public class TagHelperAttribute
     {
         private static readonly int TypeHashCode = typeof(TagHelperAttribute).GetHashCode();
```
dougbu: Get rid of this and provide a `GetHashCode()` implementation using all property values. That's possible now that `TagHelperAttribute` is immutable and may improve the efficiency of `TagHelperAttributeList`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -2,73 +2,52 @@
...
+    /// A read-only collection of <typeparamref name="TagHelperAttribute"/>s.
     /// </summary>
-    /// <typeparam name="TAttribute">
-    /// The type of <see cref="IReadOnlyTagHelperAttribute"/>s in the collection.
+    /// <typeparam name="TagHelperAttribute">
+    /// The type of <see cref="TagHelperAttribute"/>s in the collection.
     /// </typeparam>
-    public class ReadOnlyTagHelperAttributeList<TAttribute> : IReadOnlyList<TAttribute>
-        where TAttribute : IReadOnlyTagHelperAttribute
+    public class ReadOnlyTagHelperAttributeList : ReadOnlyCollection<TagHelperAttribute>
```
dougbu: Agree this shortens the class. However it prevents optimizations such as not creating the backing `List<T>` unless a value is written into the read/write subclass or this is instantiated w/ a non-empty collection. Suggest holding off on this particular part of the PR until we have our offline conversation on #683.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: Talked offline.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: This is an unintelligent poco object, not trying to make it too smart.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
NTaylorMullen: This was missed by me when I reverted the string indexer changes. Already have tests for `TryGetAttribute`

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/682#discussion_r51799530

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
NTaylorMullen: We call `AddRange` in the tests.

NTaylorMullen:
:up: :date: For unaddressed comments left over, we've talked offline.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -226,16 +146,30 @@ public bool TryGetAttributes(string name, out IReadOnlyList<TAttribute> attribut
             return matchedAttributes != null;
         }
 
-        /// <inheritdoc />
-        IEnumerator IEnumerable.GetEnumerator()
+        /// <summary>
+        /// Searches for a <see cref="TagHelperAttribute"/> who's <see cref="TagHelperAttribute.Name"/>
+        /// case-insensitively matches <param name="name"></param> and returns the zero-based index of the first
```
dougbu: `<paramref name="name"/>`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -226,16 +146,30 @@ public bool TryGetAttributes(string name, out IReadOnlyList<TAttribute> attribut
...
+        /// occurrence.
+        /// </summary>
+        /// <param name="name">The <see cref="TagHelperAttribute.Name"/> to locate in the collection.</param>
+        /// <returns>The zero-based index of the first <see cref="TagHelperAttribute"/> within the collection, if
```
dougbu: first __matching__. (Otherwise it would only ever return `0` or `-1`.)

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
dougbu: Fine now we have an agreed direction for #683 which doesn't depend on optimizing the innards of `ReadOnlyTagHelperAttributeList`.

dougbu: Could we make this class `abstract` and just _expose_ attribute collections in this way?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -2,73 +2,52 @@
...
 
         /// <summary>
-        /// Instantiates a new instance of <see cref="ReadOnlyTagHelperAttributeList{TAttribute}"/> with the specified
+        /// Instantiates a new instance of <see cref="ReadOnlyTagHelperAttributeList"/> with the specified
         /// <paramref name="attributes"/>.
         /// </summary>
         /// <param name="attributes">The collection to wrap.</param>
-        public ReadOnlyTagHelperAttributeList(IEnumerable<TAttribute> attributes)
+        public ReadOnlyTagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
+            : base(new List<TagHelperAttribute>(attributes))
```
dougbu: Is this constructor used in `ReadOnlyTagHelperAttributeList` copy scenarios? The defensive copy of the given collection (which may already be an instance of this type) looks wasteful, especially if callers have real collections and not just `IEnumerable<T>`s.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -78,106 +57,47 @@ public ReadOnlyTagHelperAttributeList(IEnumerable<TAttribute> attributes)
...
             }
-
-            return Attributes.Contains(item);
         }
 
         /// <summary>
-        /// Determines whether a <typeparamref name="TAttribute"/> with the same
-        /// <see cref="IReadOnlyTagHelperAttribute.Name"/> exists in the collection.
+        /// Determines whether a <typeparamref name="TagHelperAttribute"/> with the same
+        /// <see cref="TagHelperAttribute.Name"/> exists in the collection.
```
dougbu: The `string` indexer uses better words: `with <see cref="TagHelperAttribute.Name"/> matching <paramref name="name"/>`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -2,73 +2,52 @@
...
-        public TAttribute this[int index] => Attributes[index];
-
-        /// <summary>
-        /// Gets the first <typeparamref name="TAttribute"/> with <see cref="IReadOnlyTagHelperAttribute.Name"/>
+        /// Gets the first <typeparamref name="TagHelperAttribute"/> with <see cref="TagHelperAttribute.Name"/>
         /// matching <paramref name="name"/>.
         /// </summary>
         /// <param name="name">
-        /// The <see cref="IReadOnlyTagHelperAttribute.Name"/> of the <typeparamref name="TAttribute"/> to get.
+        /// The <see cref="TagHelperAttribute.Name"/> of the <typeparamref name="TagHelperAttribute"/> to get.
```
dougbu: `<see cref="TagHelperAttribute"/>`
Should have no `<typeparamref/>` elements in this class.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -2,83 +2,77 @@
...
+        /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>,
+        /// <paramref name="value"/> and <paramref name="minimized"/>.
         /// </summary>
-        /// <param name="name">The <see cref="Name"/> of the attribute.</param>
-        /// <param name="value">The <see cref="Value"/> of the attribute.</param>
-        public TagHelperAttribute(string name, object value)
+        /// <param name="name">The <see cref="Name"/> of the new instance.</param>
+        /// <param name="value">The <see cref="Value"/> of the new instance.</param>
+        /// <param name="minimized">The <see cref="Minimized"/> value of the new instance.</param>
+        /// <remarks>If <paramref name="minimized"/> is <c>true</c> <paramref name="value"/> is ignored when this
```
dougbu: comma after `<c>true</c>`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -98,7 +92,12 @@ public override bool Equals(object obj)
         /// <inheritdoc />
         public override int GetHashCode()
         {
-            return TypeHashCode;
+            var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(Name, StringComparer.Ordinal);
+            hashCodeCombiner.Add(Value);
+            hashCodeCombiner.Add(Minimized);
+
+            return hashCodeCombiner.CombinedHash;
```
dougbu: :+1:

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -305,10 +305,10 @@ public void AllAttributes_IgnoresCase(string originalName, string updatedName)
         {
             // Arrange
             var executionContext = new TagHelperExecutionContext("p", tagMode: TagMode.StartTagAndEndTag);
-            executionContext.AllAttributes.Add(originalName, value: false);
+            executionContext.HtmlAttributes.SetAttribute(originalName, value: false);
```
dougbu: Typo. Either don't change this line or add to the correct collection.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
dougbu: Fine to defer action on this to #683

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -51,124 +50,77 @@ public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
...
 
         /// <summary>
-        /// Gets the first <see cref="TagHelperAttribute"/> with <see cref="TagHelperAttribute.Name"/> matching
-        /// <paramref name="name"/>. When setting, replaces the first matching
-        /// <see cref="TagHelperAttribute"/> with the specified <paramref name="value"/> and removes any additional
-        /// matching <see cref="TagHelperAttribute"/>s. If a matching <see cref="TagHelperAttribute"/> is not found,
-        /// adds the specified <paramref name="value"/> to the end of the collection.
-        /// </summary>
+        /// Replaces the first <see cref="TagHelperAttribute"/> whose <see cref="TagHelperAttribute.Name"/> matches the
+        /// specified <paramref name="name"/> and removes any additional matching <see cref="TagHelperAttribute"/>s. If
```
dougbu: Stick w/ `with <see cref="TagHelperAttribute.Name"/> matching <paramref name="name"/>` i.e. `Replaces the first <see cref="TagHelperAttribute"/> with ...`.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/ReadOnlyTagHelperAttributeListTest.cs
dougbu: Shoot, :ok:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: Have discussed API changes offline. All set.

dougbu:
:watch: for last few doc comment comments and comments about comments on doc comments. Plus `abstract`ion.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
NTaylorMullen: ya gonna defer

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
NTaylorMullen: lmao look at that

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/ReadOnlyTagHelperAttributeList.cs
```diff
@@ -226,16 +146,30 @@ public bool TryGetAttributes(string name, out IReadOnlyList<TAttribute> attribut
             return matchedAttributes != null;
         }
 
-        /// <inheritdoc />
-        IEnumerator IEnumerable.GetEnumerator()
+        /// <summary>
+        /// Searches for a <see cref="TagHelperAttribute"/> who's <see cref="TagHelperAttribute.Name"/>
+        /// case-insensitively matches <param name="name"/> and returns the zero-based index of the first
```
dougbu: param __ref__

dougbu:
:shipit: 
My apologies for ending this before we hit :100: comments.

PR #683, opened 2016-01-30
Lazily initialize `TagHelperAttributeList`s.

NTaylorMullen:
- Not all `TagHelper`s have unbound HTML attributes or any attributes at all. A great example of this is MVC's input `TagHelper` which usually takes the format of `<input asp-for="..." />`. By lazily initializing we don't build extra attribute lists where not needed.

**Before:**
image

**After:**
image

**Result:** 1.4% reduction in overall allocations

#604

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -214,9 +214,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             _allAttributes.Add(name, value);
         }
 
-        /// <summary>
-        /// Internal for testing.
-        /// </summary>
+        // Internal for testing.
         internal async Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEncoder encoder)
         {
             // Get cached content for this encoder.
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -14,12 +14,17 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperExecutionContext
     {
+        private readonly string _tagName;
+        private readonly string _uniqueId;
+        private readonly TagMode _tagMode;
         private readonly List<ITagHelper> _tagHelpers;
         private readonly Func<Task> _executeChildContentAsync;
         private readonly Action<HtmlEncoder> _startTagHelperWritingScope;
         private readonly Func<TagHelperContent> _endTagHelperWritingScope;
         private TagHelperContent _childContent;
         private Dictionary<HtmlEncoder, TagHelperContent> _perEncoderChildContent;
+        private TagHelperAttributeList _htmlAttributes;
+        private TagHelperAttributeList _allAttributes;
 
         /// <summary>
         /// Internal for testing purposes only.
@@ -93,18 +98,11 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             _startTagHelperWritingScope = startTagHelperWritingScope;
             _endTagHelperWritingScope = endTagHelperWritingScope;
 
-            TagMode = tagMode;
-            HtmlAttributes = new TagHelperAttributeList();
-            AllAttributes = new TagHelperAttributeList();
-            TagName = tagName;
+            _tagMode = tagMode;
+            _tagName = tagName;
             Items = items;
-            UniqueId = uniqueId;
+            _uniqueId = uniqueId;
         }
-
-        /// <summary>
-        /// Gets the HTML syntax of the element in the Razor source.
-        /// </summary>
-        public TagMode TagMode { get; }
 
         /// <summary>
         /// Indicates if <see cref="GetChildContentAsync"/> has been called.
@@ -123,21 +121,6 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         public IDictionary<object, object> Items { get; }
 
         /// <summary>
-        /// HTML attributes.
-        /// </summary>
-        public TagHelperAttributeList HtmlAttributes { get; }
-
-        /// <summary>
-        /// <see cref="ITagHelper"/> bound attributes and HTML attributes.
-        /// </summary>
-        public TagHelperAttributeList AllAttributes { get; }
-
-        /// <summary>
-        /// An identifier unique to the HTML element this context is for.
-        /// </summary>
-        public string UniqueId { get; }
-
-        /// <summary>
         /// <see cref="ITagHelper"/>s that should be run.
         /// </summary>
         public IList<ITagHelper> TagHelpers
@@ -149,14 +132,17 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// The HTML tag name in the Razor source.
-        /// </summary>
-        public string TagName { get; }
-
-        /// <summary>
         /// The <see cref="ITagHelper"/>s' output.
         /// </summary>
         public TagHelperOutput Output { get; set; }
+
+        public TagHelperContext CreateTagHelperContext() =>
+            new TagHelperContext(_allAttributes, Items, _uniqueId);
+        public TagHelperOutput CreateTagHelperOutput() =>
+            new TagHelperOutput(_tagName, _htmlAttributes, GetChildContentAsync)
+            {
+                TagMode = _tagMode
+            };
 
         /// <summary>
         /// Tracks the given <paramref name="tagHelper"/>.
@@ -183,9 +169,12 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 throw new ArgumentNullException(nameof(name));
             }
 
+            EnsureHtmlAttributes();
+            EnsureAllAttributes();
+
             var attribute = new TagHelperAttribute(name);
-            HtmlAttributes.Add(attribute);
-            AllAttributes.Add(attribute);
+            _htmlAttributes.Add(attribute);
+            _allAttributes.Add(attribute);
         }
 
         /// <summary>
@@ -200,8 +189,12 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 throw new ArgumentNullException(nameof(name));
             }
 
-            HtmlAttributes.Add(name, value);
-            AllAttributes.Add(name, value);
+            EnsureHtmlAttributes();
+            EnsureAllAttributes();
+
+            var attribute = new TagHelperAttribute(name, value);
+            _htmlAttributes.Add(attribute);
+            _allAttributes.Add(attribute);
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -125,12 +125,12 @@ public bool ChildContentRetrieved
...
         /// HTML attributes.
         /// </summary>
-        public TagHelperAttributeList HTMLAttributes { get; }
+        public TagHelperAttributeList HTMLAttributes => _htmlAttributes;
 
         /// <summary>
         /// <see cref="ITagHelper"/> bound attributes and HTML attributes.
         /// </summary>
-        public TagHelperAttributeList AllAttributes { get; }
+        public TagHelperAttributeList AllAttributes => _allAttributes;
```
rynowak: This seems like it's going to have public surface area impact? Am I just misunderstanding?

dougbu: @rynowak you're misunderstanding slightly. `AllAttributes` and `HTMLAttributes` (yeah, naming issue as you mentioned in #682) remain `public` get-only properties.

@NTaylorMullen why aren't these bog-standard lazy-loaded properties? That is
``` c#
    public TagHelperAttributeList AllAttributes
    {
        get
        {
            if (_allAttributes == null)
            ...
            return _allAttributes;
        }
    }
```

dougbu: In particular the getter should never return `null` because the caller's next action is almost-always going to involve a dereference.

dougbu:
Suggest we'll see about the same allocation improvements with less complexity if `TagHelperAttributeList` and `ReadOnlyTagHelperAttributeList` lazy-loaded the inner `List<TagHelperAttribute>` and handled the `_attributes == null` cases as needed. This introduces sometimes-null collection properties and other oddities.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Reason why I don't put the `null` check in the getter is because of how this property is consumed when constructing a `TagHelperOutput`/`TagHelperContext` [here](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperRunner.cs#L44). 

The alternative would be to add public methods that indicate if attributes have been added `HtmlAttributesCreated`/`AllAttributesCreated`. Since `TagHelperExecutionContext`s will never be used by end-users I figured it wasn't worth the added API.

NTaylorMullen:
> Suggest we'll see about the same allocation improvements with less complexity if TagHelperAttributeList and ReadOnlyTagHelperAttributeList lazy-loaded the inner List<TagHelperAttribute> and handled the _attributes == null cases as needed. This introduces sometimes-null collection properties and other oddities.

Should talk offline about this. Not sure how `TagHelperAttributes` could special case `null` without the rest of the system above it (the system that constructs it) allowing `null`.

dougbu:
> Should talk offline about this

:ok:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Let's mix this into the https://github.com/aspnet/Razor/pull/683#issuecomment-177061807 offline conversation.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -216,24 +209,15 @@ public void AddTagHelperAttribute(string name, object value)
...
-            AllAttributes.Add(name, value);
+            EnsureAllAttributes();
+
+            _allAttributes.Add(name, value);
         }
 
         /// <summary>
-        /// Executes children asynchronously with the given <paramref name="encoder"/> in scope and returns their
-        /// rendered content.
+        /// Internal for testing.
```
dougbu: Just `//`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
```diff
@@ -25,6 +25,20 @@ public TagHelperAttributeList()
         /// </summary>
         /// <param name="attributes">The collection to wrap.</param>
         public TagHelperAttributeList(IEnumerable<TagHelperAttribute> attributes)
+            : base (new List<TagHelperAttribute>(attributes))
```
dougbu: Where is this copy actually required? I'm hoping this constructor could be `internal` and just `// Internal for testing`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
dougbu: Remove. It's unused.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -286,41 +293,44 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
                 writer.Write("<");
                 writer.Write(TagName);
 
-                // Perf: Avoid allocating enumerator
-                for (var i = 0; i < Attributes.Count; i++)
+                if (_attributes != null)
```
dougbu: Wow, not going for the #PranavPoints? Could have been `for (var i = 0; i < _attributes?.Count ?? 0; i++)`. :trollface:

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -270,51 +273,6 @@ public static TheoryData<HtmlEncoder> HtmlEncoderDataLessNull
...
-            var executionContext = new TagHelperExecutionContext("p", tagMode: TagMode.StartTagAndEndTag);
-            executionContext.AllAttributes.SetAttribute(originalName, value: false);
-
-            // Act
-            executionContext.AllAttributes.SetAttribute(updatedName, true);
-
-            // Assert
-            var attribute = Assert.Single(executionContext.AllAttributes);
-            Assert.Equal(new TagHelperAttribute(originalName, true), attribute);
-        }
```
dougbu: Restore these tests because the scenarios aren't covered otherwise. Either expose `AllAttributes` and `HtmlAttributes` as `// Internal for testing` or create a `TagHelperContext` / `TagHelperOutput` and check what that contains.

dougbu:
@NTaylorMullen please show the before and after memory numbers for this fix.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttributeList.cs
NTaylorMullen: Testing is a big part of this. We use it extensively in our tests but also when trying to copy `TagHelperAttributeList`s among each other. For instance we do this in the `Script` and `Link` `TagHelper`s in MVC.

Ultimately, I'd like to leave this in for ease of use. Without it, it'd makes the class quite annoying to test (no more returning `new[] { ... }`), it'd be over 50 changes in our own tests even.

dougbu: Tests are fine but do the MVC tag helpers require the allocations here? If yes, leave as-is. If no, make it `internal` (which shouldn't impact Razor tests).

Put another way, could the `<script>` and `<link>` tag helpers do something less obviously expensive e.g. reuse one attribute list? Or could we support the scenario with a copy constructor -- allowing the new `List<T>` to be allocated w/ a known size?

NTaylorMullen: Talked offline. Gonna leave as is.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: I like pranav points here. I approve!

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
NTaylorMullen: Nah, this is testing dictionary casing lol. These tests shouldn't have been added to begin with, it's repeatedly tested elsewhere and we never test comparers like this directly.

NTaylorMullen:
:up: :date: 

Working on an MVC reaction PR to get new numbers.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
dougbu: The "dictionary casing" you mention is implemented entirely in `ReadOnlyTagHelperAttributeList` and `TagHelperAttributeList`. The backing store isn't a dictionary and these classes do their own, non-default comparisons. These tests made sure we didn't miss a few things.

Or are you suggesting we need _more_ tests of the other dictionary-like methods in these classes -- and that'll be coming in a subsequent iteration?

dougbu: Of course, we're fine if I'm missing thorough coverage in the `ReadOnlyTagHelperAttributeList` and `TagHelperAttributeList` classes.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -149,15 +132,18 @@ public IList<ITagHelper> TagHelpers
...
-        public string TagName { get; }
-
-        /// <summary>
         /// The <see cref="ITagHelper"/>s' output.
         /// </summary>
         public TagHelperOutput Output { get; set; }
 
+        public TagHelperContext CreateTagHelperContext() =>
+            new TagHelperContext(_allAttributes, Items, _uniqueId);
+        public TagHelperOutput CreateTagHelperOutput() =>
```
dougbu: Minor but a blank line would help.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -287,9 +294,9 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
                 writer.Write(TagName);
 
                 // Perf: Avoid allocating enumerator
-                for (var i = 0; i < Attributes.Count; i++)
+                for (var i = 0; i < (_attributes?.Count ?? 0); i++)
```
dougbu: Nice. No need for `?w=0` on the GitHub URL. :smiley_cat:

dougbu:
:watch:ing but not for much beyond the numbers and reaction PR.

NTaylorMullen:
**Before**:
image


**After**:
image

PR #688, opened 2016-02-09
Enable Razor templates inside of sections.

NTaylorMullen:
- Changed the variable used to render templates (`HelperResult`s) so they can be used inside of sections.
- Updated section test file to showcase nested scenario. Also Regenerated existing test files.

#572


pranavkm:
Would this cause a perf regression?

NTaylorMullen:
lol such a :trollface:

pranavkm:
:shipit:

Issue #346, opened 2015-04-03
Escape shortcut fails when first text character is a `:`


Buildstarted:
@if(true) {
        @::Something
    }

The above code block produces this error: `"::" is not valid at the start of a code block. Only identifiers, keywords, comments, "(" and "{" are valid.`

The expected output is `:Something`. I can get around this by using `<text>:Something</text>`. My assumption is because `::` is a valid namespace qualifier in c#.

tkirill:
Hi!

I want to try to fix this, hope `up-for-grabs` bug isn't too difficult.  What branch should I choose as starting point?  `master` or `dev`?

Eilon:
@NTaylorMullen move this to the right milestone and close as Done?

NTaylorMullen:
Si :smile:

NTaylorMullen:
a0164b005d9bdd0ee59de3c90c38499fdc8ebd15

PR #689, opened 2016-02-10
(#346) Fixed case when razor fails to recognize escape shortcut

heXelium:
Fixed bug  #346. When the first character in a text is `:` for example @::sometext parser fails to trigger second colon as a part of plain text.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Parser/CSharpCodeParser.Statements.cs
+++ b/src/Microsoft.AspNetCore.Razor/Parser/CSharpCodeParser.Statements.cs
@@ -464,7 +464,10 @@ namespace Microsoft.AspNetCore.Razor.Parser
             var type = CurrentSymbol.Type;
             var loc = CurrentLocation;
 
-            var isSingleLineMarkup = type == CSharpSymbolType.Transition && NextIs(CSharpSymbolType.Colon);
+            // Both cases @: and @:: are triggered as markup, second colon in second case will be triggered as a plain text
+            var isSingleLineMarkup = type == CSharpSymbolType.Transition &&
+                (NextIs(CSharpSymbolType.Colon, CSharpSymbolType.DoubleColon));
+
             var isMarkup = isSingleLineMarkup ||
                 type == CSharpSymbolType.LessThan ||
                 (type == CSharpSymbolType.Transition && NextIs(CSharpSymbolType.LessThan));
--- a/test/Microsoft.AspNetCore.Razor.Test/Parser/CSharp/CSharpToMarkupSwitchTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Test/Parser/CSharp/CSharpToMarkupSwitchTest.cs
@@ -553,6 +553,43 @@ namespace Microsoft.AspNetCore.Razor.Test.Parser.CSharp
         }
 
         [Fact]
+        public void ParseBlockParsesMarkupStatementOnSwitchCharacterFollowedByDoubleColon()
+        {
+            // Arrange
+            ParseBlockTest("if(foo) { @::Sometext" + Environment.NewLine
+                         + "}",
+                           new StatementBlock(
+                               Factory.Code("if(foo) {").AsStatement(),
+                               new MarkupBlock(
+                                   Factory.Markup(" "),
+                                   Factory.MarkupTransition(),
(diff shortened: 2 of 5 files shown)


Eilon:
@heXelium could you also log a bug on this to make it clear exactly what scenario doesn't work, and what the expected behavior is?

Eilon:
@NTaylorMullen can you have a look?

src/Microsoft.AspNetCore.Razor/Parser/CSharpCodeParser.Statements.cs
```diff
@@ -464,7 +464,10 @@ private void Statement(Block block)
             var type = CurrentSymbol.Type;
             var loc = CurrentLocation;
 
-            var isSingleLineMarkup = type == CSharpSymbolType.Transition && NextIs(CSharpSymbolType.Colon);
+            //both cases @: and @:: are triggered as markup, second colon in second case will be triggered as a plain text
```
NTaylorMullen: Our format for comments is typically `// Hello World`. Note the space and the cap letter.

Also, if you could space out the line below: `(NextIs(CSharpSymbolType.Colon) || NextIs(CSharpSymbolType.DoubleColon));`

test/Microsoft.AspNetCore.Razor.Test/Parser/CSharp/CSharpToMarkupSwitchTest.cs
```diff
@@ -553,6 +553,24 @@ public void ParseBlockParsesMarkupStatementOnSwitchCharacterFollowedByColon()
         }
 
         [Fact]
+        public void ParseBlockParsesMarkupStatementOnSwitchCharacterFollowedByDoubleColon()
```
NTaylorMullen: Add a few more tests.

1. Triple colon. The tokenizer special cases double colon so we want to make sure triple doesn't cause any issues.
1. A code generation test that has the double and triple colon in the markup. You can add those two lines [here](https://github.com/aspnet/Razor/blob/173886bf6cda465e839b60d262faf7f614c3b2ec/test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Source/Templates.cshtml#L22). To regenerate the corresponding C# files delete the `__RemoveThisBitTo__` text from [this](https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Razor.Test/project.json#L37) file. Run `./build verify` at the root of the project.

NTaylorMullen:
:watch:

heXelium:
@NTaylorMullen added one more UT, also fixed coding style as you requested. 
Regarding code generation tests, I followed your instructions. After adding into ```Templates.cshtml```:
```html
<p>
@Repeat(10,
    @::This is line#@item of markup<br />
)
</p>

<p>
@Repeat(10,
    @:::This is line#@item of markup<br />
)
</p>
```
and executing build, I got results:
```csharp
#line 27 "Templates.cshtml"
Write(Repeat(10,
    @::This is line#@item of markup<br />
));
...
#line 33 "Templates.cshtml"
Write(Repeat(10,
    @:::This is line#@item of markup<br />
));
```
It is look like that they are incorrect. I'm not sure if this code generation was executed with fixed parser, because I don't understand what is triggering that files to be generated. Could you please explain me this points? Probably I did something wrong.

NTaylorMullen:
First off, you're a brave soul tackling the Razor parser :smile:. I took a peak and it looks like your code paths aren't properly hitting the single line transition code paths. Take a look at:

https://github.com/aspnet/Razor/blob/173886bf6cda465e839b60d262faf7f614c3b2ec/src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs#L83-L124

heXelium:
Thank you, eventually I found out how to work with this CodeGeneration tests. And the reason why it was not working it's because there are another place where this transition was not handled correctly, so I fixed that as well. Please take a look on it, and if you think of another test cases I will add them as well.

NTaylorMullen:
Looks good. :shipit: 

If you could squash your commits i'll be happy to merge them in.

heXelium:
Sure. Done

NTaylorMullen:
Merged :smile:. Thanks @heXelium for your contribution!

PR #694, opened 2016-02-17
React to changes in HTML abstractions

rynowak:
This change implements the new API for flattening content in Razor.

Also, some optimizations to avoid allocations on paths where we need
to encode HTML attribute values.

### DIFF
--- a/Razor.sln
+++ b/Razor.sln
@@ -1,7 +1,6 @@
-
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.24720.0
+VisualStudioVersion = 14.0.24711.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{3C0D6505-79B3-49D0-B4C3-176F0F1836ED}"
 EndProject
@@ -16,6 +15,10 @@ EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNetCore.Razor.Runtime.Test", "test\Microsoft.AspNetCore.Razor.Runtime.Test\Microsoft.AspNetCore.Razor.Runtime.Test.xproj", "{0535998A-E32C-4D1A-80D1-0B15A513C471}"
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNetCore.Razor.Test.Sources", "src\Microsoft.AspNetCore.Razor.Test.Sources\Microsoft.AspNetCore.Razor.Test.Sources.xproj", "{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNetCore.Html.Abstractions", "..\HtmlAbstractions\src\Microsoft.AspNetCore.Html.Abstractions\Microsoft.AspNetCore.Html.Abstractions.xproj", "{68A28E4A-3ADE-4187-9625-4FF185887CB3}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Extensions.WebEncoders", "..\HtmlAbstractions\src\Microsoft.Extensions.WebEncoders\Microsoft.Extensions.WebEncoders.xproj", "{DD2CE416-765E-4000-A03E-C2FF165DA1B6}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -43,6 +46,14 @@ Global
 		{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Release|Any CPU.Build.0 = Release|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/global.json
+++ b/global.json
@@ -1,3 +1,3 @@
 {
-    "projects": ["src"]
+    "projects": ["src", "d:\\k\\HtmlAbstractions\\src"]
 }
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -149,6 +149,29 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         }
 
         /// <inheritdoc />
+        public override void CopyTo(IHtmlContentBuilder destination)
+        {
+            for (var i = 0; i < Buffer.Count; i++)
+            {
+                var entry = Buffer[i];
+                if (entry == null)
+                {
+                    continue;
+                }
+
+                var stringValue = entry as string;
+                if (stringValue != null)
+                {
+                    destination.Append(stringValue);
+                }
+                else
+                {
+                    destination.AppendHtml((IHtmlContent)entry);
+                }
+            }
+        }
+
+        /// <inheritdoc />
         public override TagHelperContent Clear()
         {
             Buffer.Clear();
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -151,6 +151,16 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         /// <inheritdoc />
         public override void CopyTo(IHtmlContentBuilder destination)
         {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            if (_buffer == null)
+            {
+                return;
+            }
+
             for (var i = 0; i < Buffer.Count; i++)
             {
                 var entry = Buffer[i];
@@ -159,16 +169,61 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                     continue;
                 }
 
-                var stringValue = entry as string;
-                if (stringValue != null)
+                string entryAsString;
+                IHtmlContentContainer entryAsContainer;
+                if ((entryAsString = entry as string) != null)
                 {
-                    destination.Append(stringValue);
+                    destination.Append(entryAsString);
+                }
+                else if ((entryAsContainer = entry as IHtmlContentContainer) != null)
+                {
+                    entryAsContainer.CopyTo(destination);
                 }
                 else
                 {
                     destination.AppendHtml((IHtmlContent)entry);
                 }
             }
+        }
+
+        /// <inheritdoc />
+        public override void MoveTo(IHtmlContentBuilder destination)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            if (_buffer == null)
+            {
+                return;
+            }
+
+            for (var i = 0; i < Buffer.Count; i++)
+            {
+                var entry = Buffer[i];
+                if (entry == null)
+                {
+                    continue;
+                }
+
+                string entryAsString;
+                IHtmlContentContainer entryAsContainer;
+                if ((entryAsString = entry as string) != null)
+                {
+                    destination.Append(entryAsString);
+                }
+                else if ((entryAsContainer = entry as IHtmlContentContainer) != null)
+                {
+                    entryAsContainer.MoveTo(destination);
+                }
+                else
+                {
+                    destination.AppendHtml((IHtmlContent)entry);
+                }
+            }
+
+            Buffer.Clear();
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
@@ -129,6 +129,12 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         public abstract TagHelperContent Clear();
 
         /// <summary>
+        /// Copies the content to another <see cref="IHtmlContentBuilder"/>.
+        /// </summary>
+        /// <param name="destination">The destination <see cref="IHtmlContentBuilder"/>.</param>
+        public abstract void CopyTo(IHtmlContentBuilder destination);
+
+        /// <summary>
         /// Gets the content.
         /// </summary>
         /// <returns>A <see cref="string"/> containing the content.</returns>
@@ -166,6 +172,18 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         IHtmlContentBuilder IHtmlContentBuilder.Clear()
         {
             return Clear();
+        }
+
+        /// <inheritdoc />
+        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            CopyTo(destination);
+            Clear();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
@@ -128,11 +128,11 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         /// <returns>A reference to this instance after the clear operation has completed.</returns>
         public abstract TagHelperContent Clear();
 
-        /// <summary>
-        /// Copies the content to another <see cref="IHtmlContentBuilder"/>.
-        /// </summary>
-        /// <param name="destination">The destination <see cref="IHtmlContentBuilder"/>.</param>
+        /// <inheritdoc />
         public abstract void CopyTo(IHtmlContentBuilder destination);
+
+        /// <inheritdoc />
+        public abstract void MoveTo(IHtmlContentBuilder destination);
 
         /// <summary>
         /// Gets the content.
@@ -172,18 +172,6 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         IHtmlContentBuilder IHtmlContentBuilder.Clear()
         {
             return Clear();
-        }
-
-        /// <inheritdoc />
-        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
-        {
-            if (destination == null)
-            {
-                throw new ArgumentNullException(nameof(destination));
-            }
-
-            CopyTo(destination);
-            Clear();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -287,41 +287,8 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 destination.AppendHtml("<");
                 destination.AppendHtml(TagName);
 
-                // Perf: Avoid allocating enumerator
-                for (var i = 0; i < (_attributes?.Count ?? 0); i++)
-                {
-                    var attribute = _attributes[i];
-                    destination.AppendHtml(" ");
-                    destination.AppendHtml(attribute.Name);
-
-                    if (attribute.Minimized)
-                    {
-                        continue;
-                    }
-
-                    destination.AppendHtml("=\"");
-                    var value = attribute.Value;
-                    var htmlContent = value as IHtmlContent;
-                    if (htmlContent != null)
-                    {
-                        // Perf: static text in a bound attribute go down this path. Avoid allocating if possible (common case).
-                        var htmlEncodedString = value as HtmlEncodedString;
-                        if (htmlEncodedString != null && !htmlEncodedString.Value.Contains("\""))
-                        {
-                            destination.AppendHtml(htmlEncodedString);
-                        }
-                        else
-                        {
-                            destination.AppendHtml(new AttributeContent(htmlContent));
-                        }
-                    }
-                    else if (value != null)
-                    {
-                        destination.Append(value.ToString());
-                    }
-
-                    destination.AppendHtml("\"");
-                }
+                CopyAttributesTo(destination);
+                _attributes?.Clear();
 
                 if (TagMode == TagMode.SelfClosing)
                 {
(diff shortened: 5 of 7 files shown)


Razor.sln
```diff
@@ -43,6 +46,14 @@ Global
...
 		{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{E3A2A305-634D-4BA6-95DB-AA06D6C442B0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{68A28E4A-3ADE-4187-9625-4FF185887CB3}.Release|Any CPU.Build.0 = Release|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{DD2CE416-765E-4000-A03E-C2FF165DA1B6}.Release|Any CPU.Build.0 = Release|Any CPU
```
rynowak: Ignore #YOLO

global.json
```diff
@@ -1,3 +1,3 @@
 {
-    "projects": ["src"]
+    "projects": ["src", "d:\\k\\HtmlAbstractions\\src"]
```
rynowak: MOAR IGNOAR

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -149,6 +149,29 @@ public override TagHelperContent AppendHtml(IHtmlContent htmlContent)
         }
 
         /// <inheritdoc />
+        public override void CopyTo(IHtmlContentBuilder destination)
```
rynowak: Will add test

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -129,6 +129,12 @@ public TagHelperContent AppendFormat(IFormatProvider provider, string format, pa
         public abstract TagHelperContent Clear();
 
         /// <summary>
+        /// Copies the content to another <see cref="IHtmlContentBuilder"/>.
+        /// </summary>
+        /// <param name="destination">The destination <see cref="IHtmlContentBuilder"/>.</param>
```
rynowak: inheritdoc

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -310,21 +450,30 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+
+                                var stringValue = stringWriter.ToString();
+                                writer.Write(stringValue);
+                            }
                         }
                     }
                     else if (value != null)
                     {
-                        encoder.Encode(writer, value.ToString());
+                        writer.Write(value.ToString());
```
rynowak: This is a copy-paste error, will fix

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
             return _getChildContentAsync(useCachedResult, encoder);
         }
 
-        /// <inheritdoc />
-        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+        void IHtmlContentContainer.CopyTo(IHtmlContentBuilder destination)
```
rynowak: Will add tests for new functionality

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -149,6 +149,29 @@ public override TagHelperContent AppendHtml(IHtmlContent htmlContent)
...
+        {
+            for (var i = 0; i < Buffer.Count; i++)
+            {
+                var entry = Buffer[i];
+                if (entry == null)
```
dougbu: aspnet/HttpAbstractions#11 doesn't special case `null`. Should it?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -149,6 +149,29 @@ public override TagHelperContent AppendHtml(IHtmlContent htmlContent)
...
+                }
+
+                var stringValue = entry as string;
+                if (stringValue != null)
+                {
+                    destination.Append(stringValue);
+                }
+                else
+                {
+                    destination.AppendHtml((IHtmlContent)entry);
```
dougbu: Should `IHtmlContentContainer` be special-cased here?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -167,5 +173,17 @@ IHtmlContentBuilder IHtmlContentBuilder.Clear()
         {
             return Clear();
         }
+
+        /// <inheritdoc />
+        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
+        {
+            if (destination == null)
```
dougbu: Good point. `null` checks are missing in `CopyTo()` as well as aspnet/HttpAbstractions#11 in general.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+                            destination.AppendHtml(htmlEncodedString);
+                        }
+                        else
+                        {
+                            destination.AppendHtml(new AttributeContent(htmlContent));
+                        }
+                    }
+                    else if (value != null)
+                    {
+                        destination.Append(value.ToString());
```
dougbu: Why `ToString()`? IOW what can `value` be other than a `string` or an `IHtmlContent` implementation?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+                    {
+                        continue;
+                    }
+
+                    destination.AppendHtml("=\"");
+                    var value = attribute.Value;
+                    var htmlContent = value as IHtmlContent;
+                    if (htmlContent != null)
+                    {
+                        // Perf: static text in a bound attribute go down this path. Avoid allocating if possible (common case).
```
dougbu: Note we'll only go down this path if the _entire_ attribute value is one bit of static text. May need to make quote escaping more first-class to handle static text in mixed cases.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+                destination.AppendHtml(">");
+            }
+
+            _postElement?.CopyTo(destination);
+        }
+
+        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
         {
-            if (writer == null)
+            ((IHtmlContentBuilder)_preContent)?.MoveTo(destination);
```
dougbu: `_preElement`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+            {
+                destination.AppendHtml("</");
+                destination.AppendHtml(TagName);
+                destination.AppendHtml(">");
+            }
+
+            _postElement?.CopyTo(destination);
+        }
+
+        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
```
dougbu: Lots of duplication. Why not call `CopyTo()` then `Clear()` the non-`null` fields?

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -981,7 +981,9 @@ public void WriteTo_WritesFormattedTagHelper_HtmlTextWriter(TagHelperOutput outp
             // Arrange
```
dougbu: Update test name since you're no longer testing handling of `HtmlTextWriter`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -356,5 +505,30 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+            }
+
+            public IHtmlContent Inner { get; }
+
+            public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+            {
+                // There's no way of tracking the attribute value quotations in the Razor source. Therefore, we
+                // must escape any IHtmlContent double quote values in the case that a user wrote:
+                // <p name='A " is valid in single quotes'></p>
+                using (var stringWriter = new StringWriter())
```
dougbu: Suggest allocating the `StringWriter()` in the containing class (lazily), passing into the constructor, and clearing on every `WriteTo()` call. That should somewhat reduce the explosion of `char[]`, `StringBuilder` and `StringWriter` instances.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -356,5 +505,30 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
         }
+
+        private class AttributeContent : IHtmlContent
+        {
+            public AttributeContent(IHtmlContent inner)
+            {
+                Inner = inner;
+            }
+
+            public IHtmlContent Inner { get; }
```
dougbu: Surprised this isn't a `private` field. Where is this needed / used?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: Never mind. Forgot for a sec how general `TagHelperAttribute.Value` is.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,154 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+                            destination.AppendHtml(htmlEncodedString);
+                        }
+                        else
+                        {
+                            destination.AppendHtml(new AttributeContent(htmlContent));
+                        }
+                    }
+                    else if (value != null)
+                    {
+                        destination.Append(value.ToString());
```
dougbu: Slightly tangential to this PR: Some `ToString()` implementations are quite expensive. Bet we'd see real savings in any scenario involving `DateTimeOffset` or similar `value`s if `TagHelperAttribute` exposed `string ValueString { get; }` but left it `null` when `Value is IHtmlContent`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: Discussed offline about need to recursively move everything down the tree.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
rynowak: `HtmlContentBuilder` doesn't ever contain a null. The taghelper things do. I dunno why, but it's not part of this change.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
rynowak: Yes. Added

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
rynowak: Those APIs never store `null` so they don't need it.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: This proved to be an extremely profitable optimization in the case I was looking at.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: Fixed

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: Yeah, that actually seems really nice.

rynowak: :+1: awesome suggestion. Works out really well.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: I just code things that way by default. I'll change it #YOLO

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: Yeah I'd love to dig into this more. I well like the way attributes are handled in general is slightly suboptimal.

rynowak:
@dougbu updated

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -149,6 +149,84 @@ public override TagHelperContent AppendHtml(IHtmlContent htmlContent)
...
+                {
+                    entryAsContainer.MoveTo(destination);
+                }
+                else
+                {
+                    destination.AppendHtml((IHtmlContent)entry);
+                }
+            }
+
+            Buffer.Clear();
```
ajaybhargavb: `Clear()` instead of `Buffer.Clear()`?

rynowak: Why?

ajaybhargavb: It doesn't matter. I just thought it is better to do it all in one place.

rynowak: The reason why I did `Buffer.Clear` is that this method is already operating directly on the buffer. It would mix levels of abstraction to call `Clear()`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
             return _getChildContentAsync(useCachedResult, encoder);
         }
 
-        /// <inheritdoc />
-        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+        void IHtmlContentContainer.CopyTo(IHtmlContentBuilder destination)
         {
-            if (writer == null)
```
dougbu: Add `null` check for `destination`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+                destination.AppendHtml(TagName);
+                destination.AppendHtml(">");
+            }
+
+            _postElement?.CopyTo(destination);
+        }
+
+        void IHtmlContentContainer.MoveTo(IHtmlContentBuilder destination)
+        {
+            _preElement?.MoveTo(destination);
```
dougbu: Add `null` check for `destination`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+
+            var isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(TagName);
+
+            if (!isTagNameNullOrWhitespace)
+            {
+                destination.AppendHtml("<");
+                destination.AppendHtml(TagName);
+
+                CopyAttributesTo(destination);
+                _attributes?.Clear();
```
dougbu: Why call `Clear()` in the `CopyTo()` method?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -276,14 +276,87 @@ public Task<TagHelperContent> GetChildContentAsync(bool useCachedResult, HtmlEnc
...
+
+            var isTagNameNullOrWhitespace = string.IsNullOrWhiteSpace(TagName);
+
+            if (!isTagNameNullOrWhitespace)
+            {
+                destination.AppendHtml("<");
+                destination.AppendHtml(TagName);
+                
+                CopyAttributesTo(destination);
+
```
dougbu: `Clear()` the `_attributes`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -356,5 +438,76 @@ public void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                destination.AppendHtml(attribute.Name);
+
+                if (attribute.Minimized)
+                {
+                    continue;
+                }
+
+                destination.AppendHtml("=\"");
+                var value = attribute.Value;
+                var htmlContent = value as IHtmlContent;
```
dougbu: Is it impossible for `_attributes` to contain an `IHtmlContentBuilder`. Unsure this is the Right Thing:tm: in the `MoveTo()` case if not.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -972,16 +972,15 @@ public void WriteTo_WritesFormattedTagHelper(TagHelperOutput output, string expe
             Assert.Equal(expected, writer.ToString(), StringComparer.Ordinal);
         }
 
-        // This tests a separate code path that's used by THO when the writer is an HtmlTextWriter.
-        // The output should be the same, but we do some specific perf optimizations on this path.
         [Theory]
         [MemberData(nameof(WriteTagHelper_InputData))]
-        public void WriteTo_WritesFormattedTagHelper_HtmlTextWriter(TagHelperOutput output, string expected)
+        public void CopyTo_CopiesToBuilder(TagHelperOutput output, string expected)
```
dougbu: Add either a test that you can call `CopyTo()` twice and get the same result or a test that confirms `Attributes.Count` before equals `Attributes.Count` after.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -994,10 +993,48 @@ public void WriteTo_WritesFormattedTagHelper_HtmlTextWriter(TagHelperOutput outp
...
             // Assert
-            Assert.Equal(expected, inner.ToString(), StringComparer.Ordinal);
+            buffer.WriteTo(writer, testEncoder);
+
+            Assert.True(output.PreElement.IsEmpty);
+            Assert.True(output.PreContent.IsEmpty);
+            Assert.True(output.Content.IsEmpty);
+            Assert.True(output.PostContent.IsEmpty);
+            Assert.True(output.PostElement.IsEmpty);
+            Assert.Empty(output.Attributes);
```
dougbu: Odd. How does this assertion succeed?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: :+1:

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
dougbu: Thx

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: Oops!

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: I have no idea really. I'll ask whatshisface when he shows up

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
rynowak: Added assertion to this test

rynowak:
:updated: - need to discuss https://github.com/aspnet/Razor/pull/694#discussion_r54317974

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: The answer from @NTaylorMullen is that this will generally be an `HtmlEncodedString` or string + needs encoding.

The user could do something weird here like put a view buffer inside of an attribute. I think we're we're doing here is fine, we'd be suboptimal, but not incorrect if you did that.

rynowak:
Updated again - and fixed test

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
rynowak: It does now that I fixed the implementation

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: thx

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: :ok_hand:

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
dougbu: thx

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
dougbu: :smiley_cat:

dougbu:
:shipit:

rynowak:
will check in once MVC part is signed off.

Issue #674, opened 2016-01-23
Pool `TagHelperExecutionContexts`


NTaylorMullen:
Today our `TagHelperExecutionContext`s aren't pooled and end up resulting in ~2.4% of the overall allocations of a page.

![image](https://cloud.githubusercontent.com/assets/2008729/12526982/efc59f20-c128-11e5-9993-e767c45da708.png)

Eilon:
Moving to Backlog.

NTaylorMullen:
Doing this real quick. It's a 3% overall perf gain and relatively simple.

NTaylorMullen:
78451efa7fb5490a7e92ef31576c6bd0e57b30cf

PR #695, opened 2016-02-18
Pool `TagHelperExecutionContext`s within `TagHelperScopeManager`.

NTaylorMullen:
- Currently the `TagHelperScopeManager` creates a new `TagHelperExecutionContext` per `TagHelper` on a given page. With this change the max number of `TagHelperExecutionContext`s per page is the number of nested levels that exist.
- Decided to add two tests validating that specific pieces of `TagHelperExecutionContext` are updated as expected.

**Before:**
image

**After:**
image

**Result:** 3% overall reduction in allocations.

#674

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -14,14 +14,14 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperExecutionContext
     {
-        private readonly string _tagName;
-        private readonly string _uniqueId;
-        private readonly TagMode _tagMode;
         private readonly List<ITagHelper> _tagHelpers;
-        private readonly Func<Task> _executeChildContentAsync;
         private readonly Action<HtmlEncoder> _startTagHelperWritingScope;
         private readonly Func<TagHelperContent> _endTagHelperWritingScope;
         private TagHelperContent _childContent;
+        private string _tagName;
+        private string _uniqueId;
+        private TagMode _tagMode;
+        private Func<Task> _executeChildContentAsync;
         private Dictionary<HtmlEncoder, TagHelperContent> _perEncoderChildContent;
         private TagHelperAttributeList _htmlAttributes;
         private TagHelperAttributeList _allAttributes;
@@ -63,26 +63,6 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             Action<HtmlEncoder> startTagHelperWritingScope,
             Func<TagHelperContent> endTagHelperWritingScope)
         {
-            if (tagName == null)
-            {
-                throw new ArgumentNullException(nameof(tagName));
-            }
-
-            if (items == null)
-            {
-                throw new ArgumentNullException(nameof(items));
-            }
-
-            if (uniqueId == null)
-            {
-                throw new ArgumentNullException(nameof(uniqueId));
-            }
-
-            if (executeChildContentAsync == null)
-            {
-                throw new ArgumentNullException(nameof(executeChildContentAsync));
-            }
-
             if (startTagHelperWritingScope == null)
             {
                 throw new ArgumentNullException(nameof(startTagHelperWritingScope));
@@ -94,14 +74,11 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             }
 
             _tagHelpers = new List<ITagHelper>();
-            _executeChildContentAsync = executeChildContentAsync;
+
+            Reinitialize(tagName, tagMode, items, uniqueId, executeChildContentAsync);
+
             _startTagHelperWritingScope = startTagHelperWritingScope;
             _endTagHelperWritingScope = endTagHelperWritingScope;
-
-            _tagMode = tagMode;
-            _tagName = tagName;
-            Items = items;
-            _uniqueId = uniqueId;
         }
 
         /// <summary>
@@ -118,7 +95,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         /// <summary>
         /// Gets the collection of items used to communicate with other <see cref="ITagHelper"/>s.
         /// </summary>
-        public IDictionary<object, object> Items { get; }
+        public IDictionary<object, object> Items { get; private set; }
 
         /// <summary>
         /// <see cref="ITagHelper"/>s that should be run.
@@ -212,6 +189,53 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             EnsureAllAttributes();
 
             _allAttributes.Add(name, value);
+        }
+
+        /// <summary>
+        /// Clears the <see cref="TagHelperExecutionContext"/> and updates its state with the provided values.
+        /// </summary>
+        /// <param name="tagName">The tag name to use.</param>
+        /// <param name="tagMode">The <see cref="TagMode"/> to use.</param>
+        /// <param name="items">The <see cref="IDictionary{Object, Object}"/> to use.</param>
+        /// <param name="uniqueId">The unique id to use.</param>
+        /// <param name="executeChildContentAsync">The <see cref="Func{Task}"/> to use.</param>
+        public void Reinitialize(
+            string tagName,
+            TagMode tagMode,
+            IDictionary<object, object> items,
+            string uniqueId,
+            Func<Task> executeChildContentAsync)
+        {
+            if (tagName == null)
+            {
+                throw new ArgumentNullException(nameof(tagName));
+            }
+
+            if (items == null)
+            {
+                throw new ArgumentNullException(nameof(items));
+            }
+
+            if (uniqueId == null)
+            {
+                throw new ArgumentNullException(nameof(uniqueId));
+            }
+
+            if (executeChildContentAsync == null)
+            {
+                throw new ArgumentNullException(nameof(executeChildContentAsync));
+            }
+
+            _tagName = tagName;
+            _tagMode = tagMode;
+            Items = items;
+            _uniqueId = uniqueId;
+            _executeChildContentAsync = executeChildContentAsync;
+            _tagHelpers.Clear();
+            _perEncoderChildContent?.Clear();
+            _htmlAttributes = null;
+            _allAttributes = null;
+            _childContent = null;
         }
 
         // Internal for testing.
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
@@ -15,14 +15,14 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperScopeManager
     {
-        private readonly Stack<TagHelperExecutionContext> _executionScopes;
+        private readonly ExecutionContextPool _executionContextPool;
 
         /// <summary>
         /// Instantiates a new <see cref="TagHelperScopeManager"/>.
         /// </summary>
         public TagHelperScopeManager()
         {
-            _executionScopes = new Stack<TagHelperExecutionContext>();
+            _executionContextPool = new ExecutionContextPool();
         }
 
         /// <summary>
@@ -72,12 +72,13 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             }
 
             IDictionary<object, object> items;
+            var parentExecutionContext = _executionContextPool.Current;
 
             // If we're not wrapped by another TagHelper, then there will not be a parentExecutionContext.
-            if (_executionScopes.Count > 0)
+            if (parentExecutionContext != null)
             {
                 items = new CopyOnWriteDictionary<object, object>(
-                    _executionScopes.Peek().Items,
+                    parentExecutionContext.Items,
                     comparer: EqualityComparer<object>.Default);
             }
             else
@@ -85,7 +86,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 items = new Dictionary<object, object>();
             }
 
-            var executionContext = new TagHelperExecutionContext(
+            var executionContext = _executionContextPool.Rent(
                 tagName,
                 tagMode,
                 items,
@@ -93,8 +94,6 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 executeChildContentAsync,
                 startTagHelperWritingScope,
                 endTagHelperWritingScope);
-
-            _executionScopes.Push(executionContext);
 
             return executionContext;
         }
@@ -106,7 +105,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         /// <c>null</c> otherwise.</returns>
         public TagHelperExecutionContext End()
         {
-            if (_executionScopes.Count == 0)
+            if (_executionContextPool.Current == null)
             {
                 throw new InvalidOperationException(
                     Resources.FormatScopeManager_EndCannotBeCalledWithoutACallToBegin(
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -214,6 +214,46 @@ public void AddTagHelperAttribute(string name, object value)
             _allAttributes.Add(name, value);
         }
 
+        // Used to pool TagHelperExecutionContext's. With this we don't always need to recreate TagHelperExecutionContexts.
+        internal void ClearAndUpdateWith(
```
rynowak: `Reset(...)`?

rynowak: Why internal? Would user code ever see this type?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -124,5 +127,63 @@ public TagHelperExecutionContext End()
...
+
+        private class ExecutionContextPool
+        {
+            private readonly List<TagHelperExecutionContext> _unusedExecutionContexts;
+            private int _lastIndex;
+
+            public ExecutionContextPool()
+            {
+                _lastIndex = -1;
+                _unusedExecutionContexts = new List<TagHelperExecutionContext>();
```
rynowak: If you're using this like a stack isn't `_lastIndex` always equal to `Count - 1`

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -214,6 +214,46 @@ public void AddTagHelperAttribute(string name, object value)
...
+            _tagName = tagName;
+            _tagMode = tagMode;
+            Items = items;
+            _uniqueId = uniqueId;
+            _executeChildContentAsync = executeChildContentAsync;
+            _tagHelpers.Clear();
+            _perEncoderChildContent?.Clear();
+            _htmlAttributes = null;
+            _allAttributes = null;
+            _childContent = null;
```
rynowak: The next part will be to see what we can do about these things

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -14,14 +14,14 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperExecutionContext
```
rynowak: I have to admit, I was expecting to see a codegen based solution - why is this better?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Nope, user code wont see this type. As for `Reset`, that was my first naming of this but felt that the name mislead with what it really did.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Codegen would be doing roughly the same thing and would only save a single list allocation. Would have been much more complicated for very little gain.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Suggest `Reinitialize()`. Also suggest `public` because `internal` is used only for bits that would be `private` if not for testing. (All of `Microsoft.AspNetCore.Razor.Runtime.TagHelpers` could be named `.Internal` if we wanted.)

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -214,6 +214,53 @@ public void AddTagHelperAttribute(string name, object value)
...
 
+        /// <summary>
+        /// Clears the <see cref="TagHelperExecutionContext"/> and updates its state with the provided values.
+        /// </summary>
+        /// <param name="tagName">The tag name to use.</param>
+        /// <param name="tagMode">The <see cref="TagMode"/> to use.</param>
+        /// <param name="items">The <see cref="IDictionary{Object, Object}"/> to use.</param>
+        /// <param name="uniqueId">The unique id to use.</param>
+        /// <param name="executeChildContentAsync">The <see cref="Func{Task}"/> to use.</param>
+        public void Reinitialize(
```
dougbu: Suggest removing overlapping parameters from the constructors, requiring a `Reinitialize()` call for all uses. Alternatively call `Reinitialize()` from the constructors duplicating this code.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -214,6 +214,53 @@ public void AddTagHelperAttribute(string name, object value)
...
+                throw new ArgumentNullException(nameof(executeChildContentAsync));
+            }
+
+            _tagName = tagName;
+            _tagMode = tagMode;
+            Items = items;
+            _uniqueId = uniqueId;
+            _executeChildContentAsync = executeChildContentAsync;
+            _tagHelpers.Clear();
+            _perEncoderChildContent?.Clear();
```
dougbu: :+1: like reusing this dictionary.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -14,14 +14,14 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperExecutionContext
     {
-        private readonly string _tagName;
-        private readonly string _uniqueId;
-        private readonly TagMode _tagMode;
         private readonly List<ITagHelper> _tagHelpers;
-        private readonly Func<Task> _executeChildContentAsync;
         private readonly Action<HtmlEncoder> _startTagHelperWritingScope;
         private readonly Func<TagHelperContent> _endTagHelperWritingScope;
```
dougbu: Too much understanding of how MVC uses the scope manager here. The start and end bits (`Action<HtmlEncoder>` and `Func<TagHelperContent>`) are passed into the `Begin()` method and should not be treated as invariant here. Update `TagHelperScopeManager` and code gen to pass these delegates into the `TagHelperScopeManager` constructor. Then passing them into this class' constructor obviously be fine.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
     /// </summary>
     public class TagHelperScopeManager
     {
+        private readonly ExecutionContextPool _executionContextPool;
         private readonly Stack<TagHelperExecutionContext> _executionScopes;
```
dougbu: Suggest removing `_executionScopes`. The `ExecutionContextPool` contains _more_ information than the stack.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -124,5 +127,56 @@ public TagHelperExecutionContext End()
...
+        {
+            private readonly List<TagHelperExecutionContext> _executionContexts;
+            private int _nextIndex;
+
+            public ExecutionContextPool()
+            {
+                _executionContexts = new List<TagHelperExecutionContext>();
+            }
+
+            public TagHelperExecutionContext Create(
```
dougbu: Leave `Create` for methods that always create something. This is more like `Rent()` (which lines up with some other pools) or `GetContext()` (which lines up with a few other pools / buffers).

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -124,5 +127,56 @@ public TagHelperExecutionContext End()
...
+                    tagHelperExecutionContext = _executionContexts[_nextIndex];
+                    tagHelperExecutionContext.Reinitialize(tagName, tagMode, items, uniqueId, executeChildContentAsync);
+                }
+
+                _nextIndex++;
+
+                return tagHelperExecutionContext;
+            }
+
+            public void Return(TagHelperExecutionContext context)
```
NTaylorMullen: Remove param

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
NTaylorMullen: Eh, it'd just be moving the logic. I'm fine with that though.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: So add these two bits to the `Reinitialize` method?

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -13,6 +14,110 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
...
+            executionContext.AddMinimizedHtmlAttribute("Another attribute");
+
+            // Assert - 1
+            var output = executionContext.CreateTagHelperOutput();
+            Assert.Equal(updatedTagName, output.TagName);
+            Assert.Equal(updatedTagMode, output.TagMode);
+            var attribute = Assert.Single(output.Attributes);
+            Assert.Equal("Another attribute", attribute.Name);
+            Assert.Null(attribute.Value);
+            Assert.True(attribute.Minimized);
```
dougbu: Suggest these last two asserts are just noise.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -13,6 +14,110 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
...
+                updatedCallCount++;
+                return Task.FromResult(true);
+            };
+            executionContext.AddMinimizedHtmlAttribute("something");
+
+            // Act - 1
+            executionContext.Reinitialize(
+                updatedTagName,
+                updatedTagMode,
+                items: new Dictionary<object, object>(),
```
dougbu: Please something in this dictionary and confirm it exists in `// Assert - 1`.

dougbu: Never mind this one. Done in the other test.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -13,6 +14,110 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
...
+                updatedUniqueId,
+                executeChildContentAsync);
+            executionContext.AddMinimizedHtmlAttribute("Another attribute");
+
+            // Assert
+            var context = executionContext.CreateTagHelperContext();
+            var attribute = Assert.Single(context.AllAttributes);
+            Assert.Equal(attribute.Name, "Another attribute");
+            Assert.Null(attribute.Value);
+            Assert.True(attribute.Minimized);
```
dougbu: Don't bother w/ checking the rest of the `attribute`.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Has always been strange that these were passed into `TagHelperScopeManager.Begin()`. MVC doesn't have a reason to change the values from call to call. (Don't recall if it ever passed anything but a method name.) So, while we could pile more into `Reinitialize()`, better to give the values a sensible lifetime.

Bottom line, change the `TagHelperScopeManager` constructor or file a follow-up issue to do so.

dougbu:
:watch: (looking very solid)

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/696

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -93,15 +73,11 @@ internal TagHelperExecutionContext(string tagName, TagMode tagMode)
                 throw new ArgumentNullException(nameof(endTagHelperWritingScope));
             }
 
+            Reinitialize(tagName, tagMode, items, uniqueId, executeChildContentAsync);
```
dougbu: Call this after initializing `_tagHelpers` and just `.Clear()` that field in `Reinitialize()`.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
```diff
@@ -115,14 +114,61 @@ public TagHelperExecutionContext End()
...
+                        uniqueId,
+                        executeChildContentAsync,
+                        startTagHelperWritingScope,
+                        endTagHelperWritingScope);
+
+                    _executionContexts.Add(tagHelperExecutionContext);
+                }
+                else
+                {
+                    tagHelperExecutionContext = _executionContexts[_nextIndex];
```
dougbu: `Debug.Assert(_nextIndex >= 0)` to give this class some control over its own destiny. Really just double-checking `TagHelperScopeManager.End()` `throw`s when it should.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Thx

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Thx

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperScopeManager.cs
NTaylorMullen: Since `TagHelperScopeManager.End` does the throw (with a test that verifies it throws) I think it'd be a bit much to also have a `Debug.Assert` here

dougbu:
Not urgent to check but am curious whether `ExecutionContextPool` instances are smaller than the previous `Stack<TagHelperExecutionContext>` instances.

dougbu:
:shipit: w/ a couple of small fixes

Issue #691, opened 2016-02-13
Generates wrong place quotation mark of attribute in template


riwofeb:
the following code demonstrate it:
ViewComponents\SimpleViewComponent.cs
```c#
...
    public class SimpleViewComponent : ViewComponent
    {
        public IViewComponentResult Invoke(SimpleData sd)
        {
            return View(sd);
        }
    }
    public class SimpleData
    {
        public string Title { get; set; }
        public int Count { get; set; }
        public Func<dynamic, object> ContentHtml { get; set; }
    }
...
```
Views\Shared\Components\Simple\Default.cshtml
```c#
...
@model SimpleData
...
<h4>@Model.Title</h4>
<ul>
    @for (int i = 0; i < Model.Count; i++)
    {
        <li>@Model.ContentHtml($"Item #{i + 1}")</li>
    }
</ul>
...
```
Views\Home\SimplePage.cshtml
```c#
...
@{
    var extaClass = "text-success";
    @* My template *@
    Func<dynamic, object> content =
        @<text>
        <p class="text-capitalize @extaClass" style="font-size: 110%; font-weight: 600">
            @item
            <span style="font-size: 80%; color: #bbb">[Hint of title.]</span>
        </p>
        <p>...</p>
        </text>;
}
...
<hr />
@(Component.Invoke<SimpleViewComponent>(new SimpleData() { Title = "View Component title.", Count = 3, ContentHtml = content }))
<hr />
@content("This separate item!")
...
```
and when SimplePage.cshtml rendered you will see the following wrong html:
```html
...
<hr />
"""
    <h4>View Component title.</h4>
    <ul>
            <li>
                
        <p class="text-capitalize text-success style="font-size: 110%; font-weight: 600">
            Item #1
            <span style="font-size: 80%; color: #bbb">[Hint of title.]</span>
        </p>
        <p>...</p>
        
            </li>
            <li>
...
<hr />
        <p class="text-capitalize text-success" style="font-size: 110%; font-weight: 600">
            This separate item!
            <span style="font-size: 80%; color: #bbb">[Hint of title.]</span>
        </p>
        <p>...</p>
...
```
The `@(Component.Invoke<...` not rendered end quotation mark of _class_ attribute (where `text-success`), ie generator move it to wrong place  after `<hr />`.
But `@content("This separate item!")` rendered success, ie quotation mark in its place.

Eilon:
cc @NTaylorMullen in case you have thoughts on this too.

[...]

PR #699, opened 2016-02-22
Use WriteAttributeTo when TextWriter is not null

ajaybhargavb:
Issue - #691 

@NTaylorMullen 

Not sure what kind of test to write for this bug. Any suggestions?


NTaylorMullen:
If all the tests pass with this change you should add (or update) a test where there's a dynamic attribute in a section. Aka:

```
@section Foo {
    <input class="btn @extraClass" />
}
```

ajaybhargavb:
:up: :date:

NTaylorMullen:
:shipit:

Issue #698, opened 2016-02-20
Unquoted TagHelper attribute values with `=` in their value cause Debug.Assert failure.


NTaylorMullen:
If we have a `TagHelper`.

```HTML
<myth name=something=value key="@value" />
```

You get a debug assert [here](https://github.com/aspnet/Razor/blob/5ca09d0b1020c0bef535d4121def9d3a48b463d0/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs#L208).

dougbu:
@NTaylorMullen your example is invalid. Is this about catching the HTML syntax violation earlier?

The [HTML 5 spec](https://www.w3.org/TR/html5/syntax.html#attributes-0) says unquoted attribute values:
> must not contain any literal space characters, any U+0022 QUOTATION MARK characters ("), U+0027 APOSTROPHE characters ('), "=" (U+003D) characters, "<" (U+003C) characters, ">" (U+003E) characters, or U+0060 GRAVE ACCENT characters (`), and must not be the empty string.

NTaylorMullen:
@dougbu It's definitely invalid, however it shouldn't crash the parser :smile:. Instead of `Debug.Assert`ing we should be logging an error to the `ErrorSink`.

dougbu:
Well, it won't crash the parser in release mode :smiling_imp:

Thanks for the clarification.

PR #701, opened 2016-02-26
Removed unnecessary Debug.Assert in TagHelperBlockRewriter

ajaybhargavb:
Issue - #698 

@NTaylorMullen


src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -105,6 +105,11 @@ public static class TagHelperBlockRewriter
                     attributes.Add(
                         new KeyValuePair<string, SyntaxTreeNode>(result.AttributeName, result.AttributeValueNode));
                 }
+                else
+                {
+                    // Error occured while parsing the attribute. Don't try parsing the rest.
```
NTaylorMullen: `Don't try parsing the rest to avoid misleading errors.`

NTaylorMullen:
:shipit:

Issue #600, opened 2015-11-12
Consider pre-allocating TagHelperAttributes


rynowak:
Where these are read-only/const/immutable we should try to optimize by putting these into readonly static fields. The attributes and associated collections are a significant enough amount of allocations that we should do something about it.

![image](https://cloud.githubusercontent.com/assets/1430011/11107310/9bc15f88-8891-11e5-8b86-f9f0897ed6ef.png)

NTaylorMullen:
9f675abbcada6f64af954844155eafc6c7b9ede6

PR #703, opened 2016-03-01
Pre-allocate unbound `TagHelper` attributes.

NTaylorMullen:
- Hoist `TagHelperAttribute` creation into `private static readonly` fields to avoid allocations on every request. With the recent `TagHelperAttribute` change that made them immutable we can now pre-allocate them without worry of them being modified.
- Added two extra class configuration pieces to enable pre-allocation.
- Updated test files to showcase new pre-allocations.

**Before:**
image

**After:**
image

**Result:** 5.7% reduction in allocations overall.

#600

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -146,12 +146,8 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 throw new ArgumentNullException(nameof(name));
             }
 
-            EnsureHtmlAttributes();
-            EnsureAllAttributes();
-
             var attribute = new TagHelperAttribute(name);
-            _htmlAttributes.Add(attribute);
-            _allAttributes.Add(attribute);
+            AddHtmlAttribute(attribute);
         }
 
         /// <summary>
@@ -166,10 +162,24 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 throw new ArgumentNullException(nameof(name));
             }
 
+            var attribute = new TagHelperAttribute(name, value);
+            AddHtmlAttribute(attribute);
+        }
+
+        /// <summary>
+        /// Tracks the HTML attribute.
+        /// </summary>
+        /// <param name="attribute">The <see cref="TagHelperAttribute"/> to track.</param>
+        public void AddHtmlAttribute(TagHelperAttribute attribute)
+        {
+            if (attribute == null)
+            {
+                throw new ArgumentNullException(nameof(attribute));
+            }
+
             EnsureHtmlAttributes();
             EnsureAllAttributes();
 
-            var attribute = new TagHelperAttribute(name, value);
             _htmlAttributes.Add(attribute);
             _allAttributes.Add(attribute);
         }
--- a/src/Microsoft.AspNetCore.Razor/Chunks/PreallocatedTagHelperAttributeChunk.cs
+++ b/src/Microsoft.AspNetCore.Razor/Chunks/PreallocatedTagHelperAttributeChunk.cs
@@ -1,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Chunks
+{
+    /// <summary>
+    /// A <see cref="Chunk"/> that represents a pre-allocated tag helper attribute.
+    /// </summary>
+    public class PreallocatedTagHelperAttributeChunk : Chunk
+    {
+        /// <summary>
+        /// The variable holding the pre-allocated attribute.
+        /// </summary>
+        public string AttributeVariableAccessor { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
@@ -98,6 +98,34 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             }
         }
 
+        public static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
+        {
+            var parentChunk = chunk as ParentChunk;
+
+            plainText = null;
+
+            if (parentChunk == null || parentChunk.Children.Count != 1)
+            {
+                return false;
+            }
+
+            LiteralChunk literalChildChunk;
+            if ((literalChildChunk = parentChunk.Children[0] as LiteralChunk) != null)
+            {
+                plainText = literalChildChunk.Text;
+                return true;
+            }
+
+            ParentLiteralChunk parentLiteralChunk;
+            if ((parentLiteralChunk = parentChunk.Children[0] as ParentLiteralChunk) != null)
+            {
+                plainText = parentLiteralChunk.GetText();
+                return true;
+            }
+
+            return false;
+        }
+
         internal static string GetVariableName(TagHelperDescriptor descriptor)
         {
             return "__" + descriptor.TypeName.Replace('.', '_');
@@ -450,6 +478,15 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
                     .WriteStringLiteral(attributeName)
                     .WriteEndMethodInvocation();
             }
+            else if (attributeValueChunk is PreallocatedTagHelperAttributeChunk)
+            {
+                _writer
+                    .WriteStartInstanceMethodInvocation(
+                        ExecutionContextVariableName,
+                        _tagHelperContext.ExecutionContextAddHtmlAttributeMethodName)
+                    .Write(((PreallocatedTagHelperAttributeChunk)attributeValueChunk).AttributeVariableAccessor)
+                    .WriteEndMethodInvocation();
+            }
             else
             {
                 string textValue = null;
@@ -737,34 +774,6 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             if (parentChunk != null)
             {
                 return parentChunk.Children.Any(child => child is DynamicCodeAttributeChunk);
-            }
-
-            return false;
-        }
-
-        private static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
-        {
-            var parentChunk = chunk as ParentChunk;
-
-            plainText = null;
-
-            if (parentChunk == null || parentChunk.Children.Count != 1)
-            {
-                return false;
-            }
-
-            LiteralChunk literalChildChunk;
-            if ((literalChildChunk = parentChunk.Children[0] as LiteralChunk) != null)
-            {
-                plainText = literalChildChunk.Text;
-                return true;
-            }
-
-            ParentLiteralChunk parentLiteralChunk;
-            if ((parentLiteralChunk = parentChunk.Children[0] as ParentLiteralChunk) != null)
-            {
-                plainText = parentLiteralChunk.GetText();
-                return true;
             }
 
             return false;
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
@@ -38,6 +38,8 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             TagHelperOutputIsContentModifiedPropertyName = "IsContentModified";
             TagHelperOutputContentPropertyName = "Content";
             TagHelperOutputGetChildContentAsyncMethodName = "GetChildContentAsync";
+            TagHelperAttributeTypeName = "Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute";
+            EncodedHtmlStringTypeName = "Microsoft.AspNetCore.Html.HtmlEncodedString";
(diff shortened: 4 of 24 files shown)


rynowak:
>showcase 

I knew this was your PR before I even saw your gravatar

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -38,6 +38,8 @@ public GeneratedTagHelperContext()
             TagHelperOutputIsContentModifiedPropertyName = "IsContentModified";
             TagHelperOutputContentPropertyName = "Content";
             TagHelperOutputGetChildContentAsyncMethodName = "GetChildContentAsync";
+            TagHelperAttributeTypeName = "Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute";
+            HtmlStringTypeName = "Microsoft.AspNetCore.Mvc.Rendering.HtmlString";
```
pranavkm: Could we use `Microsoft.AspNetCore.Html.Abstractions.HtmlEncodedString` here?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -146,12 +146,8 @@ public void AddMinimizedHtmlAttribute(string name)
                 throw new ArgumentNullException(nameof(name));
             }
 
-            EnsureHtmlAttributes();
-            EnsureAllAttributes();
-
             var attribute = new TagHelperAttribute(name);
-            _htmlAttributes.Add(attribute);
-            _allAttributes.Add(attribute);
+            AddHtmlAttribute(attribute);
```
rynowak: These could also be cached...

src/Microsoft.AspNetCore.Razor/Chunks/PreallocatedTagHelperAttributeChunk.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Chunks
+{
+    /// <summary>
+    /// A <see cref="Chunk"/> that represents a pre-allocated tag helper attribute.
```
rynowak: whoa! correct usage of represents!

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -742,7 +751,7 @@ private static bool IsDynamicAttributeValue(Chunk attributeValueChunk)
             return false;
         }
 
-        private static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
+        internal static bool TryGetPlainTextValue(Chunk chunk, out string plainText)
```
rynowak: Should this just be public?

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
rynowak: This is a little problematic isn't it?

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +91,87 @@ protected override void Visit(TagHelperChunk chunk)
...
+                                .Write(" = ")
+                                .WriteStartNewObject(_tagHelperContext.TagHelperAttributeTypeName)
+                                .WriteStringLiteral(attribute.Key)
+                                .WriteEndMethodInvocation();
+                        }
+                    }
+                    else
+                    {
+                        string plainText;
+                        var isPlainText = CSharpTagHelperCodeRenderer.TryGetPlainTextValue(attribute.Value, out plainText);
```
rynowak: Ugh, let's not use internal for this. :hankey:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +91,87 @@ protected override void Visit(TagHelperChunk chunk)
...
+                // If there's no descriptors associated or we're hitting a bound attribute a second time.
+                if (!associatedDescriptors.Any() || !boundAttributes.Add(attribute.Key))
+                {
+                    string preAllocatedAttributeVariableName = null;
+
+                    if (attribute.Value == null)
+                    {
+                        // For minimized attributes we add the null character to the end of the key to differentiate
+                        // between name and name=""; null characters are invalid in attribute values.
+                        var preAllocatedAttributeKey = attribute.Key + '\0';
```
rynowak: Alternative here is to add a struct with two strings.

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
rynowak: This is the first case I've seen where it would actually behoove us to just have one type.

rynowak:
:watch: - left a few minor comments, overall design seems good

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +91,87 @@ protected override void Visit(TagHelperChunk chunk)
...
             Accept(chunk.Children);
         }
 
+        private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
+        {
+            var boundAttributes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            for (var i = 0; i < chunk.Attributes.Count; i++)
+            {
+                var attribute = chunk.Attributes[i];
+                var associatedDescriptors = chunk.Descriptors.Where(descriptor =>
```
pranavkm: Doesn't look like this actually needs the descriptors. You could do:
`var hasAssociatedDescriptors = chunk.Descriptors.Any(..);` and use that in the if later on.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
pranavkm: Could we create a custom cache key type instead of doing this?

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +91,87 @@ protected override void Visit(TagHelperChunk chunk)
...
+                                .WriteStringLiteral(attribute.Key)
+                                .WriteEndMethodInvocation();
+                        }
+                    }
+                    else
+                    {
+                        string plainText;
+                        var isPlainText = CSharpTagHelperCodeRenderer.TryGetPlainTextValue(attribute.Value, out plainText);
+
+                        if (isPlainText)
```
pranavkm: `if (CSharpTagHelperCodeRenderer.TryGetPlainTextValue(attribute.Value, out plainText))`

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +91,87 @@ protected override void Visit(TagHelperChunk chunk)
...
+                        {
+                            var preAllocatedAttributeKey = attribute.Key + plainText;
```
pranavkm: Is there a risk of key collisions here? (attribute.Key + plainText == attribute.Key for a bound attribute?)

pranavkm:
:watch:

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
NTaylorMullen: Ehh, we could but it'd be inconsistent with how non-complex unbound attributes are handled. Those values are handled via `Html.Raw` which wraps the value in an `HtmlString`.If we were to swap both this and the default impl of `Html.Raw` then it'd all be good.

dougbu: What difference does `Html.Raw()` make? `HtmlString` is a subclass of `HtmlEncodedString` that exists only for back-compat.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Hmm we could cache at runtime. Only downside is creating the dictionary to hold the cached objects; however, it would be a much simpler implementation (than doing the code gen approach).

Are you suggesting doing the cache at runtime or having an extra layer of caching to catch dynamicish bits?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Could, doesn't have a whole lot of value in being public though.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
NTaylorMullen: Could just compute a hash code and have the key be int.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
NTaylorMullen: :+1:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
NTaylorMullen: Hmm, ya :+1:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: The choices are `private`, the occasional `protected`, `public` or the rare `// internal for testing`. Pick from that list -- `public` here.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Talked off line. @rynowak just be spitting truths, no suggestion here.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -38,6 +38,8 @@ public GeneratedTagHelperContext()
             TagHelperOutputIsContentModifiedPropertyName = "IsContentModified";
             TagHelperOutputContentPropertyName = "Content";
             TagHelperOutputGetChildContentAsyncMethodName = "GetChildContentAsync";
+            TagHelperAttributeTypeName = "Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute";
+            EncodedHtmlStringTypeName = "Microsoft.AspNetCore.Html.Abstractions.HtmlEncodedString";
```
rynowak: Type name is actually `Microsoft.AspNetCore.Html.HtmlEncodedString`

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -86,10 +92,86 @@ protected override void Visit(TagHelperChunk chunk)
...
+                    descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key)));
+
+                // If there's no descriptors associated or we're hitting a bound attribute a second time.
+                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Key))
+                {
+                    string preAllocatedAttributeVariableName = null;
+
+                    if (attribute.Value == null)
+                    {
+                        var preAllocatedAttributeKey = StringComparer.Ordinal.GetHashCode(attribute.Key);
```
rynowak: This won't work YO #Truf

NTaylorMullen:
:up: :date:

rynowak:
:shipit:

Issue #679, opened 2016-01-27
if block inside foreach block doesn't parse ending brackets correctly


esargent:
I think this is different from https://github.com/aspnet/Razor/issues/51 but adding the link in case there is a relation.

Using RC2-16770

My view iterates through a collection of items (Practitioners) and displays some things. That works great **UNTIL** I add an if block inside the foreach. Then razor finds the closing bracket for the foreach, but associates it to the parent bracket enclosing the overall razor code block. This happens both in VS2015 code highlighting and when the code is executed. Also note, I can add an if statement with a single line - it is when I add the brackets to designate a code block that the issue occurs.

Works:

     @model DistributorPortalViewModel
     <div class="">
     <h1>Simple Customer List</h1>
     <div>Click a name to expand for details.</div>
     <hr />
    @{
        string target, targetId, currentClinic = "";
        foreach (Practitioner p in Model.Practitioners)
        {
            targetId = "collapseArea" + @p.Id;
            target = "#" + targetId;
            <div class="m5">
                <a data-toggle="collapse" href=@target>@Html.DisplayFor(modelItem => p.LastName)</a>
                <div id=@targetId class="collapse">
                    @Html.DisplayFor(modelItem => p.Name) at @Html.DisplayFor(modelItem => p.PrimaryClinic.Name)<br />
                    Is Listed in Directory? @Html.DisplayTextFor(modelItem => p.IsListed)
                </div>
            </div>
    }
    }
    <hr />
</div>

Fails:

    @model DistributorPortalViewModel
    <div class="">
    <h1>Simple Customer List</h1>
    <div>Click a name to expand for details.</div>
    <hr />
    @{
        string target, targetId, currentClinic = "";
        foreach (Practitioner p in Model.Practitioners)
        {
            targetId = "collapseArea" + @p.Id;
            target = "#" + targetId;
            if (currentClinic != p.PrimaryClinic.Name)
            {
                currentClinic = p.PrimaryClinic.Name;
                <h2>@p.PrimaryClinic.Name</h2>
                }
            <div class="m5">
                <a data-toggle="collapse" href=@target>@Html.DisplayFor(modelItem => p.LastName)</a>
                <div id=@targetId class="collapse">
                    @Html.DisplayFor(modelItem => p.Name) at @Html.DisplayFor(modelItem => p.PrimaryClinic.Name)<br />
                    Is Listed in Directory? @Html.DisplayTextFor(modelItem => p.IsListed)
                </div>
            </div>
    }
[...]

PR #707, opened 2016-03-03
Fix nested C# blocks when combined with C# @ symbols.

NTaylorMullen:
- We used to accept until markup or an ending brace which didn't allow the parser to balance nested braces.

#679


ajaybhargavb:
:shipit:

PR #713, opened 2016-03-15
Fix package metadata

Eilon:
@NTaylorMullen @dougbu @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/project.json
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/project.json
@@ -1,5 +1,10 @@
 {
-  "description": "Runtime components for rendering Razor pages.",
+  "description": "Runtime components for rendering Razor pages and implementing tag helpers.\r\nCommonly used types:\r\nMicrosoft.AspNetCore.Razor.TagHelpers.HtmlAttributeNameAttribute\r\nMicrosoft.AspNetCore.Razor.TagHelpers.HtmlTargetElementAttribute\r\nMicrosoft.AspNetCore.Razor.TagHelpers.ITagHelper\r\nMicrosoft.AspNetCore.Razor.TagHelpers.TagHelper",
+  "tags": [
+    "aspnetcore",
+    "cshtml",
+    "razor"
+  ],
   "version": "1.0.0-*",
   "compilationOptions": {
     "allowUnsafe": true,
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor/CSharpRazorCodeLanguage.cs
```diff
@@ -4,9 +4,6 @@
 using Microsoft.AspNetCore.Razor.Chunks.Generators;
 using Microsoft.AspNetCore.Razor.CodeGenerators;
 using Microsoft.AspNetCore.Razor.Parser;
-#if NET451
-
```
Eilon: Lots & found.

src/Microsoft.AspNetCore.Razor.Runtime/project.json
```diff
@@ -1,5 +1,10 @@
 {
-  "description": "Runtime components for rendering Razor pages.",
+  "description": "Runtime components for rendering Razor pages and implementing tag helpers.\r\nCommonly used types:\r\nMicrosoft.AspNetCore.Razor.TagHelpers.HtmlAttributeNameAttribute\r\nMicrosoft.AspNetCore.Razor.TagHelpers.HtmlTargetElementAttribute\r\nMicrosoft.AspNetCore.Razor.TagHelpers.ITagHelper\r\nMicrosoft.AspNetCore.Razor.TagHelpers.TagHelper",
```
dougbu: Side question: Do these Windows line endings work when the package is built on Linux / OSx?

dougbu: Separately, I'd leave `Microsoft.AspNetCore.Razor.TagHelpers.ITagHelper` out since it's used much less often than `TagHelper`.

src/Microsoft.AspNetCore.Razor.Runtime/project.json
```diff
@@ -1,5 +1,10 @@
...
+  "tags": [
+    "aspnetcore",
+    "cshtml",
+    "razor"
```
dougbu: Add `tag helpers`?

dougbu:
:shipit:

src/Microsoft.AspNetCore.Razor.Runtime/project.json
Eilon: Windows line endings don't seem to generally cause much issue on non-Windows. Plus, the main place it's ever seen is in NuGet UI, which is at least at the moment all Windows-only (or web based).

Eilon: Will remove ITagHelper.

Eilon:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/project.json
Eilon: Done

Eilon:
Merged https://github.com/aspnet/Razor/commit/0f6fb8277c954347be7c1ab031f75a7d1585ee4c. Thanks!

ajaybhargavb:
@Eilon, Don't these also need `title`?

Eilon:
The `title` by default ends up being the package id, which is I think what we normally do. You only need to specify `title` if it's different from the id.

Issue #709, opened 2016-03-08
Cannot resolve TagHelperDesignTimeDescriptors in CoreCLR


NTaylorMullen:
The impact of this bug is not having `<summary>` IntelliSense in the Razor editor.

Problem: `RazorTooling` currently only compiles for CoreCLR/netstandard and Razor `ifdefs` the design time descriptor code to only run in net45 (how it used to run).

We need to find a way to locate XML locations for assemblies in CoreCLR to make this possible.

Eilon:
Sorry, didn't feel like creating a new `taskbughancement` label :smile:

ToddGrun:
Taylor, will this also fix the missing OutputElementHint info?

NTaylorMullen:
Yes this will also fix that since it exists on the design time descriptor.

NTaylorMullen:
3d0bf621f330ad5444a5892505dc4792aa14ad03

PR #714, opened 2016-03-16
Update `TagHelperDesignTimeDescriptorFactory` to work in CoreCLR.

NTaylorMullen:
- After discussion offline to maintain consistency we decided to also remove `CodeBase` fallback in the net451 scenario. This will keep CoreCLR and net451 scenarios more consistent.
- Removed `CodeBase` validation tests.
- Added `GetAssemblyLocation` extensibility point to enable proper testing of the `TagHelperDesignTimeDescriptorFactory`.

#709

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if !NETSTANDARD1_3 // Cannot accurately resolve the location of the documentation XML file in coreclr.
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
@@ -40,10 +39,10 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             }
 
             var id = XmlDocumentationProvider.GetId(type);
-            var documentationDescriptor = CreateDocumentationDescriptor(type.Assembly, id);
+            var documentationDescriptor = CreateDocumentationDescriptor(type.GetTypeInfo().Assembly, id);
 
-            // Purposefully not using the TypeInfo.GetCustomAttributes method here to make it easier to mock the Type.
             var outputElementHintAttribute = type
+                .GetTypeInfo()
                 .GetCustomAttributes(inherit: false)
                 ?.OfType<OutputElementHintAttribute>()
                 .FirstOrDefault();
@@ -79,7 +78,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             }
 
             var id = XmlDocumentationProvider.GetId(propertyInfo);
-            var declaringAssembly = propertyInfo.DeclaringType.Assembly;
+            var declaringAssembly = propertyInfo.DeclaringType.GetTypeInfo().Assembly;
             var documentationDescriptor = CreateDocumentationDescriptor(declaringAssembly, id);
             if (documentationDescriptor != null)
             {
@@ -93,6 +92,18 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
             return null;
         }
 
+        /// <summary>
+        /// Retrieves <paramref name="assembly"/>'s location on disk.
+        /// </summary>
+        /// <param name="assembly">The assembly.</param>
+        /// <returns>The path to the given <paramref name="assembly"/>.</returns>
+        public virtual string GetAssemblyLocation(Assembly assembly)
+        {
+            var assemblyLocation = assembly.Location;
+
+            return assemblyLocation;
+        }
+
         private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
         {
             var hashCodeCombiner = HashCodeCombiner.Start();
@@ -102,15 +113,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
 
             var documentationProvider = _documentationProviderCache.GetOrAdd(cacheKey, valueFactory: _ =>
             {
-                var assemblyLocation = assembly.Location;
-
-                if (string.IsNullOrEmpty(assemblyLocation) && !string.IsNullOrEmpty(assembly.CodeBase))
-                {
-                    var uri = new UriBuilder(assembly.CodeBase);
-
-                    // Normalize the path to UNC path. This will remove things like file:// from start of uri.Path.
-                    assemblyLocation = Uri.UnescapeDataString(uri.Path);
-                }
+                var assemblyLocation = GetAssemblyLocation(assembly);
 
                 // Couldn't resolve a valid assemblyLocation.
                 if (string.IsNullOrEmpty(assemblyLocation))
@@ -235,4 +238,3 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         }
     }
 }
-#endif
--- a/src/Microsoft.AspNetCore.Razor.Runtime/project.json
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/project.json
@@ -56,9 +56,12 @@
     },
     "netstandard1.3": {
       "dependencies": {
+        "System.Collections.Concurrent": "4.0.12-*",
+        "System.Reflection": "4.1.0-*",
         "System.Reflection.Extensions": "4.0.1-*",
         "System.Reflection.TypeExtensions": "4.1.0-*",
-        "System.Text.RegularExpressions": "4.0.12-*"
+        "System.Text.RegularExpressions": "4.0.12-*",
+        "System.Xml.XDocument": "4.0.11-*"
       },
       "imports": [
         "dotnet5.4"
--- a/test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-#if !NETSTANDARDAPP1_5
 using System;
 using System.IO;
 using System.Reflection;
@@ -36,7 +35,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
 #if NET451
             Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), "..", "..", "..", ".."));
 #else
-            Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), "..", ".."));
+            Directory.GetCurrentDirectory();
 #endif
         public static readonly string DocumentedAssemblyLocation =
             Path.Combine(TestRoot, "TestFiles", "NotLocalized", "TagHelperDocumentation.dll");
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -40,10 +39,11 @@ public TagHelperDesignTimeDescriptor CreateDescriptor(Type type)
             }
 
             var id = XmlDocumentationProvider.GetId(type);
-            var documentationDescriptor = CreateDocumentationDescriptor(type.Assembly, id);
+            var documentationDescriptor = CreateDocumentationDescriptor(type.GetTypeInfo().Assembly, id);
 
             // Purposefully not using the TypeInfo.GetCustomAttributes method here to make it easier to mock the Type.
```
dougbu: Comment is no longer relevant or correct.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -93,6 +93,23 @@ public TagHelperAttributeDesignTimeDescriptor CreateAttributeDescriptor(Property
             return null;
         }
 
+        /// <summary>
+        /// Retrieves <paramref name="assembly"/>'s location on disk.
+        /// </summary>
+        /// <param name="assembly">The assembly.</param>
+        /// <returns>The path to the given <paramref name="assembly"/>.</returns>
+        public virtual string GetAssemblyLocation(Assembly assembly)
```
dougbu: Is this `virtual` for testing? Or is there a user scenario here?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -102,18 +119,10 @@ private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
 
             var documentationProvider = _documentationProviderCache.GetOrAdd(cacheKey, valueFactory: _ =>
             {
-                var assemblyLocation = assembly.Location;
-
-                if (string.IsNullOrEmpty(assemblyLocation) && !string.IsNullOrEmpty(assembly.CodeBase))
```
dougbu: Why doesn't `GetAssemblyLocation()` fallback to `assembly.CodeBase`?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
```diff
@@ -102,18 +119,10 @@ private XmlDocumentationProvider GetXmlDocumentationProvider(Assembly assembly)
...
-                    var uri = new UriBuilder(assembly.CodeBase);
-
-                    // Normalize the path to UNC path. This will remove things like file:// from start of uri.Path.
-                    assemblyLocation = Uri.UnescapeDataString(uri.Path);
-                }
+                var assemblyLocation = GetAssemblyLocation(assembly);
 
                 // Couldn't resolve a valid assemblyLocation.
-                if (string.IsNullOrEmpty(assemblyLocation))
+                if (assemblyLocation == null)
```
dougbu: Is `string.Empty` at all useful in some scenario? Suggest this line should remain as-is.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: virtual for testing. Decided to keep it public in the chance that external users need to test this as well. As it stands today it can't be unit tested for both net45/coreclr without this entry point.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: @Eilon and I discussed offline:

`After discussion offline to maintain consistency we decided to also remove CodeBase fallback in the net451 scenario. This will keep CoreCLR and net451 scenarios more consistent.`

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: `string.Empty` can no longer occur at this point. The diff is misleading, See line 105. The `GetAssemblyLocation` now does the logic that you're referring to.

src/Microsoft.AspNetCore.Razor.Runtime/project.json
```diff
@@ -56,9 +56,12 @@
     },
     "netstandard1.3": {
       "dependencies": {
+        "System.Collections.Concurrent": "4.0.12-*",
+        "System.Reflection": "4.1.0-*",
```
dougbu: Remove b/c `System.Reflection.Extensions` brings in `System.Reflection`.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
```diff
@@ -126,20 +117,9 @@ public static TheoryData CreateDescriptor_TypeDocumentationData
...
-                    },
-                    { CreateType<SingleAttributeTagHelper>(defaultLocation, defaultCodeBase), null },
-                    { CreateDocumentationTagHelperType(nonExistentLocation, codeBase: null), onlyHint },
-                    { CreateDocumentationTagHelperType(location: null, codeBase: nonExistentCodeBase), onlyHint },
-                    { CreateType<SingleAttributeTagHelper>(invalidLocation, codeBase: null), null },
-                    { CreateDocumentationTagHelperType(location: null, codeBase: invalidCodeBase), onlyHint },
+                    { typeof(SingleAttributeTagHelper), defaultLocation, null },
+                    { typeof(DocumentedTagHelper), nonExistentLocation, onlyHint },
+                    { typeof(SingleAttributeTagHelper), invalidLocation, null },
                 };
```
dougbu: Looks like the data set is a bit smaller. Why?

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
```diff
@@ -368,50 +337,35 @@ public static TheoryData CreateAttributeDescriptor_PropertyDocumentationData
...
-                    {
-                        CreateDocumentationTagHelperType(invalidLocation, codeBase: null),
-                        nameof(DocumentedTagHelper.RemarksAndSummaryProperty),
-                        null
-                    },
-                    {
-                        CreateDocumentationTagHelperType(location: null, codeBase: invalidCodeBase),
-                        nameof(DocumentedTagHelper.RemarksAndSummaryProperty),
-                        null
-                    }
```
dougbu: Guessing at the reason for the smaller data set above: Is this also about eliminating tests of invalid `CodeBase` values?

src/Microsoft.AspNetCore.Razor.Runtime/project.json
NTaylorMullen: Need `4.1.0` for `TypeInfo.Assembly`

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
dougbu: Doesn't change the fact that `assemblyLocation == string.Empty` is still invalid here. Remove the redundant check in `GetAssemblyLocation()` and then inline that one-line non-helper.

src/Microsoft.AspNetCore.Razor.Runtime/project.json
dougbu: The version of `System.Reflections.Extensions` we get is bringing in an older `System.Reflections`? Can we instead update the `System.Reflections.Extensions` version? If not, ick.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
dougbu: :ok:

dougbu:
:watch:

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
NTaylorMullen: Not testing `CodeBase` parts anymore.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactoryTest.cs
NTaylorMullen: Si

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDesignTimeDescriptorFactory.cs
NTaylorMullen: I see, :+1:

src/Microsoft.AspNetCore.Razor.Runtime/project.json
NTaylorMullen: Hmm, i'm not sure upgrading `System.Reflection.Extensions` to retrieve a new `System.Reflection` when we really don't need a higher extensions package is the right approach; however, if this is the recommended guidelines I'd be more than happy to do it.

/cc @Eilon is this something we'd typically do/recommend?

Eilon: Is there an actual problem with what's here? If not, leave it.

NTaylorMullen: Nope, no problem. Will leave it.

dougbu: Well, there is a problem: Pulls in an inconsistent set of CoreFx packages with a newer `System.Reflection` than `System.Reflection.Extensions`, `System.Reflection.TypeExtensions` and probably other packages were built against. Question is: Is this problem worth worrying about?

(Less importantly someone could come along and waste time cleaning out the transitive dependencies.)

Eilon: Talked to @davidfowl and this should be bumped up. There's no good way to know whether these packages are compatible. This is supposed to all be fixed for RTM, but for now we shouldn't mix.

pranavkm: This is going to get fixed with the newer corefx packages.

NTaylorMullen: Note: Highest version of `System.Reflection.Extensions` that we have available on cidev/volatiledev is 4.0.1 (which is what we're currently referencing). Therefore, doesn't look like there's currently any version to bump this up to.

NTaylorMullen:
:up: :date:

dougbu:
:shipit:

Issue #719, opened 2016-03-21
Pool TagHelperOutput and TagHelperContext.


NTaylorMullen:
These two objects should be pooled alongside of `TagHelperExecutionContext`. As it stands today they're allocated per `TagHelper` tag and cause additional allocations for `TagHelperAttribute[]`/`TagHelperAttributeList`.

This of course would result in a change in behavior where the lifetimes of `TagHelperOutput`/`TagHelperContext` expire once a tags `TagHelper`s have been executed.

![image](https://cloud.githubusercontent.com/assets/2008729/13926993/f691eb82-ef4b-11e5-8f54-284e731d904d.png)

After pooling these in the nimullen/experimental branch I noticed a 20% reduction in overall allocations. Granted this was for a site where every tag was a `TagHelper`

NTaylorMullen:
@Eilon / @danroth27 ran this across @DamianEdwards and he's on board.

NTaylorMullen:
3c73ae9cfeb141b743b84d9e0a2ff6bbea00ce6d

Issue #720, opened 2016-03-21
Don't allocate for method => delegate conversion when passing TagHelper infrastructure methods.


NTaylorMullen:
Today we pass `Start/EndTagHelperWritingScope` methods into `TagHelperScopeManager` at the code gen level. While this looks harmless, the C# spec requires that they be converted into delegates. In addition we also pass around `GetChildContentAsync` to flow it to the `TagHelperOutput`, again this results in delegate allocations.

We should hoist these methods to prevent the delegate allocations. They end up being a decent chunk of memory allocated (note the `Func`s):

![image](https://cloud.githubusercontent.com/assets/2008729/13927432/f10dfe5a-ef4e-11e5-9b0d-bbe72a3bb584.png)

NTaylorMullen:
f5c79b62dee1028641269357a43d9977e725271f

PR #722, opened 2016-03-24
Change `TagHelperOutput` and `TagHelperContext` lifetimes to match `TagHelperExecutionContext`.

NTaylorMullen:
- `TagHelperOutput` and `TagHelperContext` lifetimes now mach `TagHelperExecutionContext`s. This means once a set of `TagHelper`s have run on a tag the `TagHelperOutput`/`TagHelperContext` will be re-used.
- Added tests to validate the various `Reset`/`Reinitialize` methods behaved correctly.
- Updated codegen to no longer set `TagHelperExecutionContext.Output`.

**Before:**
image

**After:**
image

**Result:** 9% reduction in overall allocations.

#719


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -108,18 +111,16 @@ public IList<ITagHelper> TagHelpers
...
 
-        public TagHelperContext CreateTagHelperContext() =>
-            new TagHelperContext(_allAttributes, Items, _uniqueId);
-        public TagHelperOutput CreateTagHelperOutput() =>
-            new TagHelperOutput(_tagName, _htmlAttributes, GetChildContentAsync)
-            {
-                TagMode = _tagMode
-            };
+        /// <summary>
+        /// The <see cref="ITagHelper"/>s' context.
```
rynowak: Apostrophe in wrong place

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -74,6 +70,13 @@ internal TagHelperExecutionContext(string tagName, TagMode tagMode)
             }
 
             _tagHelpers = new List<ITagHelper>();
+            _allAttributes = new TagHelperAttributeList();
+
+            Context = new TagHelperContext(_allAttributes, items, uniqueId);
+            Output = new TagHelperOutput(tagName, new TagHelperAttributeList(), GetChildContentAsync)
```
rynowak: This will capture allocate a `Func<>` do we need to do that?

rynowak: Not a big problem if it only happens once.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -128,6 +128,11 @@ public TagHelperContent AppendFormat(IFormatProvider provider, string format, pa
         /// <returns>A reference to this instance after the clear operation has completed.</returns>
         public abstract TagHelperContent Clear();
 
+        /// <summary>
+        /// Clears the <see cref="TagHelperContent"/> so it can be reused.
+        /// </summary>
+        public abstract void Reset();
```
rynowak: How is this different from clear and why does this need to go on the base class?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -183,17 +187,26 @@ public bool IsContentModified
...
-                {
-                    _attributes = new TagHelperAttributeList();
-                }
+            TagName = tagName;
+            TagMode = tagMode;
+            Attributes.Clear();
 
-                return _attributes;
-            }
+            _preElement?.Reset();
```
rynowak: This is what I meant earlier - you construct these so you know the concrete type here

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/AttributeTargetingTagHelpers.cs
```diff
@@ -44,7 +44,7 @@ public AttributeTargetingTagHelpers()
                 __TestNamespace_CatchAllTagHelper = CreateTagHelper<global::TestNamespace.CatchAllTagHelper>();
                 __tagHelperExecutionContext.Add(__TestNamespace_CatchAllTagHelper);
                 __tagHelperExecutionContext.AddHtmlAttribute(__tagHelperAttribute_1);
-                __tagHelperExecutionContext.Output = await __tagHelperRunner.RunAsync(__tagHelperExecutionContext);
```
rynowak: This is a subtle optimization - the difference here was between `Task` and `Task<T>`. Generally this is going to run synchronously, so it can end up saving quite a bit.

rynowak:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Have another [issue](https://github.com/aspnet/Razor/issues/720) already logged to fix these things :smile:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
NTaylorMullen: `Clear` nukes content and changes `IsModified` => `true`. `Reset` nukes content and turns `IsModified` => `false`.

`IsModified` is an indicator code gen uses to determine if a user doesn't want to run a tags content. [Example](https://github.com/aspnet/Razor/blob/d04f2fd2eeba3c085e9b0044731120f8f0f91492/test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/BasicTagHelpers.cs#L111-L114).

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: The one thing stopping me from putting `Reset` only on `DefaultTagHelperContent` is the ability to set `output.Content`. Also, didn't seem horrible to have an alternative to enable reusability.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
rynowak: :ok:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
rynowak: ok

rynowak:
:shipit:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -108,18 +111,16 @@ public IList<ITagHelper> TagHelpers
             }
         }
 
+        // Internal set for testing.
         /// <summary>
-        /// The <see cref="ITagHelper"/>s' output.
+        /// The <see cref="ITagHelper"/>'s output.
         /// </summary>
-        public TagHelperOutput Output { get; set; }
+        public TagHelperOutput Output { get; internal set; }
```
dougbu: `internal` why? We have very few reasons supporting such a change.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -18,11 +18,14 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
         private List<object> _buffer;
+        private bool _isModified;
 
         private List<object> Buffer
         {
             get
             {
+                _isModified = true;
+
                 if (_buffer == null)
```
dougbu: Likely not worthwhile to lazy-init anything in this class anymore.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
dougbu: Add `<remarks>` distinguishing this method from `Clear()`. This question will come up repeatedly otherwise, especially given the word "Clears" in the `<summary>`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -52,11 +57,23 @@ public class TagHelperContext
...
         /// </summary>
-        public string UniqueId { get; }
+        public string UniqueId { get; private set; }
+
+        /// <summary>
+        /// Clears the <see cref="TagHelperContext"/> and updates its state with the provided values.
+        /// </summary>
+        /// <param name="items">The <see cref="IDictionary{Object, Object}"/> to use.</param>
+        /// <param name="uniqueId">The unique id to use.</param>
+        public void Reinitialize(IDictionary<object, object> items, string uniqueId)
```
dougbu: Why use the name `Reinitialize()` here, in `TagHelperExecutionContext` and in `TagHelperOutput` but `Reset()` in `TagHelperContent`? Think `Reinitialize()` works everywhere.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -15,7 +15,6 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
     public class TagHelperOutput : IHtmlContentContainer
     {
         private readonly Func<bool, HtmlEncoder, Task<TagHelperContent>> _getChildContentAsync;
-        private TagHelperAttributeList _attributes;
         private TagHelperContent _preElement;
         private TagHelperContent _preContent;
         private TagHelperContent _content;
```
dougbu: Wondering if we still need lazy-init for any of the properties in this class? Probably fine not to churn the code but, if lazy isn't needed, should we bother -- especially for `_content`? @rynowak said he was fine either way and I'm feeling ambivalent as well as I type this. But let's confirm the choice.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
dougbu: Does anything still use "the ability to set `output.content`"? (Thought that was needed due to the previous relationship between this class and `TagHelperExecutionContext`.) May have an unnecessary feature that shouldn't be driving our decisions here.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -17,6 +17,55 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
+            // Assert
+            Assert.Equal(expectedTagName, output.TagName);
+            Assert.Equal(expectedTagMode, output.TagMode);
+            Assert.Empty(output.Attributes);
+
+            var getChildContent = await output.GetChildContentAsync();
+            var content = getChildContent.GetContent();
+
+            // We're expecting the initial child content here because normally the TagHelper infrastructure would
+            // swap out the inner workings of GetChildContentAsync to work with it's reinitialized state.
```
dougbu: `its` :trollface:

dougbu:
:watch: for a few questions

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: See comment above :smile:. It's for testing is all.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
NTaylorMullen: Tried to use Reinitialize in places where we took parameters to update state.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: Setter still needed unless we reorganized the code gen semi-drastically to not set output.Content (it does it today to prevent IsModified from being set).

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Going to be Lazy about removing Lazy inits :smile:

NTaylorMullen:
Fixed comment. :up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
NTaylorMullen: Same laziness applies !

NTaylorMullen:
:up: :date:

dougbu:
:shipit:

NTaylorMullen:
Will get this in once the build settles down.

PR #723, opened 2016-03-25
Allow for the TagHelperTypeDescriptorResolver to be injected

javiercn:
Changes required in Razor to support https://github.com/aspnet/Mvc/issues/4087
* Add an interface for ITagHelperTypeResolver and ITagHelperDescriptorFactory
* Extract the default convention for tag helper discovery into TagHelperConventions following the same  pattern we use for ViewComponents.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperConventions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperConventions.cs
@@ -1,0 +1,30 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
+{
+    /// <summary>
+    /// Default convention for determining if a type is a tag helper.
+    /// </summary>
+    public static class TagHelperConventions
+    {
+        private static readonly TypeInfo ITagHelperTypeInfo = typeof(ITagHelper).GetTypeInfo();
+
+        /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperConventions.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Razor.TagHelpers
```
dougbu: `Microsoft.AspNetCore.Razor.Runtime.TagHelpers`

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperConventions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Default convention for determining if a type is a tag helper.
+    /// </summary>
+    public static class TagHelperConventions
```
NTaylorMullen: Is this just for consistency with the `ViewComponent` system or is there another need for this?

javiercn: I need it for AddTagHelpersAsServices and we don't want to duplicate the logic for discovering tag helper types. We do the same thing in view components

NTaylorMullen: So this is preemptive for https://github.com/aspnet/Mvc/issues/4089 ?

javiercn: Yes, for the tag helper part of it

NTaylorMullen:
:shipit:

PR #725, opened 2016-03-27
Add a couple of missing `global::` prefixes

dougbu:
- was missing when writing out full name of an `enum` value
- was missing in one case when writing out `TagHelperAttribute` type name

nit: updated `CSharpTagHelperFieldDeclarationVisitor.WritePrivateField()` to always include the prefix
- no change in generated code but centralizes the prefix addition and reduces intermediate `string`s


dougbu:
/cc @NTaylorMullen 
Will test MVC with these changes in place...

NTaylorMullen:
:shipit:

dougbu:
a8d4da9

Issue #621, opened 2015-11-23
Optimize TagHelperContent for SetContent


rynowak:
Much of the time a `TagHelperContent` has its content set by calling `SetContent`. In these usages there will generally just be one item added to the buffer. We can optimize this by eliding the creation of the buffer when `SetContent` is used.

We might want to look for another magic number here besides 1, but it's clear that inlining a single item can be a win.

![image](https://cloud.githubusercontent.com/assets/1430011/11346255/93a404d4-91d0-11e5-8bbf-35a0509606be.png)

dougbu:
There's also a bug in how MVC uses `SetContent()`: Those tag helpers currently _assume_ `Content` is empty but should be clearing what other tag helpers have done.

rynowak:
`SetContent` calls clear

dougbu:
I know but the tag helpers I'm talking about start w/ `Append()` and don't clear implicitly or explicitly.

Eilon:
@dougbu can you log a separate bug for the other issue you're discussing? I'm not sure it's 100% related to this bug, is it?

dougbu:
@Eilon 
> @dougbu can you log a separate bug

Filed aspnet/Mvc#3852.

NTaylorMullen:
https://github.com/aspnet/Razor/pull/728/commits/c400289de5f174accfd7fec3dda7a8c894189726
https://github.com/aspnet/Mvc/pull/4412/commits/ee80f5f84c9ee6b3d4f900e14e574f5b88ecec28

PR #728, opened 2016-03-31
Change `DefaultTagHelperContent` to be smart about single content entries.

NTaylorMullen:
- Today `TagHelperContent`s always allocate their underlying buffer even though they typically only ever have a single entry. Added a field to enable us to only allocate the backing buffer when we absolutely need to.
- Removed `IsEmpty` from `TagHelperContent` since it was not used in any of our `TagHelper`s for simplification. Changed `IsWhiteSpace` naming to be `IsEmptyOrWhiteSpace` since it can be used to indicate either state.
- Updated tests.

**Before:**
image

**After:**
image

**Result:** 4.1% overall reduction in allocations.

#621

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -17,18 +17,27 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
     [DebuggerDisplay("{DebuggerToString(),nq}")]
     public class DefaultTagHelperContent : TagHelperContent
     {
-        private List<object> _buffer;
+        private object _singleContent;
+        private bool _isSingleContentSet;
         private bool _isModified;
+        private bool _hasContent;
+        private List<object> _buffer;
 
         private List<object> Buffer
         {
             get
             {
-                _isModified = true;
-
                 if (_buffer == null)
                 {
                     _buffer = new List<object>();
+                }
+
+                if (_isSingleContentSet)
+                {
+                    Debug.Assert(_buffer.Count == 0);
+
+                    _buffer.Add(_singleContent);
+                    _isSingleContentSet = false;
                 }
 
                 return _buffer;
@@ -40,39 +49,27 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
 
         /// <inheritdoc />
         /// <remarks>Returns <c>true</c> for a cleared <see cref="TagHelperContent"/>.</remarks>
-        public override bool IsWhiteSpace
+        public override bool IsEmptyOrWhiteSpace
         {
             get
             {
-                if (!IsModified)
+                if (!_hasContent)
                 {
                     return true;
                 }
 
                 using (var writer = new EmptyOrWhiteSpaceWriter())
                 {
-                    foreach (var entry in _buffer)
+                    if (_isSingleContentSet)
                     {
-                        if (entry == null)
-                        {
-                            continue;
-                        }
+                        return IsEmptyOrWhiteSpaceCore(_singleContent, writer);
+                    }
 
-                        var stringValue = entry as string;
-                        if (stringValue != null)
+                    for (var i = 0; i < (_buffer?.Count ?? 0); i++)
+                    {
+                        if (!IsEmptyOrWhiteSpaceCore(Buffer[i], writer))
                         {
-                            if (!string.IsNullOrWhiteSpace(stringValue))
-                            {
-                                return false;
-                            }
-                        }
-                        else
-                        {
-                            ((IHtmlContent)entry).WriteTo(writer, HtmlEncoder.Default);
-                            if (!writer.IsWhiteSpace)
-                            {
-                                return false;
-                            }
+                            return false;
                         }
                     }
                 }
@@ -82,73 +79,20 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         }
 
         /// <inheritdoc />
-        public override bool IsEmpty
-        {
-            get
-            {
-                if (!IsModified)
-                {
-                    return true;
-                }
-
-                using (var writer = new EmptyOrWhiteSpaceWriter())
-                {
-                    foreach (var entry in _buffer)
-                    {
-                        if (entry == null)
-                        {
-                            continue;
-                        }
-
-                        var stringValue = entry as string;
-                        if (stringValue != null)
-                        {
-                            if (!string.IsNullOrEmpty(stringValue))
-                            {
-                                return false;
-                            }
-                        }
-                        else
-                        {
-                            ((IHtmlContent)entry).WriteTo(writer, HtmlEncoder.Default);
-                            if (!writer.IsEmpty)
-                            {
-                                return false;
-                            }
-                        }
-                    }
-                }
-
-                return true;
-            }
-        }
+        public override TagHelperContent Append(string unencoded) => AppendCore(unencoded);
 
         /// <inheritdoc />
-        public override TagHelperContent Append(string unencoded)
-        {
-            Buffer.Add(unencoded);
-            return this;
-        }
+        public override TagHelperContent AppendHtml(IHtmlContent htmlContent) => AppendCore(htmlContent);
 
         /// <inheritdoc />
         public override TagHelperContent AppendHtml(string encoded)
         {
             if (encoded == null)
             {
-                Buffer.Add(null);
+                return AppendCore(null);
             }
-            else
-            {
-                Buffer.Add(new HtmlEncodedString(encoded));
-            }
-            return this;
-        }
 
-        /// <inheritdoc />
-        public override TagHelperContent AppendHtml(IHtmlContent htmlContent)
-        {
-            Buffer.Add(htmlContent);
-            return this;
+            return AppendCore(new HtmlEncodedString(encoded));
         }
 
         /// <inheritdoc />
@@ -159,32 +103,20 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 throw new ArgumentNullException(nameof(destination));
             }
 
-            if (!IsModified)
+            if (!_hasContent)
             {
                 return;
             }
 
-            for (var i = 0; i < Buffer.Count; i++)
+            if (_isSingleContentSet)
             {
-                var entry = Buffer[i];
-                if (entry == null)
+                CopyToCore(_singleContent, destination);
+            }
+            else
+            {
+                for (var i = 0; i < (_buffer?.Count ?? 0); i++)
                 {
-                    continue;
-                }
-
-                string entryAsString;
-                IHtmlContentContainer entryAsContainer;
-                if ((entryAsString = entry as string) != null)
-                {
-                    destination.Append(entryAsString);
-                }
-                else if ((entryAsContainer = entry as IHtmlContentContainer) != null)
-                {
-                    entryAsContainer.CopyTo(destination);
-                }
-                else
-                {
-                    destination.AppendHtml((IHtmlContent)entry);
+                    CopyToCore(Buffer[i], destination);
                 }
             }
         }
@@ -197,62 +129,50 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 throw new ArgumentNullException(nameof(destination));
             }
 
-            if (!IsModified)
+            if (!_hasContent)
             {
                 return;
             }
 
-            for (var i = 0; i < Buffer.Count; i++)
+            if (_isSingleContentSet)
             {
-                var entry = Buffer[i];
-                if (entry == null)
+                MoveToCore(_singleContent, destination);
+            }
+            else
+            {
+                for (var i = 0; i < (_buffer?.Count ?? 0); i++)
                 {
-                    continue;
-                }
-
-                string entryAsString;
-                IHtmlContentContainer entryAsContainer;
-                if ((entryAsString = entry as string) != null)
-                {
-                    destination.Append(entryAsString);
-                }
-                else if ((entryAsContainer = entry as IHtmlContentContainer) != null)
-                {
-                    entryAsContainer.MoveTo(destination);
-                }
-                else
-                {
-                    destination.AppendHtml((IHtmlContent)entry);
+                    MoveToCore(Buffer[i], destination);
                 }
             }
 
-            Buffer.Clear();
+            Clear();
         }
 
         /// <inheritdoc />
         public override TagHelperContent Clear()
         {
-            Buffer.Clear();
+            _hasContent = false;
+            _isModified = true;
+            _isSingleContentSet = false;
+            _buffer?.Clear();
             return this;
         }
 
         /// <inheritdoc />
         public override void Reinitialize()
         {
-            _buffer?.Clear();
+            Clear();
             _isModified = false;
         }
 
         /// <inheritdoc />
-        public override string GetContent()
-        {
-            return GetContent(HtmlEncoder.Default);
-        }
+        public override string GetContent() => GetContent(HtmlEncoder.Default);
 
         /// <inheritdoc />
         public override string GetContent(HtmlEncoder encoder)
         {
-            if (_buffer == null)
+            if (!_hasContent)
             {
                 return string.Empty;
             }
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -31,16 +33,38 @@ private List<object> Buffer
                     _buffer = new List<object>();
                 }
 
+                if (_isSingleContentSet)
+                {
+                    Debug.Assert(_buffer.Count == 0);
+
+                    _buffer.Add(_singleContent);
+                    _isSingleContentSet = false;
+                }
```
dougbu: No need to check `_isSingleContentSet` except inside the `_buffer == null` block. IOW don't un-nest the code just to be able to perform the assertion.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -31,16 +33,38 @@ private List<object> Buffer
...
 
+        private object SingleContent
+        {
+            get
+            {
+                return _singleContent;
+            }
+            set
+            {
+                _isModified = true;
```
dougbu: Should `Debug.Assert(!_isSingleContentSet)` here.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -31,16 +33,38 @@ private List<object> Buffer
...
+            }
+        }
+
         /// <inheritdoc />
         public override bool IsModified => _isModified;
 
         /// <inheritdoc />
         /// <remarks>Returns <c>true</c> for a cleared <see cref="TagHelperContent"/>.</remarks>
-        public override bool IsWhiteSpace
+        public override bool IsEmptyOrWhiteSpace
```
dougbu: Why are changes to the public surface mixed into this PR? We decided long ago to support `IsEmpty` and I'm not really sure why the `<label>` and validation message tag helpers aren't using it these days.

In any case, it's hard to see what's actually changing due to the actual #621 fix. If you think this is valuable, please open an issue and submit separately.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -51,28 +75,16 @@ public override bool IsWhiteSpace
...
-                        var stringValue = entry as string;
-                        if (stringValue != null)
-                        {
-                            if (!string.IsNullOrWhiteSpace(stringValue))
-                            {
-                                return false;
-                            }
-                        }
-                        else
+                    for (var i = 0; i < Buffer.Count; i++)
```
dougbu: Since this change helps only if all tag helpers at one depth in a page add at most one entry to each `TagHelperContent`, should probably also avoid this allocation and thereby apply if all tag helpers at one depth in a page add at most one entry to each `TagHelperContent` _or `Clear()` it_. That is, avoid this allocation by exiting early if `_buffer == null` or doing the `null`-conditional dance.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -82,73 +94,37 @@ public override bool IsWhiteSpace
...
-                            ((IHtmlContent)entry).WriteTo(writer, HtmlEncoder.Default);
-                            if (!writer.IsEmpty)
-                            {
-                                return false;
-                            }
-                        }
-                    }
-                }
+        /// <inheritdoc />
+        public override TagHelperContent AppendHtml(IHtmlContent htmlContent) => AppendCore(htmlContent);
```
dougbu: Since you said offline that consistency was one of your explicit goals :smiling_imp:, move this above `AppendHtml(string)`.

NTaylorMullen: :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -282,23 +232,129 @@ public override void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+            }
+            else
+            {
+                Buffer.Add(entry);
+            }
+
+            return this;
+        }
+
+        private TagHelperContent SetContentCore(object item)
```
dougbu: Not positive this short method is worthwhile factoring out.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -82,73 +94,37 @@ public override bool IsWhiteSpace
...
+        public override TagHelperContent SetHtmlContent(IHtmlContent htmlContent) => SetContentCore(htmlContent);
 
         /// <inheritdoc />
-        public override TagHelperContent AppendHtml(string encoded)
+        public override TagHelperContent SetHtmlContent(string encoded)
         {
             if (encoded == null)
             {
-                Buffer.Add(null);
+                return SetContentCore(null);
```
dougbu: If you get rid of `SetContentCore()`, call `SetContent((string)null)` here and `SetContent(new HtmlEncodedString(encoded))` below.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -164,28 +140,15 @@ public override void CopyTo(IHtmlContentBuilder destination)
...
-            for (var i = 0; i < Buffer.Count; i++)
+            if (_isSingleContentSet)
             {
-                var entry = Buffer[i];
-                if (entry == null)
-                {
-                    continue;
-                }
+                CopyToCore(SingleContent, destination);
+                return;
```
dougbu: Give this method the same shape as `MoveTo()` i.e. remove this `return` and nest the `for` in an `else`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -82,73 +94,37 @@ public override bool IsWhiteSpace
...
-                    }
-                }
+        /// <inheritdoc />
+        public override TagHelperContent AppendHtml(IHtmlContent htmlContent) => AppendCore(htmlContent);
 
-                return true;
-            }
-        }
+        /// <inheritdoc />
+        public override TagHelperContent SetContent(string unencoded) => SetContentCore(unencoded);
```
dougbu: What is the benefit of overriding the `Set[Html]Content()` methods from the base class? Looks like the Right Thing<sup>:tm:</sup> happens without the new `virtual`ality.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -164,7 +164,7 @@ public void MoveTo_CopiesAllItems_AndClears()
             source.MoveTo(destination);
 
             // Assert
-            Assert.True(source.IsEmpty);
+            Assert.True(source.IsEmptyOrWhiteSpace);
```
dougbu: All of the tests are now looking for something different. To match previous checks, need to get the entire content and `Assert.IsEmpty(contentString)`. We really don't want some random thing sneaking whitespace in.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/TagHelperOutputTest.cs
```diff
@@ -1079,11 +1079,11 @@ public void MoveTo_MovesToBuilder(TagHelperOutput output, string expected)
...
-            Assert.True(output.PreElement.IsEmpty);
-            Assert.True(output.PreContent.IsEmpty);
-            Assert.True(output.Content.IsEmpty);
-            Assert.True(output.PostContent.IsEmpty);
-            Assert.True(output.PostElement.IsEmpty);
+            Assert.True(output.PreElement.IsEmptyOrWhiteSpace);
+            Assert.True(output.PreContent.IsEmptyOrWhiteSpace);
+            Assert.True(output.Content.IsEmptyOrWhiteSpace);
+            Assert.True(output.PostContent.IsEmptyOrWhiteSpace);
+            Assert.True(output.PostElement.IsEmptyOrWhiteSpace);
```
dougbu: Again, not asserting what was checked before.

test/Microsoft.AspNetCore.Razor.Runtime.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -386,21 +386,21 @@ public void CanIdentifyWhiteSpace_WithoutIgnoringStrings()
             tagHelperContent.Append("Hello");
 
             // Assert
-            Assert.False(tagHelperContent.IsWhiteSpace);
+            Assert.False(tagHelperContent.IsEmptyOrWhiteSpace);
         }
 
         [Fact]
-        public void IsEmpty_InitiallyTrue()
+        public void IsEmptyOrWhiteSpace_InitiallyTrue()
```
dougbu: My previous [comment](https://github.com/aspnet/Razor/pull/728/files#r58143306) about "now looking for something different" do not apply when the test is specifically about the property.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Talked offline.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Will do.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Gonna leave in for consistency and pranav points.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: Hmm, you know what. You be absolutely right. :+1:. In my prototype I didn't make append smart. In this I did. Will remove the overrides.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
NTaylorMullen: These are pooled. If you reinitialize a `TagHelperContent` the buffer may not be `null`. I could make the changes to always use the normal buffer once it's initialized but that liters the code with a lot more `if (_buffer != null)` checks. Didn't seem worth it for the same functionality.

NTaylorMullen:
:up: :date:

NTaylorMullen:
Will be getting this in tomorrow unless there are additional comments.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -282,23 +218,121 @@ public override void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                    return false;
                 }
             }
+
+            return true;
+        }
+
+        private TagHelperContent AppendCore(object entry)
+        {
+            if (!IsModified)
```
dougbu: This condition is incorrect and inefficient for `SetContent()` and similar cases -- will allocate buffers unecessarily. `IsModified` is `true` after `Clear()` is called but `Buffer` may remain `null`. Should instead have a clear `_isEmpty` field. The transitions should always be `_isEmpty` -> `_isSingleContentSet` -> `_buffer != null && _buffer.Count > 1` regardless of whether `IsModified` or `_buffer != null` initially i.e. after `Clear()` or `Reinitialize()` call.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
```diff
@@ -19,14 +19,9 @@ public abstract class TagHelperContent : IHtmlContentBuilder
...
-        /// Gets a value indicating whether the content is empty.
+        /// Gets a value indicating whether the content is empty or whitespace.
         /// </summary>
-        public abstract bool IsEmpty { get; }
-
-        /// <summary>
-        /// Gets a value indicating whether the content is whitespace.
-        /// </summary>
-        public abstract bool IsWhiteSpace { get; }
+        public abstract bool IsEmptyOrWhiteSpace { get; }
```
dougbu: Where is the reaction PR in MVC for this `public` API change? Has @Eilon or someone else agreed with this change?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
dougbu: :ok: though I think you mean "reused" when you say "pooled".

dougbu:
> Will be getting this in tomorrow unless there are additional comments.

Huh? There's an API change in here that needs approval and a signed-off MVC reaction PR.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
NTaylorMullen: @rynowak and I talked about this guy. It's a feature that we've never used or seen used and more importantly can be easily replaced. Didn't feel that this was important enough to warrant a larger conversation/decision with more people.

dougbu: We had a long discussion about what to expose on `TagHelperContent` when designing this API. In particular, the original decisions were _not_ just about what we needed in our tag helpers. So what is different now that drives removing the `IsEmpty` / `IsWhiteSpace` distinction? It's got to be something more than just avoiding fixing up one more property in this PR.

@Eilon? @DamianEdwards?

dougbu:
:watch: also waiting for https://github.com/aspnet/Razor/pull/728#discussion-diff-58301375 fix.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContent.cs
NTaylorMullen: With this PR the complexity of `DefaultTagHelperContent` increases significantly. That was the initial driving factor in removing the property. @rynowak and I discussed offline about possibly removing `IsEmpty` and `IsWhiteSpace` as part of this PR. I investigated what it would mean to remove the two properties and found that it really only made sense to remove `IsEmpty` and re-purpose the existing `IsWhiteSpace` implementation. I've left a list of reasons at the end of this response.

I'd prefer not to spend time reworking `TagHelperContent.IsEmpty` and its existing tests to undo that work one PR later. I apologize if this seemed off-base @dougbu, @rynowak and I have been working closely on our perf findings/changes in this area which is why I only cc'd him at the beginning of this PR; I've been trying to include others where it makes sense to prevent misunderstandings like this :smile:.

-------------------------------------------------------

Full list of reasons for removing `IsEmpty` as mentioned above:

- We don't use `IsEmpty` in any of our Mvc `TagHelper` implementations.
- We've never seen a user use `IsEmpty` for their own `TagHelper`s.
- `IsEmpty` only makes an extreme edge case scenario more performant. Functionally it can be replaced with `content.GetContent().Length == 0`.
- `IsEmpty` was initially designed with the intent that our Mvc `TagHelper`s would need it; and if they would need it, so would other `TagHelper` author's `TagHelper`s. Ultimately, `IsEmpty` doesn't provide enough value from a maintenance, API or functional perspective to warrant its existence.

NTaylorMullen:
Talked with @DamianEdwards offline. He approved the public API surface change here and its approval for RC2.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
rynowak: I agree, seems odd to reuse `IsModified` in this way. It would be better to just check the conditions your really care about here.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -17,8 +17,11 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
     [DebuggerDisplay("{DebuggerToString(),nq}")]
     public class DefaultTagHelperContent : TagHelperContent
     {
-        private List<object> _buffer;
+        private object _singleContent;
+        private bool _isSingleContentSet;
         private bool _isModified;
+        private bool _hasContent;
```
dougbu: Like the new field. But why does the `Buffer` setter still muck with `_isModified` now. Better to manipulate these two fields in fewer places, in particular set `_isModified` to `true` only in `Clear()` and `AppendCore()`.

NTaylorMullen: :+1:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -277,28 +199,131 @@ public override void WriteTo(TextWriter writer, HtmlEncoder encoder)
...
+                _hasContent = true;
+                _isModified = true;
+                _isSingleContentSet = true;
+                _singleContent = entry;
+            }
+            else
+            {
+                Buffer.Add(entry);
+
+                _hasContent = _buffer.Count > 0;
```
dougbu: Huh? `Buffer.Count > 1` is an invariant at this line. Just set `_hasContent` and `isModified` to `true` prior to the `if`.

NTaylorMullen: lmao :+1:

dougbu:
:shipit: once you've got the MVC reaction PR ready.

PR #730, opened 2016-04-02
Add Visual Studio specific RC1 binaries.

NTaylorMullen:
- This is needed for Visual Studio RC1 backwards compatibility.

/cc @Eilon @muratg @davidfowl


dougbu:
Curious: How does adding these two projects help VS w/ RC1 compatibility?

dougbu:
Does not seem safe to add product code with no tests at all. Think we need _something_.

dougbu:
:watch: (CodeFlow doesn't have a Waiting status :frowning:)

NTaylorMullen:
This is a temporary solution. It enables us to have RC1 binaries flow through the CI and eventually generate RC1 bits with a different name. The problem this is solving is in order for Visual Studio to support RC1 projects they need to have two sets of binaries on disk (RC1 bits and RC2 bits). Issues occurred with the two sets of binaries on disk when Visual Studio tried to resolve assembly names for assemblies that existed in two locations on disk. Example: `Microsoft.Extensions.Caching.Abstractions` has the same name, same version, different public contract between RC1 and RC2.

Once we have RC1 bits we can consider reverting this commit. This approach of checking in seemingly unneeded projects into our primary src folders was approved by @Eilon.

pranavkm:
Had a cursory look and this looks :shipit:

PR #738, opened 2016-04-26
Fix: Newline at the end of text tag in code block is rendered as markup

ajaybhargavb:
Fixes issue https://github.com/aspnet/Razor/issues/485#issuecomment-214654609

@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
+++ b/src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
@@ -1118,7 +1118,9 @@ namespace Microsoft.AspNetCore.Razor.Parser
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1118,7 +1118,9 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
                     }
                 }
 
-                if (!EndOfFile && CurrentSymbol.Type == HtmlSymbolType.NewLine)
+                if (!EndOfFile &&
+                    CurrentSymbol.Type == HtmlSymbolType.NewLine &&
```
NTaylorMullen: Since we're touching this now, lets just change it to `At(HtmlSymbolType.NewLine)`

NTaylorMullen:
:shipit:

Issue #739, opened 2016-04-28
Newline at the end of text tag should not be rendered when it is followed csharp


ajaybhargavb:
Related to #485 

Example:

In this case,
```HTML
@using(...)
{
  <div>
...
  </div>
  <input />
}
```
The newline after `</div>` should be rendered and the newline after `<input/>` should not be rendered.

cc' @dougbu @NTaylorMullen

ajaybhargavb:
Update: Nuking the newline at the end of all html tags that are followed by csharp breaks scenarios like nested `foreach` with `<ul>` and `<li>`. So doing this change only for `<text>` tags.

dougbu:
> So doing this change only for `<text>` tags.

Sounds reasonable.

PR #740, opened 2016-04-29
Newline after text tag rendered only if followed by html

ajaybhargavb:
Issue #739 

@NTaylorMullen @dougbu


src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1100,29 +1100,28 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
                 }
 
-                if (!EndOfFile &&
-                    At(HtmlSymbolType.NewLine) &&
-                    Context.LastSpan.Kind != SpanKind.Transition)
+                if (shouldAcceptWhitespaceAndNewLine)
                 {
-                    AcceptAndMoveNext();
+                    AcceptWhile(HtmlSymbolType.WhiteSpace);
+                    Optional(HtmlSymbolType.NewLine);
```
dougbu: This doesn't line up w/ the `ReadWhile()`. That'll consume multiple intervening lines but this will stop at the end of the current line. If that's intentional, need comments here explaining why.

src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1100,29 +1100,28 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
+                    if (!At(HtmlSymbolType.OpenAngle))
                     {
-                        AcceptWhile(HtmlSymbolType.WhiteSpace);
+                        // Don't accept whitespace as markup if the end text tag is followed by csharp.
+                        shouldAcceptWhitespaceAndNewLine = false;
                     }
+
+                    PutCurrentBack();
+                    PutBack(symbols);
+                    EnsureCurrent();
```
dougbu: Don't we have utility APIs to readahead that avoid these explicit fixups?

src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: This is intentional. Will add comments.

src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
```diff
@@ -1100,29 +1100,28 @@ private void EndTagBlock(Stack<Tuple<HtmlSymbol, SourceLocation>> tags, bool com
...
-                    if (Context.LastSpan.Kind == SpanKind.Transition)
-                    {
-                        // Output current span content as markup.
-                        Output(SpanKind.Markup);
+                    var symbols = ReadWhile(
+                        f => (f.Type == HtmlSymbolType.WhiteSpace) || (f.Type == HtmlSymbolType.NewLine));
 
-                        // Accept and mark the whitespace at the end of a <text> tag as code.
-                        AcceptWhile(HtmlSymbolType.WhiteSpace);
-                        Span.ChunkGenerator = new StatementChunkGenerator();
```
dougbu: Where did the `ChunkGenerator` setting move?

src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: I don't think there are any. `LookAhead` needs a count so it cannot be used.

ajaybhargavb:
:up: :date:

test/Microsoft.AspNetCore.Razor.Test/Parser/Html/HtmlDocumentTest.cs
```diff
@@ -341,6 +343,85 @@ public void ParseDocumentDoesNotRenderExtraNewlineAtTheEndTextTagInVerbatimBlock
...
+                                .Accepts(AcceptedCharacters.None)),
+                        Factory.EmptyCSharp().AsStatement(),
+                        Factory.MetaCode("}").Accepts(AcceptedCharacters.None)),
+                    BlockFactory.MarkupTagBlock("<html>")));
+        }
+
+        [Fact]
+        public void ParseDocumentRendersNewlineAtTheEndHtmlTagInACSharpBlock()
+        {
+            ParseDocumentTest("@using(var foo = new Foo()){<div>Blah</div>\r\n<div>Bleh</div>\r\n}<html>",
```
NTaylorMullen: You just changed the `<text>` functionality. This is good for confirming that the `<text>` tag works as expected but I'm sure we have an identical test elsewhere. Break this one out into two:

- `"@using(var foo = new Foo()){<div>Hello</div><text>Blah</text>\r\n\r\n<div>Bleh</div>\r\n}<html>"`
- `"@{\r\n<text>Blah</text>\r\n\r\n<input /><text>Bleh</text>\r\n    }<html>"`

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/ComplexTagHelpers.DesignTime.cs
```diff
@@ -93,24 +93,12 @@ public ComplexTagHelpers()
 #line hidden
 
 #line 18 "ComplexTagHelpers.cshtml"
-                                                              
-
-#line default
```
NTaylorMullen: Nice, this looks like it's acting more optimally.

NTaylorMullen:
:watch: for other tests

src/Microsoft.AspNetCore.Razor/Parser/HtmlMarkupParser.Block.cs
ajaybhargavb: No need to do this because we are not accepting it as code here anymore.

test/Microsoft.AspNetCore.Razor.Test/Parser/Html/HtmlDocumentTest.cs
ajaybhargavb: Both these cases are already covered by `ParseDocumentDoesNotRenderExtraNewlineAtTheEndTextTagInVerbatimBlockIfFollowedByCSharp` and `ParseDocumentRendersExtraNewlineAtTheEndTextTagInVerbatimBlockIfFollowedByHtml`. I added this case intentionally for the normal html tag case. I'll check if we have an existing test for this case and remove this.

NTaylorMullen:
:shipit: once you remove the unneeded test.

dougbu:
:shipit: w/ the additional comments in `EndTagBlock()`.

Issue #705, opened 2016-03-01
Track TagHelperAttribute quotes


NTaylorMullen:
This would enable `TagHelper` users to write things like `<input type=checkbox />` and have it output the identical code; whereas today we output `<input type="checkbox" />`.

This change would actually have a significant perf impact as well. Today in `TagHelperOutput` we have to [write an attribute to an intermediate writer](https://github.com/aspnet/Razor/blob/a3e92ab3f9656a6f2bd7504a94aa446918e62d50/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs#L316-L323) to potentially escape any double quotes in the case that a user wrote `<input type=' """" ' />` since we always output with double quotes.

This intermediate writing step in my tests have represented over 5% of allocations.

![image](https://cloud.githubusercontent.com/assets/2008729/13439271/936486fa-dfa1-11e5-99b6-d4d285865d96.png)

Eilon:
Is preserving the non-quotes actually important?

Also, what happens if a Tag Helper modifies that attribute? Does it "gain" the quotes after it's modified? E.g. let's say in my Tag Helper I set `OutputTags["type"] = "this has spaces";` (or whatever the syntax is). Does it render `<input type=this has spaces />` or `<input type="this has spaces" />`?

NTaylorMullen:
Actually, with the new immutability of `TagHelperAttribute`s the unquoted scenario isn't an issue. For instance, if a `TagHelper` author were to do `output.Attributes.SetAttribute("type", "...")` they'd be saying "create a new `TagHelperAttribute` with name `type` and value `...`" which would default to using a double quoted value. If they instead used the `string name, TagHelperAttribute attribute` overload they'd have the option of using their own quoting.

NTaylorMullen:
e14f4b095d0bd4c5a5ae079f817f9a07217907af
https://github.com/aspnet/Mvc/pull/4587/commits/cfd58f17476df9bd296a3e50f519f570e0952877

PR #743, opened 2016-05-03
Track `TagHelperAttribute` quotes.

NTaylorMullen:
- Removed `Minimized` from `TagHelperAttribute` and replaced it with `HtmlAttributeStructure`.
- Removed `AddMinimizedTagHelperAttribute` method from `TagHelperExecutionContext` since we always provide a `HtmlAttributeStructure` now.
- Stopped manually escaping double quotes because we now know how an attribute was originally written.
- Updated tests to account for new attribute format (from kvp).

#705

/cc @dougbu @pranavkm @ajaybhargavb

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
@@ -139,33 +139,19 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         }
 
         /// <summary>
-        /// Tracks the minimized HTML attribute.
-        /// </summary>
-        /// <param name="name">The minimized HTML attribute name.</param>
-        public void AddMinimizedHtmlAttribute(string name)
-        {
-            if (name == null)
-            {
-                throw new ArgumentNullException(nameof(name));
-            }
-
-            var attribute = new TagHelperAttribute(name);
-            AddHtmlAttribute(attribute);
-        }
-
-        /// <summary>
         /// Tracks the HTML attribute.
         /// </summary>
         /// <param name="name">The HTML attribute name.</param>
         /// <param name="value">The HTML attribute value.</param>
-        public void AddHtmlAttribute(string name, object value)
+        /// <param name="valueStyle">The value style of the attribute.</param>
+        public void AddHtmlAttribute(string name, object value, HtmlAttributeValueStyle valueStyle)
         {
             if (name == null)
             {
                 throw new ArgumentNullException(nameof(name));
             }
 
-            var attribute = new TagHelperAttribute(name, value);
+            var attribute = new TagHelperAttribute(name, value, valueStyle);
             AddHtmlAttribute(attribute);
         }
 
@@ -189,15 +175,16 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         /// </summary>
         /// <param name="name">The bound attribute name.</param>
         /// <param name="value">The attribute value.</param>
-        public void AddTagHelperAttribute(string name, object value)
+        /// <param name="valueStyle">The value style of the attribute.</param>
+        public void AddTagHelperAttribute(string name, object value, HtmlAttributeValueStyle valueStyle)
         {
             if (name == null)
             {
                 throw new ArgumentNullException(nameof(name));
             }
 
-
-            _allAttributes.Add(name, value);
+            var attribute = new TagHelperAttribute(name, value, valueStyle);
+            _allAttributes.Add(attribute);
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Html;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.TagHelpers
@@ -9,39 +12,40 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
     /// <summary>
     /// An HTML tag helper attribute.
     /// </summary>
-    public class TagHelperAttribute
+    public class TagHelperAttribute : IHtmlContentContainer
     {
         /// <summary>
         /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>.
-        /// <see cref="Minimized"/> is set to <c>true</c> and <see cref="Value"/> to <c>null</c>.
+        /// <see cref="ValueStyle"/> is set to <see cref="HtmlAttributeValueStyle.Minimized"/> and <see cref="Value"/> to
+        /// <c>null</c>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         public TagHelperAttribute(string name)
-            : this(name, value: null, minimized: true)
+            : this(name, value: null, valueStyle: HtmlAttributeValueStyle.Minimized)
         {
         }
 
         /// <summary>
         /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>
-        /// and <paramref name="value"/>. <see cref="Minimized"/> is set to <c>false</c>.
+        /// and <paramref name="value"/>. <see cref="ValueStyle"/> is set to <see cref="HtmlAttributeValueStyle.DoubleQuotes"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         /// <param name="value">The <see cref="Value"/> of the attribute.</param>
         public TagHelperAttribute(string name, object value)
-            : this(name, value, minimized: false)
+            : this(name, value, valueStyle: HtmlAttributeValueStyle.DoubleQuotes)
         {
         }
 
         /// <summary>
         /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>,
-        /// <paramref name="value"/> and <paramref name="minimized"/>.
+        /// <paramref name="value"/> and <paramref name="valueStyle"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the new instance.</param>
         /// <param name="value">The <see cref="Value"/> of the new instance.</param>
-        /// <param name="minimized">The <see cref="Minimized"/> value of the new instance.</param>
-        /// <remarks>If <paramref name="minimized"/> is <c>true</c>, <paramref name="value"/> is ignored when this
-        /// instance is rendered.</remarks>
-        public TagHelperAttribute(string name, object value, bool minimized)
+        /// <param name="valueStyle">The <see cref="ValueStyle"/> of the new instance.</param>
+        /// <remarks>If <paramref name="valueStyle"/> is <see cref="HtmlAttributeValueStyle.Minimized"/>,
+        /// <paramref name="value"/> is ignored when this instance is rendered.</remarks>
+        public TagHelperAttribute(string name, object value, HtmlAttributeValueStyle valueStyle)
         {
             if (name == null)
             {
@@ -50,7 +54,7 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
 
             Name = name;
             Value = value;
-            Minimized = minimized;
+            ValueStyle = valueStyle;
         }
 
         /// <summary>
@@ -64,11 +68,9 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         public object Value { get; }
 
         /// <summary>
-        /// Gets an indication whether the attribute is minimized or not.
+        /// Gets the value style of the attribute.
         /// </summary>
-        /// <remarks>If <c>true</c>, <see cref="Value"/> will be ignored.</remarks>
-        public bool Minimized { get; }
-
+        public HtmlAttributeValueStyle ValueStyle { get; }
 
         /// <inheritdoc />
         /// <remarks><see cref="Name"/> is compared case-insensitively.</remarks>
@@ -77,8 +79,147 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
             return
                 other != null &&
                 string.Equals(Name, other.Name, StringComparison.OrdinalIgnoreCase) &&
-                Minimized == other.Minimized &&
-                (Minimized || Equals(Value, other.Value));
+                ValueStyle == other.ValueStyle &&
+                (ValueStyle == HtmlAttributeValueStyle.Minimized || Equals(Value, other.Value));
+        }
+
+        /// <inheritdoc />
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
+        {
+            if (writer == null)
+            {
+                throw new ArgumentNullException(nameof(writer));
+            }
+
+            if (encoder == null)
+            {
+                throw new ArgumentNullException(nameof(encoder));
+            }
+
+            writer.Write(Name);
+
+            if (ValueStyle == HtmlAttributeValueStyle.Minimized)
+            {
+                return;
+            }
+
+            var valuePrefix = GetAttributeValuePrefix(ValueStyle);
+            if (valuePrefix != null)
+            {
+                writer.Write(valuePrefix);
+            }
+
+            var htmlContent = Value as IHtmlContent;
+            if (htmlContent != null)
+            {
+                htmlContent.WriteTo(writer, encoder);
+            }
+            else if (Value != null)
+            {
+                encoder.Encode(writer, Value.ToString());
+            }
+
+            var valueSuffix = GetAttributeValueSuffix(ValueStyle);
+            if (valueSuffix != null)
+            {
+                writer.Write(valueSuffix);
+            }
+        }
+
+        /// <inheritdoc />
+        public void CopyTo(IHtmlContentBuilder destination)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            destination.AppendHtml(Name);
+
+            if (ValueStyle == HtmlAttributeValueStyle.Minimized)
+            {
+                return;
+            }
+
+            var valuePrefix = GetAttributeValuePrefix(ValueStyle);
+            if (valuePrefix != null)
+            {
+                destination.AppendHtml(valuePrefix);
+            }
+
+            string valueAsString;
+            IHtmlContentContainer valueAsHtmlContainer;
+            IHtmlContent valueAsHtmlContent;
+            if ((valueAsString = Value as string) != null)
+            {
+                destination.Append(valueAsString);
+            }
+            else if ((valueAsHtmlContainer = Value as IHtmlContentContainer) != null)
+            {
+                valueAsHtmlContainer.CopyTo(destination);
+            }
+            else if ((valueAsHtmlContent = Value as IHtmlContent) != null)
+            {
+                destination.AppendHtml(valueAsHtmlContent);
+            }
+            else if (Value != null)
+            {
+                destination.Append(Value.ToString());
+            }
+
+            var valueSuffix = GetAttributeValueSuffix(ValueStyle);
+            if (valueSuffix != null)
+            {
+                destination.AppendHtml(valueSuffix);
+            }
+        }
+
+        /// <inheritdoc />
+        public void MoveTo(IHtmlContentBuilder destination)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            destination.AppendHtml(Name);
+
+            if (ValueStyle == HtmlAttributeValueStyle.Minimized)
+            {
+                return;
+            }
+
+            var valuePrefix = GetAttributeValuePrefix(ValueStyle);
+            if (valuePrefix != null)
+            {
+                destination.AppendHtml(valuePrefix);
+            }
+
+            string valueAsString;
+            IHtmlContentContainer valueAsHtmlContainer;
+            IHtmlContent valueAsHtmlContent;
+            if ((valueAsString = Value as string) != null)
+            {
+                destination.Append(valueAsString);
+            }
+            else if ((valueAsHtmlContainer = Value as IHtmlContentContainer) != null)
+            {
+                valueAsHtmlContainer.MoveTo(destination);
+            }
+            else if ((valueAsHtmlContent = Value as IHtmlContent) != null)
+            {
+                destination.AppendHtml(valueAsHtmlContent);
+            }
+            else if (Value != null)
+            {
+                destination.Append(Value.ToString());
+            }
+
+            var valueSuffix = GetAttributeValueSuffix(ValueStyle);
+            if (valueSuffix != null)
+            {
+                destination.AppendHtml(valueSuffix);
+            }
         }
 
         /// <inheritdoc />
@@ -94,10 +235,43 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         {
             var hashCodeCombiner = HashCodeCombiner.Start();
             hashCodeCombiner.Add(Name, StringComparer.Ordinal);
-            hashCodeCombiner.Add(Value);
-            hashCodeCombiner.Add(Minimized);
+
+            if (ValueStyle != HtmlAttributeValueStyle.Minimized)
+            {
+                hashCodeCombiner.Add(Value);
+            }
+
+            hashCodeCombiner.Add(ValueStyle);
 
             return hashCodeCombiner.CombinedHash;
+        }
+
+        private static string GetAttributeValuePrefix(HtmlAttributeValueStyle valueStyle)
+        {
+            switch (valueStyle)
+            {
+                case HtmlAttributeValueStyle.DoubleQuotes:
+                    return "=\"";
+                case HtmlAttributeValueStyle.SingleQuotes:
+                    return "='";
+                case HtmlAttributeValueStyle.NoQuotes:
+                    return "=";
+            }
+
+            return null;
+        }
+
+        private static string GetAttributeValueSuffix(HtmlAttributeValueStyle valueStyle)
+        {
+            switch (valueStyle)
+            {
+                case HtmlAttributeValueStyle.DoubleQuotes:
+                    return "\"";
+                case HtmlAttributeValueStyle.SingleQuotes:
+                    return "'";
+            }
+
+            return null;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
@@ -305,7 +306,13 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 destination.AppendHtml("<");
                 destination.AppendHtml(TagName);
 
-                CopyAttributesTo(destination);
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < Attributes.Count; i++)
+                {
+                    var attribute = Attributes[i];
+                    destination.AppendHtml(" ");
+                    attribute.CopyTo(destination);
+                }
 
                 if (TagMode == TagMode.SelfClosing)
                 {
@@ -350,7 +357,13 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 destination.AppendHtml("<");
                 destination.AppendHtml(TagName);
 
-                CopyAttributesTo(destination);
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < Attributes.Count; i++)
+                {
+                    var attribute = Attributes[i];
+                    destination.AppendHtml(" ");
+                    attribute.MoveTo(destination);
+                }
 
                 if (TagMode == TagMode.SelfClosing)
                 {
@@ -405,49 +418,11 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 writer.Write(TagName);
 
                 // Perf: Avoid allocating enumerator
-                for (var i = 0; i < (Attributes.Count); i++)
+                for (var i = 0; i < Attributes.Count; i++)
                 {
                     var attribute = Attributes[i];
                     writer.Write(" ");
-                    writer.Write(attribute.Name);
-
-                    if (attribute.Minimized)
-                    {
-                        continue;
-                    }
-
-                    writer.Write("=\"");
-                    var value = attribute.Value;
-                    var htmlContent = value as IHtmlContent;
-                    if (htmlContent != null)
-                    {
-                        // Perf: static text in a bound attribute go down this path. Avoid allocating if possible (common case).
-                        var htmlString = value as HtmlString;
-                        if (htmlString != null && !htmlString.Value.Contains("\""))
-                        {
-                            writer.Write(htmlString.Value);
-                        }
-                        else
-                        {
-                            // There's no way of tracking the attribute value quotations in the Razor source. Therefore, we
-                            // must escape any IHtmlContent double quote values in the case that a user wrote:
-                            // <p name='A " is valid in single quotes'></p>
-                            using (var stringWriter = new StringWriter())
-                            {
-                                htmlContent.WriteTo(stringWriter, encoder);
-                                stringWriter.GetStringBuilder().Replace("\"", "&quot;");
-
-                                var stringValue = stringWriter.ToString();
-                                writer.Write(stringValue);
-                            }
-                        }
-                    }
-                    else if (value != null)
-                    {
-                        encoder.Encode(writer, value.ToString());
-                    }
-
-                    writer.Write("\"");
+                    attribute.WriteTo(writer, encoder);
                 }
 
                 if (TagMode == TagMode.SelfClosing)
@@ -475,77 +450,6 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
             }
 
             _postElement?.WriteTo(writer, encoder);
-        }
-
-        private void CopyAttributesTo(IHtmlContentBuilder destination)
-        {
-            StringWriter stringWriter = null;
-
-            // Perf: Avoid allocating enumerator
-            for (var i = 0; i < (Attributes.Count); i++)
-            {
-                var attribute = Attributes[i];
-                destination.AppendHtml(" ");
-                destination.AppendHtml(attribute.Name);
-
-                if (attribute.Minimized)
-                {
-                    continue;
-                }
-
-                destination.AppendHtml("=\"");
-                var value = attribute.Value;
-                var htmlContent = value as IHtmlContent;
-                if (htmlContent != null)
-                {
-                    // Perf: static text in a bound attribute go down this path. Avoid allocating if possible (common case).
-                    var htmlString = value as HtmlString;
-                    if (htmlString != null && !htmlString.Value.Contains("\""))
-                    {
-                        destination.AppendHtml(htmlString);
-                    }
-                    else
-                    {
-                        // Perf: We'll share this writer implementation for all attributes since
-                        // they can't nest.
-                        stringWriter = stringWriter ?? new StringWriter();
-
-                        destination.AppendHtml(new AttributeContent(htmlContent, stringWriter));
-                    }
-                }
-                else if (value != null)
-                {
-                    destination.Append(value.ToString());
-                }
-
-                destination.AppendHtml("\"");
-            }
-        }
-
-        private class AttributeContent : IHtmlContent
-        {
-            private readonly IHtmlContent _inner;
-            private readonly StringWriter _stringWriter;
-
-            public AttributeContent(IHtmlContent inner, StringWriter stringWriter)
-            {
-                _inner = inner;
-                _stringWriter = stringWriter;
-            }
-
-            public void WriteTo(TextWriter writer, HtmlEncoder encoder)
-            {
-                // There's no way of tracking the attribute value quotations in the Razor source. Therefore, we
-                // must escape any IHtmlContent double quote values in the case that a user wrote:
-                // <p name='A " is valid in single quotes'></p>
-                _inner.WriteTo(_stringWriter, encoder);
-                _stringWriter.GetStringBuilder().Replace("\"", "&quot;");
-
-                var stringValue = _stringWriter.ToString();
-                writer.Write(stringValue);
-
-                _stringWriter.GetStringBuilder().Clear();
-            }
         }
     }
 }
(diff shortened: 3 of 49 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
```diff
@@ -138,33 +138,19 @@ public void Add(ITagHelper tagHelper)
         }
 
         /// <summary>
-        /// Tracks the minimized HTML attribute.
-        /// </summary>
-        /// <param name="name">The minimized HTML attribute name.</param>
-        public void AddMinimizedHtmlAttribute(string name)
```
dougbu: Keep this method. It's very explicit.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -2,46 +2,50 @@
...
+using Microsoft.AspNetCore.Html;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.TagHelpers
 {
     /// <summary>
     /// An HTML tag helper attribute.
     /// </summary>
-    public class TagHelperAttribute
+    public class TagHelperAttribute : IHtmlContentContainer
```
dougbu: Should be no need for this class to implement `IHtmlContentContainer`. `TagHelperOutput` knows what this class exposes.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -2,46 +2,50 @@
...
         }
 
         /// <summary>
         /// Instantiates a new instance of <see cref="TagHelperAttribute"/> with the specified <paramref name="name"/>
-        /// and <paramref name="value"/>. <see cref="Minimized"/> is set to <c>false</c>.
+        /// and <paramref name="value"/>. <see cref="Structure"/> is set to <see cref="HtmlAttributeStructure.DoubleQuotedValue"/>.
         /// </summary>
         /// <param name="name">The <see cref="Name"/> of the attribute.</param>
         /// <param name="value">The <see cref="Value"/> of the attribute.</param>
         public TagHelperAttribute(string name, object value)
```
dougbu: Suggest we break compat, remove this method, and force users to think about the structure -- especially if their `value` includes `HtmlString`s. Thoughts @NTaylorMullen @Eilon?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -77,8 +79,130 @@ public bool Equals(TagHelperAttribute other)
...
+        }
+
+        /// <inheritdoc />
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder) => WriteTo(Name, Value, Structure, writer, encoder);
+
+        /// <inheritdoc />
+        public void CopyTo(IHtmlContentBuilder destination) => CopyTo(Name, Value, Structure, destination);
+
+        /// <inheritdoc />
+        public void MoveTo(IHtmlContentBuilder destination) => CopyTo(destination);
```
dougbu: This is incorrect. In any case, no reason to implement `MoveTo()` since `TagHelperOutput` will throw the `TagHelperAttribute`s away in the `MoveTo()` case.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -77,8 +79,130 @@ public bool Equals(TagHelperAttribute other)
...
+        /// </summary>
+        /// <param name="name"></param>
+        /// <param name="value"></param>
+        /// <param name="structure"></param>
+        /// <param name="destination"></param>
+        public static void CopyTo(
+            string name,
+            object value,
+            HtmlAttributeStructure structure,
+            IHtmlContentBuilder destination)
```
dougbu: Do not add these `public static` methods. This is an immutable `class` that knows how to write _itself_ out.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -77,8 +79,130 @@ public bool Equals(TagHelperAttribute other)
...
+                return;
+            }
+
+            var valuePrefix = GetAttributeValuePrefix(structure);
+            if (valuePrefix != null)
+            {
+                destination.AppendHtml(valuePrefix);
+            }
+
+            var htmlContent = value as IHtmlContent;
```
dougbu: Must special case `IHtmlContentContainer` here. Have a look at `TagHelperContent.CopyToCore()`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -95,9 +219,37 @@ public override int GetHashCode()
             var hashCodeCombiner = HashCodeCombiner.Start();
             hashCodeCombiner.Add(Name, StringComparer.Ordinal);
             hashCodeCombiner.Add(Value);
```
dougbu: Minor but since `Equals` ignores `Value` if `Structure == HtmlAttributeStructure.Minimized`, so should `GetHashCode()`.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: On second thought, if we do need `MoveTo()` for realz, this is fine.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: Hmm, maybe we do need `MoveTo()` to handle a `Value` that implements `IHtmlContentContainer`. Previous `CopyAttributesTo()` code did not do that but I don't remember why. @rynowak ?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperOutput.cs
```diff
@@ -305,7 +306,13 @@ void IHtmlContentContainer.CopyTo(IHtmlContentBuilder destination)
                 destination.AppendHtml("<");
                 destination.AppendHtml(TagName);
 
-                CopyAttributesTo(destination);
+                // Perf: Avoid allocating enumerator
+                for (var i = 0; i < (Attributes.Count); i++)
```
dougbu: Ick what are the extra parentheses for?

src/Microsoft.AspNetCore.Razor/Chunks/TagHelperAttributeChunk.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Chunks
+{
+    public class TagHelperAttributeChunk : Chunk
```
dougbu: Suggest this should be an immutable `struct` a la `KeyValuePair`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -450,112 +449,116 @@ protected virtual string GenerateUniqueId()
...
+                if (attribute.Value != null)
                 {
-                    _bodyVisitor.Accept(attributeValueChunk);
+                    _bodyVisitor.Accept(attribute.Value);
                 }
 
                 return;
             }
 
-            // If we have a minimized attribute there is no value
```
dougbu: Please restore the order and condition priority of the old code. Very hard to see what's changing here or why the new order is any better than the old. If the new order is better, it's not related to tracking quotes. So, make the change in a separate PR.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -450,112 +449,116 @@ protected virtual string GenerateUniqueId()
...
+                    _tagHelperContext.EndAddHtmlAttributeValuesMethodName,
+                    ExecutionContextVariableName);
 
-                    _writer.WriteMethodInvocation(
-                        _tagHelperContext.EndAddHtmlAttributeValuesMethodName,
-                        ExecutionContextVariableName);
+                _context.TargetWriterName = currentTargetWriter;
+            }
+            else if (attribute.Structure == HtmlAttributeStructure.Minimized ||
+                TryGetPlainTextValue(attribute.Value, out textValue))
```
dougbu: Not a fan of generating more complex code for the simple `attribute.Structure == HtmlAttributeStructure.Minimized` case.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -105,16 +106,16 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
...
                 // If there's no descriptors associated or we're hitting a bound attribute a second time.
-                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Key))
+                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Name))
                 {
                     string preAllocatedAttributeVariableName = null;
 
-                    if (attribute.Value == null)
+                    if (attribute.Structure == HtmlAttributeStructure.Minimized)
                     {
-                        var preAllocatedAttributeKey = new TagHelperAttributeKey(attribute.Key, value: null);
```
dougbu: Might want a `Debug.Assert()` about here confirming `attribute.Value == null` and another confirming the opposite in the `else` block. For an `AttributeValueChunk`, `attribute.Structure == HtmlAttributeStructure.Minimized` and `attribute.Value == null` are redundant.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperAttributeNode.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using Microsoft.AspNetCore.Razor.Parser.SyntaxTree;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Parser.TagHelpers
+{
+    public class TagHelperAttributeNode
+    {
+        public TagHelperAttributeNode(string name, SyntaxTreeNode value)
+            : this(name, value, HtmlAttributeStructure.DoubleQuotedValue)
```
dougbu: Yes to a `TagHelperAttributeNode(string name)` overload; no to this one. Whenever there's an attribute value, structure choice should be explicit.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -347,9 +366,31 @@ public static class TagHelperBlockRewriter
...
+                    default:
+                        result.AttributeStructure = HtmlAttributeStructure.Minimized;
+                        break;
+                }
+            }
+
             // Remove first child i.e. foo="
             builder.Children.RemoveAt(0);
 
+
```
dougbu: Nah.

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/DuplicateAttributeTagHelpers.DesignTime.lineMappings.cs
```diff
@@ -0,0 +1,37 @@
+// !!! Do not check in. Instead paste content into test method. !!!
```
dougbu: :smiling_imp:

dougbu:
Hoping a rebase will make the changes to `TagHelperBlockRewriterTest` small enough to view on GitHub...

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
NTaylorMullen: Structure is more first class now and always provided, doesn't make sense to keep this.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: Its an interesting thought. I think I lean towards not forcing users to provide the structure because 90%+ of the time it'll be double quoted value.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
NTaylorMullen: `TagHelperAttribute`s are already a decently large portion of a `TagHelper` pages allocations. I'd like to avoid re-allocating new `TagHelperAttribute`s per your comment [here](https://github.com/aspnet/Mvc/pull/4587#discussion_r63645382).

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Could we sit down in person and review this piece of code? I'd hate to undo it and lose the optimized structure I built here. I do understand that it's super hard to follow in diff format as it is right now.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperAttributeNode.cs
NTaylorMullen: Will mark this as internal since it's used extensively in testing.

NTaylorMullen:
:up: :date: Addressed most of the comments.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: Let's talk offline.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Let's talk offline.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
NTaylorMullen: Done

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: As discussed offline, this method really has no reason for being anymore. Minimized tag helper attributes are always handled using the preallocated code path.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: Talked offline. Let's leave this as you've got it.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: Spoke w/ @rynowak and we did leave out `MoveTo()` because it's a niche of a corner case here. But there's no harm in being correct nor of handling that case.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: As discussed offline, a couple of these cases (minimized attributes and plain text attributes) fold into the preallocated case. That will allow considerable simplification of these blocks.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperExecutionContext.cs
dougbu: Note the baseline files made _zero_ use of this method: The code gen path using it was never entered.

src/Microsoft.AspNetCore.Razor/TagHelpers/HtmlAttributeStructure.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.TagHelpers
+{
+    public enum HtmlAttributeStructure
```
dougbu: When talking offline to @Eilon, I learned the word "structure" isn't helpful here. Also don't see much added value in "HTML" or "Attribute" given tag helper authors only see this type when constructing a `TagHelperAttribute` or reading the property. Suggest something between `AttributeValueQuotingStyle` and `Quoting` or `ValueStyle`.

Eilon: `HtmlAttributeValueQuotes.DoubleQuote/SingleQuote/None`?

NTaylorMullen: Talked offline, `HtmlAttributeValueStyle`

Eilon: Values:
* DoubleQuotes
* SingleQuotes
* NoQuotes
* Minimized

?

NTaylorMullen:
:up: :date: with HtmlAttributeValueStyle name change.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -64,11 +68,9 @@ public TagHelperAttribute(string name, object value, bool minimized)
         public object Value { get; }
 
         /// <summary>
-        /// Gets an indication whether the attribute is minimized or not.
+        /// TODO
```
dougbu: Ah, _that's_ why this is still a `[Design]` PR...

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -450,112 +449,90 @@ protected virtual string GenerateUniqueId()
...
 
-        private void RenderUnboundAttribute(string attributeName, Chunk attributeValueChunk)
+        private void RenderUnboundAttribute(TagHelperAttributeTracker attribute)
         {
             // Render children to provide IntelliSense at design time. No need for the execution context logic, it's
             // a runtime feature.
             if (_designTimeMode)
             {
-                if (attributeValueChunk != null)
+                if (attribute.Value != null)
```
dougbu: Does the `attribute.Value == null` case still exist at design time (even though it does not at runtime)? Put another way,

1. Should the `_bodyVisitor.Accept()` call be unconditional?
2. Should the (missing) assertion that `attribute.Value` is non-`null` before or after the `_designTimeMode` check?

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -161,6 +165,11 @@ public static class TagHelperBlockRewriter
             var capturedAttributeValueStart = false;
             var attributeValueStartLocation = span.Start;
 
+            // Default to DoubleQuotes. We purposefully do not capture NoQuotes at parse time to protect
+            // users from rendering dynamic content with spaces that can break attributes.
+            // Ex: <a my-attribute=@value /> where @value results in the test "hello world".
```
dougbu: :rocket: useful comment :rocket:

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -229,6 +238,15 @@ public static class TagHelperBlockRewriter
                     // Check for attribute start values, aka single or double quote
                     if (i < htmlSymbols.Length && IsQuote(htmlSymbols[i]))
                     {
+                        if (htmlSymbols[i].Type == HtmlSymbolType.DoubleQuote)
+                        {
+                            attributeValueStyle = HtmlAttributeValueStyle.DoubleQuotes;
```
dougbu: Why bother i.e. why not make changes only in the single quotes and minimized cases?

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -347,6 +370,28 @@ public static class TagHelperBlockRewriter
...
 
+            var firstChild = builder.Children[0] as Span;
+            if (firstChild != null && firstChild.Symbols[0] is HtmlSymbol)
+            {
+                var htmlSymbol = firstChild.Symbols[firstChild.Symbols.Count - 1] as HtmlSymbol;
+                switch (htmlSymbol.Type)
+                {
+                    // We purposefully do not capture NoQuotes at parse time to protect users from rendering
+                    // dynamic content with spaces that can break attributes.
+                    // Ex: <a my-attribute=@value /> where @value results in the test "hello world".
```
dougbu: This is the same example as in the above comment. The example likely matches this case. If that's correct, the earlier comment should say something about _not_ special-casing `my-attribute=literal-value` for consistency and for simplifying later code.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -25,8 +25,8 @@ public bool Equals(TagHelperAttribute attributeX, TagHelperAttribute attributeY)
             // Normal comparer (TagHelperAttribute.Equals()) doesn't care about the Name case, in tests we do.
```
dougbu: Old code but... shouldn't this class also "care" about the `Name` case in `GetHashCode()`?

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperExecutionContextTest.cs
```diff
@@ -448,7 +448,7 @@ public void AddMinimizedHtmlAttribute_MaintainsHtmlAttributes()
         }
 
         [Fact]
-        public void AddMinimizedHtmlAttribute_MaintainsHtmlAttributes_SomeMinimized()
+        public void AddHtmlAttribute_MaintainsHtmlAttributes_SomeMinimized()
```
dougbu: Suggest checking `SingleQuotes` too and naming this `..._VariousStyles`.

dougbu: Generally need a bit more coverage of `SingleQuotes` in this test class.

dougbu:
Slight preference: Leave off `, HtmlAttributeValueStyle.DoubleQuotes` wherever that is possible. Leaves reader wondering what's special about _some_ double quotes.

(Don't seem to be able to add comments in `TagHelperBlockRewriterTest`.)

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.AspNetCore.Razor.Text;
 using Xunit;
 using Microsoft.AspNetCore.Razor.Parser.TagHelpers.Internal;
+using Microsoft.AspNetCore.Razor.Parser.TagHelpers;
 
 namespace Microsoft.AspNetCore.Razor.Test.TagHelpers
```
dougbu: Should add a couple of tests involving `HtmlAttributeValueStyle.SingleQuotes` in this class.

src/Microsoft.AspNetCore.Razor/TagHelpers/HtmlAttributeValueStyle.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.TagHelpers
+{
+    public enum HtmlAttributeValueStyle
+    {
+        DoubleQuotes,
+        SingleQuotes,
```
dougbu: `SingleQuotes` appears only 15 times in this entire PR. In addition to needing more coverage in a couple of test classes, this suggests the generation tests should be expanded to cover more single-quoted cases. E.g. I don't think preallocated single-quoted attributes appear more than once.

dougbu:
:watch: mostly for a few more tests.

dougbu:
>                                 TagMode.StartTagAndEndTag,

Testing comment publication in this large file again.

---
Refers to: test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs:3843 in 818930d. [](commit_id = 818930df6e8972b34e01510891e1c0f8a7f61235, deletion_comment = False)

dougbu:
>                                     new TagHelperAttributeNode("bound-int", null, HtmlAttributeValueStyle.Minimized),

Hmm, that was the previous iteration and comment didn't use `code` or _emphasis_.

---
Refers to: test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperBlockRewriterTest.cs:3846 in 0c50d15. [](commit_id = 0c50d15a6821a185e5d497d8fc372f806c1a1905, deletion_comment = False)

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Check still needs to exist here because pre-allocation logic is for runtime. At design time we don't do any of the work to determine if something needs to be pre-allocated for runtime (doesn't have any value).

Will add a != null check after the `_designTimeMode` statement

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Neeed to just do single quotes here. Minimized has to be handled below. Can nuke the double quote part though :smile:

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
NTaylorMullen: That's the responsibility of the `TagHelperBlockRewriter`. There's tests in there for all the various quotes.

src/Microsoft.AspNetCore.Razor/TagHelpers/HtmlAttributeValueStyle.cs
NTaylorMullen: Lots of coverage in `TagHelperBlockRewriterTest`, GitHub is just face palming on the diff size :cry:

NTaylorMullen:
:up: :date: no longer in design.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -376,9 +373,11 @@ public static class TagHelperBlockRewriter
                 var htmlSymbol = firstChild.Symbols[firstChild.Symbols.Count - 1] as HtmlSymbol;
                 switch (htmlSymbol.Type)
                 {
-                    // We purposefully do not capture NoQuotes at parse time to protect users from rendering
-                    // dynamic content with spaces that can break attributes.
-                    // Ex: <a my-attribute=@value /> where @value results in the test "hello world".
+                    // Treat NoQuotes and DoubleQuotes equivalently. We purposefully do not persist NoQuotes
+                    // ValueStyles at code generation time to protect users from rendering dynamic content with spaces
+                    // that can break attributes.
+                    // Ex: <tag my-attribute=@value /> where @value results in the test "hello world".
```
dougbu: My earlier comment was that `<tag my-attribute=@value />` example can't be applicable both here and at [line 170](https://github.com/aspnet/Razor/pull/743/commits/e0139a4ed9c5c62d6e4c422d17aa13db8406996f#diff-72fec5f339f6e3ee604fc27d0fc6fd95R170). So what example applies here? Or, if this is for consistency with the earlier code, say that.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
```diff
@@ -347,6 +367,30 @@ public static class TagHelperBlockRewriter
...
+            var firstChild = builder.Children[0] as Span;
+            if (firstChild != null && firstChild.Symbols[0] is HtmlSymbol)
+            {
+                var htmlSymbol = firstChild.Symbols[firstChild.Symbols.Count - 1] as HtmlSymbol;
+                switch (htmlSymbol.Type)
+                {
+                    // Treat NoQuotes and DoubleQuotes equivalently. We purposefully do not persist NoQuotes
+                    // ValueStyles at code generation time to protect users from rendering dynamic content with spaces
+                    // that can break attributes.
+                    // Ex: <tag my-attribute=@value /> where @value results in the test "hello world".
```
dougbu: Sorry for the one-commit comment.

My earlier comment was that `<tag my-attribute=@value />` example can't be applicable both here and at [line 170](https://github.com/aspnet/Razor/pull/743/files#diff-72fec5f339f6e3ee604fc27d0fc6fd95R170). So what example applies here? Or, if this is for consistency with the earlier code, say that.

src/Microsoft.AspNetCore.Razor/Chunks/TagHelperAttributeChunk.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Chunks
+{
+    public struct TagHelperAttributeTracker
```
dougbu: Add `/// docs`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
dougbu: thx

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperAttributeNode.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Parser.SyntaxTree;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Parser.TagHelpers
+{
+    public class TagHelperAttributeNode
```
dougbu: Add `/// docs`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -450,112 +449,92 @@ protected virtual string GenerateUniqueId()
...
-                        _tagHelperContext.ExecutionContextAddMinimizedHtmlAttributeMethodName)
-                    .WriteStringLiteral(attributeName)
-                    .WriteEndMethodInvocation();
-            }
-            else if (attributeValueChunk is PreallocatedTagHelperAttributeChunk)
+            Debug.Assert(attribute.Value != null);
+
+            var attributeValueStyleParameter = $"global::{typeof(HtmlAttributeValueStyle).FullName}.{attribute.ValueStyle}";
+
+            // All simple text attributes will be pre-allocated.
```
dougbu: Won't minimized attributes also be pre-allocated?

Separately, does `TagHelperBlockRewriter.ReWrite()` ensure preallocated attributes won't have value style `NoQuotes`?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -450,112 +449,92 @@ protected virtual string GenerateUniqueId()
...
-                }
-                else
-                {
-                    // HTML attributes are always strings. This attribute contains C# but is not dynamic. This occurs
-                    // when the attribute is a data-* attribute.
+                _context.TargetWriterName = currentTargetWriter;
+            }
+            else
+            {
+                // HTML attributes are always strings. This attribute contains C# but is not dynamic. This occurs
```
dougbu: "contains C# but is not dynamic"? This case is very similar to the `IsDynamicAttributeValue()` case and fits perfectly into the usual notion of "dynamic".

The difference is that attribute removal isn't done for `data-` attributes. Therefore I suggest something like `This is a data-* attribute which includes C#. Do not perform the conditional attribute removal or other special cases used when IsDynamicAttributeValue(). But the attribute must still be buffered to determine its final value.`

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/CaseSensitiveTagHelperAttributeComparer.cs
dougbu: thx

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperParseTreeRewriterTest.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor/TagHelpers/HtmlAttributeValueStyle.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
```diff
@@ -77,8 +79,147 @@ public bool Equals(TagHelperAttribute other)
...
                 other != null &&
                 string.Equals(Name, other.Name, StringComparison.OrdinalIgnoreCase) &&
-                Minimized == other.Minimized &&
-                (Minimized || Equals(Value, other.Value));
+                ValueStyle == other.ValueStyle &&
+                (ValueStyle == HtmlAttributeValueStyle.Minimized || Equals(Value, other.Value));
+        }
+
+        /// <inheritdoc />
+        public void WriteTo(TextWriter writer, HtmlEncoder encoder)
```
dougbu: We have very little unit test coverage of `WriteTo()`, `CopyTo()` and `MoveTo()` except in `TagHelperOutputTest`. That class does not check any value styles other than `DoubleQuotes`. Need more coverage, at least a few more cases in the `WriteTagHelper_InputData` data set.

dougbu:
:watch: mostly for new doc comments and a few more tests.

src/Microsoft.AspNetCore.Razor/Chunks/TagHelperAttributeChunk.cs
Eilon: This is an obscure type in an even more obscure namespace. Who would read the comment?

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: Hu? Both comments apply to the same thing, one is for dealing with attributes that are built as a single span, the other is built for attributes that consist of multiple spans. In both approaches we need to detect quotage :smile: yes, just made up that awesome wordage.

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlockRewriter.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/743#discussion_r64822037

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: > Won't minimized attributes also be pre-allocated?

Yup, I'll be more explicit and state minimized instead of grouping the two as simple text.



> Separately, does TagHelperBlockRewriter.ReWrite() ensure preallocated attributes won't have value style NoQuotes?

Yup

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor/Chunks/TagHelperAttributeChunk.cs
dougbu: :ok:

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperAttributeNode.cs
dougbu: :see_no_evil: based on https://github.com/aspnet/Razor/pull/743#discussion-diff-64803618

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperAttribute.cs
dougbu: thx

dougbu:
:shipit:

PR #753, opened 2016-05-17
Fix TagHelper directive tests to fail when expected.

NTaylorMullen:
- Prior to this change the equals bits were falling back to the type comparison that `ChunkGenerator` exposes.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Razor.Parser.SyntaxTree;
+using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.Chunks.Generators
 {
@@ -34,6 +36,24 @@ namespace Microsoft.AspNetCore.Razor.Chunks.Generators
         public override void GenerateChunk(Span target, ChunkGeneratorContext context)
         {
             context.ChunkTreeBuilder.AddAddTagHelperChunk(_lookupText, target);
+        }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            var other = obj as AddTagHelperChunkGenerator;
+            return base.Equals(other) &&
+                string.Equals(_lookupText, other._lookupText, StringComparison.Ordinal);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            var combiner = HashCodeCombiner.Start();
+            combiner.Add(base.GetHashCode());
+            combiner.Add(_lookupText, StringComparer.Ordinal);
+
+            return combiner.CombinedHash;
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
```diff
@@ -35,5 +37,23 @@ public override void GenerateChunk(Span target, ChunkGeneratorContext context)
...
+            var other = obj as AddTagHelperChunkGenerator;
+            return base.Equals(obj) &&
+                string.Equals(_lookupText, other._lookupText, StringComparison.Ordinal);
+        }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            var combiner = HashCodeCombiner.Start();
+            combiner.Add(base.GetHashCode());
```
ajaybhargavb: Why is this needed?

src/Microsoft.AspNetCore.Razor/Chunks/Generators/RemoveTagHelperChunkGenerator.cs
```diff
@@ -35,5 +37,23 @@ public override void GenerateChunk(Span target, ChunkGeneratorContext context)
         {
             context.ChunkTreeBuilder.AddRemoveTagHelperChunk(_lookupText, target);
         }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            var other = obj as RemoveTagHelperChunkGenerator;
+            return base.Equals(obj) &&
```
ajaybhargavb: Wouldn't checking `other != null` be sufficient?

src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
```diff
@@ -35,5 +37,23 @@ public override void GenerateChunk(Span target, ChunkGeneratorContext context)
         {
             context.ChunkTreeBuilder.AddAddTagHelperChunk(_lookupText, target);
         }
+
+        /// <inheritdoc />
+        public override bool Equals(object obj)
+        {
+            var other = obj as AddTagHelperChunkGenerator;
+            return base.Equals(obj) &&
+                string.Equals(_lookupText, other._lookupText, StringComparison.Ordinal);
```
pranavkm: Couldn't `other` be null here?

NTaylorMullen: the base.Equals ensures it's not `null`.

src/Microsoft.AspNetCore.Razor/Chunks/Generators/RemoveTagHelperChunkGenerator.cs
NTaylorMullen: It would be functionally equivalent to what it is currently but by calling down to base we ensure that any sub-api equals changes are captured here.

src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/753#discussion_r63648995

src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
pranavkm: In the event `obj` is a `SpanChunkGenerator`  but not a `AddTagHelperChunkGenerator` this might be null. Change the first line to `base.Equals(other)` and you should be good.

NTaylorMullen: ooo :+1: good catch

dougbu: This is fine due to the `GetType()` comparison in `base.Equals()`. But obviously it is harder to read than necessary.

NTaylorMullen:
:up: :date:

ajaybhargavb:
:shipit:

pranavkm:
:shipit:

Issue #744, opened 2016-05-05
TagHelper directives don't work with optional quotes in non-import files


ryanbrandenburg:
`@addTagHelper *, Location` will work in any location, and `@addTagHelper "*, Location"` will work in an _ImportView.cshtml, but if you try to do `@addTagHelper "*, Location"` in for example Index.cshtml you'll get an error saying "The given assembly name or codebase was invalid"

To reproduce, create a new Web project, copy the `@addTagHelper` directive out of _ViewImports.cshtml into Index.cshtml and change it to `@addTagHelper "*, Microsoft.AspNetCore.Mvc.TagHelpers"`.

Eilon:
@DamianEdwards FYI moving this to RTM.

NTaylorMullen:
80ebb4a06838123599fa8132a850a141e9fc80b3

PR #757, opened 2016-05-18
Update `TagHelperDirectiveSpanVisitor` to properly capture TagHelper directive text.

NTaylorMullen:
- Added a test with optional quotes for each of the `TagHelperDirectiveDescriptor`s, add, remove and prefix.

#744

/cc @pranavkm @ajaybhargavb

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
@@ -12,8 +12,6 @@ namespace Microsoft.AspNetCore.Razor.Chunks.Generators
     /// </summary>
     public class AddTagHelperChunkGenerator : SpanChunkGenerator
     {
-        private readonly string _lookupText;
-
         /// <summary>
         /// Initializes a new instance of <see cref="AddTagHelperChunkGenerator"/>.
         /// </summary>
@@ -22,8 +20,13 @@ namespace Microsoft.AspNetCore.Razor.Chunks.Generators
         /// </param>
         public AddTagHelperChunkGenerator(string lookupText)
         {
-            _lookupText = lookupText;
+            LookupText = lookupText;
         }
+
+        /// <summary>
+        /// Gets the text used to look up <see cref="Compilation.TagHelpers.TagHelperDescriptor"/>s that should be added.
+        /// </summary>
+        public string LookupText { get; }
 
         /// <summary>
         /// Generates <see cref="AddTagHelperChunk"/>s.
@@ -35,7 +38,7 @@ namespace Microsoft.AspNetCore.Razor.Chunks.Generators
         /// the current chunk generation process.</param>
         public override void GenerateChunk(Span target, ChunkGeneratorContext context)
         {
-            context.ChunkTreeBuilder.AddAddTagHelperChunk(_lookupText, target);
+            context.ChunkTreeBuilder.AddAddTagHelperChunk(LookupText, target);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -87,26 +87,35 @@ public override void VisitSpan(Span span)
                 throw new ArgumentNullException(nameof(span));
             }
 
+            string directiveText = null;
             TagHelperDirectiveType directiveType;
-            if (span.ChunkGenerator is AddTagHelperChunkGenerator)
+
+            var addTagHelperChunkGenerator = span.ChunkGenerator as AddTagHelperChunkGenerator;
+            var removeTagHelperChunkGenerator = span.ChunkGenerator as RemoveTagHelperChunkGenerator;
+            var tagHelperPrefixChunkGenerator = span.ChunkGenerator as TagHelperPrefixDirectiveChunkGenerator;
```
dougbu: Most other places we use multiple conditional casts inline the casts into the conditions. That way only the needed casts execute.

src/Microsoft.AspNetCore.Razor/Chunks/Generators/AddTagHelperChunkGenerator.cs
```diff
@@ -22,10 +20,15 @@ public class AddTagHelperChunkGenerator : SpanChunkGenerator
         /// </param>
         public AddTagHelperChunkGenerator(string lookupText)
         {
-            _lookupText = lookupText;
+            LookupText = lookupText;
         }
 
         /// <summary>
+        /// Text used to look up <see cref="Compilation.TagHelpers.TagHelperDescriptor"/>s that should be added.
```
pranavkm: > Gets the ...

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -87,26 +87,35 @@ public override void VisitSpan(Span span)
...
+
+            if (addTagHelperChunkGenerator != null)
```
pranavkm: Was there a reason to switch it from the `If (generator is FooGenerator)`?

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
```diff
@@ -87,26 +87,35 @@ public override void VisitSpan(Span span)
                 throw new ArgumentNullException(nameof(span));
             }
 
+            string directiveText = null;
```
pranavkm: I think `string directiveText;` should suffice since it always gets intiailized

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperDirectiveSpanVisitor.cs
NTaylorMullen: Casting anyways, thought might as well just use the as syntax instead of `is` => `((AddTagHelperChunkGenerator)span.ChunkGenerator).LookupText`

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperDirectiveSpanVisitorTest.cs
```diff
@@ -18,6 +19,109 @@ public class TagHelperDirectiveSpanVisitorTest
     {
         private static readonly SpanFactory Factory = SpanFactory.CreateCsHtml();
 
+        public static TheoryData QuotedTagHelperDirectivesData
```
pranavkm: Could you add a variant with spaces in the start - since we trim those too?

pranavkm:
:shipit:

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/TagHelperDirectiveSpanVisitorTest.cs
NTaylorMullen: Talked offline, already have tests for this :smile:

PR #764, opened 2016-05-25
[Perf] Avoid ViewBuffers for writing bound TagHelper attribute values

### LINKED ISSUE #717: Write Bound String Attributes for taghelpers using a StringWriter on RazorPage
Currently writing the text for a bound TagHelper string attribute uses `BeginTagHelperWritingScope`/`EndTagHelperWritingScope`. This isn't efficient for a small amount of content, and it gets `ToString`ed anyway.

We should make this use similar infrastructure to `BeginWriteAttributeTo` which uses a cached `StringWriter` as part of the page.

mnltejaswini:
Fixes #717 
This fix has dependency on https://github.com/aspnet/Mvc/pull/4732 
The scenario is https://github.com/aspnet/Mvc/issues/4593 

Currently, writing TagHelper attribute uses the same code as the TagHelper content writing. This uses StartTagHelperWritingScope and EndTagHelperWritingScope which is expensive for writing attributes as these methods uses ViewBuffer, ViewBufferTextWriter. We an use simple StringWriter for writing attributes. Also, as attribute writing is not nested, we can use a shared writer within the RazorPage. The RazorPage has two new methods special cased for writing TagHelper attributes using the shared StringWriter. Razor uses these methods to determine the attribute values.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
@@ -29,6 +29,8 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             MarkAsHtmlEncodedMethodName = "Html.Raw";
             StartTagHelperWritingScopeMethodName = "StartTagHelperWritingScope";
             EndTagHelperWritingScopeMethodName = "EndTagHelperWritingScope";
+            BeginWriteTagHelperAttributeMethodName = "BeginWriteTagHelperAttribute";
+            EndWriteTagHelperAttributeMethodName = "EndWriteTagHelperAttribute";
             RunnerTypeName = "Microsoft.AspNetCore.Razor.Runtime.TagHelperRunner";
             ScopeManagerTypeName = "Microsoft.AspNetCore.Razor.Runtime.TagHelperScopeManager";
             ExecutionContextTypeName = "Microsoft.AspNetCore.Razor.Runtime.TagHelperExecutionContext";
@@ -161,6 +163,16 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
         /// The name of the method used to end a writing scope.
         /// </summary>
         public string EndTagHelperWritingScopeMethodName { get; set; }
+
+        /// <summary>
+        /// The name of the method used to begin an attribute writing scope.
+        /// </summary>
+        public string BeginWriteTagHelperAttributeMethodName { get; set; }
+
+        /// <summary>
+        /// The name of the method used to end an attribute writing scope.
+        /// </summary>
+        public string EndWriteTagHelperAttributeMethodName { get; set; }
 
         /// <summary>
         /// The name of the type used to run tag helpers.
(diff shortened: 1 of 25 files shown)


mnltejaswini:
@NTaylorMullen

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -29,6 +29,8 @@ public GeneratedTagHelperContext()
             MarkAsHtmlEncodedMethodName = "Html.Raw";
             StartTagHelperWritingScopeMethodName = "StartTagHelperWritingScope";
             EndTagHelperWritingScopeMethodName = "EndTagHelperWritingScope";
+            BeginWriteTagHelperAttributeMethodName = "BeginWriteTagHelperAttribute";
```
NTaylorMullen: Add these setters to the MvcRazorHost. [Example](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs#L107-L108)

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -53,8 +53,10 @@ protected override void Visit(TagHelperChunk chunk)
                     // Need to disable the warning "X is assigned to but never used." for the value buffer since
                     // whether it's used depends on how a TagHelper is used.
                     Writer.WritePragma("warning disable 0414");
-                    WritePrivateField(
-                        _tagHelperContext.TagHelperContentTypeName,
+                    Writer
+                        .Write("private ")
+                        .WriteVariableDeclaration(
+                        "string",
```
NTaylorMullen: nit: Indent the parameters so it's clear they're being passed into the WriteVariableDeclartion method.

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/NestedScriptTagTagHelpers.cs
```diff
@@ -57,16 +57,16 @@ public NestedScriptTagTagHelpers()
...
 #line 8 "NestedScriptTagTagHelpers.cshtml"
                                             Write(ViewBag.DefaultInterval);
 
 #line default
 #line hidden
                 WriteLiteral(" + 1");
-                __tagHelperStringValueBuffer = EndTagHelperWritingScope();
-                __tagHelperExecutionContext.AddHtmlAttribute("data-interval", Html.Raw(__tagHelperStringValueBuffer.GetContent(HtmlEncoder)));
+                __tagHelperStringValueBuffer = EndWriteTagHelperAttribute();
+                __tagHelperExecutionContext.AddHtmlAttribute("data-interval", Html.Raw(__tagHelperStringValueBuffer));
```
NTaylorMullen: Q: Do `HtmlString`s like these (`Html.Raw`) have a significant impact in the profile?

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
mnltejaswini: It is strange. They are already indented using the default VS settings.

NTaylorMullen:
:shipit:

PR #765, opened 2016-05-27
[Perf] Preallocate TagHelperAttributes for bound string typed attributes

### LINKED ISSUE #704: Pre-allocate TagHelper attributes that are String typed, bound but not complex.
In issue #600 we did the work to pre-allocate unbound non-complex `TagHelper` attributes. We could do the same work for bound attributes that are of type `string` and are not complex. This would save us the extra `TagHelperAttribute` allocation's that are used to fill `TagHelperContext.AllItems`.

For instance, in our `BasicViews` performance scenario there's only a single unbound `TagHelper` attribute but there are tons of bound, string based attributes.

mnltejaswini:
Fixes #704 
Scenario: https://github.com/aspnet/Mvc/issues/4593 

The number of TagHelperAttributes allocated  reduced by 50% before and after optimization. 
Before optimization
-------------------------
image

After optimization
---------------------
image

Total savings is number of bytes allocated is 1.3%

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
@@ -331,6 +331,15 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             // If this attribute value has not been seen before, need to record its value.
             if (previousValueAccessor == null)
             {
+                var preallocatedAttributeValueChunk = attribute.Value as PreallocatedTagHelperAttributeChunk;
+
+                if (preallocatedAttributeValueChunk != null)
+                {
+                    RenderBoundPreAllocatedAttribute(preallocatedAttributeValueChunk, currentValueAccessor);
+
+                    return currentValueAccessor;
+                }
+
                 // Bufferable attributes are attributes that can have Razor code inside of them. Such
                 // attributes have string values and may be calculated using a temporary TextWriter or other
                 // buffer.
@@ -374,6 +383,37 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
 
                 return previousValueAccessor;
             }
+        }
+
+        // Render bound attributes that are of string type and are preallocated.
+        private void RenderBoundPreAllocatedAttribute(
+            PreallocatedTagHelperAttributeChunk preallocatedAttributeValueChunk,
+            string valueAccessor)
+        {
+            var attributeValueAccessor = string.Format(
+                CultureInfo.InvariantCulture,
+                "{0}.{1}",
+                preallocatedAttributeValueChunk.AttributeVariableAccessor,
+                _tagHelperContext.TagHelperAttributeValuePropertyName);
+
+            _writer
+                .WriteStartAssignment(valueAccessor)
+                .Write("(string)")
+                .Write(attributeValueAccessor)
+                .WriteLine(";");
+
+            if (_designTimeMode)
+            {
+                // Execution contexts are a runtime feature.
+                return;
+            }
+
+            _writer
+                .WriteStartInstanceMethodInvocation(
+                    ExecutionContextVariableName,
+                    _tagHelperContext.ExecutionContextAddTagHelperAttributeMethodName)
+                .Write(preallocatedAttributeValueChunk.AttributeVariableAccessor)
+                .WriteEndMethodInvocation();
         }
 
         // Render assignment of attribute value to the value accessor.
@@ -735,7 +775,7 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             return false;
         }
 
-        private static bool IsDynamicAttributeValue(Chunk attributeValueChunk)
+        public static bool IsDynamicAttributeValue(Chunk attributeValueChunk)
         {
             var parentChunk = attributeValueChunk as ParentChunk;
             if (parentChunk != null)
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
@@ -41,6 +41,7 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
             ExecutionContextSetOutputContentAsyncMethodName = "SetOutputContentAsync";
             TagHelperAttributeTypeName = "Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute";
             EncodedHtmlStringTypeName = "Microsoft.AspNetCore.Html.HtmlString";
+            TagHelperAttributeValuePropertyName = "Value";
         }
 
         /// <summary>
@@ -229,5 +230,10 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
         /// The name of the type used to represent encoded content.
         /// </summary>
         public string EncodedHtmlStringTypeName { get; set; }
+
+        /// <summary>
+        /// The name of the Value property of <c>TagHelperAttribute</c>.
+        /// </summary>
+        public string TagHelperAttributeValuePropertyName { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
+++ b/src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.AspNetCore.Razor.Chunks;
+using Microsoft.AspNetCore.Razor.Compilation.TagHelpers;
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using Microsoft.Extensions.Internal;
 
@@ -95,24 +96,27 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
 
             if (!Context.Host.DesignTimeMode)
             {
-                PreAllocateUnboundTagHelperAttributes(chunk);
+                PreAllocateTagHelperAttributes(chunk);
             }
 
             // We need to dive deeper to ensure we pick up any nested tag helpers.
             Accept(chunk.Children);
         }
 
-        private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
+        private void PreAllocateTagHelperAttributes(TagHelperChunk chunk)
         {
             var boundAttributes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             for (var i = 0; i < chunk.Attributes.Count; i++)
             {
                 var attribute = chunk.Attributes[i];
-                var hasAssociatedDescriptors = chunk.Descriptors.Any(descriptor =>
-                    descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Name)));
+                var associatedAttributeDescriptors = chunk.Descriptors.SelectMany(descriptor => descriptor.Attributes)
+                    .Where(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Name));
 
                 // If there's no descriptors associated or we're hitting a bound attribute a second time.
-                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Name))
+                var isUnBoundAttribute = !associatedAttributeDescriptors.Any() || !boundAttributes.Add(attribute.Name);
+
+                // Perf: We will preallocate TagHelperAttribute for unbound attributes and simple bound string valued attributes.
+                if (isUnBoundAttribute || CanPreallocateBoundAttribute(associatedAttributeDescriptors, attribute))
                 {
                     string preAllocatedAttributeVariableName = null;
 
@@ -120,7 +124,11 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
                     {
                         Debug.Assert(attribute.Value == null);
 
-                        var preAllocatedAttributeKey = new TagHelperAttributeKey(attribute.Name, value: null, valueStyle: attribute.ValueStyle);
+                        var preAllocatedAttributeKey = new TagHelperAttributeKey(
+                            attribute.Name, 
+                            value: null, 
+                            unBoundAttribute: isUnBoundAttribute, 
+                            valueStyle: attribute.ValueStyle);
                         if (TryCachePreallocatedVariableName(preAllocatedAttributeKey, out preAllocatedAttributeVariableName))
                         {
                             Writer
@@ -141,7 +149,7 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
                         string plainText;
                         if (CSharpTagHelperCodeRenderer.TryGetPlainTextValue(attribute.Value, out plainText))
                         {
-                            var preAllocatedAttributeKey = new TagHelperAttributeKey(attribute.Name, plainText, attribute.ValueStyle);
+                            var preAllocatedAttributeKey = new TagHelperAttributeKey(attribute.Name, plainText, isUnBoundAttribute, attribute.ValueStyle);
                             if (TryCachePreallocatedVariableName(preAllocatedAttributeKey, out preAllocatedAttributeVariableName))
                             {
                                 Writer
@@ -152,10 +160,23 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
                                     .Write(" = ")
                                     .WriteStartNewObject("global::" + _tagHelperContext.TagHelperAttributeTypeName)
                                     .WriteStringLiteral(attribute.Name)
-                                    .WriteParameterSeparator()
-                                    .WriteStartNewObject("global::" + _tagHelperContext.EncodedHtmlStringTypeName)
-                                    .WriteStringLiteral(plainText)
-                                    .WriteEndMethodInvocation(endLine: false)
+                                    .WriteParameterSeparator();
+
+                                if (isUnBoundAttribute)
+                                {
+                                    // For unbound attributes, we need to create HtmlString.
+                                    Writer
+                                        .WriteStartNewObject("global::" + _tagHelperContext.EncodedHtmlStringTypeName)
+                                        .WriteStringLiteral(plainText)
+                                        .WriteEndMethodInvocation(endLine: false);
+                                }
+                                else
+                                {
+                                    Writer.WriteStringLiteral(plainText);
+                                    
+                                }
+
+                                Writer
                                     .WriteParameterSeparator()
                                     .Write($"global::{typeof(HtmlAttributeValueStyle).FullName}.{attribute.ValueStyle}")
                                     .WriteEndMethodInvocation();
@@ -175,6 +196,24 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
                     }
                 }
             }
+        }
+
+        private static bool CanPreallocateBoundAttribute(
+            IEnumerable<TagHelperAttributeDescriptor> associatedAttributeDescriptors,
+            TagHelperAttributeTracker attribute)
+        {
+            // If the attribute value is a Dynamic value, it cannot be preallocated.
+            if (CSharpTagHelperCodeRenderer.IsDynamicAttributeValue(attribute.Value))
+            {
+                return false;
+            }
+
+            // Only attributes that are associated with string typed properties can be preallocated.
+            var attributeName = attribute.Name;
+            var allStringProperties = associatedAttributeDescriptors
+                .All(attributeDescriptor => attributeDescriptor.IsStringProperty);
+
+            return allStringProperties;
         }
 
         public override void Accept(Chunk chunk)
@@ -224,16 +263,19 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
 
         private struct TagHelperAttributeKey : IEquatable<TagHelperAttributeKey>
         {
-            public TagHelperAttributeKey(string name, string value, HtmlAttributeValueStyle valueStyle)
+            public TagHelperAttributeKey(string name, string value, bool unBoundAttribute, HtmlAttributeValueStyle valueStyle)
             {
                 Name = name;
                 Value = value;
+                UnBoundAttribute = unBoundAttribute;
                 ValueStyle = valueStyle;
             }
 
             public string Name { get; }
 
             public string Value { get; }
+
+            public bool UnBoundAttribute { get; }
 
             public HtmlAttributeValueStyle ValueStyle { get; }
 
@@ -242,6 +284,7 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators.Visitors
                 var hashCodeCombiner = HashCodeCombiner.Start();
                 hashCodeCombiner.Add(Name, StringComparer.Ordinal);
                 hashCodeCombiner.Add(Value, StringComparer.Ordinal);
+                hashCodeCombiner.Add(UnBoundAttribute);
                 hashCodeCombiner.Add(ValueStyle);
 
                 return hashCodeCombiner.CombinedHash;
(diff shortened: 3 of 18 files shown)


mnltejaswini:
@NTaylorMullen  Once you validate the fix, I will regenerate the baselines for the tests to pass.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -287,6 +287,12 @@ protected virtual string GenerateUniqueId()
                         continue;
                     }
 
+                    if (attributeValueChunk is PreallocatedTagHelperAttributeChunk)
+                    {
+                        RenderPreallocatedBoundAttribute(attributeName, attributeValueChunk);
+                        continue;
```
dougbu: This means preallocated bound attributes are added to the `AllAttributes` collection but their values are never assigned to the associated tag helper properties. Suggest this special case needs to live inside `RenderNewAttributeValueAssignment()`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -287,6 +287,12 @@ protected virtual string GenerateUniqueId()
                         continue;
                     }
 
+                    if (attributeValueChunk is PreallocatedTagHelperAttributeChunk)
```
dougbu: Use `as` rather than `is` followed by an unconditional cast i.e. change the `RenderPreallocatedBoundAttribute()` parameter to be `PreallocatedTagHelperAttributeChunk`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
mnltejaswini: You are correct. Initially we thought it would be sufficient to just add the preallocated alltributes to the AllAttributes list and it is not need to set the associated tag helper properties. But after debugging, I found that we need to set the tag helper properties to the preallocated attribute values. I will make the change.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Any reason this whole bit can't be inside `RenderBoundAttribute`?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -559,6 +565,22 @@ private void RenderUnboundAttribute(string attributeName, Chunk attributeValueCh
             }
         }
 
+        private void RenderPreallocatedBoundAttribute(string attributeName, Chunk attributeValueChunk)
```
NTaylorMullen: Assuming [this](https://github.com/aspnet/Razor/pull/765/files#r64953504) is true. Inline this method.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -168,6 +188,23 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
             }
         }
 
+        private static bool IsLiteralAttributeValue(Chunk attributeValueChunk)
```
NTaylorMullen: Don't write your own, expose what's in the `CSharpTagHelperCodeRenderer` and use that.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -151,6 +156,21 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
                                     .WriteStringLiteral(plainText)
                                     .WriteEndMethodInvocation(endLine: false)
                                     .WriteEndMethodInvocation();
+                                }
+                                else
+                                {
+                                    Writer
```
NTaylorMullen: Only difference between this set of writes and the above is the value being surrounded with `Html.Raw`/`HtmlString` suggest unifying the pre and post logic and just having an if unbound for the value..

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
mnltejaswini: CSharpTagHelperCodeRender is checking for IsDynamicValue. Here we might need to check for IsLiteralChunk. So I added this method

NTaylorMullen: `!IsDynamicValue` is simple text though.

mnltejaswini:
:up::date:

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/PrefixedAttributeTagHelpers.Reversed.cs
```diff
@@ -220,15 +198,11 @@ public PrefixedAttributeTagHelpers()
...
-
-#line default
-#line hidden
-            __tagHelperExecutionContext.AddTagHelperAttribute("int-prefix-value", __TestNamespace_InputTagHelper2.IntDictionaryProperty["value"]);
+            __TestNamespace_InputTagHelper2.IntDictionaryProperty["value"] = __tagHelperAttribute_12.Value.ToString();
+            __tagHelperExecutionContext.AddTagHelperAttribute(__tagHelperAttribute_12);
             __TestNamespace_InputTagHelper1.IntDictionaryProperty["value"] = __TestNamespace_InputTagHelper2.IntDictionaryProperty["value"];
-            __TestNamespace_InputTagHelper2.StringDictionaryProperty["thyme"] = "string";
-            __tagHelperExecutionContext.AddTagHelperAttribute("string-prefix-thyme", __TestNamespace_InputTagHelper2.StringDictionaryProperty["thyme"]);
+            __TestNamespace_InputTagHelper2.StringDictionaryProperty["thyme"] = __tagHelperAttribute_13.Value.ToString();
```
NTaylorMullen: This will result in new unintended allocations.

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/PrefixedAttributeTagHelpers.Reversed.cs
```diff
@@ -220,15 +198,11 @@ public PrefixedAttributeTagHelpers()
             {
                 throw new InvalidOperationException(FormatInvalidIndexerAssignment("string-prefix-thyme", "TestNamespace.InputTagHelper1", "StringDictionaryProperty"));
             }
-#line 22 "PrefixedAttributeTagHelpers.cshtml"
-__TestNamespace_InputTagHelper2.IntDictionaryProperty["value"] = 37;
-
-#line default
-#line hidden
-            __tagHelperExecutionContext.AddTagHelperAttribute("int-prefix-value", __TestNamespace_InputTagHelper2.IntDictionaryProperty["value"]);
+            __TestNamespace_InputTagHelper2.IntDictionaryProperty["value"] = __tagHelperAttribute_12.Value.ToString();
```
NTaylorMullen: This is not equivalent to what was there before. It's ToString()ing an int when the value should actually be an int

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -377,13 +386,60 @@ protected virtual string GenerateUniqueId()
...
+        }
+
         // Render assignment of attribute value to the value accessor.
         private void RenderNewAttributeValueAssignment(
             TagHelperAttributeDescriptor attributeDescriptor,
             bool bufferableAttribute,
             Chunk attributeValueChunk,
             string valueAccessor)
         {
+            var preallocatedAttributeValueChunk = attributeValueChunk as PreallocatedTagHelperAttributeChunk;
```
NTaylorMullen: This bit of code is never called.

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
mnltejaswini: ok. I saw some other chunks like ExpressionChunks. Don't we need to worry about them?

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/PrefixedAttributeTagHelpers.Reversed.cs
mnltejaswini: If it a string values, there will be no allocations. But as you pointed in another comment, this is functionally incorrect if the value is other than string type. I will fix that.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
mnltejaswini: Sorry, my bad, this should have been cleaned up.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
NTaylorMullen: `ExpressionChunk`s are typically due to `@Foo` in an attribute. If you point me to where you're seeing them I can clarify further.

test/Microsoft.AspNetCore.Razor.Test/TestFiles/CodeGenerator/Output/PrefixedAttributeTagHelpers.Reversed.cs
mnltejaswini: For string typed, there will be no allocations of new strings when we do ToString()

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -377,6 +386,36 @@ protected virtual string GenerateUniqueId()
...
 
+        private void RenderPreallocatedAttributeValues(
+            PreallocatedTagHelperAttributeChunk preallocatedAttributeValueChunk,
+            string valueAccessor)
+        {
+            var attributeValueAccessor = string.Format(
+                    CultureInfo.InvariantCulture,
+                    "{0}.{1}",
+                    preallocatedAttributeValueChunk.AttributeVariableAccessor,
+                    _tagHelperContext.TagHelperAttributeValuePropertyName);
```
dougbu: De-indent

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -377,6 +386,36 @@ protected virtual string GenerateUniqueId()
...
+
+            _writer
+                .WriteStartAssignment(valueAccessor)
+                .WriteInstanceMethodInvocation(
+                    attributeValueAccessor,
+                    nameof(attributeValueAccessor.ToString));
```
dougbu: Perhaps more obvious to unconditionally cast `preallocatedAttribute.Value` to `string`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -235,5 +236,7 @@ public GeneratedTagHelperContext()
         /// The name of the type used to represent encoded content.
         /// </summary>
         public string EncodedHtmlStringTypeName { get; set; }
+
+        public string TagHelperAttributeValuePropertyName { get; set; }
```
dougbu: Add `/// docs`

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
mnltejaswini: Discussed about this with Taylor and understood that calling the method is preferable to casting

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -377,6 +386,36 @@ protected virtual string GenerateUniqueId()
             }
         }
 
+        private void RenderPreallocatedAttributeValues(
```
NTaylorMullen: `RenderPreAllocatedAttribute`

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
```diff
@@ -377,6 +386,36 @@ protected virtual string GenerateUniqueId()
...
+                preallocatedAttributeValueChunk.AttributeVariableAccessor,
+                _tagHelperContext.TagHelperAttributeValuePropertyName);
+
+            _writer
+                .WriteStartAssignment(valueAccessor)
+                .WriteInstanceMethodInvocation(
+                    attributeValueAccessor,
+                    nameof(attributeValueAccessor.ToString));
+
+            if (_designTimeMode)
```
NTaylorMullen: You could probably re-use the checks in the calling method instead of re-writing this code. Ultimately this method would just contain everything prior to this `if (_designTimeMode)`

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Add a comment stating when bound attributes are pre-allocated.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -168,6 +182,34 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
...
+            // If the attribute value is a Dynamic value, it cannot be preallocated.
+            if (CSharpTagHelperCodeRenderer.IsDynamicAttributeValue(attribute.Value))
+            {
+                return false;
+            }
+
+            // Only attributes that are associated with string typed properties can be preallocated.
+            var attributeName = attribute.Key;
+            var attributeValueChunk = attribute.Value;
+            var associatedDescriptors = tagHelperDescriptors.Where(descriptor =>
```
NTaylorMullen: You did this in the calling method. This will always result in true.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -168,6 +182,34 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
...
+                descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attributeName)));
+            foreach (var associatedDescriptor in associatedDescriptors)
```
NTaylorMullen: Once you do the above you can just do 
```C#
var allStringProperties = associatedDescriptors.All(attributeDescriptor => attributeDescriptor.IsStringProperty);
return allStringProperties;
```

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: I was just worried that this would allocate, but it doesn't as you clarified. I kind of like the casting approach now that I look at it. Ensures that the system is working end-to-end.

NTaylorMullen:
:watch:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
mnltejaswini: I did not understand. The code generated is completely different from the regular path and the only check I am doing is design time mode. Can you please clarify.

mnltejaswini:
:up::date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpTagHelperCodeRenderer.cs
NTaylorMullen: Oh shoot, you're write, scratch that. It's good as is :+1:

NTaylorMullen:
Missed this guy: https://github.com/aspnet/Razor/pull/765#discussion_r65230522

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -93,30 +94,33 @@ protected override void Visit(TagHelperChunk chunk)
...
                 var attribute = chunk.Attributes[i];
-                var hasAssociatedDescriptors = chunk.Descriptors.Any(descriptor =>
+                var associatedDescriptors = chunk.Descriptors.Where(descriptor =>
                     descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key)));
 
                 // If there's no descriptors associated or we're hitting a bound attribute a second time.
-                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Key))
+                var isUnBoundAttribute = !associatedDescriptors.Any() || !boundAttributes.Add(attribute.Key);
+
+                // Perf: We will preallocate TagHelperAttribute for unbound attributes and bound string valued attributes.
```
NTaylorMullen: `and simple bound string valued attributes.`

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -93,30 +94,33 @@ protected override void Visit(TagHelperChunk chunk)
...
 
-        private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
+        private void PreAllocateTagHelperAttributes(TagHelperChunk chunk)
         {
             var boundAttributes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             for (var i = 0; i < chunk.Attributes.Count; i++)
             {
                 var attribute = chunk.Attributes[i];
-                var hasAssociatedDescriptors = chunk.Descriptors.Any(descriptor =>
+                var associatedDescriptors = chunk.Descriptors.Where(descriptor =>
```
NTaylorMullen: Sorry, I failed at explaining before. I be :tired_face:. You don't utilize the full descriptors from here on down. Just do:

```C#
var associatedDescriptors = chunk.Descriptors.SelectMany(descriptor.Attributes).Where(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Key));
```

You can then pass the attribute descriptors through and verify them without having to query again for name matches.

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -168,6 +183,25 @@ private void PreAllocateUnboundTagHelperAttributes(TagHelperChunk chunk)
...
+            IEnumerable<TagHelperDescriptor> associatedDescriptors,
+            KeyValuePair<string, Chunk> attribute)
+        {
+            // If the attribute value is a Dynamic value, it cannot be preallocated.
+            if (CSharpTagHelperCodeRenderer.IsDynamicAttributeValue(attribute.Value))
+            {
+                return false;
+            }
+
+            // Only attributes that are associated with be string typed properties can be preallocated.
```
NTaylorMullen: `associated with be string` => `associated with string`

NTaylorMullen:
Just small things left. You may need to react to this change in MVC. Double check that / PR and then

:shipit: for this.

mnltejaswini:
@NTaylorMullen  There were bunch of merge conflicts between your PR and mine. I resolved them. Can you please do a quick scan to check that I did not do any weird during resolving merge conflicts.

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -229,5 +230,10 @@ public GeneratedTagHelperContext()
         /// The name of the type used to represent encoded content.
         /// </summary>
         public string EncodedHtmlStringTypeName { get; set; }
+
+        /// <summary>
+        /// The name of the Value property of <c>TagHelperAttribute</c>
```
dougbu: Add a period at the end of this comment, like all doc comments.

src/Microsoft.AspNetCore.Razor/CodeGenerators/GeneratedTagHelperContext.cs
```diff
@@ -229,5 +230,10 @@ public GeneratedTagHelperContext()
...
+        /// </summary>
+        public string TagHelperAttributeValuePropertyName { get; set; }
```
NTaylorMullen: You'll also need to add this bit to the MvcRazorHost FYI

NTaylorMullen:
Looks good after merge conflicts :+1:

src/Microsoft.AspNetCore.Razor/CodeGenerators/Visitors/CSharpTagHelperFieldDeclarationVisitor.cs
```diff
@@ -95,32 +96,39 @@ protected override void Visit(TagHelperChunk chunk)
...
             {
                 var attribute = chunk.Attributes[i];
-                var hasAssociatedDescriptors = chunk.Descriptors.Any(descriptor =>
-                    descriptor.Attributes.Any(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Name)));
+                var associatedAttributeDescriptors = chunk.Descriptors.SelectMany(descriptor => descriptor.Attributes)
+                    .Where(attributeDescriptor => attributeDescriptor.IsNameMatch(attribute.Name));
 
                 // If there's no descriptors associated or we're hitting a bound attribute a second time.
-                if (!hasAssociatedDescriptors || !boundAttributes.Add(attribute.Name))
+                var isUnBoundAttribute = !associatedAttributeDescriptors.Any() || !boundAttributes.Add(attribute.Name);
```
dougbu: Given the use of this `bool` later, probably better to move `!boundAttributes.Add(attribute.Name)` into the `if` and have this variable just mean the attribute isn't bound.

If on the other hand `TagHelperAttributeKey` needs this meaning, suggest renaming the variable to describe the actual condition e.g. `isUnboundOrRepeatedAttribute`.

NTaylorMullen: @dougbu and I talked offline. This is good as is. If you could leave a comment describing what qualifies an unbound attribute that'd be awesome. Aka, no associated descriptors or is a second attribute which has the same name as a bound attribute.

mnltejaswini: Yes. TagHelperAttributeKey needs its meaning because we need to create a HTMLString for the repeated bound attribute. I can rename it.

Issue #760, opened 2016-05-23
NotImplementedException on rendering attributes with non-latin letters added from IClientModelValidator


dubtar:
Implementing IClientModeValidator 
```C#
    public class MyAttr: ValidationAttribute, IClientModelValidator  {
    ...
    public void AddValidation(ClientModelValidationContext context)
    {
        string message = "Привет!"; // message in Russian throws on render
        // string messge = "Test"; is rendered fine
        context.Attributes.Add("data-val-myval", message);
    }
```

Throws exception if message contains non-latin letters.

>     An unhandled exception occurred while processing the request.
> 
>     NotImplementedException: The method or operation is not implemented.
>     Microsoft.AspNetCore.Razor.TagHelpers.DefaultTagHelperContent.EmptyOrWhiteSpaceWriter.Write(Char value)
>     System.Text.Encodings.Web.TextEncoder.Write(TextWriter output, Char* input, Int32 inputLength)
>     System.Text.Encodings.Web.TextEncoder.EncodeCore(TextWriter output, Char* value, Int32 valueLength)
>     System.Text.Encodings.Web.TextEncoder.Encode(TextWriter output, String value, Int32 startIndex, Int32 characterCount)
>     Microsoft.AspNetCore.Mvc.Rendering.TagBuilder.AppendAttributes(TextWriter writer, HtmlEncoder encoder)
>     Microsoft.AspNetCore.Mvc.Rendering.TagBuilder.WriteTo(TextWriter writer, HtmlEncoder encoder)
>     Microsoft.AspNetCore.Mvc.ViewFeatures.Internal.ViewBuffer.WriteTo(TextWriter writer, HtmlEncoder encoder)
>     Microsoft.AspNetCore.Razor.TagHelpers.DefaultTagHelperContent.IsEmptyOrWhiteSpaceCore(Object entry, EmptyOrWhiteSpaceWriter writer)
>     Microsoft.AspNetCore.Razor.TagHelpers.DefaultTagHelperContent.get_IsEmptyOrWhiteSpace()
>     Microsoft.AspNetCore.Mvc.TagHelpers.LabelTagHelper.d__15.MoveNext()

dougbu:
Both a bug and a test gap.

dougbu:
44f0aa6

PR #771, opened 2016-05-31
Do not HTML encode while checking for whitespace

dougbu:
- #760
- using a `NullHtmlEncoder` improves `TagHelperContent.IsEmptyOrWhiteSpace` semantics
 - e.g. a tab is still a tab after encoding
 - also avoids `NotImplementedException`

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
@@ -288,6 +288,7 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
             var stringValue = entry as string;
             if (stringValue != null)
             {
+                // Do not encode the string because encoded value remains whitespace from user's POV.
                 if (!string.IsNullOrWhiteSpace(stringValue))
                 {
                     return false;
@@ -295,7 +296,8 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
             }
             else
             {
-                ((IHtmlContent)entry).WriteTo(writer, HtmlEncoder.Default);
+                // Use NullHtmlEncoder to avoid treating encoded whitespace as non-whitespace e.g. "\t" as "&#x9;".
+                ((IHtmlContent)entry).WriteTo(writer, NullHtmlEncoder.Default);
                 if (!writer.IsEmptyOrWhiteSpace)
                 {
                     return false;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -288,14 +288,16 @@ private bool IsEmptyOrWhiteSpaceCore(object entry, EmptyOrWhiteSpaceWriter write
             var stringValue = entry as string;
             if (stringValue != null)
             {
+                // Do not encode the string because encoded value remains whitespace from user's POV.
```
dougbu: Note this branch also avoids HTML encoding. My change below just makes things consistent for strings and everything else.

dougbu:
thx @NTaylorMullen

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -342,7 +344,7 @@ public override Encoding Encoding
             public bool IsEmptyOrWhiteSpace { get; private set; } = true;
 
 #if NETSTANDARD1_5
-            // This is an abstract method in DNXCore
+            // This is an abstract method in .NET Core but NullHtmlEncoder doesn't need it.
             public override void Write(char value)
```
NTaylorMullen: Can't this still get called if a users `IHtmlContent` calls `writer.Write((char)...)` in its `WriteTo(writer, encoder)` method?

dougbu: Bizarre but `true`. :ok: I'll implement it and write the bizarre test.

dougbu:
:up::date:

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -342,10 +344,13 @@ public override Encoding Encoding
             public bool IsEmptyOrWhiteSpace { get; private set; } = true;
 
 #if NETSTANDARD1_5
```
NTaylorMullen: Nuke so it works the same on desktop and then add a test to ensure it properly picks up `!IsEmptyOrWhiteSpace`

NTaylorMullen:
:shipit: once you've covered https://github.com/aspnet/Razor/pull/771#discussion_r65282467

dougbu:
:up::date: as a FYI

dougbu:
44f0aa6

PR #776, opened 2016-06-05
Update Json.NET to 9.0.1-beta1

javiercn:


NTaylorMullen:
Waiting for the additional commit on the removal of the imports

NTaylorMullen:
Talked offline, keeping the imports in for now to verify things build even when packages are being built inconsistently.

:shipit:

Issue #778, opened 2016-06-08
Remove unused classes


NTaylorMullen:
Remove XmlHelper, CharUtils, EnumeratorExtensions, EnumUtil, BufferingTextReader, SourceSpan and TextBufferReader.

pranavkm:
Also https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor/RazorDebugHelpers.cs which I think lets you remove some of the project dependencies in Razor.

NTaylorMullen:
6e4f15bad4252fcf477c7b806ff459b558f77674
ccba161447f132d86e5714f90af8d49689c3eb9b

PR #781, opened 2016-06-09
Remove unused/unpurposeful classes.

NTaylorMullen:
- Several classes are legacy and left over from many changes in the Razor code base. This changeset removes those classes.
- Also moved several corresponding test cases

#778 

/cc @pranavkm


pranavkm:
:shipit:

PR #797, opened 2016-06-29
Perf: Reduce string allocations from TagHelperDescriptorFactory during startup 

mnltejaswini:
Below are some string allocations identified during startup of an Customer app (https://github.com/aspnet/Hosting/issues/781) 

image

Refactor the ValidateMethod to make these string allocations only when needed

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
@@ -190,10 +190,9 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
 
             foreach (var name in allowedChildren)
             {
-                var valid = TryValidateName(
-                    name: name,
-                    tagHelperName: tagHelperName,
-                    restrictChildrenAttribute: true,
+                var valid = TryValidateRestrictedChildrenAttributeName(
+                    name,
+                    tagHelperName,
                     errorSink: errorSink);
 
                 if (valid)
@@ -274,7 +273,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         internal static bool ValidateParentTagName(string parentTag, ErrorSink errorSink)
         {
             return parentTag == null ||
-                TryValidateName(
+                TryValidateAttributeName(
                     parentTag,
                     Resources.TagHelperDescriptorFactory_ParentTag,
                     errorSink: errorSink);
@@ -306,7 +305,8 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 Resources.TagHelperDescriptorFactory_Attribute :
                 Resources.TagHelperDescriptorFactory_Tag;
 
-            var validName = TryValidateName(
+
+            var validName = TryValidateAttributeName(
                 name,
                 targetName,
                 errorSink: errorSink);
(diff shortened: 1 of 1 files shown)


dougbu:
What are the `string` allocations for the same scenario w/ this change in place?

mnltejaswini:
Haven't tested for the customer app. But collected some numbers for BasicViews scnenario.
The memory profile for the startup of BasicViews scenario before and after the change is
Before

image

After

image

dougbu:
This saves ~100 K bytes out of ~27600 K bytes collected in the BasicViews scenario -- less than 0.4%. Am I reading the images correctly?

Interested in the customer numbers for this PR...

dougbu:
Not an incredibly significant performance issue but I agree the internal factoring of this class can be improved. We can probably do a bit better than where this PR is at. I'll make comments in the code toward that end.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -365,6 +368,30 @@ private static bool ValidateName(string name, bool targetingAttributes, ErrorSin
...
+                            tagHelperName,
+                            invalidCharacter);
+            }
+            else
+            {
+                return (invalidCharacter) =>
+                        Resources.FormatHtmlTargetElementAttribute_InvalidName(
+                            tagHelperName.ToLower(), 
+                            name, 
+                            invalidCharacter);
```
dougbu: De-intent everything after the `return` by 4 chars.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -322,27 +308,40 @@ private static bool ValidateName(string name, bool targetingAttributes, ErrorSin
...
 
             return validName;
         }
 
         private static bool TryValidateName(
             string name,
-            string whitespaceError,
-            Func<char, string> characterErrorBuilder,
+            string tagHelperName,
+            ErrorSink errorSink)
```
dougbu: No need for this overload. Have all callers pass the `restrictChildrenAttribute` argument.

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -322,27 +308,40 @@ private static bool ValidateName(string name, bool targetingAttributes, ErrorSin
...
+        {
+            return TryValidateName(
+                name,
+                tagHelperName,
+                false,
+                errorSink);
+            
+        }
+
+        private static bool TryValidateName(
```
dougbu: It's time to split this in two or more e.g. `TryValidateAttributeName()` and `TryValidateRestrictChildrenAttributeName()`. Each method can then use the appropriate resources when errors occur. Not that much of the code is truly separate from the error formatting.

Alternatively, separate the whitespace check (two methods or inline at the 3 call sites) from an invalid character check e.g. `private int FindInvalidCharacter(string name)` which returns the index of an invalid `char` or `-1` when all is good. This alternative nicely separates looking for low-level problems from adding errors about those problems.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
mnltejaswini: @dougbu  The code actually find all the invalid characters in the name and adds errors for each of the invalid character in the ErrorSink. So, the second may not be applicable.

mnltejaswini:
:up::date:

dougbu:
:shipit:

Eilon:
@NTaylorMullen @dougbu @rynowak can you re-review this to see what we should do here? Is this still good to merge?

NTaylorMullen:
This PR is no longer valid. @pranavkm ended up fixing this with some precompilation perf fixes [here](https://github.com/aspnet/Razor/commit/5944dd6f8738c1bfcc5e0d0a21fc5c59036ac5d4).

Issue #800, opened 2016-07-08
Remove unused ToString in partial parsing call path


NTaylorMullen:
Currently Razor has a `TextChange.ToString()` call that is invoked for Razor editor tracing. All trace calls are emitted when the EDITOR_TRACING flags are absent (always) which leaves this unused ToString() which happens to makeup 87% of the successful partial parsing code path.

[ToString/Format in question](https://github.com/aspnet/Razor/blob/05a52f445fcc49151ef082af6580e4a6ffbe592d/src/Microsoft.AspNetCore.Razor/RazorEditorParser.cs#L170).

![image](https://cloud.githubusercontent.com/assets/2008729/16673914/680d7a4a-4467-11e6-98aa-e7f464291e77.png)

NTaylorMullen:
Going to start working on this since it makes https://github.com/aspnet/Razor/issues/626 more visible.

NTaylorMullen:
d337bacc69fe4ca39532d958712738c71f07c091

PR #802, opened 2016-07-11
Add additional ifdefs to partial parsing tracing logic.

NTaylorMullen:
- Prior to this change we would be ToStringing the TextChange which resulted in 87% of successful partial parsing execution time.

#800

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/RazorEditorParser.cs
+++ b/src/Microsoft.AspNetCore.Razor/RazorEditorParser.cs
@@ -145,10 +145,10 @@ namespace Microsoft.AspNetCore.Razor
         /// <returns>A <see cref="PartialParseResult"/> value indicating the result of the incremental parse.</returns>
         public virtual PartialParseResult CheckForStructureChanges(TextChange change)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Razor/RazorEditorParser.cs
```diff
@@ -161,14 +161,17 @@ public virtual PartialParseResult CheckForStructureChanges(TextChange change)
...
-
+#if EDITOR_TRACING
             // If there isn't already a parse underway, try partial-parsing
             var changeString = string.Empty;
+#endif
             using (_parser.SynchronizeMainThreadState())
             {
+#if EDITOR_TRACING
                 // Capture the string value of the change while we're synchronized
                 changeString = change.ToString();
```
pranavkm: Remove?

NTaylorMullen: It's needed for tracing below.

ajaybhargavb:
:shipit:

pranavkm:
:shipit:

Issue #780, opened 2016-06-09
Mark EditorHints as obsolete, it's no longer used by tooling.


NTaylorMullen:

Eilon:
You mean the `EditorBrowsable` stuff?

Eilon:
Also what does "no longer used" mean? Was it ever?

NTaylorMullen:
Nope, not `EditorBrowsable`, [this](https://github.com/aspnet/Razor/blob/3d0bf621f330ad5444a5892505dc4792aa14ad03/src/Microsoft.AspNetCore.Razor/Editor/EditorHints.cs). Talked to @anurse a while back and they were originally added to indicate if an attribute value was virtual; at which point Todd would detect that and provide appropriate `~/` IntelliSense. Thing is, Todds HTML editor is a bit more decoupled from the Razor side of things so he ended up taking his own stab at making that possible. End result, they were never used.

dougbu:
So is this a suggestion to `[Obsolete]` the unread / unset `SpanEditHandler.EditorHints` property? Or do you want to treat that class as if it were `internal`?

NTaylorMullen:
I'd prefer to treat it as if it was internal but that'd be a binary breaking change. I'm assuming that's a no go. Therefore, obsolete until we can remove it in a bigger release would suffice.

Eilon:
Bug is approved as a breaking change. Please mark things obsolete (but remain compatible) in this bug, then log a new breaking-change bug for 2.0.0.

Eilon:
Breaking change guidelines: https://github.com/aspnet/Home/wiki/Engineering-guidelines#breaking-changes

NTaylorMullen:
fc4a61586e2773f4b9836c60d01a19229a290e3c

PR #810, opened 2016-07-29
Obsolete `EditorHints`.

NTaylorMullen:
- For tests added grouped pragma warning disables to not pollute the test files with too many duplicate warning disable/restores.

#780

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Editor/SpanEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor/Editor/SpanEditHandler.cs
@@ -29,8 +29,9 @@ namespace Microsoft.AspNetCore.Razor.Editor
         public AcceptedCharacters AcceptedCharacters { get; set; }
 
         /// <summary>
-        /// Provides a set of hints to editors which may be manipulating the document in which this span is located.
+        /// This property is obsolete and will be removed in a future version.
         /// </summary>
+        [Obsolete("This property is obsolete and will be removed in a future version.")]
         public EditorHints EditorHints { get; set; }
 
         public Func<string, IEnumerable<ISymbol>> Tokenizer { get; set; }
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Razor/Editor/SpanEditHandler.cs
```diff
@@ -29,8 +29,9 @@ public SpanEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer, AcceptedCha
         public AcceptedCharacters AcceptedCharacters { get; set; }
 
         /// <summary>
-        /// Provides a set of hints to editors which may be manipulating the document in which this span is located.
+        /// This property is obsolete and will be removed in a future version.
```
NTaylorMullen: Didn't feel like any extra information would be beneficial since this property is not currently set by the Razor system. This comment transitively applies to the above `EditorHints` class.

If you feel differently I can add a little snippet stating that it's not set by Razor.

src/Microsoft.AspNetCore.Razor/Editor/SpanEditHandler.cs
```diff
@@ -158,7 +159,10 @@ internal static bool IsAdjacentOnLeft(Span target, Span other)
 
         public override string ToString()
         {
-            return GetType().Name + ";Accepts:" + AcceptedCharacters + ((EditorHints == EditorHints.None) ? string.Empty : (";Hints: " + EditorHints.ToString()));
+            return GetType().Name + ";Accepts:" + AcceptedCharacters +
+#pragma warning disable // For EditorHints
```
Eilon: Please disable just the one exact warning number, not all warnings.

Eilon: And in the comment of the pragma, say which warning it is (e.g. ``// ignore obsolete warning for EditorHints`)

src/Microsoft.AspNetCore.Razor/Editor/SpanEditHandler.cs
Eilon: I think this is a sufficiently obscure part of the system that this comment is just fine.

NTaylorMullen:
:up: :date:

Eilon:
:shipit:

Issue #815, opened 2016-08-04
Debug.Assert failure when expression block has malformed tags


ajaybhargavb:
Example:
`<div>@(</div>)`

This causes this Debug.Assert to fail
https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/Internal/TagHelperParseTreeRewriter.cs#L130

@NTaylorMullen

PR #816, opened 2016-08-04
Fix Debug.Assert failure for malformed tags in expression

ajaybhargavb:
Issue - #815 

@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/Internal/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/Internal/TagHelperParseTreeRewriter.cs
@@ -60,12 +60,12 @@ namespace Microsoft.AspNetCore.Razor.Parser.TagHelpers.Internal
 
         public void Rewrite(RewritingContext context)
         {
-            RewriteTags(context.SyntaxTree, context);
+            RewriteTags(context.SyntaxTree, context, depth: 0);
 
             context.SyntaxTree = _currentBlock.Build();
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/Internal/TagHelperParseTreeRewriter.cs
```diff
@@ -802,15 +803,22 @@ private TagBlockTracker PopTrackerStack()
 
         private class TagBlockTracker
         {
-            public TagBlockTracker(string tagName, bool isTagHelper)
+            public TagBlockTracker(string tagName, bool isTagHelper) : this (tagName, isTagHelper, depth: 0)
```
NTaylorMullen: No need for the extra ctor, this is a private clss.

test/Microsoft.AspNetCore.Razor.Test/Parser/TagHelpers/Internal/TagHelperParseTreeRewriterTest.cs
```diff
@@ -4486,5 +4486,32 @@ public static IEnumerable<object[]> NestedBlockData
         {
             RunParseTreeRewriterTest(documentContent, expectedOutput, "p", "div");
         }
+
+        [Fact]
+        public void Rewrite_HandlesMalformedNonTagHelperTags_Correctly()
```
NTaylorMullen: MalformedNested

NTaylorMullen:
:shipit: assuming existing tests pass.

Issue #808, opened 2016-07-26
Capture parse setup exceptions to prevent Visual Studio crashes


NTaylorMullen:
Visual Studio being extensible results in third parties writing Razor code to enable new functionality. When there's mismatches between the two, exceptions can be thrown which end up crashing visual studio.

We should capture exceptions and translate them into parser errors for VS to consume/show without crashing.

This was originally reported here: https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_workItems?id=240220&_a=edit

Eilon:
cc @rynowak for FYI information.

NTaylorMullen:
82c9c40709283fd3c2547020c6b41c81766fa849

PR #817, opened 2016-08-04
Capture exceptions when trying to parse files and return parse errors.

NTaylorMullen:
#808

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/RazorTemplateEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor/RazorTemplateEngine.cs
@@ -2,14 +2,19 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 using System.Threading;
+using Microsoft.AspNetCore.Razor.Chunks;
 using Microsoft.AspNetCore.Razor.Chunks.Generators;
 using Microsoft.AspNetCore.Razor.CodeGenerators;
+using Microsoft.AspNetCore.Razor.Compilation.TagHelpers;
 using Microsoft.AspNetCore.Razor.Parser;
+using Microsoft.AspNetCore.Razor.Parser.SyntaxTree;
 using Microsoft.AspNetCore.Razor.Text;
 
 namespace Microsoft.AspNetCore.Razor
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor/RazorTemplateEngine.cs
```diff
@@ -363,26 +368,48 @@ public GeneratorResults GenerateCode(TextReader input, CancellationToken? cancel
...
+
+                // Collect results and return
+                return new GeneratorResults(results, codeGeneratorResult, codeGeneratorContext.ChunkTreeBuilder.Root);
+            }
+            catch (Exception ex)
+            {
+                var errorSink = new ErrorSink();
+                errorSink.OnError(
+                    SourceLocation.Undefined,
+                    RazorResources.FormatFatalException(sourceFileName, Environment.NewLine, ex.Message),
```
ajaybhargavb: Does `ex.Message` also include information about the type of the exception?

NTaylorMullen: It does not.  I saw @pranavkm avoid using the entire `Exception.ToString()` unless built in Debug so I followed the same pattern minus the Debug flag fork.

ajaybhargavb: Got it. But I still feel it would be nice if the parser error shows the actual type of the exception.

ajaybhargavb:
:shipit:

src/Microsoft.AspNetCore.Razor/RazorTemplateEngine.cs
pranavkm: This is kinda bad. Mvc takes parse errors and turns them in to something the compilation error page can display. We'd end up showing a "We can't parse this page due to these user errors" and end up showing a pretty empty page (while also making it pretty crummy for users to give us any useful stack trace). Shouldn't the consumer trap this exception rather than us?

NTaylorMullen: The costumer could. However, the issue that this is solving is to prevent VS from crashing when bits of Razor are replaced via extensions. 

Is there an alternative you'd prefer?

pranavkm: Couldn't VS catch it at that point? They might be able to show a better error too. If there isn't a way for them to catch it, could we catch a specific exception rather than this being catch all?

NTaylorMullen: @ToddGrun would VS be able to catch this or do you guys instantiate this logic on a background thread?

ToddGrun: It looks like this method is only called on a background thread, the VS editor doesn't own the calling code.

pranavkm: Fair enough. We'll live with this.

pranavkm: Could VS call a different method or do we want to keep the method names the same?

NTaylorMullen: VS uses the Razor editor parser which spins this up on its own (on the background thread) when a full parse is needed.

NTaylorMullen: Talked offline. Going to conditionally try-catch here based on if we're design time.

NTaylorMullen:
:up: :date:

src/Microsoft.AspNetCore.Razor/RazorTemplateEngine.cs
```diff
@@ -363,6 +368,50 @@ public GeneratorResults GenerateCode(TextReader input, CancellationToken? cancel
...
+                // During runtime we want code generation explosions to flow up into the calling code. At design time
+                // we want to capture these exceptions to prevent IDEs from crashing.
+                return GenerateCodeCore(input, className, rootNamespace, sourceFileName, checksum);
+            }
+
+            try
+            {
+                return GenerateCodeCore(input, className, rootNamespace, sourceFileName, checksum);
+            }
+            catch (Exception ex)
```
pranavkm: `catch (Exception ex) when (Host.DesignTimeMode)` is an alternative in case you don't want to duplicate the call.

pranavkm:
:shipit:

Issue #792, opened 2016-06-20
String typed TagHelper attributes don't provide instant IntelliSense for trailing periods


NTaylorMullen:
For instance in file => new web application if you type the following you don't get "codification" of the trailing period, it's handled as markup until you type more (note the background color of the code):

![image](https://cloud.githubusercontent.com/assets/2008729/16205659/a74168d4-36d9-11e6-837a-f3738e5b2866.png)

This means if you try to invoke IntelliSense at the period location you don't get any. If you then type any additional code it then highlights:

![image](https://cloud.githubusercontent.com/assets/2008729/16205697/ca6c91b2-36d9-11e6-87c8-35cc999abcb3.png)

This does not happen for non-TagHelper tags; they have a different functionality where the period is captured as code for a brief period of time and then transitioned into markup.

/cc @BrennanConroy

NTaylorMullen:
Also tracked in TFS by https://devdiv.visualstudio.com/defaultcollection/DevDiv/_workitems?id=232792&_a=edit

NTaylorMullen:
Grabbing this since I'm low on issues to work on :smile:. Feel free to point me towards something else more pressing if it exists.

NTaylorMullen:
8d7c51bd600a180fdbd112d7c65c56f3bcc17079

PR #820, opened 2016-08-12
Change TagHelpers to work in partial parsing.

NTaylorMullen:
- Prior to this change any query for ownership within a TagHelper would never succeed. More specifically, when a user would try to write C# continuation statements like `@DateTime.` (note the `.`) they would not get IntelliSense for the `.` in string based TagHelper attributes because the `.` would immediately be handled as a markup period.
- TagHelpers structure can be influenced greatly by changes to the tags start/end body; therefore, only allowed modifications to TagHelper attribute values to avoid complexity and enable the more widespread scenario of a user typing C# code in an attribute value.
- Updated existing tests to reflect the new edit handlers that were added to TagHelper attributes.
- Added partial parsing tests to verify partial parses succeed/fail when expected.
- Verified this change works by replacing binaries in VS.
- Did not add doc comments since we've previously decided that Razor is a corner enough sub-system of ASP.NET that they don't add a whole lot of value. This is consistent with the classes changed.

#792

**Background**:
- `PartialParseResult.Accepted` = partial parse succeeded
- `PartialParseResult.Accepted | PartialParseResult.Provisional` = partial parse succeeded but when the user stops typing perform a full re-parse.
- `PartialParseResult.Rejected` = Can't accurately determine the state of the document, do a full re-parse to ensure correctness.

Before this change `Rejected` would always be returned which would force full re-parses on TagHelper attribute changes; in the case of `DateTime.` the period is always handled as a markup period on full re-parse. The fix here was to make that period be Accepted | Provisional so the `.` would be treated as C# until the next full re-parse occurred.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlock.cs
+++ b/src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlock.cs
@@ -5,9 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using Microsoft.AspNetCore.Razor.Compilation.TagHelpers;
 using Microsoft.AspNetCore.Razor.Parser.SyntaxTree;
 using Microsoft.AspNetCore.Razor.TagHelpers;
-using Microsoft.AspNetCore.Razor.Compilation.TagHelpers;
+using Microsoft.AspNetCore.Razor.Text;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.Parser.TagHelpers
@@ -121,6 +122,45 @@ namespace Microsoft.AspNetCore.Razor.Parser.TagHelpers
                     yield return childSpan;
                 }
             }
+        }
+
+        public override Span LocateOwner(TextChange change)
+        {
+            var oldPosition = change.OldPosition;
+            if (oldPosition < Start.AbsoluteIndex)
+            {
+                // Change occurs prior to the TagHelper.
+                return null;
+            }
+
+            var bodyEndLocation = SourceStartTag?.Start.AbsoluteIndex + SourceStartTag?.Length + base.Length;
+            if (oldPosition > bodyEndLocation)
+            {
+                // Change occurs after the TagHelpers body. End tags for TagHelpers cannot claim ownership of changes
+                // because any change to them impacts whether or not a tag is a TagHelper.
+                return null;
+            }
+
+            var startTagEndLocation = Start.AbsoluteIndex + SourceStartTag?.Length;
+            if (oldPosition < startTagEndLocation)
+            {
+                // Change occurs in the start tag.
+
+                var attributeElements = Attributes
+                    .Select(attribute => attribute.Value)
+                    .Where(value => value != null);
+
+                return LocateOwner(change, attributeElements);
+            }
+
+            if (oldPosition < bodyEndLocation)
+            {
+                // Change occurs in the body
+                return base.LocateOwner(change);
+            }
+
+            // TagHelper does not contain a Span that can claim ownership.
+            return null;
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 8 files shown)


test/Microsoft.AspNetCore.Razor.Test/RazorEditorParserTest.cs
```diff
@@ -23,6 +27,279 @@ public class RazorEditorParserTest : PartialParsingTestBase<CSharpRazorCodeLangu
...
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(TagHelperPartialParseRejectData))]
+        public void TagHelperTagBodiesRejectPartialChanges(
+            TextChange change,
+            MarkupBlock expectedDocument,
+            PartialParseResult partialParseResult)
```
pranavkm: Wouldn't this, by definition of the test method name, always be `Rejected`?

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNetCore.Razor.TagHelpers;
 using Microsoft.AspNetCore.Razor.Compilation.TagHelpers;
 using Microsoft.Extensions.Internal;
+using Microsoft.AspNetCore.Razor.Text;
```
pranavkm: Sort usings

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -123,6 +124,38 @@ public override IEnumerable<Span> Flatten()
...
+        {
+            var oldPosition = change.OldPosition;
+            if (oldPosition < Start.AbsoluteIndex)
+            {
+                // Change occurs prior to the TagHelper.
+                return null;
+            }
+
+            var startTagEndLocation = Start.AbsoluteIndex + SourceStartTag?.Length;
+            if (oldPosition < startTagEndLocation)
```
pranavkm: Might be more in line with the previous if check to see if the change is outside bounds here.

```C#
var bodyEndLocation = base.Start.AbsoluteIndex + base.Length;
if (oldPosition > bodyEndLocation)
{
   return null;
}
```

src/Microsoft.AspNetCore.Razor/Parser/TagHelpers/TagHelperBlock.cs
```diff
@@ -123,6 +124,38 @@ public override IEnumerable<Span> Flatten()
...
+
+                var attributeElements = Attributes
+                    .Select(attribute => attribute.Value)
+                    .Where(value => value != null);
+
+                return LocateOwner(change, attributeElements);
+            }
+
+            var bodyEndLocation = base.Start.AbsoluteIndex + base.Length;
+            if (oldPosition < bodyEndLocation)
```
pranavkm: `<=` if you're not making the change from before.

test/Microsoft.AspNetCore.Razor.Test/Framework/TestSpanBuilder.cs
```diff
@@ -139,7 +139,27 @@ public static SpanConstructor Markup(this SpanFactory self, params string[] cont
 
         public static SpanConstructor CodeMarkup(this SpanFactory self, params string[] content)
         {
-            return self.Span(SpanKind.Code, content, markup: true).With(new MarkupChunkGenerator());
+            return self
+                .Span(SpanKind.Code, content, markup: true)
+                .AsCodeMarkup();
+        }
+
+        public static SpanConstructor CSharpCodeMarkup(this SpanFactory self, string content)
```
pranavkm: Isn't all CodeMarkup CSharpCodeMarkup?

pranavkm:
:watch: for the few questions.

test/Microsoft.AspNetCore.Razor.Test/Framework/TestSpanBuilder.cs
NTaylorMullen: Nope, this is the very sad part about the Razor syntax tree. There's 2 different mechanisms to define what an element is:

1. `Span.Kind`, can be `Code` or `Markup`.
2. `Span.Symbols` type (HTML or CSharp). In this case `CSharpCodeMarkup` only differs from `CodeMarkup` in the symbol type.

NTaylorMullen:
:up: :date:

pranavkm:
:shipit:

ajaybhargavb:
:shipit:

PR #823, opened 2016-08-22
Added view component tag helper updates.

cjqian:
- Exposed `ToHtmlCase` for use in creating tag names for view component tag helper descriptors.
- Added a `WriteAutoPropertyDeclaration` method.
- Added `DecorateChunks` as an opportunity to modify chunks before parsing in code generation, and `BuildAfterExecuteContent` to append extra generated code. Check these out in action [here](https://github.com/cjqian/Mvc/blob/35c2a375ac707ea99aa100e9684162dccc0821dc/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs)! (Still in progress, but hopefully you get the idea.)


davidfowl:
No tests?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,22 @@ public override CodeGeneratorResult Generate()
...
+
+                    BuildAfterExecuteContent(writer, Tree.Children);
                 }
             }
 
             return new CodeGeneratorResult(writer.GenerateCode(), writer.LineMappingManager.Mappings);
         }
 
+        protected virtual void BuildAfterExecuteContent(CSharpCodeWriter writer, IList<Chunk> chunks)
+        {
```
davidfowl: What happens if you mutate the chunks list here (add/remove)?

cjqian:
@davidfowl Tests for which part? We're just exposing `ToHtmlCase`, and there didn't seem to be tests for CSharpCodeWriter. I plan on writing tests for `MvcCSharpCodeGenerator` (which implements `DecorateChunks` and `BuildAfterExecuteContent`) when that gets pushed.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
cjqian: Do you mean that we would `DecorateChunks` at `BuildAfterExecuteContent`?

For view component tag helpers, we need to mutate their chunks such that the type name includes the name of the page (ex. `AspNetCore._Views_Index_Cshtml`) since we'll be appending the after execute content within that class. 

This needs to be done before

davidfowl: Both here and in `DecorateChunks`, what happens if I add a new chunk to the list of chunks? I know the view component tag helpers story required mutating the existing chunks, but what happens if new ones are added or removed. Should this list be readonly or are there valid things you could do.

Similarly, what happens if chunks are removed. Will code that runs later blow up?

davidfowl:
@cjqian Tests verifying that the new public method you exposed does what you expect. Also that the new virtual method gets called with the right arguments.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
cjqian: Oh, understood. I don't think anything would blow up...  It would be similar to adding/deleting parts of the .cshtml file, maybe. Worst case, you could change the content of the chunks and rewrite the view? If the chunks were invalid, perhaps things would just not render? (I'm not sure on this, what do you think?) The chunk list has to be modifiable in our current implementation so we can modify the type names. Is this valid?

src/Microsoft.AspNetCore.Razor.Runtime/Runtime/TagHelpers/TagHelperDescriptorFactory.cs
```diff
@@ -711,7 +711,7 @@ private static bool IsAccessibleProperty(PropertyInfo property)
         /// ONE1TWO2THREE3 => one1two2three3
         /// First_Second_ThirdHi => first_second_third-hi
         /// </example>
```
dougbu: Need to fix the `<example>` since this is now `public`. Right now all the lines will be shown as one messy line. Easiest to surround the lines w/ `<code>...</code>`.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,22 @@ public override CodeGeneratorResult Generate()
...
                     }
+
+                    BuildAfterExecuteContent(writer, Tree.Children);
                 }
             }
 
             return new CodeGeneratorResult(writer.GenerateCode(), writer.LineMappingManager.Mappings);
         }
 
+        protected virtual void BuildAfterExecuteContent(CSharpCodeWriter writer, IList<Chunk> chunks)
```
dougbu: `/// docs` here and for `DecorateChunks()` since the next dev (e.g. in MVC repo) won't know what's up. I know this class has few doc comments but it's something we already use in other repos.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
dougbu: Well, `DecorateChunks()` is _about_ mucking w/ the `chunks` collection and this method is about writing stuff out. Suggest this method use `IReadOnlyList<Chunk> chunks` to make things more clear.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/BuildAfterExecuteContentTestCodeGenerator.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Chunks;
+using Microsoft.AspNetCore.Razor.CodeGenerators;
+
+namespace Microsoft.AspNetCore.Razor.Test.CodeGenerators
+{
+    public class BuildAfterExecuteContentTestCodeGenerator : CSharpCodeGenerator
```
dougbu: Could this be a `private class` in `CSharpCodeGeneratorTest`?

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/DecorateChunksTestCodeGenerator.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Chunks;
+using Microsoft.AspNetCore.Razor.CodeGenerators;
+
+namespace Microsoft.AspNetCore.Razor.Test.CodeGenerators
+{
+    public class DecorateChunksTestCodeGenerator : CSharpCodeGenerator
```
dougbu: Could this also be a `private class` in `DecorateChunksTestCodeGenerator`?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -319,6 +319,20 @@ public CSharpCodeWriter WriteMethodInvocation(string methodName, bool endLine, p
                 .WriteEndMethodInvocation(endLine);
         }
 
+        public CSharpCodeWriter WriteAutoPropertyDeclaration(string accessibility, string typeName, string name)
```
dougbu: I assume the view component as tag helper feature needs easy auto-property generation. Correct?

But why does this method needs to exist in Razor? MVC or any external code could implement this as an extension method:
``` c#
public static CSharpCodeWriter WriteAutoPropertyDeclaration(
    this CSharpCodeWriter writer,
    string accessibility,
    string typeName,
    string name)
{
    return writer
        .Write(accessibility)
        // ...
        .WriteLine();
}
```

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -30,7 +36,7 @@ public void ChunkTreeWithUsings()
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace1", syntaxTreeNode.Object);
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace2.SubNamespace", syntaxTreeNode.Object);
             var codeGenerator = new CodeGenTestCodeGenerator(codeGeneratorContext);
-            var testFile = TestFile.Create("TestFiles/CodeGenerator/Output/CSharpCodeGenerator.cs");
+            var testFile = TestFile.Create($"{_testOutputDirectory}/ChunkTreeWithUsings.cs");
```
dougbu: `Path.Combine()` the directory and file.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -14,9 +17,12 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
 {
     public class CSharpCodeGeneratorTest
     {
+        private static readonly string _testOutputDirectory = "TestFiles/CodeGenerator/Output/CSharpCodeGenerator";
```
dougbu: `Path.Combine()` the folders.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/DecorateChunksTestCodeGenerator.cs
```diff
@@ -0,0 +1,42 @@
...
+        public enum GeneratorType
+        {
+            Add = 0,        // Add duplicates the first chunk in the list.
+            Clear = 1,      // Clear removes all chunks.
+            Default = 2     // Default does not modify chunks.
+        }
+
+        private readonly int _generatorType;
+
+        public DecorateChunksTestCodeGenerator(CodeGeneratorContext context, int type) : base(context)
```
dougbu: Why isn't `type` of type `GeneratorType`?

That said, there isn't so much boilerplate here that 3 separate classes would be unworkable. That should hopefully align with `BuildAfterExecuteContentTestCodeGenerator` and allow you to add just one `[Theory]` and data set to `CSharpCodeGeneratorTest`. (From a test perspective, they're all just `CSharpCodeGenerator` that need to be run and have their output checked.)

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/DecorateChunksTestCodeGenerator.cs
```diff
@@ -0,0 +1,42 @@
...
+            switch (_generatorType)
+            {
+                case (int)GeneratorType.Add:
+                    var firstChunk = chunks.First();
+                    chunks.Add(firstChunk);
+                    break;
+                case (int)GeneratorType.Clear:
+                    chunks.Clear();
+                    break;
+                case (int)GeneratorType.Default:
```
dougbu: Why bother subclassing `CSharpCodeGenerator` if the overridden `DecorateChunks()` is a no-op? `CSharpCodeGenerator` isn't `abstract`.

dougbu:
Like the coverage of the new tests and the new extension points. My comments suggest alternatives but I don't such anything that's Just Wrong<sup>:tm:</sup>.

dougbu:
:watch:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeWriter.cs
cjqian: Correct. Do you recommend taking this out? We could maybe start a `MvcCSharpCodeWriter` class, or not have it at all since we only really use it once or twice. Just thought it might be useful for others in the future.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/DecorateChunksTestCodeGenerator.cs
cjqian: We subclass `CSharpCodeGenerator` because ultimately, we want these parts to be implemented in [`MvcCSharpCodeGenerator`](https://github.com/cjqian/Mvc/blob/d9e7ef9c1eb279480a81880c685e8e9d8fe0a793/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcCSharpCodeGenerator.cs) without overriding `Generate`. Can you think of a more graceful way to do this without adding no-op methods in Razor?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeWriter.cs
davidfowl: It's generic enough to leave in the base. Are there other similar methods?

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
             }
 
             return new CodeGeneratorResult(writer.GenerateCode(), writer.LineMappingManager.Mappings);
         }
 
+        /// <summary>
+        /// Provides an entry point to append code (after execute content) to a generator razor class.
+        /// </summary>
+        /// <param name="writer"></param>
+        /// <param name="chunks"></param>
```
dougbu: Pls fill in the blanks.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
dougbu: Discussed offline. This is actually used to find resources and `'\\'` isn't handled in that code. Fine as-is.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -55,13 +65,123 @@ public void ChunkTreeWithUsings()
             if (!string.Equals(expectedOutput, result.Code, StringComparison.Ordinal))
             {
                 BaselineWriter.WriteBaseline(
-                    @"test\Microsoft.AspNetCore.Razor.Test\TestFiles\CodeGenerator\Output\CSharpCodeGenerator.cs",
+                    Path.Combine(_testOutputDirectory, "Output", "CSharpCodeGenerator.cs"),
```
dougbu: `"ChunkTreeWithUsings.cs"`

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -30,7 +38,9 @@ public void ChunkTreeWithUsings()
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace1", syntaxTreeNode.Object);
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace2.SubNamespace", syntaxTreeNode.Object);
             var codeGenerator = new CodeGenTestCodeGenerator(codeGeneratorContext);
-            var testFile = TestFile.Create("TestFiles/CodeGenerator/Output/CSharpCodeGenerator.cs");
+            
+            var path = "TestFiles/CodeGenerator/Output/CSharpCodeGenerator/ChunkTreeWithUsings.cs";
```
dougbu: Minor but could this be `_testOutputDirectory + "/ChunkTreeWithUsings.cs"`?

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
dougbu: But do we want the `CSharpCodeGenerator` folder here i.e. put the files / resources in the `Output` folder?

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
dougbu: Also looks like `"test"` and `"Microsoft.AspNetCore.Razor.Test"` are missing from the start of the `params` array and `"Output"` is already in `_testOutputDirectory`.

Probably best to enable baseline generation (temporary edit of the `project.json` file) and make sure things are written where they're expected.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -55,13 +65,123 @@ public void ChunkTreeWithUsings()
...
+                    {commentGenerator, commentResult }
+                };
+                
+                return theoryData;
+            }
+        }
+        
+        [Theory]
+        [MemberData(nameof(ModifyOutputData))]
+        public void BuildAfterExecuteContent_DecorateChunks_ModifiesOutput(
```
dougbu: Suggest `BuildAfterExecuteContent_DecorateChunks_ModifyOutput()` since this name mentions two methods.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
dougbu: On second thought, leave the final folder there. Just make sure the files are written correctly w/ `GENERATE_BASELINES`.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -55,13 +65,123 @@ public void ChunkTreeWithUsings()
...
+            CSharpCodeGenerator generator, 
+            CodeGeneratorResult expectedResult)
+        {
+            // Arrange
+            var result = generator.Generate();
+
+            // Act
+            Assert.Equal(result.Code, expectedResult.Code);
```
dougbu: Do the `GENERATE_BASELINES` stuff here too.

dougbu:
:watch: almost done!

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
+
+                    BuildAfterExecuteContent(writer, Tree.Children);
                 }
             }
 
             return new CodeGeneratorResult(writer.GenerateCode(), writer.LineMappingManager.Mappings);
         }
 
+        /// <summary>
+        /// Provides an entry point to append code (after execute content) to a generator razor class.
```
dougbu: `Razor`

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
+        /// </summary>
+        /// <param name="writer">writer writes the additional content</param>
+        /// <param name="chunks">a list of chunks representing the generated page</param>
```
dougbu: Sentences with an initial cap and a final period please. Also suggest `The <see cref="CSharpCodeWriter"/> to receive the additional content.` and `The list of <see cref="Chunk"/>s for the generated program.`

Similar comments for the `DecorateChunks()` parameters.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -30,7 +40,9 @@ public void ChunkTreeWithUsings()
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace1", syntaxTreeNode.Object);
             codeGeneratorContext.ChunkTreeBuilder.AddUsingChunk("FakeNamespace2.SubNamespace", syntaxTreeNode.Object);
             var codeGenerator = new CodeGenTestCodeGenerator(codeGeneratorContext);
-            var testFile = TestFile.Create("TestFiles/CodeGenerator/Output/CSharpCodeGenerator.cs");
+
+            var path = "TestFiles/CodeGenerator/Output/CSharpCodeGenerator/ChunkTreeWithUsings.cs";
```
dougbu: Only if making other changes: `$"{_testOutputDirectory}/ChunkTreeWithUsings.cs"`.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -54,14 +66,151 @@ public void ChunkTreeWithUsings()
...
+        }
+        
+        [Theory]
+        [MemberData(nameof(ModifyOutputData))]
+        public void BuildAfterExecuteContent_DecorateChunks_ModifyOutput(
+            CSharpCodeGenerator generator, 
+            CodeGeneratorResult expectedResult,
+            string fileName)
+        {
+            // Arrange
```
dougbu: Minor: This is also `, Act`

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeWriter.cs
dougbu: Discussed offline. This is fine as-is.

dougbu:
:watch: for one last touch-up of the doc comments.

cjqian:
:up: :date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
+        /// <param name="writer">The <see cref="CSharpCodeWriter"/> to receive the additional content.</param>
+        /// <param name="chunks">The list of <see cref="Chunk"/>s for the generated program.</param>
+        protected virtual void BuildAfterExecuteContent(CSharpCodeWriter writer, IList<Chunk> chunks)
+        {
+        }
+
+        /// <summary>
+        /// Provides an entry point to modify chunk data before code generation.
+        /// </summary>
+        /// <param name="context">The <see cref="CodeGeneratorContext"/> of the chunks."/> </param>
```
dougbu: Last `"/>` is a typo. Remove those 3 characters.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
 
             return new CodeGeneratorResult(writer.GenerateCode(), writer.LineMappingManager.Mappings);
         }
 
+        /// <summary>
+        /// Provides an entry point to append code (after execute content) to a generated Razor class.
+        /// </summary>
+        /// <param name="writer">The <see cref="CSharpCodeWriter"/> to receive the additional content.</param>
+        /// <param name="chunks">The list of <see cref="Chunk"/>s for the generated program.</param>
+        protected virtual void BuildAfterExecuteContent(CSharpCodeWriter writer, IList<Chunk> chunks)
```
dougbu: Forgot: Does `IReadOnlyList<Chunk>` not work for your feature?

dougbu:
:shipit:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
```diff
@@ -82,12 +83,32 @@ public override CodeGeneratorResult Generate()
...
+        {
+        }
+
+        /// <summary>
+        /// Provides an entry point to modify chunk data before code generation.
+        /// </summary>
+        /// <param name="context">The <see cref="CodeGeneratorContext"/> of the chunks."/> </param>
+        /// <param name="chunks">The list of <see cref="Chunk"/>s for the generated program.</param>
+        protected virtual void DecorateChunks(CodeGeneratorContext context, IList<Chunk> chunks)
```
NTaylorMullen: So we don't technically need this piece. MVC already semi-decorates the chunk tree by merging in _ViewImports chunks.

See: https://github.com/aspnet/Mvc/blob/c942eab6e268714f23f0d3b31e35036901d3b2e0/src/Microsoft.AspNetCore.Mvc.Razor.Host/MvcRazorHost.cs#L328-L331

cjqian:
:up: :date:

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeWriter.cs
```diff
@@ -319,6 +319,20 @@ public CSharpCodeWriter WriteMethodInvocation(string methodName, bool endLine, p
...
+        {
+            return Write(accessibility)
+                .Write(" ")
+                .Write(typeName)
+                .Write(" ")
+                .Write(name)
+                .Write(" {")
```
NTaylorMullen: .WriteLine(" { get; set; }"); :+1:

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -2172,6 +2172,37 @@ public void CreateDescriptors_BuildsDescriptorsFromSimpleTypes()
                 TagHelperAttributeDescriptorComparer.Default);
         }
 
+        public static TheoryData<string, string> HtmlConversionData
+        {
+            get
+            {
+                var theoryData = new TheoryData<string, string>
```
NTaylorMullen: For consistency, just do `return new TheoryData<string, string>` instead of setting it to a variable.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Runtime/TagHelpers/TagHelperDescriptorFactoryTest.cs
```diff
@@ -2172,6 +2172,37 @@ public void CreateDescriptors_BuildsDescriptorsFromSimpleTypes()
                 TagHelperAttributeDescriptorComparer.Default);
         }
 
+        public static TheoryData<string, string> HtmlConversionData
```
NTaylorMullen: `TheoryData` here for consistency, don't need to specify the generics

NTaylorMullen: @dougbu makes it easier to change when modifying existing tests. It's also what we do most elsewhere in the Razor code base.

src/Microsoft.AspNetCore.Razor/CodeGenerators/CSharpCodeGenerator.cs
dougbu: Decided offline to leave this as is b/c `Tree.Children` is an `IList<Chunk>` and creating another `List<Chunk>` isn't worth it.

test/Microsoft.AspNetCore.Razor.Test/CodeGenerators/CSharpCodeGeneratorTest.cs
```diff
@@ -14,9 +18,15 @@ namespace Microsoft.AspNetCore.Razor.CodeGenerators
 {
     public class CSharpCodeGeneratorTest
     {
+#if GENERATE_BASELINES
+        private static readonly string _baselinePathStart = "test/Microsoft.AspNetCore.Razor.Test";
```
NTaylorMullen: If the field is `private` you can use `const` instead of `static readonly`.

cjqian:
:up: :date:

dougbu:
:shipit:

NTaylorMullen:
:shipit:

PR #828, opened 2016-09-04
Increase .travis.yml consistency between repos

dougbu:
- aspnet/Universe#349
- minimize `dotnet` setup time; no need for caching
- `MONO_THREADS_PER_CPU` env variable isn't relevant since we aren't running Mono tests
- no need for icu4c w/ current `dotnet` builds


dougbu:
aa3bcc3

PR #833, opened 2016-10-20
Add Razor Evolution Engine and primitives

rynowak:
/cc @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultItemCollection.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultItemCollection.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultItemCollection : ItemCollection
+    {
+        private readonly Dictionary<object, object> _items;
+
+        public DefaultItemCollection()
+        {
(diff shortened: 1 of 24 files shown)


daveaglick:
This looks interesting - is there a corresponding issue, or is it something you can discuss yet?

rynowak:
hi @daveaglick - We don't have a good high level issue for this right now (should create some).

Some information that I can share, we're doing some **long-lead** work to improve Razor's extensibility. This was prototyped roughly as part of the effort building https://github.com/aspnet/RazorPages and is being adding to the mainline codebase of Razor and MVC so that we can ship packages.

Unfortunately there's not a particular release version or date that this is associated with, but I can say that it **won't** be part of the upcoming 1.1.0 release. 

When we **do** release this, we will make the new Razor functionality the default experience in MVC but provide a shim to keep using the old host until we reach a **major** version release. There isn't much extensibility in current Razor, but we're committed to not removing what is there until it's a major version.

===============

Without committing to any specific goals, I'll say that pretty much anything you can imagine is in scope. That includes investment in our VS, and lighting up high quality Razor support in tools that don't have it today. We think that Razor has a lot more to give, and this is a chance for us to clean up the basement so we can start delivering those features.

daveaglick:
@rynowak Thanks a lot for the detailed info. The vision sounds fantastic. I'll watch with interest as this is built out.

GuardRex:
> We don't have a good high level issue for this right now
> pretty much anything you can imagine is under consideration by the team

I'd like to suggest the establishment of an issue for feature requests (or point us to UserVoice).

rynowak:
Thanks for the feedback, I opened an issue to track the high level goals of this work.

PR #837, opened 2016-10-27
Add skeleton of syntax phases

rynowak:


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorParsingPhase.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultRazorParsingPhase : RazorEnginePhaseBase, IRazorParsingPhase
+    {
+        protected override void ExecuteCore(RazorCodeDocument codeDocument)
+        {
+            var syntaxTree = RazorSyntaxTree.Parse(codeDocument.Source);
```
NTaylorMullen: So if a user wants to provide their own SyntaxTree to work on they need to remove the parsing phase from the builder. Might be worth trying to retrieve the syntax tree and then parsing if there wasn't one already provided.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -16,10 +16,24 @@ public static RazorEngine Create()
...
         }
 
+        public static RazorEngine CreateEmpty(Action<IRazorEngineBuilder> configure)
+        {
+            var builder = new DefaultRazorEngineBuilder();
+            configure?.Invoke(builder);
+            return builder.Build();
+        }
+
+        internal static void AddDefaults(IRazorEngineBuilder builder)
```
NTaylorMullen: Will probably want to eventually fork this into `AddRuntimeDefaults` and `AddDesignTimeDefaults`. I feel like it'd be better to put the fork in "is runtime vs design time" logic at the engine builder level and not in the individual phases/features (that's where it currently is in todays Razor, has been very painful/bug farm).

rynowak: This is a good idea, I'm banking it along with my other good ideas for later. Not relevant until we need it.

PR #838, opened 2016-10-28
Moved location of and changed CaseSensitiveTagHelperAttributeComparer…

jbagga:
… to a more generic one

Addresses [Mvc Issue #5323](https://github.com/aspnet/Mvc/issues/5323)

cc @dougbu @NTaylorMullen


test/Microsoft.AspNetCore.Razor.Test/project.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "dotnet-test-xunit": "2.2.0-*",
     "Microsoft.AspNetCore.Razor": "1.1.0-*",
+    "Microsoft.AspNetCore.Razor.Runtime": "1.1.0-*",
```
NTaylorMullen: Ya this sucks. Not a lot we can do though unless we made another runtime test sources package. Not sure that's worth it since it'd only have a single class.

src/Microsoft.AspNetCore.Razor.Test.Sources/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Test.Internal
+{
+    public class CaseSensitiveTagHelperAttributeComparer : IEqualityComparer<TagHelperAttribute>
```
NTaylorMullen: internal

test/Microsoft.AspNetCore.Razor.Test/project.json
jbagga: CaseSensitiveTagHelperAttributeComparer needs TagHelperAttribute and NullHtmlEncoder which are in Razor.Runtime. Is there another way to reference them?

NTaylorMullen: Nope. It just makes me sad lol.

jbagga:
cc @dougbu and @NTaylorMullen 
Updated. Also look at [the Mvc PR](https://github.com/aspnet/Mvc/pull/5469)

src/Microsoft.AspNetCore.Razor.Test.Sources/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor.Test.Internal
```
dougbu: This is an ugly namespace. We almost-always avoid `.Test` segments and only use `.Internal` in test class namespaces when the subject `class` is in the same `.Internal` namespace. (Well, both MVC and Razor violate the coding guidelines and use `.Test` more than they should.)

In any case, this is a test utility so I understand the `.Test` portion but the namespace should be much closer to `TagHelperAttribute`'s. Suggest `Microsoft.AspNetCore.Razor.TagHelpers.Test` or &hellip;`.Testing`.

jbagga: I think I matched it with other classes in Razor.Test.Sources. So it's okay to not have the same namespaces as other files in that project?

NTaylorMullen: I'd argue it should be consistent.

dougbu: @NTaylorMullen sure, if you feel strongly enough, we can (or @jbagga can) fix the rest of the classes in this project. `Microsoft.AspNetCore.Razor.TagHelpers.Test` would be fine for all.

NTaylorMullen: @dougbu Seeing that it's currently all based off of sharing TagHelper test logic I think that sounds good. If we take that approach we should probably swap the package name too.

I think the namespace/package name swapping could done in a separate PR if you didn't want to do it right away (or you can do it here if you want). Leaving this as `Razor.Test.Internal` for now at least maintains the consistency.

dougbu: :ok: @NTaylorMullen 
@jbagga please file a bug (label Task) to fix the namespace and package name up here or do it as part of this PR.

jbagga: If I do this here, should the package name also be what @dougbu suggested for the namespace?

NTaylorMullen: Yup. Note, I'd do a search of all the repos for the name because i believe we have special entries for it within our build system 😄 .

jbagga:
@dougbu @NTaylorMullen Updated. Working on changing references to the namespace/package in other repositories

pranavkm:
Did we consider the name `Microsoft.AspNetCore.TagHelpers.Testing`? It goes well with `Microsoft.AspNetCore.Testing`, `Microsoft.Extensions.Logging.Testing`, and `Microsoft.AspNetCore.Server.IntegrationTesting`

jbagga:
@Eilon @dougbu Should I change it to Microsoft.AspNetCore.Razor.TagHelpers.Testing? Or just Microsoft.AspNetCore.TagHelpers.Testing?

NTaylorMullen:
@pranavkm didn't consider. Not bad though. I think at this point it becomes a time investment thing. My opinion: if you feel like it'll be quick go for it; otherwise don't worry about it. In the end we're not shipping the package. 

As a side note, if the package is going to continue to be shared sources we'll need to have a `.Sources` suffix to indicate it's build time only. Other alternative is to nuke the shared *.cs entry in project.json and have it be a normal package reference.

jbagga:
If adding Sources is the right way to go, might as well change it to Microsoft.AspNetCore.TagHelpers.Testing.Sources. But the package is in Razor/src so should Razor be somewhere in the name?

jbagga:
Changing package name and namespace to Microsoft.AspNetCore.Razor.TagHelpers.Testing.Sources

jbagga:
@NTaylorMullen @dougbu Updated

NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Razor.TagHelpers.Testing.Sources/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,63 @@
...
+                string.Equals(attributeX.Name, attributeY.Name, StringComparison.Ordinal) &&
+                attributeX.ValueStyle == attributeY.ValueStyle &&
+                (attributeX.ValueStyle == HtmlAttributeValueStyle.Minimized ||
+                 string.Equals(GetString(attributeX.Value), GetString(attributeY.Value)));
+        }
+
+        public int GetHashCode(TagHelperAttribute attribute)
+        {
+            // Manually combine hash codes here. We can't reference HashCodeCombiner because we have internals visible
+            // from Mvc.Core and Mvc.TagHelpers; both of which reference HashCodeCombiner.
```
dougbu: Is this still true? If yes, clarify who "we" is because Mvc.Core and Mvc.TagHelpers aren't in this repo. If no, use the `HashCodeCombiner`.

jbagga: Missed that. I'll look into it.

jbagga:
@dougbu Updated.

src/Microsoft.AspNetCore.Razor.TagHelpers.Testing.Sources/CaseSensitiveTagHelperAttributeComparer.cs
```diff
@@ -0,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.AspNetCore.Html;
+using Microsoft.DotNet.InternalAbstractions;
```
dougbu: Oh, wait. Isn't `HashCodeCombiner` in the `Microsoft.Extensions.Internal` namespace?

PR #840, opened 2016-10-31
Add abstractions for Razor IR

rynowak:
This is an API skeleton for the IR data model that we'll be using as a
spiritual continuation of the 'chunks' API. Currently missing a lot of
detail which will be filled in as needed.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/IRNode.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/IRNode.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public abstract class IRNode
+    {
+        internal static readonly IRNode[] EmptyArray = new IRNode[0];
+
+        public abstract IList<IRNode> Children { get; }
+
+        public abstract IRNode Parent { get; set; }
+
+        public abstract void Accept(IRNodeVisitor visitor);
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/IRNodeVisitor.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public abstract class IRNodeVisitor 
+    {
+        public virtual void Visit(IRNode node)
+        {
+            node.Accept(this);
+        }
+
+        public virtual void VisitDefault(IRNode node)
```
NTaylorMullen: lol Roslyn

NTaylorMullen:
Also, not sure I'm 100% convinced on the names of IRX because we typically prefix interfaces with "I" in the asp.net repos.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/IRNode.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public abstract class IRNode
+    {
+        internal static readonly IRNode[] EmptyArray = new IRNode[0];
+
+        public abstract IList<IRNode> Children { get; }
```
NTaylorMullen: Part of me thinks we should have two classes. Parent and leaf nodes.

rynowak:
If you have a better suggestion we do that.. I don't want to just block on naming right now though

NTaylorMullen:
> If you have a better suggestion we do that.. I don't want to just block on naming right now though

Agree which is why I signed off already 😄

PR #860, opened 2016-11-06
Update CSharpCodeParser.cs

PeterNawrot:
Code cleanup


src/Microsoft.AspNetCore.Razor/Parser/CSharpCodeParser.cs
```diff
@@ -476,7 +476,8 @@ private bool MethodCallOrArrayIndex(AcceptedCharacters acceptedCharacters)
                         Accept(dot);
                     }
                 }
-                else if (!At(CSharpSymbolType.WhiteSpace) && !At(CSharpSymbolType.NewLine))
+                // If not white space nor new line
```
NTaylorMullen: No need for comment. The code says this 😄

NTaylorMullen:
Thanks @PetarDimov for the PR!

PR #861, opened 2016-11-09
Add indexability to the RazorSourceDocument.

NTaylorMullen:
- Removed the `CreateReader` API in favor of a `CopyTo`, `Length`, `[index]` and `Encoding`.
- Updated existing APIs to react to the change.

/cc @pranavkm @rynowak 

If the design looks good then i'll write/update remaining tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
@@ -2,38 +2,67 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 using System.Text;
 
 namespace Microsoft.AspNetCore.Razor.Evolution
 {
     internal class DefaultRazorSourceDocument : RazorSourceDocument
     {
-        private MemoryStream _stream;
+        private readonly string _content;
 
-        public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string filename)
+        public DefaultRazorSourceDocument(string content, Encoding encoding, string filename)
         {
-            if (stream == null)
+            if (content == null)
             {
-                throw new ArgumentNullException(nameof(stream));
+                throw new ArgumentNullException(nameof(content));
             }
 
-            _stream = stream;
+            if (encoding == null)
+            {
+                throw new ArgumentNullException(nameof(encoding));
+            }
+
+            _content = content;
             Encoding = encoding;
             Filename = filename;
         }
 
-        public Encoding Encoding { get; }
+        public override char this[int position] => _content[position];
+
+        public override Encoding Encoding { get; }
 
         public override string Filename { get; }
 
-        public override TextReader CreateReader()
-        {
-            var copy = new MemoryStream(_stream.ToArray());
+        public override int Length => _content.Length;
 
-            return Encoding == null
-                ? new StreamReader(copy, detectEncodingFromByteOrderMarks: true)
-                : new StreamReader(copy, Encoding);
+        public override void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count)
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            if (sourceIndex < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(sourceIndex));
+            }
+
+            if (destinationIndex < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(destinationIndex));
+            }
+
+            if (count < 0 || count > Length - sourceIndex || count > destination.Length - destinationIndex)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
+            }
+
+            if (count == 0)
+            {
+                return;
+            }
+
+            _content.CopyTo(sourceIndex, destination, destinationIndex, count);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/LineTrackingStringBuffer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/LineTrackingStringBuffer.cs
@@ -15,6 +15,11 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
         private TextLine _endLine;
 
         public LineTrackingStringBuffer(string content)
+            : this(content.ToCharArray())
+        {
+        }
+
+        public LineTrackingStringBuffer(char[] content)
         {
             _endLine = new TextLine(0, 0);
             _lines = new List<TextLine>() { _endLine };
@@ -43,7 +48,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
             return new CharacterReference(line.Content[idx], new SourceLocation(absoluteIndex, line.Index, idx));
         }
 
-        private void Append(string content)
+        private void Append(char[] content)
         {
             for (int i = 0; i < content.Length; i++)
             {
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
@@ -13,17 +13,13 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
 
         public bool DesignTimeMode { get; set; }
 
-        public virtual RazorSyntaxTree Parse(TextReader input)
-        {
-            var reader = new SeekableTextReader(input);
+        public virtual RazorSyntaxTree Parse(TextReader input) => Parse(input.ReadToEnd());
 
-            return Parse((ITextDocument)reader);
-        }
+        public virtual RazorSyntaxTree Parse(string input) => Parse(((ITextDocument)new SeekableTextReader(input)));
 
-        public virtual RazorSyntaxTree Parse(ITextDocument input)
-        {
-            return ParseCore(input);
-        }
+        public virtual RazorSyntaxTree Parse(char[] input) => Parse(((ITextDocument)new SeekableTextReader(input)));
+
+        public virtual RazorSyntaxTree Parse(ITextDocument input) => ParseCore(input);
 
         private RazorSyntaxTree ParseCore(ITextDocument input)
         {
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
@@ -9,6 +9,16 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 {
     public abstract class RazorSourceDocument
     {
+        public abstract Encoding Encoding { get; }
+
+        public abstract string Filename { get; }
+
+        public abstract char this[int position] { get; }
+
+        public abstract int Length { get; }
+
+        public abstract void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count);
+
         public static RazorSourceDocument ReadFrom(Stream stream, string filename)
         {
             if (stream == null)
@@ -36,14 +46,31 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         private static RazorSourceDocument ReadFromInternal(Stream stream, string filename, Encoding encoding)
         {
-            var memoryStream = new MemoryStream();
-            stream.CopyTo(memoryStream);
+            var reader = new StreamReader(
+                stream,
+                encoding ?? Encoding.UTF8,
+                detectEncodingFromByteOrderMarks: true,
+                bufferSize: (int)stream.Length,
+                leaveOpen: true);
 
-            return new DefaultRazorSourceDocument(memoryStream, encoding, filename);
+            using (reader)
+            {
+                var content = reader.ReadToEnd();
+
+                if (encoding == null)
+                {
+                    encoding = reader.CurrentEncoding;
+                }
+                else if (encoding != reader.CurrentEncoding)
+                {
+                    throw new InvalidOperationException(
+                        Resources.FormatMismatchedContentEncoding(
+                            encoding.EncodingName,
+                            reader.CurrentEncoding.EncodingName));
+                }
+
+                return new DefaultRazorSourceDocument(content, encoding, filename);
+            }
         }
-
-        public abstract string Filename { get; }
-
-        public abstract TextReader CreateReader();
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorSyntaxTree.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorSyntaxTree.cs
@@ -32,11 +32,10 @@ namespace Microsoft.AspNetCore.Razor.Evolution
             }
 
             var parser = new RazorParser();
+            var sourceContent = new char[source.Length];
+            source.CopyTo(0, sourceContent, 0, source.Length);
 
-            using (var reader = source.CreateReader())
-            {
-                return parser.Parse(reader);
-            }
+            return parser.Parse(sourceContent);
         }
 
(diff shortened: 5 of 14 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,46 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
+
+        private void InitializeContent()
         {
-            var copy = new MemoryStream(_stream.ToArray());
+            _stream.Seek(0, SeekOrigin.Begin);
+            var length = (int)_stream.Length;
+
+            using (var reader = new StreamReader(_stream, Encoding, detectEncodingFromByteOrderMarks: true, bufferSize: length, leaveOpen: true))
+            {
+                var text = reader.ReadToEnd();
```
NTaylorMullen: Decided to not do it "yet" 😄

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,46 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
+        public override char this[int position] => Content[position];
+
+        public override Encoding Encoding { get; }
 
         public override string Filename { get; }
 
-        public override TextReader CreateReader()
+        public override int Length => Content.Length;
+
+        public override void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count) =>
```
pranavkm: Does this force reading the whole thing as a string all the time? We won't be able to seek using a `StreamReader` and there wouldn't be a way to seek n-chars over a raw Stream. In this case, is it simpler to expose `Content` as is?

NTaylorMullen: Talked offline. This is to expose a common contract that any `RazorSourceDocument` can fulfill.

NTaylorMullen:
🆙 📅 (out of design)

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,47 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
+            Content.CopyTo(sourceIndex, destination, destinationIndex, count);
```
rynowak: Please don't use `=>` where you're going to put a newline. It makes the code read like a different language.

src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
```diff
@@ -9,6 +9,16 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 {
     public abstract class RazorSourceDocument
     {
+        public abstract Encoding Encoding { get; }
```
pranavkm: Are we not doing docs?

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
     internal class DefaultRazorSourceDocument : RazorSourceDocument
     {
         private MemoryStream _stream;
+        private string _content;
 
         public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string filename)
```
rynowak: This constructor should change. In the original implementation the `stream` was the buffer holding the data. In your example it is the `content`. We don't need both. Do the reading in the factory method before calling this ctor.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,47 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
+            var reader = Encoding == null
                 ? new StreamReader(copy, detectEncodingFromByteOrderMarks: true)
                 : new StreamReader(copy, Encoding);
+            using (reader)
+            {
+                var text = reader.ReadToEnd();
+
+                if (Encoding != null && Encoding != reader.CurrentEncoding)
+                {
+                    throw new InvalidOperationException($"The set {nameof(Encoding)} does not match the provided content's encoding.");
```
rynowak: resources

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,47 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
 
-        public override TextReader CreateReader()
+        public override int Length => Content.Length;
+
+        public override void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count) =>
+            Content.CopyTo(sourceIndex, destination, destinationIndex, count);
+
+        private void InitializeContent()
         {
             var copy = new MemoryStream(_stream.ToArray());
```
pranavkm: `_stream` is a MemoryStream. Shouldn't need to copy it. You could require that it is not closed in the ctor.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
```diff
@@ -13,17 +13,13 @@ public RazorParser()
...
-        }
+        public virtual RazorSyntaxTree Parse(string input) => Parse(((ITextDocument)new SeekableTextReader(input)));
 
-        public virtual RazorSyntaxTree Parse(ITextDocument input)
-        {
-            return ParseCore(input);
-        }
+        public virtual RazorSyntaxTree Parse(char[] input) => Parse(((ITextDocument)new SeekableTextReader(input)));
+
+        public virtual RazorSyntaxTree Parse(ITextDocument input) => ParseCore(input);
```
rynowak: Are these changes just for PranavPoints?

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorSourceDocumentTest.cs
```diff
@@ -10,6 +10,38 @@ namespace Microsoft.AspNetCore.Razor.Evolution
...
+            // Act
+            for (var i = 0; i < document.Length; i++)
+            {
+                indexerBuffer[i] = document[i];
+            }
+
+            // Assert
+            var output = new string(indexerBuffer);
+            Assert.Equal(expectedContent, output);
+        }
```
rynowak: Range checks?

src/Microsoft.AspNetCore.Razor.Evolution/RazorSyntaxTree.cs
```diff
@@ -32,11 +32,10 @@ public static RazorSyntaxTree Parse(RazorSourceDocument source)
...
 
             var parser = new RazorParser();
+            var sourceContent = new char[source.Length];
+            source.CopyTo(0, sourceContent, 0, source.Length);
 
-            using (var reader = source.CreateReader())
-            {
-                return parser.Parse(reader);
-            }
+            return parser.Parse(sourceContent);
```
rynowak: It would be even better to see the tokenizer use `SourceDocument` directly. May not be feasible in one move

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -23,17 +24,47 @@ public DefaultRazorSourceDocument(MemoryStream stream, Encoding encoding, string
...
-
-            return Encoding == null
+            var reader = Encoding == null
```
pranavkm: Since we have the bytes, couldn't we just use `Encoding.GetString(bytes)`?

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/LineTrackingStringBuffer.cs
```diff
@@ -14,7 +14,9 @@ internal class LineTrackingStringBuffer
         private TextLine _currentLine;
         private TextLine _endLine;
 
-        public LineTrackingStringBuffer(string content)
+        public LineTrackingStringBuffer(string content) : this(content.ToCharArray()) { }
```
pranavkm: Formatting, `: this(content.ToCharArray()) { }` on a new line

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorSourceDocumentTest.cs
```diff
@@ -10,6 +10,38 @@ namespace Microsoft.AspNetCore.Razor.Evolution
     public class DefaultRazorSourceDocumentTest
     {
         [Fact]
+        public void Indexer()
```
pranavkm: Need better names

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
pranavkm: In the custom implementations we have in Mvc, we explicitly do the range checks rather than letting the underlying APIs throw. We should keep it consistent.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
NTaylorMullen: I prefer the readability when providing overloaded methods.

src/Microsoft.AspNetCore.Razor.Evolution/RazorSyntaxTree.cs
NTaylorMullen: It'd probably be possible but lets let that sit for when we revisit the parser.

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorSourceDocumentTest.cs
NTaylorMullen: We want to test if it explodes? I thought we wanted to avoid doing that.

src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
NTaylorMullen: I feel like it'd be too early to do docs until we know the final form of the system. Also, in the past we've avoided documenting Razor bits because of their limited use.

When things are more complete I don't think it'd be a bad idea to break the latter rule since the whole purpose of Razor.Evolution is to make Razor more consumable by users.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
NTaylorMullen: It's to ensure the memory stream isn't closed when we read it.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
NTaylorMullen: Sure

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
pranavkm: > The overloads we're using don't enable us to pass in "leave open"?

Which overloads are these?

NTaylorMullen: `new StreamReader(copy, detectEncodingFromByteOrderMarks: true)`.

If I expand it to take in the current `Encoding` it explodes when `Encoding` is `null`. All this doesn't matter if you're [suggestion below](https://github.com/aspnet/Razor/pull/861#discussion_r88122755) works.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
rynowak: Yeah, this change is pretty fundamental to the rest of the changes. Keep the static factory methods around on the abstract class so it's still easy to construct one.

Just bear in mind that this will likely need to change again in the future. We can't really use strings long term because of LOH.

NTaylorMullen: We can take a Roslyn approach and have different types of `RazorSourceDocument`s based on how they're constructed. Aka `LargeStringDocument`, `StringDocument`, `FileDocument`. Garbage names but you get the idea 😄

NTaylorMullen:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceDocument.cs
```diff
@@ -2,38 +2,72 @@
...
+
+            if (count < 0 || count > Length - sourceIndex || count > destination.Length - destinationIndex)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
+            }
 
-            return Encoding == null
-                ? new StreamReader(copy, detectEncodingFromByteOrderMarks: true)
-                : new StreamReader(copy, Encoding);
+            return count > 0;
```
rynowak: Why is this done this way instead of just putting these validations in the enclosing method? We generally want the throwing callstack to be exactly what the user called.

src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
```diff
@@ -39,11 +49,23 @@ private static RazorSourceDocument ReadFromInternal(Stream stream, string filena
...
             stream.CopyTo(memoryStream);
 
-            return new DefaultRazorSourceDocument(memoryStream, encoding, filename);
-        }
+            memoryStream.Seek(0L, SeekOrigin.Begin);
 
-        public abstract string Filename { get; }
+            var reader = encoding == null
+                ? new StreamReader(memoryStream, detectEncodingFromByteOrderMarks: true)
+                : new StreamReader(memoryStream, encoding);
```
rynowak: There's no need for the memory stream here, you can read directly from the provided stream as long as you don't close it.

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorSourceDocumentTest.cs
rynowak: Yeah we can skip it.

NTaylorMullen:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/Resources.resx
```diff
@@ -120,6 +120,9 @@
   <data name="IRBuilder_PopInvalid" xml:space="preserve">
     <value>The '{0}' operation is not valid when the builder is empty.</value>
   </data>
+  <data name="MismatchedContentEncoding" xml:space="preserve">
+    <value>The provided encoding does not match the content's encoding.</value>
```
pranavkm: Do you want to include the actual values in the message 

> The specified encoding '{0}' does not match the content's encoding '{1}'.

pranavkm: Also a test for the encoding detection? Unless I missed it someplace.

NTaylorMullen: @rynowak suggested avoiding explosion test cases. This one probably makes sense to test though since we're purposefully exploding in an unexpected scenario. Can also dd the names; that'd be useful 👍

NTaylorMullen:
Will merge once @rynowak signs off.

Issue #845, opened 2016-11-02
Core: Implement TagHelper discovery


NTaylorMullen:
When restructuring the way Razor's parsing pipeline works we did not port the TagHelper discovery mechanisms in the same way as we did in the past. We need to implement this ontop of our various pass extensibility pieces.

NTaylorMullen:
26a1cf3cff57cc57694c70ef01e9dfce036ff23e
6c8ef157b4487cb50f7e3b4021a6b0192d9ed37d

PR #863, opened 2016-11-15
Add TagHelper binding abstractions.

NTaylorMullen:
- Added a TagHelperFeature to hold TagHelper specific pieces that can be replaced.
- Built a syntax tree pass that applies the ported TagHelper bits.
- Updated tests that expected different RazorEngine defaults.

When design is signed off on i'll add tests.

#845

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
@@ -32,6 +32,8 @@ namespace Microsoft.AspNetCore.Razor.Evolution
         {
             builder.Phases.Add(new DefaultRazorParsingPhase());
             builder.Phases.Add(new DefaultRazorSyntaxTreePhase());
+
+            builder.Features.Add(new TagHelperBinderSyntaxTreePass());
         }
 
         public abstract IReadOnlyList<IRazorEngineFeature> Features { get; }
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineTest.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution;
 using Moq;
 using Xunit;
 
@@ -73,15 +74,17 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         private static void AssertDefaultFeatures(IEnumerable<IRazorEngineFeature> features)
         {
-            Assert.Empty(features);
-        }
-
-        private static void AssertDefaultPhases(IReadOnlyList<IRazorEnginePhase> features)
-        {
             Assert.Collection(
                 features,
-                f => Assert.IsType<DefaultRazorParsingPhase>(f),
-                f => Assert.IsType<DefaultRazorSyntaxTreePhase>(f));
+                feature => Assert.IsType<TagHelperBinderSyntaxTreePass>(feature));
+        }
+
+        private static void AssertDefaultPhases(IReadOnlyList<IRazorEnginePhase> phases)
+        {
+            Assert.Collection(
+                phases,
+                phase => Assert.IsType<DefaultRazorParsingPhase>(phase),
+                phase => Assert.IsType<DefaultRazorSyntaxTreePhase>(phase));
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Evolution/TagHelperBinderSyntaxTreePass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/TagHelperBinderSyntaxTreePass.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class TagHelperBinderSyntaxTreePass : IRazorSyntaxTreePass
+    {
+        public RazorEngine Engine { get; set; }
+
+        public int Order => 100;
+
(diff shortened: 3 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/TagHelpers/TagHelperFeature.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.TagHelpers
+{
+    internal class TagHelperFeature : IRazorEngineFeature
```
NTaylorMullen: This needs to be public but we need to discuss/finalize our `ITagHelperDescriptorResolver` contracts to do that (A legacy piece that's internal).

Also, horribad name but couldn't think of anything better from the original prototype.

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineTest.cs
```diff
@@ -73,15 +74,17 @@ public void Create_Lambda_AddsFeaturesAndPhases()
...
         private static void AssertDefaultFeatures(IEnumerable<IRazorEngineFeature> features)
         {
-            Assert.Empty(features);
+            Assert.Collection(
+                features,
+                feature => Assert.IsType<TagHelperBinderSyntaxTreePass>(feature));
         }
 
-        private static void AssertDefaultPhases(IReadOnlyList<IRazorEnginePhase> features)
+        private static void AssertDefaultPhases(IReadOnlyList<IRazorEnginePhase> phases)
```
NTaylorMullen: Just fixed naming.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelpers/TagHelperBinderSyntaxTreePass.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.TagHelpers
+{
+    internal class TagHelperBinderSyntaxTreePass : IRazorSyntaxTreePass
+    {
+        public RazorEngine Engine { get; set; }
+
+        public int Order => 100;
```
NTaylorMullen: Up for debate 😄

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -13,6 +14,9 @@ public static RazorEngine Create()
             return Create(configure: null);
         }
 
+        // TODO: These should probably exist as extension methods in a different library to add features that don't exist
+        // in the parser. I.E. TagHelperDescriptorResolver implementation.
```
rynowak: No todos. 👎

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution.TagHelpers;
```
rynowak: Please hold off on adding new namespaces for now. I want to see the shape of things first.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelpers/TagHelperFeature.cs
rynowak: Why does this need to be public?

NTaylorMullen: Because it takes in an `ITagHelperDescriptorResolver` which is currently `internal` as well. Eventually we'll have a runtime implementation of the `TagHelperDescriptorResolver` that'll need to implement `ITagHelperDescriptorResolver`. It can't do that unless the interface is public or if it's in the same project; however, we can't put the `ITagHelperDescriptoResolver` implementation in this project because it'll depend on Roslyn.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
NTaylorMullen: Will do.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelpers/TagHelperFeature.cs
rynowak: You described why it's currently `internal, but why does it need to be `public`? I think ideally it would be, but it's not a blocker for now.

I would also not assume that this class will be tied to `ITHDR` when we're done.

Also the naming is fine.

NTaylorMullen: Interesting, you envisioning flowing `ITHDR` in a different manner? Only reason why I felt that it needed to be public was its coupling to `ITHDR`.

NTaylorMullen:
Small 🆙📅 

Still need to talk about [this](https://github.com/aspnet/Razor/pull/863#discussion_r88170937).

src/Microsoft.AspNetCore.Razor.Evolution/TagHelpers/TagHelperFeature.cs
rynowak: Yes. I think it's find to go ahead with this as internal.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution;
```
rynowak: This shouldn't be needed

NTaylorMullen:
🆙 📅 

Added tests for the binder syntax tree pass. Tried not too be too specific with the syntax tree output. More or less verified that the legacy bits were running 👍

NTaylorMullen:
Merged.

PR #865, opened 2016-11-17
Implement IR lowering phase

rynowak:
This is a first cut of IR lowering and includes the basic node types and
some tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -1,0 +1,215 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Evolution.Intermediate;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultRazorIRLoweringPhase : RazorEnginePhaseBase, IRazorIRLoweringPhase
+    {
+        protected override void ExecuteCore(RazorCodeDocument codeDocument)
+        {
+            var syntaxTree = codeDocument.GetSyntaxTree();
+            ThrowForMissingDependency<RazorSyntaxTree>(syntaxTree);
+
+            var visitor = new Visitor();
+
+            visitor.VisitBlock(syntaxTree.Root);
+
+            var irDocument = (DocumentIRNode)visitor.Builder.Build();
+            codeDocument.SetIRDocument(irDocument);
+        }
+
+        private class Visitor : ParserVisitor
+        {
+            public Visitor()
+            {
+                Builder = RazorIRBuilder.Document();
+
+                Namespace = new NamespaceDeclarationIRNode();
+                NamespaceBuilder = RazorIRBuilder.Create(Namespace);
+                Builder.Push(Namespace);
+
+                Class = new ClassDeclarationIRNode();
+                ClassBuilder = RazorIRBuilder.Create(Class);
+                Builder.Push(Class);
+
+                Method = new MethodDeclarationIRNode();
+                Builder.Push(Method);
+            }
+
+            public RazorIRBuilder Builder { get; }
+
+            public NamespaceDeclarationIRNode Namespace { get; }
+
+            public RazorIRBuilder NamespaceBuilder { get; }
+
+            public ClassDeclarationIRNode Class { get; }
+
+            public RazorIRBuilder ClassBuilder { get; }
+
+            public MethodDeclarationIRNode Method { get; }
+
+            public override void VisitStartAttributeBlock(AttributeBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Push(new HtmlAttributeIRNode()
+                {
+                    Name = chunk.Name,
+                    Prefix = chunk.Prefix,
+                    Suffix = chunk.Prefix,
+
+                    SourceLocation = block.Start,
+                });
+            }
+
+            public override void VisitEndAttributeBlock(AttributeBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Pop();
+            }
+
+            public override void VisitStartDynamicAttributeBlock(DynamicAttributeBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Push(new CSharpAttributeValueIRNode()
+                {
+                    Prefix = chunk.Prefix,
+                    SourceLocation = block.Start,
+                });
+            }
+
+            public override void VisitEndDynamicAttributeBlock(DynamicAttributeBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Pop();
+            }
+
+            public override void VisitLiteralAttributeSpan(LiteralAttributeChunkGenerator chunk, Span span)
+            {
+                Builder.Add(new HtmlAttributeValueIRNode()
+                {
+                    Prefix = chunk.Prefix,
+                    SourceLocation = span.Start,
+                });
+            }
+
+            public override void VisitStartTemplateBlock(TemplateBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Push(new TemplateIRNode());
+            }
+
+            public override void VisitEndTemplateBlock(TemplateBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Pop();
+            }
+
+            // CSharp expressions are broken up into blocks and spans because Razor allows Razor comments
+            // inside an expression.
+            // Ex:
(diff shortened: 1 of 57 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/HtmlContentIRNode.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public override IList<RazorIRNode> Children { get; } = new List<RazorIRNode>();
+
+        public override RazorIRNode Parent { get; set; }
+
+        internal override SourceLocation SourceLocation { get; set; }
+
+        public override void Accept(RazorIRNodeVisitor visitor)
+        {
+            if (visitor == null)
```
NTaylorMullen: Should be consistent with this. Lots of IRNodes where you don't null check.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/HtmlContentIRNode.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public sealed class HtmlContentIRNode : RazorIRNode
```
NTaylorMullen: Why sealed?

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/RazorIRNode.cs
```diff
@@ -13,6 +14,8 @@ public abstract class RazorIRNode
 
         public abstract RazorIRNode Parent { get; set; }
 
+        internal abstract SourceLocation SourceLocation { get; set; }
```
NTaylorMullen: Why internal?

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/RazorIRNodeVisitorOfT.cs
```diff
@@ -15,7 +15,86 @@ public virtual TResult VisitDefault(RazorIRNode node)
             return default(TResult);
         }
```
NTaylorMullen: After seeing this come together more I'm not convinced we need this class.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/HtmlContentIRNode.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public string Content { get; set; }
+
+        public override IList<RazorIRNode> Children { get; } = new List<RazorIRNode>();
```
NTaylorMullen: Will want to comb through these IR Nodes and not always allocate a list for the children. In the 99% case this node will have 0 children.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/BlockDirectiveIRNode.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public class BlockDirectiveIRNode : RazorIRNode
```
NTaylorMullen: How do you know which bits are the body and which bits are the token declaration? By fixing that you may be able to have a single directive node since RazorIRNode always has children available.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/ParserVisitor.cs
```diff
@@ -0,0 +1,132 @@
...
+            
+            for (var i = 0; i < block.Children.Count; i++)
+            {
+                block.Children[i].Accept(this);
+            }
+
+            VisitEndBlock(block);
+        }
+
+        public virtual void VisitStartBlock(Block block)
```
NTaylorMullen: Seems wasteful/overly complex to have all these start/end mechanics given how we're using them (to create a container object of a specific IR Node type).

Why not have each of the call points be responsible for diving their children? Aka:
```C#
Builder.Push(new CustomIRNodeType() { ... })
visitor.VisitBlock(block); // Dive children
Builder.Pop();
```

Could also take a block creation approach where the call to accept is expected to output some container type that's used as a parent to dive the children.

Lots of choices available. Just really don't like start/end 😄

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -0,0 +1,215 @@
...
+
+                Namespace = new NamespaceDeclarationIRNode();
+                NamespaceBuilder = RazorIRBuilder.Create(Namespace);
+                Builder.Push(Namespace);
+
+                Class = new ClassDeclarationIRNode();
+                ClassBuilder = RazorIRBuilder.Create(Class);
+                Builder.Push(Class);
+
+                Method = new MethodDeclarationIRNode();
```
rynowak: Discussed this in person. I don't like execute because it's too much like MVC. I like better that we'd call it something neutral like `RazorMethodDeclarationIRNode`. That makes it clear that it's a "Razor" method without tying it to a framework

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -0,0 +1,215 @@
...
+                Builder.Push(Method);
+            }
+
+            public RazorIRBuilder Builder { get; }
+
+            public NamespaceDeclarationIRNode Namespace { get; }
```
rynowak: Yeah I think we could skip this

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -0,0 +1,215 @@
...
+            public override void VisitEndDynamicAttributeBlock(DynamicAttributeBlockChunkGenerator chunk, Block block)
+            {
+                Builder.Pop();
+            }
+
+            public override void VisitLiteralAttributeSpan(LiteralAttributeChunkGenerator chunk, Span span)
+            {
+                Builder.Add(new HtmlAttributeValueIRNode()
+                {
+                    Prefix = chunk.Prefix,
```
rynowak: Sure

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            public override void VisitSetBaseTypeSpan(SetBaseTypeChunkGenerator chunk, Span span)
+            {
+                Class.BaseType = span.Content;
+            }
+
+            public override void VisitMarkupSpan(MarkupChunkGenerator chunk, Span span)
+            {
+                Builder.Add(new HtmlContentIRNode()
```
rynowak: Not sure I understand this. This is a span. I didn't change how the chunks work

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/HtmlContentIRNode.cs
rynowak: Sealed by default

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/RazorIRNode.cs
rynowak: This is legacy

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/RazorIRNodeVisitorOfT.cs
rynowak: This is for 3rd party extensibility

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/ParserVisitor.cs
rynowak: As discussed this is the existing pattern for chunk generators

rynowak:
@NTaylorMullen addressed the feedback here and added some tests and merged 👍

Issue #849, opened 2016-11-02
Core: Port Syntax Tree pases from prototype


NTaylorMullen:
In todays Razor we have the [ConditionalAttributeCollapser](https://github.com/aspnet/Razor/blob/77ed9f22fc8894fbce796bb8a704d6cd03a3b226/src/Microsoft.AspNetCore.Razor/Parser/ConditionalAttributeCollapser.cs) and the [WhitespaceRewriter](https://github.com/aspnet/Razor/blob/77ed9f22fc8894fbce796bb8a704d6cd03a3b226/src/Microsoft.AspNetCore.Razor/Parser/WhitespaceRewriter.cs) to make the final SyntaxTree lighter. In the Razor evolution project we've yet to port these two concepts ontop of SyntaxTree passes.

NTaylorMullen:
cf7489e600a2a9f57eb3babef65a8b0fa773f2e6

PR #867, opened 2016-11-19
Port WhiteSpaceRewriter and ConditionalAttributeCollapser.

NTaylorMullen:
- Add an `HtmlNodeOptimizationPass` that does all of the `RazorSyntaxTreeRewriting` that legacy used to achieve outside of `TagHelper`s.
- Add tests for `HtmlNodeOptimizationPass` to verify it's executing appropriate bits.

#849 

/cc @rynowak @pranavkm


test/Microsoft.AspNetCore.Razor.Evolution.Test/HtmlNodeOptimizationPassTest.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        [Fact]
+        public void Execute_CollapsesConditionalAttributes()
+        {
+            // Assert
+            var content = "<input value='hello world' />";
+            var sourceDocument = TestRazorSourceDocument.Create(content);
+            var originalTree = RazorSyntaxTree.Parse(sourceDocument);
+            var pass = new HtmlNodeOptimizationPass();
+            var codeDocument = RazorCodeDocument.Create(sourceDocument);
```
rynowak: You can use `TestRazorCodeDocument.Create(...)` here and it will accept the content directly

Issue #853, opened 2016-11-02
Core: Implement extensible directives


NTaylorMullen:
Make Razor directives extensible much like what we did in the Razor pages prototype.

NTaylorMullen:
518378f4990efff05ae4122a9206f6d8485fe669

PR #869, opened 2016-11-23
Add extensible directive abstractions

NTaylorMullen:
- Based generic directive implementation off of descriptors.
- Added parsing logic to consume descriptors and parse content that's expected.
- Added parsing errors to automagically detect unexpected directive pieces.
- Updated visitor implementations to understand the directive bits.
- Added a builder abstraction to easily create descriptors. Had to maintain the ability to manually construct a descriptor to enable convenient serialization/deserialization.

#853

Will add tests once design looks good. 

/cc @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -206,6 +206,30 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                 Namespace.Children.Insert(i, @using);
             }
 
+            public override void VisitDirectiveToken(DirectiveTokenChunkGenerator chunkGenerator, Span span)
+            {
+                Builder.Add(new DirectiveTokenIRNode()
+                {
+                    Content = span.Content,
+                    Descriptor = chunkGenerator.Descriptor,
+                    SourceLocation = span.Start,
+                });
+            }
+
+            public override void VisitStartDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
+            {
+                Builder.Push(new DirectiveIRNode()
+                {
+                    Name = chunkGenerator.Descriptor.Name,
+                    Descriptor = chunkGenerator.Descriptor,
+                });
+            }
+
+            public override void VisitEndDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
+            {
+                Builder.Pop();
+            }
+
             private class ContainerRazorIRNode : RazorIRNode
             {
                 private SourceLocation? _location;
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
@@ -1,0 +1,16 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public class DirectiveDescriptor
+    {
+        public string Name { get; set; }
+
+        public DirectiveDescriptorKind Kind { get; set; }
+
+        public IReadOnlyList<DirectiveTokenDescriptor> Tokens { get; set; }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptorComparer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptorComparer.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DirectiveTokenDescriptorComparer : IEqualityComparer<DirectiveTokenDescriptor>
+    {
+        public static readonly DirectiveTokenDescriptorComparer Default = new DirectiveTokenDescriptorComparer();
+
+        protected DirectiveTokenDescriptorComparer()
+        {
+        }
+
+        public bool Equals(DirectiveTokenDescriptor descriptorX, DirectiveTokenDescriptor descriptorY)
+        {
+            if (descriptorX == descriptorY)
+            {
+                return true;
+            }
+
+            return descriptorX != null &&
+                string.Equals(descriptorX.Value, descriptorY.Value, StringComparison.Ordinal) &&
+                descriptorX.Kind == descriptorY.Kind;
+        }
+
+        public int GetHashCode(DirectiveTokenDescriptor descriptor)
+        {
+            if (descriptor == null)
+            {
+                throw new ArgumentNullException(nameof(descriptor));
+            }
+
+            var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(descriptor.Value, StringComparer.Ordinal);
+            hashCodeCombiner.Add(descriptor.Kind);
+
+            return hashCodeCombiner.CombinedHash;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.Evolution/IDirectiveDescriptorBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/IDirectiveDescriptorBuilder.cs
@@ -1,0 +1,18 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public interface IDirectiveDescriptorBuilder
+    {
+        IDirectiveDescriptorBuilder AddType();
+
+        IDirectiveDescriptorBuilder AddMember();
+
+        IDirectiveDescriptorBuilder AddString();
+
+        IDirectiveDescriptorBuilder AddLiteral(string literal);
+
+        DirectiveDescriptor Build();
+    }
+}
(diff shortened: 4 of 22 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -206,6 +206,44 @@ public override void VisitImportSpan(AddImportChunkGenerator chunkGenerator, Spa
...
+                    Descriptor = chunkGenerator.Descriptor,
+                });
+            }
+
+            public override void VisitEndDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
+            {
+                var directiveNode = (DirectiveIRNode)Builder.Pop();
+
+                var tokens = directiveNode.Children.OfType<DirectiveTokenIRNode>().ToList();
+                var exceptTokens = directiveNode.Children.Except(tokens).ToList();
```
NTaylorMullen: This is pretty crappy but for the IR passes to come that will light up directives it'll be important to have easy access to the Tokens.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -39,11 +39,16 @@ internal class CSharpCodeParser : TokenizerBackedParser<CSharpTokenizer, CSharpS
         private Dictionary<CSharpKeyword, Action<bool>> _keywordParsers = new Dictionary<CSharpKeyword, Action<bool>>();
 
         public CSharpCodeParser(ParserContext context)
+            : this (directiveDescriptors: Enumerable.Empty<DirectiveDescriptor>(), context: context)
```
NTaylorMullen: Should talk about how we want to replace existing directives with the directive descriptors. Aka:

1. Added in the CSharpCodeParser as "additional" directives?
2. Added in the person who constructs CSharpCodeParser in a manner that enables users to remove default directives from the system.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -1404,8 +1409,13 @@ private void AwaitExpression(bool topLevel)
...
 
-        private void SetupDirectives()
+        private void SetupDirectives(IEnumerable<DirectiveDescriptor> directiveDescriptors)
         {
+            foreach (var directiveDescriptor in directiveDescriptors)
+            {
+                MapDirectives(() => HandleDirective(directiveDescriptor), directiveDescriptor.Name);
+            }
+
             MapDirectives(TagHelperPrefixDirective, SyntaxConstants.CSharp.TagHelperPrefixKeyword);
```
NTaylorMullen: These will be able to go away once they're implemented ontop of the directive descriptors. Will file an issue.

src/Microsoft.AspNetCore.Razor.Evolution/IDirectiveDescriptorBuilder.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public interface IDirectiveDescriptorBuilder
+    {
+        IDirectiveDescriptorBuilder AddType();
+
+        IDirectiveDescriptorBuilder AddMember();
+
+        IDirectiveDescriptorBuilder AddString();
+
+        IDirectiveDescriptorBuilder AddLiteral(string literal, bool optional);
```
NTaylorMullen: Don't love the "optional" part of this but this comes into play when doing `@addTagHelper "..."` where the quotes are optional. Also works for ending optional semicolons which exist on many of our directives.

If you can think of something better I'm all ears 😄

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
rynowak: I get **why** you're doing this, but wouldn't it be better to just leave the `Children` as is and make `Tokens` do something like:

```
public IEnumerable<DirectiveTokenIRNode> Tokens => Children.OfType<DirectiveTokenIRNode>();
```

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
```diff
@@ -0,0 +1,103 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public class DirectiveDescriptor
```
rynowak: Add a private constructor

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
```diff
@@ -0,0 +1,103 @@
...
+        public static IDirectiveDescriptorBuilder CreateCodeBlock(string name)
+        {
+            return new DefaultDirectiveDescriptorBuilder(name, DirectiveDescriptorType.CodeBlock);
+        }
+
+        public string Name { get; set; }
+
+        public DirectiveDescriptorType Type { get; set; }
+
+        public IList<DirectiveTokenDescriptor> Tokens { get; set; }
```
rynowak: Wait, what? Is this an implementation of the builder or a factory of the builder?

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptorType.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public enum DirectiveDescriptorType
```
rynowak: IMO we should suffix enums like this with `Kind` instead of `Type`. It's easy to overload the meaning of type.

src/Microsoft.AspNetCore.Razor.Evolution/IDirectiveDescriptorBuilder.cs
rynowak: My two ideas

1. Semicolons are always optional in directives (you don't have a choice)
2. Don't build `@addTagHelper` on top of extensible directives unless we want the "optional" part to become popular

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
rynowak: I don't think there's any value in making the "built in" directives **unparsable** by removing them. Users have the ability to make "built in" directives invalid in the IR if they don't want them to be supported.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
NTaylorMullen: Hmm, sure we can do that 😄

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
NTaylorMullen: That'd break deserialization

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveDescriptor.cs
rynowak: Then how about putting these factory methods on another type - `static class DirectiveDescriptorBuilder`

NTaylorMullen:
🆙 📅 added tests

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptorComparer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public bool Equals(DirectiveTokenDescriptor descriptorX, DirectiveTokenDescriptor descriptorY)
+        {
+            if (descriptorX == descriptorY)
+            {
+                return true;
+            }
+
+            return descriptorX != null &&
+                string.Equals(descriptorX.Value, descriptorY.Value, StringComparison.Ordinal) &&
```
rynowak: This can nullref if `x != null && y == null`

NTaylorMullen: Nope. The if above prevents that 😄 

```C#
if (descriptorX == descriptorY)
```

dougbu: The previous `if` is a short-circuit for the case where `descriptorX` and `descriptorY` are references to the same instance or both `null`. Need something like the following before this expression:
``` c#
if (descriptorX == null || descriptorY == null)
{
   // Not reachable if both descriptors are null.
   return false;
}
```

NTaylorMullen: If descriptorX != null then descriptorY != null because the == statement above would have short-circuited.

What you're suggesting is that we return `false` instead of `true` if both the descriptors being compared are `null`. Probably smart. We should do this for our other comparers because they fall into the same trap.

dougbu: Huh?

If `descriptorX != descriptorY` and `descriptorX != null`, `descriptorY` could be anything except the same reference as `descriptoryX` -- including `null`. That is, the code currently excludes only a single non-`null` value for `descriptorY`. The code you checked in can easily null ref.

Second, two `null` references should compare equal. I was suggesting an _addition_:
``` c#
if (descriptorX == descriptorY)
{
    return true;
}

if (descriptorX == null || descriptorY == null)
{
    return false;
}

return (string.Equals(descriptorX.Name, descriptorY.Name, StringComparison.Ordinal) &&
    descriptorX.Kind == descriptorY.Kind &&
    Enumerable.SequenceEqual( ... );
```

I had a look at the other comparers in Razor. Primarily the test ones are `null`-safe; they often `Assert.NotNull()` for both parameters.

rynowak: What he said

rynowak: And by that I mean, that I agree with my countryman :canada: :ice_hockey: 🍁 :canada:

NTaylorMullen: I be dumb 👍

PR #872, opened 2016-11-30
Add IR phase

rynowak:
This change adds a phase which runs IR passes. Design and code are almost
exactly the same as the existing SyntaxTree phase. However all of this is
public because the IR is public API.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/IRazorIRPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/IRazorIRPhase.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public interface IRazorIRPhase : IRazorEnginePhase
+    {
+    }
+}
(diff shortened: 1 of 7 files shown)


rynowak:
FYI only (unless you're @NTaylorMullen) - going to merge this, but leave open for him in case there is feedback

rynowak:
Ok, well github won't let me. Aw shucks 😭

davidfowl:
I don't know how it all hangs together yet but I approve.

src/Microsoft.AspNetCore.Razor.Evolution/IRazorIRPhase.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public interface IRazorIRPhase : IRazorEnginePhase
```
NTaylorMullen: This isn't used. Also not convinced we need it.

NTaylorMullen: Yup unless that's done in a different PR

rynowak:
OK, I'll remedy that.

PR #874, opened 2016-12-04
Add end-to-end support for extensible directives

rynowak:
This change adds a way to actually configure the RazorEngine to use
extensible directives (previously buried behind legacy API). As part of
this feature adds the RazorParserOptions class to encapsulate anything
else that becomes a parser options (ahem taghelpers).

Now we have a pattern for this when we get there.

Options are propagated as part of the RazorSyntaxTree for
testability/sanity and this was actually responsible for the bulk of the
changes.

Also added some extension methods for adding directives to the
IRazorEngineBuilder and an end to end integration test.

//cc @NTaylorMullen


rynowak:
FYI Only

NTaylorMullen:
Love it 👍

PR #879, opened 2016-12-06
Nodes that have children should use .Children

rynowak:
This removes special casing for nodes that contain children that were
hiding them in a .Content or .Value property.

/cc @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/CSharpAttributeValueIRNode.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/CSharpAttributeValueIRNode.cs
@@ -9,15 +9,13 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
 {
     public class CSharpAttributeValueIRNode : RazorIRNode
     {
-        public override IList<RazorIRNode> Children { get; } = EmptyArray;
+        public override IList<RazorIRNode> Children { get; } = new List<RazorIRNode>();
 
         public override RazorIRNode Parent { get; set; }
 
         internal override SourceLocation SourceLocation { get; set; }
 
         public string Prefix { get; set; }
-
-        public RazorIRNode Content { get; set; }
 
         public override void Accept(RazorIRNodeVisitor visitor)
         {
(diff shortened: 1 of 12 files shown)


test/Microsoft.AspNetCore.Razor.Evolution.Test/Intermediate/DefaultRazorIRLoweringPhaseIntegrationTest.cs
```diff
@@ -133,7 +133,7 @@ public void Lower_WithUsing()
         }
 
         [Fact]
-        public void Lower_WithFunctions()
```
rynowak: These names were backwards 😆

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/CSharpAttributeValueIRNode.cs
```diff
@@ -9,16 +9,14 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
 {
     public class CSharpAttributeValueIRNode : RazorIRNode
     {
-        public override IList<RazorIRNode> Children { get; } = EmptyArray;
+        public override IList<RazorIRNode> Children { get; } = new List<RazorIRNode>();
```
pranavkm: Would it make sense to do this in the base type and change this to a lazy List if we find that we're creating too many of these? Seems super repetitive to have ` = new List<RazorIRNode>();` everywhere.

rynowak: That's something we could consider. I don't want to get bogged down into that level of detail at this point.

NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/CSharpAttributeValueIRNode.cs
rynowak: Also @pranavkm - we were thinking about just splitting the IR back into Tokens + Nodes. Only tokens could represent code with a source mapping, and Nodes could represent context.

But we're not going to block on that right now 👍

hishamco:
@rynowak I think this should be closed after your last commit https://github.com/aspnet/Razor/commit/026f9ffb69fc5f34fc8ca823a14bd4fa84a1e757

rynowak:
thanks!

Issue #884, opened 2016-12-07
Transition IR nodes SourceLocation => DocumentLocation


NTaylorMullen:
IRNodes currently track the source location of the original source that resulted in their creation. We need more information than this in order to generate code for IR. There's a pre-existing structure that was used to do this calling MappingLocation.

NTaylorMullen:
2639fad8ab35d46a7f4955d66ce3c81409a267ce

PR #886, opened 2016-12-07
Transitioned SourceLocation => MappingLocation on IRNodes.

NTaylorMullen:
- Also modified the property name from `SourceLocation` => `DocumentLocation` to avoid ambiguity.
- Updated IR baselines
- Updated IR baseline infrastructure to conditionally render the document location.

#884 

/cc @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -62,8 +62,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                     Name = chunkGenerator.Name,
                     Prefix = chunkGenerator.Prefix,
                     Suffix = chunkGenerator.Suffix,
-
-                    SourceLocation = block.Start,
+                    SourceRange = new MappingLocation(block.Start, block.Length),
                 });
             }
 
@@ -81,7 +80,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                 Builder.Push(new CSharpAttributeValueIRNode()
                 {
                     Prefix = chunkGenerator.Prefix,
-                    SourceLocation = block.Start,
+                    SourceRange = new MappingLocation(block.Start, block.Length),
                 });
             }
 
@@ -96,7 +95,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                 {
                     Prefix = chunkGenerator.Prefix,
                     Content = chunkGenerator.Value,
-                    SourceLocation = span.Start,
+                    SourceRange = new MappingLocation(span.Start, span.Length),
                 });
             }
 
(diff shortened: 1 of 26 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -120,14 +120,32 @@ public override void VisitEndTemplateBlock(TemplateBlockChunkGenerator chunkGene
...
+                {
+                    var startLocation = value.Children[0].DocumentLocation;
+                    var contentLength = 0;
+
+                    for (var i = 0; i < value.Children.Count; i++)
+                    {
+                        contentLength += value.Children[i].DocumentLocation.ContentLength;
+                    }
+
+                    expressionNode.DocumentLocation = new MappingLocation(
```
NTaylorMullen: We were generating document locations incorrectly because the expression was saying its start was at the "@" symbol when that didn't exist as part of the IR.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/RazorIRNodeWriter.cs
```diff
@@ -147,7 +147,10 @@ protected void WriteName(RazorIRNode node)
 
         protected void WriteLocation(RazorIRNode node)
         {
-            _writer.Write(node.SourceLocation.ToString());
+            if (node.DocumentLocation != null)
```
NTaylorMullen: MappingLocation isn't a struct currently. When working on the code gen we frequently check if `SourceLocation` is set on a particular `RazorIRNode` which makes it difficult to differentiate between `SourceLocation` 0,0,0 and one just not being set. We'd have to force all source locations to initially be set to -1,-1,-1 and then check against that repeatedly.

Long story short I feel that having a nullable document location works well here; especially given how much IRNodes are passed around.

NTaylorMullen:
This is a pre-req for the code gen stuff and thought it'd be simple enough to break out as its own thing.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -60,7 +60,7 @@ public override void VisitStartAttributeBlock(AttributeBlockChunkGenerator chunk
                     Value = value,
                     Suffix = chunkGenerator.Suffix,
 
-                    SourceLocation = block.Start,
+                    DocumentLocation = new MappingLocation(block.Start, block.Length),
```
rynowak: This is a span or a range not a location. The terminology is wrong 😆

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -231,27 +249,27 @@ public override void VisitEndDirectiveBlock(DirectiveChunkGenerator chunkGenerat
 
             private class ContainerRazorIRNode : RazorIRNode
```
rynowak: You'll want to rebase, this is gone in `dev`

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/RazorIRNodeWriter.cs
rynowak: This is still just legacy for now, we haven't yet written the location/range abstractions we're going to use.

rynowak: So I guess what I mean to say is that we should do something simple that works 👍 until we have more time to look at it.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
NTaylorMullen: Ahh. Instead of DocumentLocation i'll use `SourceRange`.

Issue #888, opened 2016-12-08
IR Integration tests break on linux


NTaylorMullen:
It looks like we're escaping newlines when generating baselines but the baselines are then tied to Windows (\r\n).

NTaylorMullen:
f09fd291a74a1b0221227fea9ff5257d6456b71b

PR #889, opened 2016-12-08
Make `HtmlContentIRNode` content newlines platform agnostic.

NTaylorMullen:
- Prior to this the platform that the newlines were escaped on would be the platform the baselines would pass on.

#888

/cc @rynowak - Mostly just curious about your opinion of the `[NEWLINE]` identifier.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/RazorIRNodeWriter.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/RazorIRNodeWriter.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/RazorIRNodeWriter.cs
```diff
@@ -155,9 +156,10 @@ protected void WriteContent(string content)
                 return;
             }
 
-            // We explicitly escape newlines in node content so that the IR can be compared line-by-line.
+            // We explicitly escape newlines in node content so that the IR can be compared line-by-line. The escaped
+            // newline cannot be platform specific so we need to replace it with a custom identifier.
             // Also, escape our separator so we can search for ` - `to find delimiters.
-            _writer.Write(content.Replace("\r", "\\r").Replace("\n", "\\n").Replace("-", "\\-"));
+            _writer.Write(content.Replace(Environment.NewLine, "[NEWLINE]").Replace(" - ", "\\-"));
```
rynowak: Why not just normalize it to `\\n` - We have code elsewhere that does this already, it replaces `\r` with `""` (empty)

Issue #847, opened 2016-11-02
Core: Implement CSharp lowering pass (run time)


NTaylorMullen:
Convert IR created in [this](https://github.com/aspnet/Razor/issues/844) issue into runtime CSharp code.

NTaylorMullen:
5d4c4e1ccf67dc81118e7ef1da5c5c29020bd9d7
eb230e0408a5d7b087c102ae6f8e83293cfec7a1

PR #892, opened 2016-12-10
Add Razor runtime code generation.

NTaylorMullen:
- Added TabSize,IsIndentingWithTabs and NamespaceImports to the RazorParser options. These are replacements for the existing RazorEngineHost abstraction.
- Added RazorParserOptions consumption pattern to more than just the parsing phase.
- Added a ChecksumIRNode to ensure Debugging can work.
- Updated tests to to react to new Checksum and Namespace nodes in the IR tree.
- Added code generation test infrastructure to validate ported tests.

#847


src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/ChecksumIRNode.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            visitor.VisitChecksum(this);
+        }
+
+        public override TResult Accept<TResult>(RazorIRNodeVisitor<TResult> visitor)
+        {
+            return visitor.VisitChecksum(this);
+        }
+
+        public static ChecksumIRNode Create(RazorSourceDocument sourceDocument)
```
NTaylorMullen: Instead of having this on the Checksum node we could have this as part of the source document. This felt cleaner though.

NTaylorMullen:
Only reason why this PR is semi-design is I didn't get the chance to verify every ported code gen test yet. If it's easier to review I broke the PR up into two commits; ported code and non-ported code.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/ChecksumIRNode.cs
rynowak: I think we'll want both maybe. We'll want to be able to run the checksum on a `SourceDocument`. So maybe we propagate it? Keep doing what you're doing for now. This is just a design note.

src/Microsoft.AspNetCore.Razor.Evolution/RazorParserOptions.cs
```diff
@@ -15,10 +17,17 @@ public static RazorParserOptions CreateDefaultOptions()
         private RazorParserOptions()
         {
             Directives = new List<DirectiveDescriptor>();
+            NamespaceImports = new HashSet<string>(StringComparer.Ordinal) { nameof(System), typeof(Task).Namespace };
         }
 
         public bool DesignTimeMode { get; set; }
 
+        public int TabSize { get; set; } = 4;
```
rynowak: I think what you've done here is good enough for right now.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/BasicIntegrationTest.cs
```diff
@@ -97,8 +97,8 @@ public void Process_CustomDirective()
             Assert.Equal("test_directive", directiveSpan.Content);
 
             var irDocument = document.GetIRDocument();
-            var irNamespace = irDocument.Children[0];
-            var irClass = irNamespace.Children[0];
```
rynowak: I think we could just remove this test, baselining is a better strategy for this anyway.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -109,5 +109,36 @@ protected void AssertIRMatchesBaseline(DocumentIRNode document)
...
+
+        protected void AssertCSharpDocumentMatchesBaseline(RazorCSharpDocument document)
+        {
+            if (Filename == null)
+            {
+                var message = $"{nameof(AssertCSharpDocumentMatchesBaseline)} should only be called from an integration test ({nameof(Filename)} is null).";
+                throw new InvalidOperationException(message);
+            }
+
+            var baselineFilename = Path.ChangeExtension(Filename, ".codegen.cs");
```
rynowak: Should we put something in here about designtime vs runtime?

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/BasicIntegrationTest/HelloWorld.ir.txt
```diff
@@ -1,5 +1,8 @@
 Document - 
+    Checksum - 
     NamespaceDeclaration -  - 
+        UsingStatement -  - System
+        UsingStatement -  - System.Threading.Tasks
         ClassDeclaration -  -  -  -  - 
             RazorMethodDeclaration -  -  -  -  - 
-                HtmlContent - (0:0,0 [13] ) - Hello, World!
+                HtmlContent - (0:0,0 [13] TestFiles/IntegrationTests/BasicIntegrationTest/HelloWorld.cshtml) - Hello, World!
```
rynowak: I'm kinda wondering if we should 'alias' this for readability when it's based on the 'primary' file. Or maybe just truncate it down to the last segment.

NTaylorMullen: Will truncate

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/IntegrationTestBase.cs
NTaylorMullen: Was thinking we could describe that via the test class name. In this case we have a razor runtime csharp code generator test. Some duplication but very clear.

NTaylorMullen:
Merged.

Issue #894, opened 2016-12-12
Move Inherits, Functions and Section to sit ontop of extensible directive infrastructure


NTaylorMullen:
All three of these can sit ontop of the extensible directive bits. This is needed to fully satisfy code generation.

NTaylorMullen:
2db4985c214909e50cb66f865d158d19d8bf377c

PR #895, opened 2016-12-13
Modify functions, section and inherits to use extensible directives.

NTaylorMullen:
- Removed existing type names used to track `@functions`, `@section` and `@inherits`.
- Updated parsing logic to reflect existing directive behaviors.
- Added additional IR and syntax tree pass in order to fulfill the default directive expectations.
- Updated tests to to expect new extensible directives parse structure.

#894


NTaylorMullen:
Will add tests for new IR and syntax tree passes if design looks good.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
```diff
@@ -0,0 +1,61 @@
...
+                }
+
+                VisitDefault(node);
+            }
+
+            public override void VisitDirective(DirectiveIRNode node)
+            {
+                if (string.Equals(node.Name, CSharpCodeParser.FunctionsDirectiveDescriptor.Name, StringComparison.Ordinal))
+                {
+                    foreach (var child in node.Children.Except(node.Tokens))
```
NTaylorMullen: So I'm not removing the functions directive from where it originally was because it needs to have some presence as a `DirectiveIRNode` so it can be consumed at `DesignTime` to provide proper IntelliSense.

Options:
1. Do what we're doing now, let it exist in two places. This means we have to choose not to `VisitDefault` when we code gen.
2. Remove from razor method declaration and extract its token ir nodes and the functions body to the class level. I believe we can make due with just the tokens to provide IntelliSense.
3. Move the entire directive node to the class level and have code gen special case the functions directive.
4. `DirectiveIRNodes` are not allowed to make it to the code gen phase. There's a new pass that extracts all tokens into a special location and nukes the DirectiveIRNodes from the IR tree. This approach assumes IR passes have transformed directives into their final form and left behind artifacts.

I'm leaning towards 4 but it's more work.

rynowak: This seems little more complicated than I expected.
  - Add the code block to the class 
  - `if DesignTime` -> Transform the directive into something we can use intellisense for
  - Remove the node since you processed it.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveSyntaxTreePass.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultDirectiveSyntaxTreePass : IRazorSyntaxTreePass
```
rynowak: This seems fine, but will we want to do this on the IR instead?

NTaylorMullen: Chose to make this a syntax tree pass so by the time we get to generating IR errors are already logged for nested sections. Had some tests verifying that syntax tree parsing output the errors.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
NTaylorMullen: Going to leave the Directive nodes in both locations to start. Will add an IR pass when I do design time code generation to harvest all DirectiveIRNodes to provide IntelliSense. This way users don't have to worry about providing design time IntelliSense for their directives tokens.

NTaylorMullen:
:up::date: - No longer in design.

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultDirectiveIRPassTest.cs
```diff
@@ -0,0 +1,95 @@
...
+                node => CSharpStatement(" var value = true; ", node));
+            var method = (RazorMethodDeclarationIRNode)@class.Children[0];
+            Children(method,
+                node => Html(string.Empty, node),
+                node => Directive("functions", node,
+                    directiveChild => CSharpStatement(" var value = true; ", directiveChild)),
+                node => Html(string.Empty, node));
+        }
+
+        private static DocumentIRNode Lower(string content)
```
rynowak: IIRC I made a base class for this so it doesn't have to be copy pasted all over. If I didn't, please do 😆

test/Microsoft.AspNetCore.Razor.Evolution.Test/Intermediate/DefaultRazorIRLoweringPhaseIntegrationTest.cs
```diff
@@ -157,7 +157,7 @@ private DocumentIRNode Lower(RazorCodeDocument codeDocument)
                 var phase = engine.Phases[i];
                 phase.Execute(codeDocument);
 
-                if (phase is IRazorIRLoweringPhase)
+                if (phase is IRazorIRPhase)
```
rynowak: I guess I didn't 😆 - just make a base class instead of copying

test/Microsoft.AspNetCore.Razor.Evolution.Test/Intermediate/RazorIRAssert.cs
```diff
@@ -73,6 +73,34 @@ public static void Html(string expected, RazorIRNode node)
...
+        }
+
+        public static void Directive(string expectedName, RazorIRNode node, params Action<RazorIRNode>[] childValidators)
+        {
+            try
+            {
+                var directive = Assert.IsType<DirectiveIRNode>(node);
+                Assert.Equal(expectedName, directive.Name);
+
+                Children(directive, childValidators);
```
rynowak: This should be outside the `try` because it will already throw a good exception

PR #899, opened 2016-12-14
Add model for differentiating design time and runtime parsing.

NTaylorMullen:
- If needed, a phase/feature can always retrieve the syntax tree to lookup whether the parse tree was made in a "design time" fashion.
- Future DesignTime / Runtime specific bits will be added to their corresponding `AddRuntimeDefaults`/`AddDesignTimeDefaults` methods.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/IRazorEngineBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/IRazorEngineBuilder.cs
@@ -11,6 +11,8 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         IList<IRazorEnginePhase> Phases { get; }
 
+        bool DesignTime { get; set; }
+
         RazorEngine Build();
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/IRazorEngineBuilder.cs
```diff
@@ -11,6 +11,8 @@ public interface IRazorEngineBuilder
 
         IList<IRazorEnginePhase> Phases { get; }
 
+        bool DesignTime { get; set; }
```
rynowak: should be `get` only

NTaylorMullen: Talked offline.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -17,6 +17,24 @@ public static RazorEngine Create(Action<IRazorEngineBuilder> configure)
         {
             var builder = new DefaultRazorEngineBuilder();
             AddDefaults(builder);
+            AddRuntimeDefaults(builder);
+            configure?.Invoke(builder);
+            return builder.Build();
+        }
+
+        public static RazorEngine CreateDesignTime()
```
NTaylorMullen: Talked offline.

Issue #901, opened 2016-12-15
Make directive and directive tokens only exist at the IR level


NTaylorMullen:
This will enable simpler code generation logic.

NTaylorMullen:
c6150ba2878d827b23b1b6f31d2cfe299c88e6b5

PR #902, opened 2016-12-15
Move section support into IR.

NTaylorMullen:
- Previously we'd special case `@section` at code generation time; now we transform the directive into an IR node.
- Changed the expectations of `DefineSection` to not take in a section writer. It's now expected to modify what `Write`, `WriteLiteral` etc. write to when inside of the lambda. This is done today in TagHelpers via `StartTagHelperWritingScope`.
- Updated baseline files to reflect new `DefineSection` expectations.
- Updated IR tests since we no longer leave around `DirectiveIRNode`s.

#901

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
@@ -8,16 +8,25 @@ using Microsoft.AspNetCore.Razor.Evolution.Legacy;
 
 namespace Microsoft.AspNetCore.Razor.Evolution
 {
-    internal class DefaultDirectiveIRPass : IRazorIRPass
+    internal class DefaultDirectiveIRPass : RazorIRPassBase
     {
-        public RazorEngine Engine { get; set; }
+        RazorParserOptions _parserOptions;
 
-        public int Order => 150;
+        public override int Order => 150;
 
-        public DocumentIRNode Execute(RazorCodeDocument codeDocument, DocumentIRNode irDocument)
+        protected override void OnIntialized(RazorCodeDocument codeDocument)
         {
-            var walker = new DirectiveWalker();
-            walker.VisitDefault(irDocument);
+            var syntaxTree = codeDocument.GetSyntaxTree();
+            ThrowForMissingDocumentDependency(syntaxTree);
+
+            _parserOptions = syntaxTree.Options;
+        }
+
+        public override DocumentIRNode ExecuteCore(DocumentIRNode irDocument)
+        {
+            var designTime = _parserOptions.DesignTimeMode;
+            var walker = new DirectiveWalker(designTime);
+            walker.VisitDocument(irDocument);
 
             return irDocument;
         }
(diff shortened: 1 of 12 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
```diff
@@ -16,15 +16,31 @@ internal class DefaultDirectiveIRPass : IRazorIRPass
 
         public DocumentIRNode Execute(RazorCodeDocument codeDocument, DocumentIRNode irDocument)
         {
-            var walker = new DirectiveWalker();
-            walker.VisitDefault(irDocument);
+            var syntaxTree = codeDocument.GetSyntaxTree();
+            if (syntaxTree == null)
```
rynowak: Now might be a good time to introduce that base class we decided against earlier.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
```diff
@@ -16,15 +16,31 @@ internal class DefaultDirectiveIRPass : IRazorIRPass
...
+            {
+                throw new InvalidOperationException(Resources.FormatFeatureDependencyMissing(
+                    typeof(DefaultDirectiveIRPass).Name,
+                    syntaxTree.GetType().Name,
+                    typeof(RazorCodeDocument).Name));
+            }
+
+            var designTime = syntaxTree.Options.DesignTimeMode;
+            var walker = new DirectiveWalker(designTime);
+            walker.VisitDocument(irDocument);
```
rynowak: It might be simpler in the long run to propagate the options.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
```diff
@@ -48,13 +66,40 @@ public override void VisitDirective(DirectiveIRNode node)
...
+                    }
+
+                    var lambdaContent = _designTime ? "__razor_section_writer" : string.Empty;
+                    var sectionName = node.Tokens.FirstOrDefault()?.Content;
+                    var defineSectionStartStatement = new CSharpStatementIRNode()
+                    {
+                        Content = /* ORIGINAL: DefineSectionMethodName */ $"DefineSection(\"{sectionName}\", async ({lambdaContent}) => {{",
+                    };
+
+                    node.Parent.Children.Insert(sectionIndex, defineSectionStartStatement);
```
rynowak: I think in the future we'll want to represent this as a single `CSharpStatementIRNode` but for now it seems fine.

What I mean is that this goes into the tree as three nodes, but it's really one. The first and last node are not valid on their own.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
NTaylorMullen: Propogate them onto the irDocument?

src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
NTaylorMullen: I could make a custom node that acted as a container and had 3 children. Thoughts?

PR #903, opened 2016-12-16
Upgrade to VS 2017

natemcmaster:
Mostly done. Still getting a test failure in VS, not sure if it's an issue with Test Explorer or with the upgrade

Microsoft.AspNetCore.Razor.TagHelpers.TagHelperOutputTest.GetChildContentAsync_CallsGetChildContentAsync(useCachedResult: True)
Test FullName:	Microsoft.AspNetCore.Razor.TagHelpers.TagHelperOutputTest.GetChildContentAsync_CallsGetChildContentAsync (37e201474d2b402cac97d06002d3f02b7abcef2b)
Test Source:	 : line 0
Test Outcome:	Failed

Result Message:	System.InvalidOperationException : The test method expected 0 parameter values, but 1 parameter value was provided.


rynowak:
@natemcmaster - are the failures only in VS?

src/Microsoft.AspNetCore.Razor.Evolution/Microsoft.AspNetCore.Razor.Evolution.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <PackageTags>aspnetcore;cshtml;razor</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="**\*.cs" />
+    <EmbeddedResource Include="**\*.resx" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" Version="1.2.0-*" PrivateAssets="All" /> -->
```
rynowak: should this be commented out?

rynowak: NVM saw your other TODO 👍

test/Microsoft.AspNetCore.Razor.Test/Microsoft.AspNetCore.Razor.Test.csproj
```diff
@@ -0,0 +1,59 @@
...
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">
+    <PackageReference Include="Microsoft.NETCore.App" Version="1.1.0" />
+    <PackageReference Include="System.Diagnostics.TraceSource" Version="4.3.0-*" />
+  </ItemGroup>
+
+  <Target Name="CleanTestProjects" BeforeTargets="CoreCompile" AfterTargets="Clean">
+    <RemoveDir Directories="$(TargetDir)TestFiles" Condition="Exists('$(TargetDir)TestFiles')" />
+  </Target>
```
rynowak: Why is this needed? Does this copy test files to the output? Let's discuss when you have a minute.

natemcmaster:
:watch: may want to hold off. It seems this project doesn't build on command line, only within VS. https://github.com/dotnet/sdk/issues/518

natemcmaster:
> We can probably rename these to work around the issue.

Looks like we **should** rename. See https://github.com/xunit/xunit/issues/21 and https://github.com/xunit/xunit/issues/650. What's odd is that this worked fine in dotnet-test-xunit but fails in xunit.runner.visualstudio.

test/Microsoft.AspNetCore.Razor.Test/Microsoft.AspNetCore.Razor.Test.csproj
natemcmaster: Yes, the tests copy to output. Part of the problem I was trying to workaround is this: https://developercommunity.visualstudio.com/content/problem/7956/current-working-directory-of-net-core-tests-is-set.html

rynowak: Yeah we can't use that. 

The point of  `__RemoveThisBitTo__GENERATE_BASELINES` is that we can run the tests in a special mode to update the baseline files if we expect the codegen to have changed. The hardcoded path from `pwd` to the root of the project was intentional so that we can update the files in place.

rynowak: We could implement this in a smarter way than hardcoding a path, but it's important that we be able to update the original files in an automatic way. (PS we will run into this again in MVC).

natemcmaster: Maybe this will still work if we use AppContext.BaseDirectory/AppDomain.CurrentDomain.BaseDirectory instead of Directory.GetCurrentDirectory().

natemcmaster:
:up: :date: resolve the issues. It passes on command line and in VS now.

test/Microsoft.AspNetCore.Razor.Runtime.Test/Microsoft.AspNetCore.Razor.Runtime.Test.csproj
```diff
@@ -7,10 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="**\*.cs" />
+    <Compile Include="**\*.cs" Exclude="TestFiles\**\*" />
     <Compile Include="..\..\shared\Microsoft.AspNetCore.Razor.TagHelpers.Testing.Sources\**\*.cs" />
-    <Content Include="TestFiles\**\*" CopyToOutputDirectory="Always" />
-    <EmbeddedResource Include="TestFiles\**\*" />
```
rynowak: This is supposed to be here, I don't know how this could be passing without it....

rynowak: Maybe this project doesn't use them? /cc @NTaylorMullen

natemcmaster: It isn't in the PJ version: https://github.com/aspnet/Razor/blob/e2dd09c918cfc9e77ae8f31042b1c65f94935e5a/test/Microsoft.AspNetCore.Razor.Runtime.Test/project.json

Also, this is actually part of what was causing https://github.com/dotnet/sdk/issues/518

NTaylorMullen: Nope not needed 👍

src/Microsoft.AspNetCore.Razor.Evolution/Properties/AssemblyInfo.cs
```diff
@@ -1,14 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Reflection;
-using System.Resources;
 using System.Runtime.CompilerServices;
 
-[assembly: AssemblyMetadata("Serviceable", "True")]
```
NTaylorMullen: Curious: Where are these guys going for all the repos?

src/Microsoft.AspNetCore.Razor.Runtime/Sources/ClosedGenericMatcher.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Reflection;
+
+// TODO remove this file and use sources packages https://github.com/aspnet/Common/issues/180
+
+namespace Microsoft.Extensions.Internal
+{
+    /// <summary>
+    /// Helper related to generic interface definitions and implementing classes.
+    /// </summary>
+    internal static class ClosedGenericMatcher
```
NTaylorMullen: Can we throw all of these shared sources into a more unappealing folder aside from `Sources`? AKa. `__TemporarySources__`. If you can think of a better name go for it. 

`Sources` seems to general and I can envision people adding bits to the folder even when they don't need to because it exists.

src/Microsoft.AspNetCore.Razor.Evolution/Properties/AssemblyInfo.cs
natemcmaster: These attributes are all generated by the SDK when MSBuild attributes are present. See https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets

PR #905, opened 2016-12-19
Add line mapping support to source document

rynowak:
Ignore the project changes, I'll rebase

The intent here is that the tokenizer/parser just need to index into the document. Any part of the codebase can do math with the absolute position, and we have one component that knows about newlines 👍

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
@@ -17,6 +17,8 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         public abstract int Length { get; }
 
+        public abstract RazorSourceLineCollection Lines { get; }
+
         public abstract void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count);
 
         public static RazorSourceDocument ReadFrom(Stream stream, string filename)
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceLineCollection.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public abstract class RazorSourceLineCollection
+    {
+        public abstract int Count { get; }
+
+        public abstract int GetLineLength(int index);
+
+        internal abstract SourceLocation GetLocation(int position);
```
rynowak: I'm working on a series of changes to 'de-legacy' this type 😆 🚀

src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
```diff
@@ -17,6 +17,8 @@ public abstract class RazorSourceDocument
 
         public abstract int Length { get; }
 
+        public abstract RazorSourceLineCollection Lines { get; }
```
NTaylorMullen: Seems odd for this guy to be on the `RazorSourceDocument`.  Seems more suited for its methods to be exposed as extension methods. Aka `GetSourceLocation(this RazorSourceDocument document, int position)` etc.

rynowak: How would that work? You have to expose **something** that those extension methods calls. 

I don't see why this is odd. The source document is a container for text, and it also provides information about how that text is broken up into lines.

NTaylorMullen: We already expose it as a collection of characters though. I feel like it'd be overkill to add yet another abstraction on how to read the same document on the abstract base type. Maybe the approach of `SourceDocumentReader` that linified the document and enabled retrieval of source locations?

NTaylorMullen: Talked offline. :shipit:

PR #907, opened 2016-12-20
Getting rid of Advance (part 1)

rynowak:
Removes one use of advance for taghelper directives that use the *string literal* form.


rynowak:
c088a8871df6a7986873b1a563592c3a494c0e8a

Issue #873, opened 2016-12-01
Optimize string allocation in the Tokenizer


Yves57:
I have seen that many strings are allocated in the `Microsoft.AspNetCore.Razor.Tokenizer.Tokenizer.EndSymbol()` method.
But some strings are in fact often the same. So I modified the end of the method like that:

```
.....
string tokenContent;
if (Buffer.Length != 1)
{
   tokenContent = Buffer.ToString();
}
else
{
   if (!_oneCharTokenCache.TryGetValue(Buffer[0], out tokenContent))
   {
      tokenContent = Buffer.ToString();
      _oneCharTokenCache.Add(Buffer[0], tokenContent);
   }
}
sym = CreateSymbol(start, tokenContent, type, errors);
```

I have made a test with an extrem case ([LargePageMvc page](https://github.com/aspnet/Performance/blob/dev/testapp/LargePageMvc/Views/Home/Index.cshtml)).

Before the modification, 215643 strings where allocated. After the optimization, 131577 strings allocations where saved (more than half). And the `_oneCharTokenCache` dictionary had only 17 items inside: some symbols like <, !, >, =, and all digits.

Two additional remarks:
- As a bonus, it improves performances from 587ms to call `MarkupParser.ParseDocument()` down to 510ms on my (old!) computer. It can improve a little bit F5 startup scenarios with large pages...
- Of course, we can extend the concept to any `string` lengths, but there is less redundancy, and it will be probably less performant.

If you think that it is interesting, I can propose a PR...

Yves57:
Other optimization to propose (but smaller).
In [CSharpTokenizer.Identifier](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor/Tokenizer/Internal/CSharpTokenizer.cs#L477), the buffer is transformed to string two times.

Yves57:
Still around `Tokenizer.EndSymbol()` and still to avoid redundant string allocations, I propose to add a `protected virtual string GetSymbolContent(TSymbolType type)` that manages the suggested dictionary proposed in my first issue comment.
But it would be overriden in `HtmlTokenizer` like that:

```
protected override string GetSymbolContent(HtmlSymbolType type)
{
    if (type == HtmlSymbolType.NewLine && Buffer.Length == 1)
    {
       return "\r";
    }
    else if (type == HtmlSymbolType.NewLine && Buffer.Length == 2)
    {
       return "\r\n";
    }
[...]

PR #908, opened 2016-12-20
Optimize HTML symbol content string allocation

Yves57:
See #873.
The `Tokenizer` class may be frequently instancied for small pieces of text (if there is C#/HTML mix). So I don't want to allocate a new dictionary for few symbol redundancy. That's why I changed my design to a simplier solution.
I cannot give Visual Studio 2017RC Profiling result (issue declared) or dotMemory screenshot (too few contributions the free licence :-). But I have added manually counters to evaluate the same [LargePageMvc page](https://github.com/aspnet/Performance/blob/dev/testapp/LargePageMvc/Views/Home/Index.cshtml). It goes from ~215000 to ~73500 `Buffer.ToString()` calls. So the optimization saves ~142500 string allocations for this file.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/Tokenizer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/Tokenizer.cs
@@ -215,10 +215,15 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
                     errors[i] = CurrentErrors[i];
                 }
 
-                sym = CreateSymbol(start, Buffer.ToString(), type, errors);
+                sym = CreateSymbol(start, GetSymbolContent(type), type, errors);
             }
             StartSymbol();
             return sym;
+        }
+
+        protected virtual string GetSymbolContent(TSymbolType type)
+        {
+            return Buffer.ToString();
         }
 
         protected bool TakeUntil(Func<char, bool> predicate)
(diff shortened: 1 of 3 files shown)


Yves57:
All continuous integration checks have failed. But I think that the errors are not specific to my request. Do I need to do something?

rynowak:
Yeah I don't see anything in the logs that has to do with your changes. Once the PR is signed off we can kick those off again

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/Tokenizer.cs
```diff
@@ -215,12 +215,17 @@ protected TSymbol EndSymbol(SourceLocation start, TSymbolType type)
...
+                sym = CreateSymbol(start, GetSymbolContent(type), type, errors);
             }
             StartSymbol();
             return sym;
         }
 
+        protected virtual string GetSymbolContent(TSymbolType type)
+        {
+            return Buffer.ToString();
+        }
```
rynowak: Are there similarly worthwhile optimizations for C#? Granted there's typically a lot more HTML than CSharp

Yves57: After a quick look in `CSharpTokenizer`, I have identified many similar possible optimizations:
- New line (one or two chars)
- White space (optimization only for one char white space)
- '@' (transitions)
- Operators
- Brackets
- Integer (when only one digit, but it is maybe too much, even '0' is frequently used)
If you think that it makes sense, I can create another PR later

rynowak:
Agreed, let's merge this for now - @Yves57 if you think there is low hanging fruit in the C# parser let's do that too.

rynowak:
@NTaylorMullen can you try rerunning appveyor + travis and then merge it?

NTaylorMullen:
> @NTaylorMullen can you try rerunning appveyor + travis and then merge it?

I reran Travis and Eilon is taking care of appveyor.

Eilon:
@NTaylorMullen assigning to you to do the needful.

NTaylorMullen:
Merged.

Issue #878, opened 2016-12-06
Nuke SyntaxTree residual marker symbols from IR


NTaylorMullen:
The current IR generation doesn't weed out Html/C# marker symbols from the syntax tree.

Background: marker symbols are used to tell the editor when it should provide IntelliSense for a specific language at a syntax barrier..

PR #909, opened 2016-12-22
Don't create IR nodes for marker symbols

ajaybhargavb:
Issue - https://github.com/aspnet/Razor/issues/878
- Also setting the correct ContentLength from HtmlContentIRNode

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -126,7 +126,26 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
             public override void VisitEndTemplateBlock(TemplateBlockChunkGenerator chunkGenerator, Block block)
             {
-                Builder.Pop();
+                var templateNode = Builder.Pop();
+                if (templateNode.Children.Count > 0)
+                {
+                    var sourceRangeStart = templateNode
+                        .Children
+                        .FirstOrDefault(child => child.SourceRange != null)
+                        ?.SourceRange;
+
+                    if (sourceRangeStart != null)
+                    {
+                        var contentLength = templateNode.Children.Sum(child => child.SourceRange?.ContentLength ?? 0);
+
+                        templateNode.SourceRange = new MappingLocation(
+                            sourceRangeStart.AbsoluteIndex,
+                            sourceRangeStart.LineIndex,
+                            sourceRangeStart.CharacterIndex,
+                            contentLength,
+                            sourceRangeStart.FilePath ?? _codeDocument.Source.Filename);
+                    }
+                }
             }
 
             // CSharp expressions are broken up into blocks and spans because Razor allows Razor comments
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -181,11 +221,26 @@ public override void VisitStatementSpan(StatementChunkGenerator chunkGenerator,
 
             public override void VisitMarkupSpan(MarkupChunkGenerator chunkGenerator, Span span)
             {
+                if (span.Symbols.Count == 1 &&
+                    span.Symbols[0] is HtmlSymbol symbol &&
```
NTaylorMullen: Bajillion pranav points. On another note though we should stay away from C# 7 until we move to it as a team 😄

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -181,11 +221,26 @@ public override void VisitStatementSpan(StatementChunkGenerator chunkGenerator,
...
                 var currentChildren = Builder.Current.Children;
                 if (currentChildren.Count > 0 && currentChildren[currentChildren.Count - 1] is HtmlContentIRNode)
                 {
                     var existingHtmlContent = (HtmlContentIRNode)currentChildren[currentChildren.Count - 1];
                     existingHtmlContent.Content = string.Concat(existingHtmlContent.Content, span.Content);
+                    existingHtmlContent.SourceRange = new MappingLocation(
+                        existingHtmlContent.SourceRange.AbsoluteIndex,
+                        existingHtmlContent.SourceRange.LineIndex,
+                        existingHtmlContent.SourceRange.CharacterIndex,
+                        existingHtmlContent.SourceRange.ContentLength + span.Content.Length,
```
NTaylorMullen: Amazing catch. 👍 👍 👍 👍

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -126,7 +126,30 @@ public override void VisitStartTemplateBlock(TemplateBlockChunkGenerator chunkGe
 
             public override void VisitEndTemplateBlock(TemplateBlockChunkGenerator chunkGenerator, Block block)
             {
-                Builder.Pop();
+                var templateNode = Builder.Pop();
+                if (templateNode.Children.Count > 0)
```
NTaylorMullen: Talked offline about this. We still want to ensure what the user wrote is flowed through the IR.

ajaybhargavb: Talked offline. Will remove node removal change but keeping the SourceRange setting.

ajaybhargavb:
🆙 📅

PR #910, opened 2016-12-22
Optimize 'Span.Content' memory allocation

Yves57:
See #873 
Tested with the [LagePageMvc](https://github.com/aspnet/Performance/blob/dev/testapp/LargePageMvc/Views/Home/Index.cshtml) page. It goes in the fast path in ~5100 cases (so I suppose it should save ~15300 allocations: `new StringBuilder()`, the initial StringBuilder chunk, and the `StringBuilder.ToString()`).


Yves57:
@rynowak - Problem with AppVeyor, but not related to the PR I think...

rynowak:
Nice!

Eilon:
@rynowak parking on your plate to do the needful (https://github.com/aspnet/Razor/issues/873 is also assigned to you).

rynowak:
Merged e9a688be15302a874814ff18557f01a223067fe0

Thanks!

Issue #848, opened 2016-11-02
Core: Implement CSharp lowering pass (design time)


NTaylorMullen:
Convert IR created in [this](https://github.com/aspnet/Razor/issues/844) issue into design time CSharp code.

NTaylorMullen:
f47a40a4a7f1ce51e0f9689e3810360cb332419f

PR #911, opened 2016-12-23
Add Razor design time code generation.

NTaylorMullen:
- Added a common csharp rendering phase base to put shared assets of runtime and design time code gen.
- Added a new `DesignTimeIRPass` to setup the IR bits to provide accurate intellisense.
- Added a `CodeGenerationIntegrationTest` and moved the RuntimeCodeGenerationTests into it. This way we can re-use the cshtml files and it makes searching/running the tests easy.
- Updated how line mappings are calculated for some nodes.

#848


src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/MethodDeclarationIRNode.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public class MethodDeclarationIRNode : RazorIRNode
```
rynowak: I think we shouldn't create this, it should just be code.

src/Microsoft.AspNetCore.Razor.Evolution/RazorCSharpLoweringPhaseBase.cs
```diff
@@ -0,0 +1,294 @@
...
+                var spaces = resolvedPadding % context.Options.TabSize;
+                var tabs = resolvedPadding / context.Options.TabSize;
+
+                return new string('\t', tabs) + new string(' ', spaces);
+            }
+            else
+            {
+                return new string(' ', resolvedPadding);
+            }
+        }
```
rynowak: This should really be part of the SourceDocument.

PR #913, opened 2016-12-28
Get rid of locations on Tokens.

rynowak:


Razor.sln
```diff
@@ -24,8 +22,6 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Razor.
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Razor.Evolution.Test", "test\Microsoft.AspNetCore.Razor.Evolution.Test\Microsoft.AspNetCore.Razor.Evolution.Test.csproj", "{969357A4-CCF1-46D9-B002-9AA072AFC75C}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Razor.Runtime.Test", "test\Microsoft.AspNetCore.Razor.Runtime.Test\Microsoft.AspNetCore.Razor.Runtime.Test.csproj", "{277AB67E-9C8D-4799-A18C-C628E70A8664}"
-EndProject
```
rynowak: Ignore these, I'll revert

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/ConditionalAttributeCollapser.cs
```diff
@@ -39,7 +39,7 @@ protected override SyntaxTreeNode RewriteBlock(BlockBuilder parent, Block block)
             }
 
             // Create a new span containing this content
-            var span = new SpanBuilder();
+            var span = new SpanBuilder(block.Children[0].Start);
```
rynowak: Note, span builders take a start location now.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/HtmlMarkupParser.cs
```diff
@@ -219,10 +219,13 @@ private void OtherParserBlock()
         {
             AddMarkerSymbolIfNecessary();
             Output(SpanKind.Markup);
+
             using (PushSpanConfig())
             {
                 CodeParser.ParseBlock();
             }
+
+            Span.Start = CurrentLocation;
```
rynowak: We need to initialize the 'start' of the spanbuilder each time we transition between tokenizers. This means we 'catch up' each time the other tokenizer did some work.

Putting this state in a single object would really help, but I didn't want to pull that into this PR.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/HtmlMarkupParser.cs
```diff
@@ -305,7 +310,7 @@ public override void ParseBlock()
                     else
                     {
                         Context.ErrorSink.OnError(
-                            CurrentSymbol.Start,
+                            CurrentStart,
```
rynowak: In general errors all **start** at the start of the current symbol. 

Note that in this PR I changed the meaning of current location to track the end of `CurrentSymbol`.  I did a bulk find-replace on these and then corrected the ones that should still be `CurrentLocation`. The differences are mostly down to whether or not the current symbol will be accepted or put back.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/HtmlMarkupParser.cs
```diff
@@ -789,8 +794,8 @@ private void BeforeAttribute()
             IEnumerable<HtmlSymbol> whitespaceAfterAttributeName)
         {
             // First, determine if this is a 'data-' attribute (since those can't use conditional attributes)
-            var name = nameSymbols.GetContent(Span.Start);
-            var attributeCanBeConditional = !name.Value.StartsWith("data-", StringComparison.OrdinalIgnoreCase);
+            var name = string.Concat(nameSymbols.Select(s => s.Content));
```
rynowak: I got rid of `GetContent`. All of these are good optimization opportunities for us.

For instance I don't think it's possible that the attribute name is multiple tokens here, but I didn't want to do too much.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/HtmlMarkupParser.cs
```diff
@@ -1729,22 +1742,15 @@ private int HandleNestingSequence(string sequence, int position, int currentNest
...
-                        if (postSequence != null)
-                        {
-                            Context.Source.Position = postSequence.Start.AbsoluteIndex;
-                        }
-                        else
-                        {
-                            Context.Source.Position = bookmark;
-                        }
+                        // Position at the start of the postSequence symbol, which might be null.
+                        Context.Source.Position = postSequenceBookmark;
```
rynowak: This is simpler now because we don't have to dereference `postSequence`. Both

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/HtmlMarkupParser.cs
```diff
@@ -1706,7 +1718,7 @@ private int HandleNestingSequence(string sequence, int position, int currentNest
                 if (string.Equals(possibleStart, sequence, Comparison))
                 {
                     // Capture the current symbol and "put it back" (really we just want to clear CurrentSymbol)
-                    var bookmark = Context.Source.Position;
+                    var bookmark = Context.Source.Position - CurrentSymbol.Content.Length;
```
rynowak: This defines bookmark as the start the current symbol. I guess this could be just `CurrentStart`.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/ISymbol.cs
```diff
@@ -5,10 +5,10 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
...
     internal interface ISymbol
     {
-        SourceLocation Start { get; }
+        Span Parent { get; set; }
+
         string Content { get; }
 
-        void OffsetStart(SourceLocation documentStart);
-        void ChangeStart(SourceLocation newStart);
+        SourceLocation Start { get; }
```
rynowak: This value is always computed. I'm not sure if we'll want to leave this here long term or not but it was nice for migrating the code.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SpanBuilder.cs
```diff
@@ -74,15 +101,9 @@ public void Accept(ISymbol symbol)
...
-            {
-                Start = symbol.Start;
-                symbol.ChangeStart(SourceLocation.Zero);
-                _tracker.CurrentLocation = SourceLocation.Zero;
-            }
-            else
+            if (Start.Equals(SourceLocation.Undefined))
             {
-                symbol.ChangeStart(_tracker.CurrentLocation);
+                throw new InvalidOperationException("SpanBuilder must have a valid location");
```
rynowak: I'll add a resource

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SymbolBase.cs
```diff
@@ -21,52 +20,65 @@ internal abstract class SymbolBase<TType> : ISymbol where TType : struct
...
             // Hash code should include only immutable properties.
-            var hashCodeCombiner = HashCodeCombiner.Start();
-            hashCodeCombiner.Add(Content, StringComparer.Ordinal);
-            hashCodeCombiner.Add(Type);
+            var hash = HashCodeCombiner.Start();
+            hash.Add(Content, StringComparer.Ordinal);
+            hash.Add(Type);
 
-            return hashCodeCombiner;
+            return hash;
```
rynowak: #YOLO

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TagHelperBlockRewriter.cs
```diff
@@ -190,7 +190,7 @@ internal static class TagHelperBlockRewriter
                     {
                         capturedAttributeValueStart = true;
 
-                        attributeValueStartLocation = span.Start + symbol.Start;
+                        attributeValueStartLocation = symbol.Start;
```
rynowak: These are now absolute positions not offsets

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/Tokenizer.cs
```diff
@@ -193,19 +193,15 @@ protected TSymbol Single(TSymbolType type)
 
         protected void StartSymbol()
         {
-            Buffer.Clear();
+            Debug.Assert(Buffer.Length == 0);
+            Debug.Assert(CurrentErrors.Count == 0);
```
rynowak: I moved clearing these to `EndSymbol`. It turns out that `CurrentStart` was never correct

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TokenizerView.cs
```diff
@@ -34,15 +34,6 @@ public bool Next()
 
         public void PutBack(TSymbol symbol)
         {
-            Debug.Assert(Source.Position == symbol.Start.AbsoluteIndex + symbol.Content.Length);
-            if (Source.Position != symbol.Start.AbsoluteIndex + symbol.Content.Length)
```
rynowak: I can't really verify this anymore easily so I had to remove it. I want to introduce a real put-back implementation.

test/Microsoft.AspNetCore.Razor.Evolution.Test/Legacy/CSharpAutoCompleteTest.cs
```diff
@@ -121,7 +121,7 @@ public void VerbatimBlockAutoCompleteAtStartOfFile()
                                         Factory.Markup("<p>").Accepts(AcceptedCharacters.None)),
                                     new MarkupTagBlock(
                                         Factory.Markup("</p>").Accepts(AcceptedCharacters.None))),
-                               Factory.Span(SpanKind.Code, new CSharpSymbol(Factory.LocationTracker.CurrentLocation, string.Empty, CSharpSymbolType.Unknown))
```
rynowak: The location is now tracked by default. You don't need to pass it around as much

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -643,12 +648,15 @@ private void SectionBlock(string left, string right, bool caseSensitive)
...
+
             var wasNested = IsNested;
             IsNested = true;
             using (PushSpanConfig())
             {
                 ParseBlock();
             }
+
+            Span.Start = CurrentLocation;
             Initialize(Span);
```
NTaylorMullen: Every time you call Initialize it looks like you're setting `Span.Start` prior. Any reason to not throw that into the Initialize method? Maybe i'm missing something.

rynowak: Initialize is also used in some other places where we don't want to set `Span.Start`. I thought about this and didn't think it was a good plan.

rynowak:
This was merged as f191cc62a2be23dade4b3bd12a3a91f72c14533d

Issue #880, opened 2016-12-06
Add instrumentation end-to-end.


NTaylorMullen:
With the new IR system in Razor the instrumentation bits were left out because they can be implemented as an IR pass and code gen renderer. We should re-add these based on the prototype.

rynowak:
This is done.

ajaybhargavb:
Thanks @rynowak

PR #916, opened 2016-12-29
Added Instrumentation support

ajaybhargavb:
Issue - #880

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineBuilderExtensions.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineBuilderExtensions.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public static class RazorEngineBuilderExtensions
+    {
+        public static IRazorEngineBuilder AddTagHelpers(this IRazorEngineBuilder builder, params TagHelperDescriptor[] tagHelpers)
+        {
+            var resolver = (TestTagHelperDescriptorResolver)builder.Features.OfType<TagHelperFeature>().FirstOrDefault()?.Resolver;
+            if (resolver == null)
+            {
+                resolver = new TestTagHelperDescriptorResolver();
+                builder.Features.Add(new TagHelperFeature(resolver));
+            }
+
+            resolver.TagHelpers.AddRange(tagHelpers);
+            return builder;
+        }
+
+        private class TestTagHelperDescriptorResolver : ITagHelperDescriptorResolver
+        {
+            public List<TagHelperDescriptor> TagHelpers { get; } = new List<TagHelperDescriptor>();
+
+            public IEnumerable<TagHelperDescriptor> Resolve(TagHelperDescriptorResolutionContext resolutionContext)
+            {
+                return TagHelpers;
+            }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultInstrumentationPass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultInstrumentationPass.cs
@@ -1,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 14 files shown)


NTaylorMullen:
Talked offline. Going to take a CSharpStatementIRNode approach instead.

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EscapedTagHelpers_Runtime.codegen.cs
```diff
@@ -29,13 +29,19 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_EscapedTag
         #pragma warning disable 1998
         public async System.Threading.Tasks.Task ExecuteAsync()
         {
+            Instrumentation.BeginContext(25, 72, true);
             WriteLiteral("\r\n<div class=\"randomNonTagHelperAttribute\">\r\n    <p class=\"Hello World\" ");
+            Instrumentation.EndContext();
 #line 4 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EscapedTagHelpers.cshtml"
-                       Write(DateTime.Now);
+                       Write(Instrumentation.BeginContext(100, 12, false);
```
NTaylorMullen: Broken

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/RazorInstrumentationIRPass.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            var walker = new InstrumentationIRNodeWalker();
+            walker.VisitDocument(irDocument);
+
+            return irDocument;
+        }
+
+        internal class InstrumentationIRNodeWalker : RazorIRNodeWalker
+        {
+            private bool _skipNext = false;
```
rynowak: It's not clear from looking at this exactly what the 'skipNext' semantic is for...

ajaybhargavb:
@rynowak, Are you saying we should disable instrumentation when running most of our tests and have new tests with it enabled?

src/Microsoft.AspNetCore.Razor.Evolution/RazorInstrumentationIRPass.cs
ajaybhargavb: This is just here to prevent this code from going in to an infinite loop. I agree this needs some sort of comment.

dougbu:
@ajaybhargavb are you working still? I think @rynowak (who's a bit sick) was planning to take over this PR.

In any case, instrumentation should _not_ be on by default. No way; no how.

ajaybhargavb:
@dougbu, I am not. I was just curious. I meant having instrumentation enabled for all our code generation tests. We do the same for our old razor code generation tests.

dougbu:
> I meant having instrumentation enabled for all our code generation tests. We do the same for our old razor code generation tests.

I agree `RazorCodeGeneratorTest<TLanguage>` sets `RazorEngineHost.EnableInstrumentation` and nothing in its subclass hierarchy (a fair number of tests) overrides that. If that's what you meant, I'm less concerned. (Though we really should at least run a subset of the tests without instrumentation -- old tests too.)

rynowak:
> Are you saying we should disable instrumentation when running most of our tests and have new tests with it enabled?

I am saying that the instrumentation feature pass should not be included by default. It's just noise unless you're specifically opting in to browerlink, and MVC should do that when configuring the engine.

We should one small set of tests that verify this relatively simple feature.

src/Microsoft.AspNetCore.Razor.Evolution/RazorInstrumentationIRPass.cs
rynowak: We probably need to do some lateral thinking about this, but the idea is that you don't rewrite the tree while visiting. We should make that more obvious.

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorEngineBuilderExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+            resolver.TagHelpers.AddRange(tagHelpers);
+            return builder;
+        }
+
+        private class TestTagHelperDescriptorResolver : ITagHelperDescriptorResolver
+        {
+            public List<TagHelperDescriptor> TagHelpers { get; } = new List<TagHelperDescriptor>();
+
+            public IEnumerable<TagHelperDescriptor> Resolve(TagHelperDescriptorResolutionContext resolutionContext)
```
NTaylorMullen: Pranav points!

rynowak:
Merged as ffd455b77eb88f21584a19a2401a3b63c6af1dbf

PR #917, opened 2016-12-29
Adds stronger verification for spans

rynowak:
This change ensures that spans are contiguous and that all source is part
of a span. This means that a character can't be 'lost' and not a member of
any span.

And guess what? We have a bug like that. So now a few tests are skipped
due to that bug.

Also made some changes to tests that construct invalid spans or spans
without correct locations as their expected input. This allows us to add
the above verification to all parser tests.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/Legacy/ParserTestBase.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/Legacy/ParserTestBase.cs
@@ -14,14 +14,21 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
     {
         internal static Block IgnoreOutput = new IgnoreOutputBlock();
 
-        internal SpanFactory Factory { get; private set; }
-        internal BlockFactory BlockFactory { get; private set; }
-
         internal ParserTestBase()
         {
             Factory = CreateSpanFactory();
             BlockFactory = CreateBlockFactory();
         }
+
+        /// <summary>
+        /// Set to true to autocorrect the locations of spans to appear in document order with no gaps.
+        /// Use this when spans were not created in document order.
+        /// </summary>
+        protected bool FixupSpans { get; set; }
+
+        internal SpanFactory Factory { get; private set; }
+
+        internal BlockFactory BlockFactory { get; private set; }
 
         internal abstract RazorSyntaxTree ParseBlock(string document, bool designTime);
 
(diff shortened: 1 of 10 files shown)


test/Microsoft.AspNetCore.Razor.Evolution.Test/Legacy/ParserTestBase.cs
```diff
@@ -547,5 +568,37 @@ private class IgnoreOutputBlock : Block
         {
             public IgnoreOutputBlock() : base(BlockType.Template, new SyntaxTreeNode[0], null) { }
         }
+
+        // Corrects the parents and previous/next information for spans
+        internal class SpancestryCorrector : ParserVisitor
```
NTaylorMullen: Add private ctor so we don't accidentally start newing this thing up

test/Microsoft.AspNetCore.Razor.Evolution.Test/SyntaxTreeVerifier.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    // Verifies recursively that a syntax tree has no gaps in terms of position/location.
+    internal class SyntaxTreeVerifier : ParserVisitor
+    {
+        private SourceLocationTracker _tracker = new SourceLocationTracker(SourceLocation.Zero);
```
NTaylorMullen: readonly

test/Microsoft.AspNetCore.Razor.Evolution.Test/SyntaxTreeVerifier.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    // Verifies recursively that a syntax tree has no gaps in terms of position/location.
+    internal class SyntaxTreeVerifier : ParserVisitor
```
NTaylorMullen: private ctor

rynowak:
Addressed comments and merged 0835de17ecde4fe0560ad649918e47b6e0a3e8da

Issue #915, opened 2016-12-28
Add Preallocated property optimization for TagHelpers


ajaybhargavb:
Follow-up to https://github.com/aspnet/Razor/issues/881, https://github.com/aspnet/Razor/issues/882 and  https://github.com/aspnet/Razor/issues/912

PR #919, opened 2016-12-30
Added TagHelper preallocated attribute optimization pass

ajaybhargavb:
Issue - #915 

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorRuntimeCSharpLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorRuntimeCSharpLoweringPhase.cs
@@ -332,6 +332,16 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                     tagHelperVariableName);
             }
 
+            internal override void VisitAddPreallocatedTagHelperHtmlAttribute(AddPreallocatedTagHelperHtmlAttributeIRNode node)
+            {
+                Context.Writer
+                    .WriteStartInstanceMethodInvocation(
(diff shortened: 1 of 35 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorRuntimeCSharpLoweringPhase.cs
```diff
@@ -332,6 +332,16 @@ internal override void VisitCreateTagHelper(CreateTagHelperIRNode node)
                     tagHelperVariableName);
             }
 
+            internal override void VisitAddPreallocatedTagHelperHtmlAttribute(AddPreallocatedTagHelperHtmlAttributeIRNode node)
```
rynowak: FYI will will need to remove all of these kinds of specialized nodes before we ship. This is ok as a temporary thing while we figure out the design in this space.

rynowak: The intent is that IR nodes have some kind of semantic meaning, not that they stand in for codegen.

rynowak:
Merged as eb0a2a18335c624c1cf6a0c395cc311ac10f0589

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorRuntimeCSharpLoweringPhase.cs
NTaylorMullen: Agree. We can move nearly all of the TagHelper IR nodes into `CSharpStatementIRNode`s which would make these pre-allocated IR bits unneeded.

PR #921, opened 2017-01-03
Clean up SourceLocation

rynowak:
This change makes SourceLocation public and moves some of the undesirable
functions off of the class and into the .Legacy namespace.


pranavkm:
:shipit:

PR #923, opened 2017-01-07
Optimize C# symbol content string allocation

Yves57:
See #873 
Follows the optimization #908 done for the HTML symbols.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpTokenizer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpTokenizer.cs
@@ -168,6 +168,181 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
             }
         }
 
+        // Optimize memory allocation by returning constants for the most frequent cases
+        protected override string GetSymbolContent(CSharpSymbolType type)
+        {
+            var symbolLength = Buffer.Length;
+
+            if (symbolLength == 1)
+            {
+                switch (type)
+                {
+                    case CSharpSymbolType.IntegerLiteral:
+                        switch (Buffer[0])
+                        {
+                            case '0':
+                                return "0";
+                            case '1':
+                                return "1";
+                            case '2':
+                                return "2";
+                            case '3':
+                                return "3";
+                            case '4':
+                                return "4";
+                            case '5':
+                                return "5";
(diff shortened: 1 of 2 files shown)


Yves57:
Continuous Integration problems seems to be out my my PR (seems to be here before my PR). Do you agree?

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpTokenizer.cs
```diff
@@ -168,6 +168,181 @@ protected override StateResult Dispatch()
...
+                    case CSharpSymbolType.Increment:
+                        return "++";
+                    case CSharpSymbolType.LessThanEqual:
+                        return "<=";
+                    case CSharpSymbolType.LeftShift:
+                        return "<<";
+                    case CSharpSymbolType.Equals:
+                        return "==";
+                    case CSharpSymbolType.GreaterThanEqual:
+                        if (Buffer[0] == '=')
```
NTaylorMullen: Why do we need to check this here? Does the tokenizer also treat `=>` as Greater Than Equal?

NTaylorMullen:
> Continuous Integration problems seems to be out my my PR (seems to be here before my PR). Do you agree?

Possible. It's on my plate to investigate that today so we'll avoid merging this until those are resolved whether this causes it or not.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpTokenizer.cs
Yves57: [Yes](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpTokenizer.cs#L111).
I don't know why it is the same token type but...

NTaylorMullen: 🤦‍♂️

rynowak:
@NTaylorMullen is this good to go?

NTaylorMullen:
> @NTaylorMullen is this good to go?

Yup, will merge it

NTaylorMullen:
Merged. 5fa4672c99d9c7ce6ecb6ed72e0c841cc8287095

Issue #851, opened 2016-11-02
Tooling Standalone: Implement TagHelper discovery on top of ISymbol


rynowak:
We need an implementation of TagHelper discovery on top of Roslyn's symbol metadata so that we can do it without building the project.

NTaylorMullen:
8215d28ada0a949f7a43ac3da926d9145d984c05

PR #931, opened 2017-01-13
Port TagHelperDescriptorFactory tests.

NTaylorMullen:
- Ported the existing descriptor factory tests and fixed issues with the current implementation.
- Ported documentation tests with the exception of the localization variants.
- Updated the DefaultTagHelperResolver to filter TagHelper types based on accessibility.
- Added DefaultTagHelperResolver tests.

#851


src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -180,13 +217,21 @@ private IEnumerable<string> GetAllowedChildren(INamedTypeSymbol type, ErrorSink
 
             foreach (var name in allowedChildren)
             {
-                var valid = TryValidateName(
+                if (string.IsNullOrWhiteSpace(name))
```
NTaylorMullen: These changes (and a few other minor ones) were added after the prototype was done.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -180,13 +217,21 @@ private IEnumerable<string> GetAllowedChildren(INamedTypeSymbol type, ErrorSink
...
+                }
+                else if (TryValidateName(
                     name,
-                    whitespaceError: "invalid",
-                    characterErrorBuilder: (invalidCharacter) => "invalid",
-                    errorSink: errorSink);
-
-                if (valid)
+                    invalidCharacter => Workspaces.Resources.FormatTagHelperDescriptorFactory_InvalidRestrictChildrenAttributeName(
+                        nameof(TagHelperTypes.RestrictChildrenAttribute),
```
NTaylorMullen: These bits are iffy but I wasn't sure if it was worth representing these values in the `TagHelperTypes`

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -719,14 +882,34 @@ private bool IsAccessibleProperty(IPropertySymbol property)
...
+                case SpecialType.System_Int16:
+                case SpecialType.System_UInt16:
+                case SpecialType.System_Int32:
+                case SpecialType.System_UInt32:
+                case SpecialType.System_Int64:
+                case SpecialType.System_UInt64:
+                case SpecialType.System_Decimal:
+                case SpecialType.System_Single:
+                case SpecialType.System_Double:
+                case SpecialType.System_String:
```
NTaylorMullen: `ToDisplayString()` has `string`, `long` etc. stay as `string`, `long`. In the old variant of the descriptor factories all types always had their full names returned. I couldn't find any `IsPrimitive` equivalent on the `ITypeSymbol` so feel free to suggest if you know something better.

src/Microsoft.CodeAnalysis.Razor.Workspaces/Microsoft.CodeAnalysis.Razor.Workspaces.csproj
```diff
@@ -7,9 +7,6 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;cshtml;razor</PackageTags>
   </PropertyGroup>
-  <PropertyGroup Label="Configuration">
-    <RootNamespace>Microsoft.CodeAnalysis.Razor</RootNamespace>
```
NTaylorMullen: Resources don't embed correctly with our `./build resx` script.

PR #933, opened 2017-01-13
Add parsing of includes and imports

rynowak:
Also adds the source document to the RazorSyntaxTree and does some cleanup
related to this. This lets us verify which tree goes to which document and
that seems important.

Added basic tests to verify that parsing happens, though it's not being
used for anything right now.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.IO;
 
 namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
 {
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RazorParser.cs
```diff
@@ -25,17 +24,19 @@ public RazorParser(RazorParserOptions options)
...
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
 
-        public virtual RazorSyntaxTree Parse(char[] input) => Parse(((ITextDocument)new SeekableTextReader(input)));
+            var chars = new char[source.Length];
+            source.CopyTo(0, chars, 0, source.Length);
 
-        public virtual RazorSyntaxTree Parse(ITextDocument input) => ParseCore(input);
+            var reader = new SeekableTextReader(chars);
```
rynowak: Going to send another PR when we implement this on top of `RazorSourceDocument` instead of making a copy

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperBinderSyntaxTreePass.cs
```diff
@@ -31,7 +31,7 @@ public RazorSyntaxTree Execute(RazorCodeDocument document, RazorSyntaxTree synta
                 if (errorSink.Errors.Count > 0)
                 {
                     var combinedErrors = CombineErrors(syntaxTree.Diagnostics, errorSink.Errors);
-                    var erroredTree = RazorSyntaxTree.Create(syntaxTree.Root, combinedErrors, syntaxTree.Options);
+                    var erroredTree = RazorSyntaxTree.Create(syntaxTree.Root, syntaxTree.Source, combinedErrors, syntaxTree.Options);
```
rynowak: Going to send another PR which factors this into a convenience method.

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorParsingPhaseTest.cs
```diff
@@ -46,6 +46,46 @@ public void Execute_UsesConfigureParserFeatures()
...
+                TestRazorSourceDocument.Create(),
+            };
+
+            var codeDocument = TestRazorCodeDocument.Create(TestRazorSourceDocument.Create(), imports, includes);
+
+            // Act
+            phase.Execute(codeDocument);
+
+            // Assert
+            Assert.Collection(
```
NTaylorMullen: Eh, this is a bit too simple to assume that the import/include trees are being parsed independently.  I'd add a single line content to each and verify that we get blocks/spans etc. for them.

PR #934, opened 2017-01-13
Modifications to allow building VSIX from build.

pranavkm:


pranavkm:
In the interim where we don't have VS 2017 on all the build agents, we'll run this on a dedicated agent. The version numbers won't align, but it'll unblock the build + signing process.

PR #936, opened 2017-01-17
Proposition to introduce "Symbol factories" in the parser

Yves57:
Proposition to introduce "Symbol factories" in the parser:
- it is a first step to allow memory symbol allocation optimization
- it is a first step to allow to reuse allocated symbol for a next parsing

It may me not definitive (I have different ideas to optimize the factories), but I have made some easy optimizations to show the symbol factories interest: these few dedicated HTML Symbol classes allow symbol allocated memory to go from ~10.1Mb to ~6.5Mb in the "large_mvc_view.cshtml" example.


rynowak:
The idea here is good, but I think this area is 3-4 big things away from taking this kind of PR. We're also focused right now on some more practical matters, namely bootstrapping our tooling experience and getting the set of features completed in `.Evolution` that MVC needs to start using the new codebase. 

=================

**Some notes on our specific proposal**

The `LanguageCharacteristics` and everything with it is a carryover from the olden days when we had VB support. We plan to remove it, as well as most of the other `TSymbol` parameterized types

We're more likely to go super low-abstraction in this area. We don't need more concepts, we need fewer.

Statics are fine when this code is testable.

We'll consider adopting something like [Roslyn's red-green tree](https://blogs.msdn.microsoft.com/ericlippert/2012/06/08/persistence-facades-and-roslyns-red-green-trees/) in this area.

It's not clear yet to what degree we'll make any of this public. Since we don't fully parse the C# code in our 'syntax tree' it's just not the useful to manipulate. IR is public surface area for extensibility.

Yves57:
> Oh and I actually forgot the most important one. Symbol is not the correct terminology. It's a Token

I agree, but it is terminology in Razor :-).

Anyway no problem, it was just a proposition, and thank you for the explanation! And if I can help for any issue, it would be a pleasure for me...

Issue #932, opened 2017-01-13
Add temporary support for ViewComponentTagHelpers


NTaylorMullen:
Seeing that we haven't finalized our extensibility story for Razor.Evolution we need to add a work around to support `ViewComponentTagHelper` discovery. This will need to be painfully hardcoded to make it possible but alas it's only temporary.

NTaylorMullen:
6e647854faa7e28f1a4b3b96e1e6147a72c8cf68

PR #937, opened 2017-01-18
Add support for `ViewComponentTagHelpers`.

NTaylorMullen:
- Hardcoded `ViewComponent` discovery.
- Hardcoded `ViewComponentTagHelperDescriptor` creation.
- Added test to validate that ViewComponents are discovered and transitioned into TagHelpers properly.
- Avoided adding a reference to MVC to prevent circular references. This resulted in custom marker attributes to represent `ViewComponent`s. Also made a lot of use of `ViewComponent` conventions (ending in "ViewComponent").

#932


src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -19,46 +21,82 @@ public DefaultTagHelperResolver(bool designTime)
         public override IReadOnlyList<TagHelperDescriptor> GetTagHelpers(Compilation compilation)
         {
             var results = new List<TagHelperDescriptor>();
+            var errors = new ErrorSink();
```
NTaylorMullen: We don't flow this error sink out today. It'd probably be valuable to do that to inform tooling on bits that go wrong.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -19,46 +21,82 @@ public DefaultTagHelperResolver(bool designTime)
...
+
+            VisitCompilation(visitor, compilation);
+
+            var factory = new ViewComponentTagHelperDescriptorFactory(compilation);
+
+            foreach (var type in types)
+            {
+                try
+                {
+                    var descriptor = factory.CreateDescriptor(type);
```
NTaylorMullen: These bits will be replaced by MVC invocations. Didn't feel that it was worth the time to convert each exception clause into an `errorSink.OnError` call/flow the ErrorSink throughout the factory.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
rynowak: Yep. We need to do a refactor on `THD` or least evaluate the design. Let's plan on doing this today.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -19,46 +21,82 @@ public DefaultTagHelperResolver(bool designTime)
...
             {
                 var descriptors = factory.CreateDescriptors(type, errors);
                 results.AddRange(descriptors);
             }
+        }
 
-            return results;
+        private static void VisitViewComponents(Compilation compilation, List<TagHelperDescriptor> results, ErrorSink errors)
+        {
+            var @interface = compilation.GetTypeByMetadataName(ViewComponentTypes.ViewComponentAttribute);
```
rynowak: should we no-op if this type isn't defined?

rynowak: Have you given any thought to how we'll determine if the user is on `MVC >= 1.1`

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -88,5 +126,68 @@ internal bool IsTagHelper(INamedTypeSymbol symbol)
...
+            internal bool IsViewComponent(INamedTypeSymbol symbol)
+            {
+                if (symbol.DeclaredAccessibility != Accessibility.Public ||
+                    symbol.IsAbstract ||
+                    symbol.IsGenericType)
+                {
+                    return false;
+                }
+
+                return symbol.Name.EndsWith(ViewComponentTypes.ViewComponentSuffix) ||
```
rynowak: There's also `NonViewComponent` attribute to consider, which makes the entire hierarchy not a VC.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ViewComponentTagHelperDescriptorFactory.cs
```diff
@@ -0,0 +1,238 @@
...
+                        componentType.ToDisplayString(FullNameTypeDisplayFormat),
+                        nameof(Task)));
+                }
+            }
+            else if (returnType != null)
+            {
+                // Will invoke synchronously. Method must not return void, Task or Task<T>.
+                if (returnType.SpecialType == SpecialType.System_Void)
+                {
+                    throw new InvalidOperationException(ViewComponentResources.FormatViewComponent_SyncMethod_ShouldReturnValue(
```
rynowak: Are these the errors the VC would report at runtime when called?

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
NTaylorMullen: > should we no-op if this type isn't defined?

If there's no attribute the classes could be defined in a class library using the ViewComponent suffix couldn't they?



> Have you given any thought to how we'll determine if the user is on MVC >= 1.1

Great question. Hmm, I could look for public types added in 1.1 to check for its existence. What do you think?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ViewComponentTagHelperDescriptorFactory.cs
NTaylorMullen: Yes

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
NTaylorMullen: Oh shoot, good catch. I missed that bit.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
NTaylorMullen: Hmm, depending on the approach for detecting 1.1 MVC maybe nooping would be the right approach

PR #938, opened 2017-01-18
Remove support for includes

rynowak:
I was mistaken, this isn't a feature we will need for _ViewStart.


rynowak:
be10bb5f7bbdbe4e3efbe8d84ccfaf495617350d

PR #946, opened 2017-01-24
Add support for new style Roslyn dotless commits.

NTaylorMullen:
- Roslyn swapped the way they performed dotless commit insertions. They went from:
date => date. => DateTime.  to
date => date. => date => DateTime => DateTime.
The problem with the new approach is that date => DateTime would be rejected and therefore force the editor to reparse and reclassify any dots as HTML giving improper IntelliSense.
- Updated Razor implicit expression edit handling to allow identifier => identifier replacements as long as the identifiers didn't result in keyword or directives.
- Added tests to verify the scenarios impacted.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
@@ -85,6 +85,11 @@ namespace Microsoft.AspNetCore.Razor.Editor
                 return HandleDotlessCommitInsertion(target);
             }
 
+            if (IsAcceptableIdentifierReplacement(target, normalizedChange))
+            {
+                return TryAcceptChange(target, normalizedChange);
+            }
+
             if (IsAcceptableReplace(target, normalizedChange))
             {
                 return HandleReplacement(target, normalizedChange);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -85,6 +85,11 @@ protected override PartialParseResult CanAcceptChange(Span target, TextChange no
                 return HandleDotlessCommitInsertion(target);
```
NTaylorMullen: Some of the inbetween states of Roslyn's new dotless commits are handled by the existing dotless commit code; therefore, these are still necessary.

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -149,10 +154,65 @@ private static bool IsSecondaryDotlessCommitInsertion(Span target, TextChange ch
...
+                var symbolEndIndex = symbolStartIndex + symbol.Content.Length;
+
+                // We're looking for the first symbol that contains the TextChange.
+                if (symbolEndIndex > change.OldPosition)
+                {
+                    if (symbolEndIndex < change.OldPosition + change.OldLength)
+                    {
+                        // Change spans multiple symbols.
+                    }
+                    else if (symbol.Type == CSharpSymbolType.Identifier)
```
ToddGrun: Are there any gaps in target.Symbols or are they adjacent to eachother?

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -149,10 +154,65 @@ private static bool IsSecondaryDotlessCommitInsertion(Span target, TextChange ch
...
+        }
+
+        private bool IsAcceptableIdentifierReplacement(Span target, TextChange change)
+        {
+            if (!change.IsReplace)
+            {
+                return false;
+            }
+
+            for (var i = 0; i < target.Symbols.Count; i++)
```
ToddGrun: I assume this is typically a fairly small list?

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -149,10 +154,65 @@ private static bool IsSecondaryDotlessCommitInsertion(Span target, TextChange ch
...
+            {
+                var symbol = target.Symbols[i] as CSharpSymbol;
+
+                if (symbol == null)
+                {
+                    break;
+                }
+
+                var symbolStartIndex = target.Start.AbsoluteIndex + symbol.Start.AbsoluteIndex;
+                var symbolEndIndex = symbolStartIndex + symbol.Content.Length;
```
ToddGrun: nit: symbol.Length might be more clear

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
NTaylorMullen: Should always be adjacent.

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
NTaylorMullen: Ehh, depends on your definition of small. I'd be shocked to see more than a few thousand. 

I thought about binary searching this but figured it'd be a premature optimization.

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
NTaylorMullen: There is no Length property. We need to inspect the Content and then look at the length for symbols. You're probably thinking of Spans 😄

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
anurse: Yeah, a binary search should work since it should be ordered. But I agree we should see how this is to start with and get the tests right. I think a binary search is a pretty simple optimization to add in once we've got the code correct though, and it'll pay off quite a bit.

src/Microsoft.AspNetCore.Razor/Editor/ImplicitExpressionEditHandler.cs
```diff
@@ -149,10 +154,65 @@ private static bool IsSecondaryDotlessCommitInsertion(Span target, TextChange ch
...
+
+                // We're looking for the first symbol that contains the TextChange.
+                if (symbolEndIndex > change.OldPosition)
+                {
+                    if (symbolEndIndex < change.OldPosition + change.OldLength)
+                    {
+                        // Change spans multiple symbols.
```
anurse: Can you just invert this condition and add it to the `else if` clause? Then it can just be a simple `if`

PR #948, opened 2017-01-24
Remove typeof from codegen

rynowak:
The C# codegen was using some of the new names for various TagHelper
constructs instead of the old namespace.

This change reverts to the old names.


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorRuntimeCSharpLoweringPhase.cs
```diff
@@ -524,7 +524,7 @@ public override void VisitSetTagHelperProperty(SetTagHelperPropertyIRNode node)
                     .WriteParameterSeparator()
                     .Write(propertyValueAccessor)
                     .WriteParameterSeparator()
-                    .Write($"global::{typeof(HtmlAttributeValueStyle).FullName}.{node.ValueStyle}")
+                    .Write($"global::Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.{node.ValueStyle}")
```
pranavkm: `const string`

NTaylorMullen:
:shipit:

PR #950, opened 2017-01-26
Enabled Razor.Evolution tests on linux

ajaybhargavb:
Normalized the newlines for the input file being read to CRLF.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorDesignTimeCSharpLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorDesignTimeCSharpLoweringPhase.cs
@@ -10,6 +10,8 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 {
     internal class DefaultRazorDesignTimeCSharpLoweringPhase : RazorCSharpLoweringPhaseBase
     {
+        internal static readonly object NewLineString = "NewLineString";
+
         protected override void ExecuteCore(RazorCodeDocument codeDocument)
         {
             var irDocument = codeDocument.GetIRDocument();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorDesignTimeCSharpLoweringPhase.cs
```diff
@@ -18,9 +20,16 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
             var syntaxTree = codeDocument.GetSyntaxTree();
             ThrowForMissingDependency(syntaxTree);
 
+            var codeWriter = new CSharpCodeWriter();
+            var newLineString = codeDocument.Items[NewLineString];
+            if (newLineString != null)
+            {
+                codeWriter.NewLine = (string)newLineString;
+            }
```
rynowak: Leave a comment explaining why this is necessary.

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestRazorSourceDocument.cs
```diff
@@ -14,25 +16,34 @@ private TestRazorSourceDocument(string content, Encoding encoding, string filena
         {
         }
 
-        public static RazorSourceDocument CreateResource(string path, Encoding encoding = null)
+        public static RazorSourceDocument CreateResource(string path, Encoding encoding = null, bool normalizeNewLines = false)
```
rynowak: Should this just be the default when you're reading a resource? We want **ALL** of our integration tests to do this unless otherwise specified.

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestRazorSourceDocument.cs
```diff
@@ -41,9 +52,24 @@ public static MemoryStream CreateStreamContent(string content = "Hello, World!",
...
+            {
+                content = NormalizeNewLines(content);
+            }
+
             return new TestRazorSourceDocument(content, encoding ?? Encoding.UTF8, "test.cshtml");
         }
+
+        private static string NormalizeNewLines(string content)
+        {
+            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
rynowak: Why is this only on non-windows? We want this to be normalized ALWAYS

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestRazorSourceDocument.cs
ajaybhargavb: I don't think we should do that because a lot of the parser tests call these create methods and we shouldn't be normalizing new lines for those tests.

rynowak: Ok fair enough

rynowak:
:shipit: after addressing minor comments

PR #953, opened 2017-01-26
Fix lowering of taghelpers inside a section

rynowak:
The IR lowering phase was attaching the 'tag helper fields' node to the
builder instead of to the top-level node (document). This meant that
things wouldn't be where we expect when the first tag helper occurrence is
inside a directive block (section).

Found this porting MVC to use the new Razor codebase.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
             ThrowForMissingDependency(syntaxTree);
 
             var builder = RazorIRBuilder.Document();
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -416,8 +418,8 @@ private void DeclareTagHelperFields(TagHelperBlock block)
             {
                 if (_tagHelperFields == null)
                 {
-                    _tagHelperFields = new DeclareTagHelperFieldsIRNode();
-                    _builder.Add(_tagHelperFields);
+                    _tagHelperFields = new DeclareTagHelperFieldsIRNode() { Parent = _document, };
+                    _document.Children.Add(_tagHelperFields);
```
rynowak: This is the fix here, the rest is just plumbing.

rynowak:
6a95002f0e2e7e58382087ec87a7e9128aba1577

PR #955, opened 2017-01-26
Use the TagHelperBinder to dedupe taghelpers

rynowak:
This change does deduplication of taghelpers during the binding/rewriting
phase. This is needed when a taghelper has multiple sets of html
attributes that are required (behaves like an OR). This is used lots in
MVC.

The old codebase used to do this in the codegen phase, but it seems
beneficial to do as early as possible.


rynowak:
d187edbd76a460a64b4c9af5268ed757ed9a968b

PR #957, opened 2017-01-27
Correct an issue with codegen for taghelperprefix

rynowak:
The issue here is that when a taghelper prefix is in use it will be
including in the HTML output, when it should be chopped off.

See the diff in the codegen for examples.


rynowak:
2b2cf6efec02210a16ed72aaadfff03025adf537

Issue #958, opened 2017-01-27
Add support for @addTagHelper and friends


rynowak:
See title. We need to do a pass and make sure we are actually testing all of these.

PR #960, opened 2017-01-27
Added support for AddTagHelper, RemoveTagHelper and TagHelperPrefix directives

ajaybhargavb:
Issue - #958 

~~Didn't add a TagHelperPrefix test because it is done in https://github.com/aspnet/Razor/pull/957~~

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -118,14 +118,59 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
             public override void VisitAddTagHelperSpan(AddTagHelperChunkGenerator chunkGenerator, Span span)
             {
+                _builder.Push(new DirectiveIRNode()
+                {
+                    Name = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Name,
+                    Descriptor = CSharpCodeParser.AddTagHelperDirectiveDescriptor,
+                    Source = BuildSourceSpanFromNode(span),
+                });
+
+                _builder.Add(new DirectiveTokenIRNode()
+                {
+                    Content = chunkGenerator.LookupText,
+                    Descriptor = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Tokens.First(),
+                    Source = BuildSourceSpanFromNode(span),
+                });
+
+                _builder.Pop();
             }
 
             public override void VisitRemoveTagHelperSpan(RemoveTagHelperChunkGenerator chunkGenerator, Span span)
             {
+                _builder.Push(new DirectiveIRNode()
+                {
+                    Name = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Name,
+                    Descriptor = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor,
+                    Source = BuildSourceSpanFromNode(span),
+                });
+
+                _builder.Add(new DirectiveTokenIRNode()
+                {
+                    Content = chunkGenerator.LookupText,
+                    Descriptor = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Tokens.First(),
+                    Source = BuildSourceSpanFromNode(span),
+                });
+
+                _builder.Pop();
             }
 
             public override void VisitTagHelperPrefixDirectiveSpan(TagHelperPrefixDirectiveChunkGenerator chunkGenerator, Span span)
             {
+                _builder.Push(new DirectiveIRNode()
+                {
+                    Name = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor.Name,
+                    Descriptor = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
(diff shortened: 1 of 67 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -118,14 +118,44 @@ public override void VisitEndDirectiveBlock(DirectiveChunkGenerator chunkGenerat
 
             public override void VisitAddTagHelperSpan(AddTagHelperChunkGenerator chunkGenerator, Span span)
             {
+                _builder.Add(new DirectiveTokenIRNode()
+                {
+                    Content = chunkGenerator.LookupText,
+                    Descriptor = new DirectiveTokenDescriptor()
+                    {
+                        Kind = DirectiveTokenKind.String,
+                        Value = chunkGenerator.LookupText
```
NTaylorMullen: Don't set Value, it's used by the Literal `DirectiveTokenKind`. We should nuke the property entirely.

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AddTagHelperDirective_DesignTime.mappings.txt
```diff
@@ -0,0 +1,10 @@
+Source Location: (14:0,14 [17] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AddTagHelperDirective.cshtml)
+|"*, TestAssembly"|
+Generated Location: (423:10,30 [17] )
+|*, TestAssembly";|
```
NTaylorMullen: Line mappings are wrong.

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AddTagHelperDirective_DesignTime.mappings.txt
```diff
@@ -0,0 +1,10 @@
...
+
+Source Location: (14:0,14 [17] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AddTagHelperDirective.cshtml)
```
NTaylorMullen: Looks like you have two duplicate line mappings.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
rynowak: This should be a directive node with a children that are directive tokens.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -31,6 +31,21 @@ internal class CSharpCodeParser : TokenizerBackedParser<CSharpTokenizer, CSharpS
                 .AddType()
                 .Build();
 
+        internal static readonly DirectiveDescriptor AddTagHelperDirectiveDescriptor =
+            DirectiveDescriptorBuilder
+                .Create(SyntaxConstants.CSharp.AddTagHelperKeyword)
+                .Build();
```
NTaylorMullen: We should be adding a string token right? That way we'd be able to just re-use the descriptors tokens instead of creating them new

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AddTagHelperDirective_DesignTime.mappings.txt
NTaylorMullen: Still wrong.

Should also add a non-quoted case to the AddTagHelperDirective.cshtml file.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
rynowak: I agree with @NTaylorMullen - if we're going to use these descriptors to describe the directives, we should make them accurate.

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Evolution.Test/TagHelperBinderSyntaxTreePassTest.cs
```diff
@@ -55,8 +55,16 @@ public void Execute_RewritesTagHelpers()
             Assert.Equal("input", inputTagHelper.TagName);
         }
 
-        [Fact]
-        public void Execute_RewritesTagHelpers_TagHelperMatchesElementTwice()
+        [Theory]
```
NTaylorMullen: No theories, just re-write the test. Can always store the boiler plate descriptors in a property.

Issue #961, opened 2017-01-30
Investigate Blocking Issue with Instrumentation


rynowak:
Adding `DefaultInstrumentationPass` to MVC results in errors like the following:

```
Error @ (60:16,17)(17) - [Code blocks (e.g. @{{var variable = 23;}}) must not appear in non-string tag helper attribute values.
 Already in an expression (code) context. If necessary an explicit expression (e.g. @(@readonly)) may be used.] (16,17) Code blocks (e.g. @{{var variable = 23;}}) must not appear in non-string tag helper attribute values.
 Already in an expression (code) context. If necessary an explicit expression (e.g. @(@readonly)) may be used.
Error @ (60:16,17)(17) - [Code blocks (e.g. @{{var variable = 23;}}) must not appear in non-string tag helper attribute values.
 Already in an expression (code) context. If necessary an explicit expression (e.g. @(@readonly)) may be used.] (16,17) Code blocks (e.g. @{{var variable = 23;}}) must not appear in non-string tag helper attribute values.
 Already in an expression (code) context. If necessary an explicit expression (e.g. @(@readonly)) may be used.
```

We need to understand the issue, add a unit test to Razor, and then re-enable instrumentation and the test `InstrumentedViews_RenderAsExpected` in MVC

Generated Code:
```
#pragma checksum "D:\k\Mvc\test\WebSites\HtmlGenerationWebSite\Views\Catalog_CacheTagHelper\Deals.cshtml" "{ff1816ec-aa5e-4d10-87f7-6f4963833460}" "b1c00e8a952bafc654a066f435da8da8043b8b08"
namespace AspNetCore
{
    #line hidden
    using System;
    using System.Threading.Tasks;
    using System.Linq;
    using System.Collections.Generic;
    using Microsoft.AspNetCore.Mvc;
    using Microsoft.AspNetCore.Mvc.Rendering;
    using Microsoft.AspNetCore.Mvc.ViewFeatures;
    using Microsoft.Extensions.Caching.Memory;
    public class Generated_Deals : Microsoft.AspNetCore.Mvc.Razor.RazorPage<dynamic>
    {
        #line hidden
        #pragma warning disable 0414
        private string __tagHelperStringValueBuffer = null;
        #pragma warning restore 0414
        private global::Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperExecutionContext __tagHelperExecutionContext = null;
[...]

PR #967, opened 2017-02-01
Disable instrumentation for TagHelper attribute values

ajaybhargavb:
Issue - #961 

@rynowak @NTaylorMullen

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultInstrumentationPassTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultInstrumentationPassTest.cs
@@ -112,6 +112,100 @@ namespace Microsoft.AspNetCore.Razor.Evolution
         }
 
         [Fact]
+        public void InstrumentationPass_SkipsCSharpExpression_InsideTagHelperAttribute()
+        {
+            // Arrange
+            var builder = RazorIRBuilder.Document();
+            builder.Push(new TagHelperIRNode()
+            {
+                Source = CreateSource(3)
+            });
+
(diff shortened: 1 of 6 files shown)


test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultInstrumentationPassTest.cs
```diff
@@ -112,6 +112,53 @@ public void InstrumentationPass_SkipsCSharpExpression_WithoutLocation()
         }
 
         [Fact]
+        public void InstrumentationPass_SkipsCSharpExpression_InsideTagHelperAttribute()
```
rynowak: What about a test for a taghelper property?

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/InstrumentationPassIntegrationTest.cs
```diff
@@ -37,6 +37,12 @@ public void BasicTest()
                             Name = "value",
                             PropertyName = "FooProp",
                             TypeName = "System.String"
+                        },
+                        new TagHelperAttributeDescriptor
```
rynowak: Add a comment here explaining why this is significant

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultInstrumentationPassTest.cs
ajaybhargavb: Added

Issue #969, opened 2017-02-02
Remove the Value property from DirectiveTokenDescriptor


NTaylorMullen:
The extensible directive system has the ability to capture a specific token value and provide appropriate validation. In the end this type of token isn't needed in the current scope of directives and more importantly shouldn't impact every token descriptor

NTaylorMullen:
3f5d1bb2d6f5699b45689c73b05b1a69a32c413a

PR #970, opened 2017-02-03
Remove literal directive tokens.

NTaylorMullen:
- Literal directive tokens acted as a way for a user to provide markup bits to be required when parsing a directive.
- Removed source implementations.
- Removed tests validating the feature.

#969

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptor.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptor.cs
@@ -6,7 +6,5 @@ namespace Microsoft.AspNetCore.Razor.Evolution
     public class DirectiveTokenDescriptor
     {
         public DirectiveTokenKind Kind { get; set; }
-
-        public string Value { get; set; }
     }
 }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -1551,20 +1551,6 @@ private void HandleDirective(DirectiveDescriptor descriptor)
...
-                    case DirectiveTokenKind.Literal:
-                        if (string.Equals(CurrentSymbol.Content, tokenDescriptor.Value, StringComparison.Ordinal))
-                        {
-                            AcceptAndMoveNext();
-                        }
-                        else
-                        {
-                            Context.ErrorSink.OnError(
-                                CurrentStart,
-                                LegacyResources.FormatUnexpectedDirectiveLiteral(descriptor.Name, tokenDescriptor.Value),
```
ajaybhargavb: Is this used anywhere else?

NTaylorMullen: Yup

src/Microsoft.AspNetCore.Razor.Evolution/DirectiveTokenDescriptor.cs
```diff
@@ -6,7 +6,5 @@ namespace Microsoft.AspNetCore.Razor.Evolution
     public class DirectiveTokenDescriptor
```
pranavkm: Btw, this feels a bit weird now. It's a class with one enum field. Do we plan on sticking more things on this?

pranavkm: I say this as I realize I need to add more stuff to this type.

Issue #973, opened 2017-02-03
Clear up TagHelperDescriptor API


rynowak:
We need to do some cleaning of the TagHelperDescriptor apis and integrate those changes into MVC and our tooling projects.

ajaybhargavb:
8f9ff1abd9b03838e967358de583fbae2c9fb693

PR #975, opened 2017-02-07
[Design] Clean up TagHelperDescriptor APIs.

NTaylorMullen:
#- Removed all design time descriptors and put their API surface into their corresponding descriptor. Part of removing the design time API surface was removing the tracking of `<Remarks>`, it wasn't used so there's on need to track it until we need it.
- Removed the Type requirement from `TagHelperDescriptor`. With this separation we'll be able to have abstract `TagHelper`s that aren't based on a class implementation.
- Removed Prefix from the `TagHelperDescriptor` API surface. It was a legacy requirement based on how the Razor parser was put together. We can work around this now.
- Stripped correlation information from the immediate `TagHelperDescriptor` API surface. Instead this information is now tracked in `CorrelationRequirement`s. This change means that you will not have multiple `TagHelperDescriptor`s per `TagHelper`; instead it's all tracked in a single descriptor. A side effect of this change waws the renaming of `IsIndexer` => `IsDictionary`.
- Added `ITagHelperDescriptor` to represent `TagHelper`s that implement the `ITagHelper` class. It re-introduces the TypeName association of a `TagHelper`.
- Added `PropertyTagHelperAttributeDescriptor` to represent that an attribute was associated with a property.

#973


NTaylorMullen:
Trying to get feedback on the new API surface. Will wire everything together/fix tests etc. if the surface looks good.

rynowak:
> Removed Prefix from the TagHelperDescriptor API surface. It was a legacy requirement based on how the Razor parser was put together. We can work around this now.

I'm glad you did this, this was something I wanted to mention during requirements gathering, but forgot. 

The prefix is a property of the context, not of an individual taghelper

rynowak:
I'm going to avoid feedback on naming for the first round so we can get the concepts and shape right.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptor.cs
```diff
@@ -0,0 +1,54 @@
...
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+
+                _typeName = value;
+            }
+        }
```
rynowak: This should probably also include the assembly name

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperAttributeDescriptor.cs
```diff
@@ -127,30 +113,8 @@ public string TypeName
...
-            if (IsIndexer)
-            {
-                return name.StartsWith(Name, StringComparison.OrdinalIgnoreCase);
-            }
-            else
-            {
-                return string.Equals(name, Name, StringComparison.OrdinalIgnoreCase);
-            }
-        }
+        public string Summary { get; set; }
```
rynowak: I would just make this `Documentation` + `DocumentationId`, and have us return the entire doc blob. Let VS parse it if that's what they want to do.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperAttributeDescriptor.cs
```diff
@@ -127,30 +113,8 @@ public string TypeName
...
         /// <summary>
-        /// The <see cref="TagHelperAttributeDesignTimeDescriptor"/> that contains design time information about
-        /// this attribute.
-        /// </summary>
-        public TagHelperAttributeDesignTimeDescriptor DesignTimeDescriptor { get; set; }
-
-        /// <summary>
-        /// Determines whether HTML attribute <paramref name="name"/> matches this
-        /// <see cref="TagHelperAttributeDescriptor"/>.
+        /// A summary of how to use a tag helper.
```
rynowak: This should probably have a display name

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptor.cs
```diff
@@ -166,71 +102,23 @@ public IEnumerable<TagHelperAttributeDescriptor> Attributes
...
         /// <remarks><c>null</c> indicates all children are allowed.</remarks>
         public IEnumerable<string> AllowedChildren { get; set; }
 
         /// <summary>
-        /// Get the name of the HTML element required as the immediate parent.
+        /// A summary of how to use a tag helper.
         /// </summary>
-        /// <remarks><c>null</c> indicates no restriction on parent tag.</remarks>
-        public string RequiredParent { get; set; }
+        public string Summary { get; set; }
```
rynowak: Is this the docs? Let's give it the same treatment here.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptor.cs
```diff
@@ -166,71 +102,23 @@ public IEnumerable<TagHelperAttributeDescriptor> Attributes
...
 
         /// <summary>
```
rynowak: Add DisplayName

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptor.cs
NTaylorMullen: I was thinking `AssemblyName` made more sense at the `TagHelperDescriptor` level to coincide with `@addTagHelper`/`@removeTagHelper`. Both concepts that are at the core of Razor.

If I moved `AssemblyName` to this level that'd mean that anyone building their own type of `TagHelperDescriptor` would then also need to handle its discovery mechanism. Aka `ViewTagHelperDescriptor` would need to also plug into parsing / correlation of `TagHelper`s.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperAttributeDescriptor.cs
NTaylorMullen: Interesting... Hmmm. If we did that then we wouldn't need a `DisplayName`. I think I like this approach.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptor.cs
rynowak: Gross. Let's discuss more

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperAttributeDescriptor.cs
NTaylorMullen: Scratch that. Would still need display name.

NTaylorMullen:
Playing with adding `DisplayName` a bit and it got me thinking about taking a descriptor builder approach like we have with razor directives. For instance a `DisplayName` of a `PropertyTagHelperAttributeDescriptor` should be: `{SimpleTypeName} {ContainingTypeName}.{PropertyName}`. I think what I'm trying to get at is it's incredibly easy to fall into pits of failure when constructing `TagHelperDescriptor`s (purpose of this PR is to make `TagHelperDescriptor`s more understandable). It'd be a decent amount of work but I think a builder approach makes a lot of sense and will make our lives easier in the future.

With a descriptor builder approach we could have something like:

```C#
var anchorTagHelper = ITagHelperDescriptorBuilder.Create(assemblyName)
    .Requirement(requirement => requirement
        .TagName("a")
        .Attribute(builder => builder.Name("asp-action"))
        .TagStructure(TagStructure.SelfClosing))
    .Requirement(requirement => requirement
        .TagName("a")
        .Attribute(builder => builder.NamePrefix("asp-route-"))
        .Parent("li"));
    .BoundAttribute(attributeName: "asp-action", propertyName: "Action", typeName: "System.String")
    .BoundAttribute(builder => builder
        .Name("asp-route-")
        .PropertyName("RouteValues")
        .TypeName("RouteValueDictionary``")
        .AsDictionary())
    .Metadata("SomethingInPropertyBag", "123")
    .AllowChildren(...) // Doesn't really make sense on an anchor tag helper but you get the point
    .Build();
```

Thoughts?

rynowak:
> Thoughts?

It's a non-public detail so we can do whatever makes us happy. What's important with this work is that we stabilize the public API that will be providing to VS. Creating taghelper descriptors is not an extensibility scenario right now.

NTaylorMullen:
> It's a non-public detail so we can do whatever makes us happy. What's important with this work is that we stabilize the public API that will be providing to VS. Creating taghelper descriptors is not an extensibility scenario right now.

How important is `DisplayName` to you? VS would technically have all the info necessary to create a custom display name given we provide all the documentation.

rynowak:
If we don't need display name on the API surface we can leave it out for now.

src/Microsoft.AspNetCore.Razor.Evolution/CorrelationRuleBuilder.cs
```diff
@@ -0,0 +1,104 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public class CorrelationRuleBuilder
```
rynowak: sealed or abstract 😆

rynowak:
Here's the gist from the naming review:
https://gist.github.com/rynowak/5f26e0922a5579853a041b2a26dd151f

PR #979, opened 2017-02-08
Parse string directive tokens correctly

pranavkm:


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -1553,7 +1553,8 @@ private void HandleDirective(DirectiveDescriptor descriptor)
                         outputKind = SpanKind.Code;
                         break;
                     case DirectiveTokenKind.String:
-                        AcceptAndMoveNext();
+                        AcceptUntil(CSharpSymbolType.WhiteSpace, CSharpSymbolType.NewLine);
+                        NextToken();
```
NTaylorMullen: This skips unneeded whitespacce inbetween directive tokens.

test/Microsoft.AspNetCore.Razor.Evolution.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -751,6 +751,29 @@ public void OptionalDirectiveTokens_AreSkipped()
                     Factory.Span(SpanKind.Markup, " ", markup: false).Accepts(AcceptedCharacters.WhiteSpace)));
         }
 
+        [Theory]
```
NTaylorMullen: - no theories.
- Should ensure there's a test that validates that the whitespace after your string token doesn't get ignored.

pranavkm:
🆙 📅

PR #982, opened 2017-02-08
Handle single line directives properly

ajaybhargavb:
Without this change it throws a parser error if there is whitespace after the semicolon.

@NTaylorMullen @pranavkm @rynowak


rynowak:
solid 👍

Issue #947, opened 2017-01-24
RazorSourceDocument throws when stream is empty


rynowak:
Hit while testing MVC

```
Result StackTrace:	
at System.IO.StreamReader..ctor(Stream stream, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize, Boolean leaveOpen)
   at Microsoft.AspNetCore.Razor.Evolution.RazorSourceDocument.ReadFromInternal(Stream stream, String filename, Encoding encoding)
   at Microsoft.AspNetCore.Razor.Evolution.RazorSourceDocument.ReadFrom(Stream stream, String filename, Encoding encoding)
   at Microsoft.AspNetCore.Mvc.Razor.Internal.RazorCompilationService..ctor(ICompilationService compilationService, RazorEngine engine, RazorProject project, IRazorViewEngineFileProviderAccessor fileProviderAccessor, ILoggerFactory loggerFactory) in D:\k\Mvc\src\Microsoft.AspNetCore.Mvc.Razor\Internal\RazorCompilationService.cs:line 69
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
[...]

PR #983, opened 2017-02-09
Allow RazorSourceDocument to consume empty streams.

NTaylorMullen:
- Added tests to validate how encoding flows when used with empty streams in addition to their usability (not throwing).

#947

### DIFF
--- a/Razor.sln
+++ b/Razor.sln
@@ -14,6 +14,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Razor.
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{F8C12DD6-659D-405A-AA27-FB22AD92A010}"
 	ProjectSection(SolutionItems) = preProject
+		build\common.props = build\common.props
 		NuGet.config = NuGet.config
 	EndProjectSection
 EndProject
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
@@ -50,31 +50,38 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         private static RazorSourceDocument ReadFromInternal(Stream stream, string filename, Encoding encoding)
         {
-            var reader = new StreamReader(
-                stream,
-                encoding ?? Encoding.UTF8,
-                detectEncodingFromByteOrderMarks: true,
-                bufferSize: (int)stream.Length,
-                leaveOpen: true);
+            var streamLength = (int)stream.Length;
+            var content = string.Empty;
+            var contentEncoding = encoding ?? Encoding.UTF8;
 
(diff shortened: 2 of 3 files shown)


Razor.sln
```diff
@@ -35,9 +36,9 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.VisualStudio.Razo
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.CodeAnalysis.Razor.Workspaces.Test", "test\Microsoft.CodeAnalysis.Razor.Workspaces.Test\Microsoft.CodeAnalysis.Razor.Workspaces.Test.csproj", "{053B72C3-67BB-4C0D-BF94-34D435E2E42F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.CodeAnalysis.Razor", "src\Microsoft.CodeAnalysis.Razor\Microsoft.CodeAnalysis.Razor.csproj", "{42403DAF-F0BC-4F3A-B7F2-46D7013345D8}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.CodeAnalysis.Razor", "src\Microsoft.CodeAnalysis.Razor\Microsoft.CodeAnalysis.Razor.csproj", "{42403DAF-F0BC-4F3A-B7F2-46D7013345D8}"
```
NTaylorMullen: No idea...

rynowak: #YOLO

src/Microsoft.AspNetCore.Razor.Evolution/RazorSourceDocument.cs
```diff
@@ -50,31 +50,38 @@ public static RazorSourceDocument ReadFrom(Stream stream, string filename, Encod
...
-                    encoding = reader.CurrentEncoding;
-                }
-                else if (encoding != reader.CurrentEncoding)
+            if (streamLength > 0)
+            {
+                var reader = new StreamReader(
+                    stream,
+                    contentEncoding,
+                    detectEncodingFromByteOrderMarks: true,
+                    bufferSize: streamLength,
```
rynowak: We could probably just solve this issue as well with a `4096` here. This code also has the problem that it will try to read the buffer in one whole go which is wicked bad.

NTaylorMullen: Hmm, fair. I'll update.

NTaylorMullen: Actually, looking at Roslyn's implementation, they do the same thing. If length 0 they utilize `string.Empty`. Segmenting the buffer into 4096 pieces would be a separate issue.

rynowak: Ok fine log an issue for this then :-p

This is not black magic, it's just a streamreader.

NTaylorMullen: https://github.com/aspnet/Razor/issues/997

PR #984, opened 2017-02-09
Remove DirectiveIRNode from the IR document after the default directi…

ajaybhargavb:
…ve classifier has run

@NTaylorMullen @rynowak 

This pass has the order `DefaultDirectiveClassifierOrder = 2900` which runs after any custom directive classifier which will have the order `DirectiveClassifierOrder = 2100`. So it should be safe the remove the nodes here.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
@@ -50,8 +50,6 @@ namespace Microsoft.AspNetCore.Razor.Evolution
             {
                 if (string.Equals(node.Name, CSharpCodeParser.FunctionsDirectiveDescriptor.Name, StringComparison.Ordinal))
                 {
-                    node.Parent.Children.Remove(node);
-
                     foreach (var child in node.Children.Except(node.Tokens))
                     {
                         child.Parent = _classNode;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultDirectiveIRPass.cs
```diff
@@ -94,18 +89,8 @@ public override void VisitDirective(DirectiveIRNode node)
...
-                else if (string.Equals(node.Name, CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Name, StringComparison.Ordinal))
-                {
-                    node.Parent.Children.Remove(node);
-                }
-                else if (string.Equals(node.Name, CSharpCodeParser.TagHelperPrefixDirectiveDescriptor.Name, StringComparison.Ordinal))
-                {
-                    node.Parent.Children.Remove(node);
-                }
+
+                node.Parent.Children.Remove(node);
```
NTaylorMullen: Hmm, this kinda overloads this passes purpose to also clean up `DirectiveIRNodes`. This could be fine for now but we may want to move it to a different pass later; that way users can get our default directive support without us nuking every `DirectiveIRNode` under the sun 😄

ajaybhargavb: Understood. Leaving this here for now to unblock porting Mvc tests

ajaybhargavb:
🆙 📅

PR #987, opened 2017-02-09
Implement a simple base for document classifiers

rynowak:
This should allow us to de-dupe a lot of code in MVC.


rynowak:
The next PR will introduce some better defaults for the `DefaultDocumentClassifierPass`. I wanted to separate that out so we're not looking at IR/codegen diffs in the same PR as a refactor.

PR #988, opened 2017-02-10
Fix a failing test

rynowak:
This poor poor test was left behind at some point when `[ConditionalTheory]` broke and the baseline file got out of date.

Now that we've largely fixed this it's running again and failing since it wasn't updated when we made changes to the codegen.

I changed this to `[ConditionalFact]` as it's not a theory.


rynowak:
FYI only

PR #998, opened 2017-02-13
A gentle introduction to RuntimeTarget

rynowak:
This is a new abstraction that represents the api surface available for
codegen to target. Every kind of document should have an associated
RuntimeTarget or just use the default.

To prevent breakage, our DocumentClassifierBase class will provide a
default API set to implementors (like MVC).

I haven't fundamentally changed how codegen is done yet, I've just hidden
it behind a new abstraction. The RuntimeTarget now is also responsible for
selecting between design time and runtime.

The bulk of the noise here is from splitting a lot of the codegen stuff
into its own files.


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpLiteralCodeConventions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
+{
+    internal class CSharpLiteralCodeConventions : CSharpRenderingConventions
```
rynowak: Most of the files in this directory and in particular all of the *conventions* classes aren't being changed here just move to their own files.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DefaultRuntimeTarget.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public override TExtension GetExtension<TExtension>()
+        {
+            throw new NotImplementedException();
+        }
+
+        public override bool HasExtension<TExtension>()
+        {
+            throw new NotImplementedException();
+        }
```
rynowak: I just wanted to sketch out some of the API in this PR. The next change will actually use it.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/IRuntimeTargetBuilder.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
+{
+    public interface IRuntimeTargetBuilder
```
rynowak: There's nothing to configure yet.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RuntimeTarget.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            else
+            {
+                AddRuntimeDefaults(builder);
+            }
+
+            if (configure != null)
+            {
+                configure?.Invoke(builder);
+            }
```
rynowak: I couldn't pass up the PranavPoints

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorCSharpLoweringPhase.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            var combinedErrors = syntaxTree.Diagnostics.Concat(renderingContext.ErrorSink.Errors).ToList();
+            var csharpDocument = new RazorCSharpDocument()
+            {
+                GeneratedCode = renderingContext.Writer.GenerateCode(),
+                LineMappings = renderingContext.LineMappings,
+                Diagnostics = combinedErrors
+            };
+
+            codeDocument.SetCSharpDocument(csharpDocument);
```
rynowak: The code that's here is a merger of the old two implementations. The new functionality is to check for the `RuntimeTarget` and then use it.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RuntimeTarget.cs
NTaylorMullen: lol you don't need the `if != null` then 😄

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DefaultRuntimeTarget.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                return new DesignTimeCSharpRenderer(context);
+            }
+            else
+            {
+                return new RuntimeCSharpRenderer(context);
+            }
+        }
+
+        public override TExtension GetExtension<TExtension>()
```
NTaylorMullen: I might be missing something but shouldn't these have where clauses?

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -0,0 +1,355 @@
...
+        {
+            const string ItemParameterName = "item";
+            const string TemplateWriterName = "__razor_template_writer";
+
+            Context.Writer
+                .Write(ItemParameterName).Write(" => ")
+                .WriteStartNewObject("Microsoft.AspNetCore.Mvc.Razor.HelperResult" /* ORIGINAL: TemplateTypeName */);
+
+            var initialRenderingConventions = Context.RenderingConventions;
+            var redirectConventions = new CSharpRedirectRenderingConventions(TemplateWriterName, Context.Writer);
```
NTaylorMullen: Shouldn't need these bits at design time since all write calls result in `__o = ...`

rynowak:
Yeah I didn't change any of the codegen features, I just moved them around.

rynowak:
0dbf62196caf49431aaec465bc280ef8cf6c8d9e

PR #1002, opened 2017-02-14
Part 2 of RuntimeTarget

rynowak:
Introducing ExtensionIRNode and an implementation of templates based on
the new feature set.

Now TemplateIRNode is-a ExtensionIRNode. It's implemented using just
extensibility and isn't part of the standard razor codegen. I'm adding it
to the RazorEngine so that it's still there by default.

I've also included a pattern for visitors to special case
ExtensionIRNode-derived classes that they know about. This requires a
little bit of boilerplate but makes it easy to traverse just the nodes you
care about while keeping the set of nodes open.

For now the general codegen feature still hasn't had a refactor, but this
opens things up for us to start finishing things like MVC's @inject
directive.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRedirectRenderingConventions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRedirectRenderingConventions.cs
@@ -7,22 +7,22 @@ namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
 {
     internal class CSharpRedirectRenderingConventions : CSharpRenderingConventions
     {
-        private readonly string _redirectWriter;
-
         public CSharpRedirectRenderingConventions(string redirectWriter, CSharpCodeWriter writer) 
             : base(writer)
         {
-            _redirectWriter = redirectWriter;
+            RedirectWriter = redirectWriter;
         }
 
-        public override string StartWriteMethod => "WriteTo(" + _redirectWriter + ", " /* ORIGINAL: WriteToMethodName */;
+        public string RedirectWriter { get; }
 
-        public override string StartWriteLiteralMethod => "WriteLiteralTo(" + _redirectWriter + ", " /* ORIGINAL: WriteLiteralToMethodName */;
+        public override string StartWriteMethod => "WriteTo(" + RedirectWriter + ", " /* ORIGINAL: WriteToMethodName */;
 
-        public override string StartBeginWriteAttributeMethod => "BeginWriteAttributeTo(" + _redirectWriter + ", " /* ORIGINAL: BeginWriteAttributeToMethodName */;
+        public override string StartWriteLiteralMethod => "WriteLiteralTo(" + RedirectWriter + ", " /* ORIGINAL: WriteLiteralToMethodName */;
 
-        public override string StartWriteAttributeValueMethod => "WriteAttributeValueTo(" + _redirectWriter + ", " /* ORIGINAL: WriteAttributeValueToMethodName */;
+        public override string StartBeginWriteAttributeMethod => "BeginWriteAttributeTo(" + RedirectWriter + ", " /* ORIGINAL: BeginWriteAttributeToMethodName */;
 
-        public override string StartEndWriteAttributeMethod => "EndWriteAttributeTo(" + _redirectWriter /* ORIGINAL: EndWriteAttributeToMethodName */;
+        public override string StartWriteAttributeValueMethod => "WriteAttributeValueTo(" + RedirectWriter + ", " /* ORIGINAL: WriteAttributeValueToMethodName */;
+
+        public override string StartEndWriteAttributeMethod => "EndWriteAttributeTo(" + RedirectWriter /* ORIGINAL: EndWriteAttributeToMethodName */;
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRenderingContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRenderingContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using Microsoft.AspNetCore.Razor.Evolution.Intermediate;
 
 namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
 {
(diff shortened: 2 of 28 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRedirectRenderingConventions.cs
```diff
@@ -7,22 +7,22 @@ namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
...
-
         public CSharpRedirectRenderingConventions(string redirectWriter, CSharpCodeWriter writer) 
             : base(writer)
         {
-            _redirectWriter = redirectWriter;
+            RedirectWriter = redirectWriter;
         }
 
-        public override string StartWriteMethod => "WriteTo(" + _redirectWriter + ", " /* ORIGINAL: WriteToMethodName */;
+        public string RedirectWriter { get; }
```
rynowak: Testing 👍

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRenderingContext.cs
```diff
@@ -43,5 +44,7 @@ public CSharpRenderingConventions RenderingConventions
         public RazorParserOptions Options { get; set; }
 
         public TagHelperRenderingContext TagHelperRenderingContext { get; set; }
+
+        public Action<RazorIRNode> RenderChildren { get; set; }
```
rynowak: This is how you recurse when writing out a node.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DefaultRuntimeTarget.cs
```diff
@@ -1,39 +1,61 @@
...
         private readonly RazorParserOptions _options;
 
-        public DefaultRuntimeTarget(RazorParserOptions options)
+        public DefaultRuntimeTarget(RazorParserOptions options, IEnumerable<IRuntimeTargetExtension> extensions)
         {
             _options = options;
+            Extensions = extensions.ToArray();
         }
 
+        public IRuntimeTargetExtension[] Extensions { get; }
```
rynowak: Testing 👍

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/ExtensionIRNode.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Evolution.CodeGeneration;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    public abstract class ExtensionIRNode : RazorIRNode
+    {
+        internal abstract void WriteNode(RuntimeTarget target, CSharpRenderingContext context);
```
rynowak: This is internal right now because the rendering context is internal. We'll need a follow up to polish up this API and finalize it.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/TemplateIRNode.cs
```diff
@@ -21,7 +22,7 @@ public override void Accept(RazorIRNodeVisitor visitor)
                 throw new ArgumentNullException(nameof(visitor));
             }
 
-            visitor.VisitTemplate(this);
+            AcceptExtensionNode<TemplateIRNode>(this, visitor);
```
rynowak: This is the standard boilerplate for the extension node pattern.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -54,6 +55,10 @@ internal static void AddDefaults(IRazorEngineBuilder builder)
             builder.Phases.Add(new DefaultRazorIRPhase());
             builder.Phases.Add(new DefaultRazorCSharpLoweringPhase());
 
+            // General extensibility
+            builder.Features.Add(new DefaultRazorDirectiveFeature());
+            builder.Features.Add(new DefaultRazorTargetExtensionFeature());
```
rynowak: Adding these by default because practically speaking they are always gonna be there.

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngineBuilderExtensions.cs
```diff
@@ -12,7 +13,7 @@ public static IRazorEngineBuilder AddDirective(this IRazorEngineBuilder builder,
         {
             if (builder == null)
             {
-                throw new ArgumentNullException(nameof(directive));
+                throw new ArgumentNullException(nameof(builder));
```
rynowak: This was wrong

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngineBuilderExtensions.cs
```diff
@@ -26,6 +27,24 @@ public static IRazorEngineBuilder AddDirective(this IRazorEngineBuilder builder,
             return builder;
         }
 
+        public static IRazorEngineBuilder AddTargetExtension(this IRazorEngineBuilder builder, IRuntimeTargetExtension extension)
```
NTaylorMullen: I'm starting to think that "Target" might not be the best name. I fail at picking names but `RuntimeTarget`/`AddTargetExtension` don't make me think of "here's a place to tie into code generation"

rynowak: Let's quickly come up with a better name, or we can file an issue to revisit. I want to get this done to unblock use of it in MVC.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/TemplateIRNode.cs
rynowak: No you can't. You don't know what `TNode` is. If this were C++ sure :trollface:

rynowak:

src/Microsoft.AspNetCore.Razor.Evolution/RazorEngineBuilderExtensions.cs
NTaylorMullen: Talked offline.

rynowak:
7a1a6dd1d6726aa9a7b531bfa405d6c829eead1e

PR #1009, opened 2017-02-16
DefaultRazorIRLoweringPhase is too agressive

rynowak:
This change fixes a bug where DefaultRazorIRLoweringPhase is too
aggressive in merging HTML spans. You can hit the bug by delimiting two
html spans with a metacode character like:
```
<foo>@{ <bar/> }</foo>
```
The lowering phase will combine these HTML nodes, which is invalid as they
don't have contiguous spans.

The change here is to merge spans only when they both have an invalid
location or are contiguous.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Razor.Evolution.Intermediate;
 using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using System.Diagnostics;
 
 namespace Microsoft.AspNetCore.Razor.Evolution
 {
@@ -176,16 +177,21 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                 _builder.Pop();
             }
 
-            protected SourceSpan BuildSourceSpanFromNode(SyntaxTreeNode node)
+            protected SourceSpan? BuildSourceSpanFromNode(SyntaxTreeNode node)
             {
                 var location = node.Start;
-                var sourceRange = new SourceSpan(
+                if (location == SourceLocation.Undefined)
+                {
+                    return null;
+                }
+
+                var span = new SourceSpan(
                     node.Start.FilePath ?? Filename,
                     node.Start.AbsoluteIndex,
                     node.Start.LineIndex,
                     node.Start.CharacterIndex,
                     node.Length);
-                return sourceRange;
+                return span;
             }
         }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -399,27 +404,41 @@ public override void VisitMarkupSpan(MarkupChunkGenerator chunkGenerator, Span s
...
                 {
                     var existingHtmlContent = (HtmlContentIRNode)currentChildren[currentChildren.Count - 1];
-                    existingHtmlContent.Content = string.Concat(existingHtmlContent.Content, span.Content);
 
-                    if (existingHtmlContent.Source != null)
+                    var source = BuildSourceSpanFromNode(span);
+                    if (existingHtmlContent.Source == null && source == null)
                     {
-                        var contentLength = existingHtmlContent.Source.Value.Length + span.Content.Length;
+                        Combine(existingHtmlContent, span);
```
NTaylorMullen: I'd be curious if this is ever hit in our code gen tests. If so, what case?

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -399,27 +404,41 @@ public override void VisitMarkupSpan(MarkupChunkGenerator chunkGenerator, Span s
...
+                    Source = BuildSourceSpanFromNode(span),
+                });
+            }
+            private void Combine(HtmlContentIRNode node, Span span)
+            {
+                node.Content = node.Content + span.Content;
+                if (node.Source != null)
+                {
+                    node.Source = new SourceSpan(
+                        node.Source.Value.FilePath ?? Filename,
```
NTaylorMullen: So we don't have this magic in two places we could re-purpose BuildSourceSpanFromNode to have an alternative that takes in a SourceSpan and a length.

It'd be nice to ensure there's only 1 location doing this file path logic is all.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
rynowak: No idea, I'm just being through. I can take it out if you don't like it.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
rynowak: Sure

rynowak: Actually this really shouldn't be here. We should always have a filename on `node.Source`.

rynowak: Talked in person and agreed to resolve this by putting a `Debug.Assert` here.

PR #1010, opened 2017-02-16
Prkrishn/1003

pranavkm:


test/Microsoft.CodeAnalysis.Razor.Test/DefaultTagHelperDescriptorFactoryTest.cs
```diff
@@ -2098,11 +2098,11 @@ public void CreateDescriptors_BuildsDescriptorsFromSimpleTypes()
         {
             // Arrange
             var errorSink = new ErrorSink();
-            var objectAssemblyName = typeof(object).GetTypeInfo().Assembly.GetName().Name;
+            var objectAssemblyName = typeof(Enumerable).GetTypeInfo().Assembly.GetName().Name;
             var expectedDescriptor =
-                CreateTagHelperDescriptor("object", "System.Object", objectAssemblyName);
+                CreateTagHelperDescriptor("enumerable", "System.Linq.Enumerable", objectAssemblyName);
```
pranavkm: System.object on netcoreapp1.1 gets type forwarded to a different assembly (`System.PrivateCoreLib`). Consequently, it differs from the name in the deps \ reference assembly viz. `System.Runtime`. Using the `typeof(object).Assembly` doesn't work because it doesn't cover enough assemblies. I just picked up some random type instead

natemcmaster:
Don't forget to remove the exclusions from repo.props.

PR #1012, opened 2017-02-16
Add `RazorLanguageServiceException` to capture unexpected exceptions.

NTaylorMullen:
- Consumers will then be able to catch the `RazorLanguageServiceException`s directly and log them as needed.


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorEngineAssemblyResolver.cs
```diff
@@ -15,15 +15,25 @@ internal class DefaultRazorEngineAssemblyResolver : IRazorEngineAssemblyResolver
...
-            var assemblies = GetRazorCustomizationAssemblies(compilation);
+            try
+            {
+                var compilation = await project.GetCompilationAsync().ConfigureAwait(false);
+                var assemblies = GetRazorCustomizationAssemblies(compilation);
+
+                if (assemblies.Count == 0)
+                {
+                    Console.WriteLine();
+                }
```
rynowak: Get rid of this trash 👍

PR #1013, opened 2017-02-16
Refactor ParserVisitor

rynowak:
We need the visitor to allow control over whether to recurse or not into
something. This change makes the old ParserVisitor class behave much more
like the newer IR Walkers.

We need this for the tokens refactor because IR lowering will not be just
a trivial visitor anymore in the future.


rynowak:
bbeb485405911351f590f7389d7f4f461f445125

PR #1017, opened 2017-02-17
Replace RazorError with RazorDiagnostics in public API

ajaybhargavb:
@rynowak


test/Microsoft.AspNetCore.Razor.Test/Framework/ParserTestBase.cs
```diff
@@ -492,7 +492,7 @@ public static string FormatErrors(IEnumerable<RazorError> errors)
             }
 
             var builder = new StringBuilder();
-            foreach (RazorError err in errors)
+            foreach (var err in errors)
```
rynowak: `error` - dont' be weird 👎

PR #1020, opened 2017-02-17
Introducing RazorIRToken

rynowak:
This IR node will be part of the new token model for IR. It will be used
by all nodes that contain user content. Going forward, tokens will be the
thing that contains text and produces line mappings.

This commit just introduces the class.


rynowak:
965ae5490f4105fd934e0c6038f529fbe61ec118

Issue #1022, opened 2017-02-17
Filter TagHelperDescriptors based on passed in assembly in Razor language service.


NTaylorMullen:
This is for the initial TagHelper tooling release.

NTaylorMullen:
7a2f89b5de348f1205d3808fd70ca9c34bda8174

PR #1025, opened 2017-02-17
Add assembly name filtering to GetTagHelpers calls.

NTaylorMullen:
- Updated the Razor extension to properly discover all assembly names available to a project and use those as a TagHelper filter.

#1022

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Razor.Evolution;
 using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using System.Linq;
 
 namespace Microsoft.CodeAnalysis.Razor
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -17,12 +18,12 @@ public DefaultTagHelperResolver(bool designTime)
...
 
-        public override TagHelperResolutionResult GetTagHelpers(Compilation compilation)
+        public override TagHelperResolutionResult GetTagHelpers(Compilation compilation, IEnumerable<string> assemblyNameFilters)
         {
             var descriptors = new List<TagHelperDescriptor>();
             var errors = new ErrorSink();
 
-            VisitTagHelpers(compilation, descriptors, errors);
+            VisitTagHelpers(compilation, assemblyNameFilters, descriptors, errors);
             VisitViewComponents(compilation, descriptors, errors);
```
rynowak: Shouldn't VC's have the same filtering logic?

tooling/Microsoft.VisualStudio.RazorExtension/RazorInfo/RazorInfoViewModel.cs
```diff
@@ -14,6 +14,7 @@
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.TextManager.Interop;
+using System.IO;
```
rynowak: Sort

NTaylorMullen:
> Let's resolve the issue with VCs.

WIll do and then i'll merge this.

PR #1026, opened 2017-02-17
Replace CSharpIRToken with RazorIRToken

rynowak:
Deletes CSharpIRToken to use the more general RazorIRToken class.

Rather than using the visitor to visit tokens, now writing a
CSharpExpresionIRNode is an 'atom', and will write its tokens itself.


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -278,9 +276,15 @@ public override void VisitDeclareTagHelperFields(DeclareTagHelperFieldsIRNode no
...
-            var lineMapping = new LineMapping(sourceLocation, generatedLocation);
+            if (node.Source == null)
+            {
+                return;
+            }
+
+            var source = node.Source.Value;
+            
+            var generatedLocation = new SourceSpan(Context.Writer.GetCurrentSourceLocation(), source.Length);
+            var lineMapping = new LineMapping(source, generatedLocation);
```
rynowak: The callers don't consistently check whether or not they have a `.source` which would make this throw if handed a user-synthesized token

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -38,14 +33,17 @@ public override void VisitCSharpExpression(CSharpExpressionIRNode node)
...
+                        if (token != null && token.IsCSharp)
                         {
-                            AddLineMappingFor(childNode);
+                            AddLineMappingFor(token);
+                            Context.Writer.Write(token.Content);
+                        }
+                        else
+                        {
+                            // There may be something else inside the expression like a Template or another extension node.
+                            Visit(node.Children[i]);
```
rynowak: This looks like our opportunity to tighten up the set of allowed/valid nodes. For instance a template is implemented as an extension node, but it's expected to appear in an expession context.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
NTaylorMullen: Is that a new behavior?

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -305,14 +309,14 @@ private void AddLineMappingFor(RazorIRNode node)
 
                 Context.Writer.Write(((HtmlContentIRNode)node).Content);
             }
-            else if (node is CSharpTokenIRNode)
+            else if (node is RazorIRToken token && token.IsCSharp)
```
NTaylorMullen: 👏

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -376,11 +376,19 @@ public override void VisitExpressionSpan(ExpressionChunkGenerator chunkGenerator
                     }
                 }
 
-                _builder.Add(new CSharpTokenIRNode()
+                if (_builder.Current is CSharpExpressionIRNode)
```
NTaylorMullen: Why the new check?

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -376,11 +376,19 @@ public override void VisitExpressionSpan(ExpressionChunkGenerator chunkGenerator
...
+                    _builder.Add(new RazorIRToken()
+                    {
+                        Content = span.Content,
+                        Kind = RazorIRToken.TokenKind.CSharp,
+                        Source = BuildSourceSpanFromNode(span),
+                    });
+                }
+                else
+                {
+                    throw null;
```
NTaylorMullen: O_o

test/Microsoft.AspNetCore.Razor.Evolution.Test/Intermediate/RazorIRAssert.cs
```diff
@@ -203,7 +203,10 @@ public static void CSharpExpression(string expected, RazorIRNode node)
                 var content = new StringBuilder();
                 for (var i = 0; i < cSharp.Children.Count; i++)
                 {
-                    content.Append(((CSharpTokenIRNode)cSharp.Children[i]).Content);
+                    if (cSharp.Children[i] is RazorIRToken token && token.IsCSharp)
```
NTaylorMullen: We casted before, lets just cast now and Assert.True(token.IsCSharp) so we know when we make a behavior change in the future.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
rynowak: No, but users could do things that will make this code throw a nullref.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
rynowak: This was just some debugging code I left in.

test/Microsoft.AspNetCore.Razor.Evolution.Test/Intermediate/RazorIRAssert.cs
rynowak: Yeah this should have an `Assert.IsType` instead.

rynowak:
:shipit:

rynowak:
1ae0b216306e72c128c91b116775e6fa3a88a8ea

PR #1027, opened 2017-02-18
Add baselines for codegen tests

rynowak:
This will make is much easier to investigate failures that bubble up to
the codegen level. You'll be able to see if there was a change to the IR
rather than just the final code.


rynowak:
FYI only

PR #1031, opened 2017-02-22
Replace makefile.shade with a PowerShell script

natemcmaster:
Replace the custom build steps for building a VSIX with a powershell script.

cref aspnet/Korebuild#174 and https://github.com/aspnet/KoreBuild/pull/180


build/repo.targets
```diff
@@ -1,8 +1,12 @@
 <Project>
+  <!--
+    TODO include VSIX build in default lifecycle builds
```
natemcmaster: I chose not to as we only have one CI agent with compatible build tools installed.

pranavkm: The plan is to have VS 2017 installed on all our build agents. We're avoiding the hassle of updating them later by installing a RC build now.

pranavkm: See https://github.com/aspnet/Coherence-Signed/issues/409

natemcmaster: Cool. When we finish that, we can come back and take care of this TODO.

PR #1033, opened 2017-02-22
Fixing support for XML docs in OOP

rynowak:
The issue here is that the OOP host doesn't yet have support for
documentation, it will just return null. Fixing this code to look for the
documentation after we get the descriptors back into VS.

I tested this and confirmed that it works with TagHelpers in dlls + xml
file documentation as well as TagHelpers compiled in the project itself.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
@@ -3,9 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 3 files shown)


rynowak:
That's what I'm doing? Or are you just saying I don't need to add a property?

NTaylorMullen:
> That's what I'm doing? Or are you just saying I don't need to add a property?

Asking if you need to add a property.

rynowak:
Yeah I can probably get by without it, I'll try that.

rynowak:
Merging this to unblock Alex.

rynowak:
94d21e03f590e84a28f9fa313c45f35bcffba28c

PR #1035, opened 2017-02-23
Add .editorconfig

pranavkm:
Based on https://github.com/dotnet/roslyn/blob/master/.editorconfig with some AspNet specific tweaks

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,88 @@
+# EditorConfig is awesome:http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Don't use tabs for indentation.
+[*]
+indent_style = space
+# (Please don't specify an indent_size here; that has too many unintended consequences.)
+
+# Code files
+[*.{cs,csx,vb,vbx}]
+indent_size = 4
+
+# Xml project files
+[*.{csproj,vbproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
+indent_size = 2
+
+# Xml config files
+[*.{props,targets,ruleset,config,nuspec,resx,vsixmanifest,vsct}]
+indent_size = 2
+
+# JSON files
+[*.json]
+indent_size = 2
+
+# Dotnet code style settings:
+[*.cs]
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
+
+# Don't use this. qualifier
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+
+# use int x = .. over Int32
+dotnet_style_predefined_type_for_locals_parameters_members = true:error
+
+# use int.MaxValue over Int32.MaxValue
+dotnet_style_predefined_type_for_member_access = true:error
+
+# Prefer object and collection initializers
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:suggestion
+
+# Prefer tuple names to ItemX properties.
+dotnet_style_explicit_tuple_names = true:suggestion
+
+# Prefer to use null-conditional operator where possible.
+dotnet_style_null_propagation = true:suggestion
+
+# Prefer method-like constructs to have a block body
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,88 @@
...
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
+
+# Prefer out variables to be declared inline when possible.
+csharp_style_inlined_variable_declaration = true:suggestion
+
+# Disallow throw expressions.
+csharp_style_throw_expression = false:error
+
+# Prefer to use conditional coalescing operation (?.) when invoking a lambda instead of performing a null check.
+csharp_style_conditional_delegate_call = true:suggestion
```
pranavkm: Sadly, this doesn't work as yet (https://github.com/dotnet/roslyn/issues/17304)

pranavkm:
We could standardize this across our repos

.editorconfig
```diff
@@ -0,0 +1,88 @@
...
+dotnet_style_qualification_for_property = false:suggestion
+
+# use int x = .. over Int32
+dotnet_style_predefined_type_for_locals_parameters_members = true:error
+
+# use int.MaxValue over Int32.MaxValue
+dotnet_style_predefined_type_for_member_access = true:error
+
+# Prefer object and collection initializers
+dotnet_style_object_initializer = true:suggestion
```
rynowak: I don't think we should have suggestions or warnings for anything that isn't part of our engineering guidelines.

Also, we should turn off suggestions that aren't part of our engineering guidelines (like the suggestion to use throw expresssion)

rynowak: /cc @Eilon

.editorconfig
```diff
@@ -0,0 +1,88 @@
...
+
+# Prefer property-like constructs to have an expression-body
+csharp_style_expression_bodied_properties = true:none
+csharp_style_expression_bodied_indexers = true:none
+csharp_style_expression_bodied_accessors = true:none
+
+# Require var all the time.
+csharp_style_var_for_built_in_types = true:error
+csharp_style_var_when_type_is_apparent = true:error
+csharp_style_var_elsewhere = true:error
```
rynowak: Does error prevent you from running your code?

pranavkm: It actually produces a compiler error with red squiggly lines when building inside VS. We could make it a warning, but our guideline seems pretty rigid about it: https://github.com/aspnet/Home/wiki/Engineering-guidelines#usage-of-the-var-keyword

.editorconfig
pranavkm: If we're neutral about a behavior, I think we should probably remove these entries. Specifying a value here including saying `true: none` (https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference) overrides the value in VS. If we don't care for it, we'll let the user's IDE decide.

.editorconfig
```diff
@@ -0,0 +1,88 @@
...
+indent_size = 2
+
+# JSON files
+[*.json]
+indent_size = 2
+
+# Dotnet code style settings:
+[*.cs]
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
```
pranavkm: Sadly, no way to produce a erorr for this. However, it does override the VS behavior.

.editorconfig
pranavkm: image

rynowak: I'm hugely not a fan of this then. Trying to hack something? Nah bro better deal with these stylecop errors first.

pranavkm: Warnings are a bit less in your face:

image

.editorconfig
rynowak: Then we should remove these entries.

pranavkm:
Updated to provide 'suggestions' for things that we care about. This doesn't produce errors due to stylistic choices. `warnings` become errors when `TreatWarningsAsErrors` is enabled (which for some reason isn't on in our projects by default - need to investigate).

PR #1038, opened 2017-02-23
Add activity log for TagHelper exceptions

rynowak:
This will log to the VS activity log any time we fail to discovery
taghelpers due to an exception. There's no real user experience around
this... if someone is having an issue we can ask them to send us the log
or post the relevant portion.

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
@@ -10,6 +10,8 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.Evolution;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
@@ -18,6 +20,9 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor
     {
         [Import]
         public VisualStudioWorkspace Workspace { get; set; }
+
+        [Import]
+        public SVsServiceProvider Services { get; set; }
 
         public async Task<TagHelperResolutionResult> GetTagHelpersAsync(Project project, IEnumerable<string> assemblyNameFilters)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -90,5 +105,11 @@ private void AddXmlDocumentation(Compilation compilation, IReadOnlyList<TagHelpe
                 }
             }
         }
+
+        private IVsActivityLog GetActivityLog()
+        {
+            var services = (IServiceProvider)Services;
```
pranavkm: Is the cast needed? (Going by https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.shell.svsserviceprovider.aspx, it shouldn't be).

rynowak: It's not strictly necessary in this case, but this is the general pattern. VS types have a 'service interface' which like a marker.

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
```diff
@@ -22,6 +22,12 @@
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="2.0.0-rc3-61306-01" />
     <PackageReference Include="Microsoft.VisualStudio.LanguageServices" Version="2.0.0-rc3-61306-01" />
     <PackageReference Include="Microsoft.VisualStudio.LanguageServices.Razor.RemoteClient" Version="2.0.0-rc3-61306-01" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.14.0" Version="14.3.25407" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Immutable.10.0" Version="10.0.30319" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Immutable.11.0" Version="11.0.50727" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Immutable.12.0" Version="12.0.21003" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Immutable.14.0" Version="14.3.25407" />
+    <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.12.0" Version="12.0.30110" />
```
rynowak: The various APIs are spread across these packages. We're not using them all, but it makes sense to have them regardless.

Issue #997, opened 2017-02-13
Prevent LOH allocations when constructing Razor source documents


NTaylorMullen:
Today we Create our Razor documents one way. We should inspect the size of the stream we're reading and buffer it incrementally to avoid LOH allocations. This is similar to what Roslyn does: https://github.com/dotnet/roslyn/blob/4799250ab7adf1e8ca39d79961fcd4cbcb96fa1f/src/Compilers/Core/Portable/Text/SourceText.cs#L127-L131

rynowak:
Fixed by dee8d86. Thanks again @Yves57

PR #1049, opened 2017-03-05
Prevent LOH allocations when constructing large Razor Source Documents.

Yves57:
See #997.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceLineCollection.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceLineCollection.cs
@@ -1,0 +1,123 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultRazorSourceLineCollection : RazorSourceLineCollection
+    {
+        private readonly RazorSourceDocument _document;
+        private readonly int[] _lineStarts;
+
+        public DefaultRazorSourceLineCollection(RazorSourceDocument document)
+        {
+            _document = document;
+            _lineStarts = GetLineStarts();
+        }
+
+        public override int Count => _lineStarts.Length;
+
+        public override int GetLineLength(int index)
+        {
+            if (index < 0 || index >= _lineStarts.Length)
+            {
+                throw new IndexOutOfRangeException(nameof(index));
+            }
+
+            if (index == _lineStarts.Length - 1)
+            {
+                // Last line is special.
+                return _document.Length - _lineStarts[index];
+            }
+
+            return _lineStarts[index + 1] - _lineStarts[index];
+        }
+
+        internal override SourceLocation GetLocation(int position)
+        {
+            if (position < 0 || position >= _document.Length)
+            {
+                throw new IndexOutOfRangeException(nameof(position));
+            }
+
+            var index = Array.BinarySearch<int>(_lineStarts, position);
+            if (index >= 0)
+            {
+                // We have an exact match for the start of a line.
+                Debug.Assert(_lineStarts[index] == position);
+
+                return new SourceLocation(_document.Filename, position, index, characterIndex: 0);
+            }
+
+
+            // Index is the complement of the line *after* the one we want, because BinarySearch tells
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceLineCollection.cs
```diff
@@ -0,0 +1,123 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class DefaultRazorSourceLineCollection : RazorSourceLineCollection
```
rynowak: @Yves57 - is this a copy/paste of the original class? what changed here?

src/Microsoft.AspNetCore.Razor.Evolution/LargeTextRazorSourceDocument.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (encoding == null)
+            {
+                throw new ArgumentNullException(nameof(encoding));
+            }
+
+            _chunkMaxLength = chunkMaxLength;
+            Encoding = encoding;
+            Filename = filename;
+
+            ReadChunks(reader, _chunkMaxLength, out _length, out _chunks);
```
rynowak: Just have this return the chunks rather than using `_out`

src/Microsoft.AspNetCore.Razor.Evolution/LargeTextRazorSourceDocument.cs
```diff
@@ -0,0 +1,136 @@
...
+            _lines = new DefaultRazorSourceLineCollection(this);
+        }
+
+        public override char this[int position]
+        {
+            get
+            {
+                var chunkIndex = position / _chunkMaxLength;
+                var insideChunkPosition = position - chunkIndex * _chunkMaxLength;
```
rynowak: Is there a reason to avoid `%` here?

src/Microsoft.AspNetCore.Razor.Evolution/LargeTextRazorSourceDocument.cs
```diff
@@ -0,0 +1,136 @@
...
+                insideChunkPosition = 0;
+            }
+        }
+
+        private static void ReadChunks(StreamReader reader, int chunkMaxLength, out int length, out List<char[]> chunks)
+        {
+            length = 0;
+            chunks = new List<char[]>();
+
+            for (;;)
```
rynowak: use `while (true)`

rynowak: It would be nice to avoid a potentially infinite loop here, we get a bit nervous about `for(;;)` or `while(true)`. 

This can be simplified because `ReadBlock` will only read fewer than `count` characters when it's reached the end. https://msdn.microsoft.com/en-us/library/system.io.streamreader.readblock(v=vs.110).aspx#Anchor_2

```

char[] chunk;
int read;

do
{
    chunk = new char[chunkMaxLength];
    read = reader.ReadBlock(chunk, index, chunkMaxLength);
    length += read;
    if (read > 0)
    {
        chunks.Add(chunk);
    }
}
while (read == chunkMaxLength);
```

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorSourceLineCollection.cs
Yves57: A cut/paste to this new class in fact (because it is used in both `RazorSourceDocument` derived classes).
The only difference is in `GetLineStarts()`. The new code uses `RazorSourceDocument.Length` instead of `string.Length` and `RazorSourceDocument[index]` instead of `string[index]`.

src/Microsoft.AspNetCore.Razor.Evolution/LargeTextRazorSourceDocument.cs
Yves57: No good reason 😄 .
Changed...

src/Microsoft.AspNetCore.Razor.Evolution/LargeTextRazorSourceDocument.cs
Yves57: Simplified...

Issue #1006, opened 2017-02-15
Review Naming


rynowak:
These are the items we want to review in a later naming / API review pass of Razor.Evolution.

For now we're just filing items here and not getting blocked on minutiae if we can't come to an agreement on something trivial.

-------------------------

- [x] IR Node classes
- [x] `RuntimeTarget`/`IRuntimeTargetExtension`
- [x] `BasicWriter`/`TagHelperWriter`
- [x] `ITagHelperDescriptorBuilder` / `ITagHelperBoundAttributeDescriptorBuilder`
- [x] `TagHelperDocumentContext` `GetTagHelperContext`

rynowak:
This work is now tracked elsewhere. Closing this

PR #1053, opened 2017-03-07
Clean up TagHelperDescriptor APIs.

NTaylorMullen:

- Removed all design time descriptors and put their API surface into their corresponding descriptor. Part of removing the design time API surface was removing the tracking of `<Remarks>`, it wasn't used so there's on need to track it until we need it.
- Removed the Type requirement from `TagHelperDescriptor`. With this separation we'll be able to have abstract `TagHelper`s that aren't based on a class implementation.
- Removed Prefix from the `TagHelperDescriptor` API surface. It was a legacy requirement based on how the Razor parser was put together. We can work around this now.
- Stripped correlation information from the immediate `TagHelperDescriptor` API surface. Instead this information is now tracked in `TagMatchingRule`s. This change means that you will not have multiple `TagHelperDescriptor`s per `TagHelper`; instead it's all tracked in a single descriptor. A side effect of this change was the transformation of `IsIndexer` => 3 new properties.
- Renamed many descriptor types and property names.
- Added builder APIs to construct TagHelpers since they're inherently immutable in their API surface.
- Added `ITagHelperDescriptorBuilder` to represent `TagHelper`s that are built from an `ITagHelper` implementing class. It re-introduces the `TypeName` association of a `TagHelper`.
- Added `ITagHelperBoundAttributeDescriptorBuilder` to represent that an attribute was associated with a property.
- Added validation methods to the descriptor builders to enable consumers to validate the current state of the builder and add diagnostics as necessary.
- Moved descriptors away from RazorError.
- Updated the various comparers to understand the descriptors new API.

#973


src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptor.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public string TypeName { get; protected set; }
+
+        public string IndexerTypeName { get; protected set; }
+
+        public string Documentation { get; protected set; }
+
+        public string DisplayName { get; protected set; }
+
+        public IEnumerable<RazorDiagnostic> Diagnostics { get; protected set; }
```
rynowak: is there a reason why we wouldn't want to make this `IReadOnlyList<...>`?

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptor.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    /// <summary>
+    /// A metadata class describing a tag helper attribute.
+    /// </summary>
+    public abstract class BoundAttributeDescriptor
+    {
+        protected BoundAttributeDescriptor(string kind)
+        {
+            Kind = kind;
+        }
```
rynowak: Why is `Kind` required here and nothing else? If we're going to make it required we should also null-check it. I'd be fine with just not making it required. Extending this set of classes isn't something you do casually.

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptorComparer.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+        }
+
+        public virtual bool Equals(BoundAttributeDescriptor descriptorX, BoundAttributeDescriptor descriptorY)
+        {
+            if (descriptorX == descriptorY)
+            {
+                return true;
+            }
+
```
rynowak: You need to null-check `descriptorY` here.

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptorComparer.cs
```diff
@@ -0,0 +1,63 @@
...
+            return descriptorX != null &&
+                descriptorX.Kind == descriptorY.Kind &&
```
rynowak: string.Equals

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptorComparer.cs
rynowak: I usually like to do `if (object.ReferenceEquals(x, y)) { return true; } ... if (x == null ^ y == null) { return false; }`

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,260 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    public sealed class ITagHelperBoundAttributeDescriptorBuilder
```
rynowak: Some of these names are yuck 🤢

rynowak: You could also consider making this a nested class/interface in `BoundAttributeDescriptor` so that it reads like `BoundAttributeDescriptor.IBuilder`. 

The nice thing about this as it stands is that only the `Create` method is what ties it to an `ITagHelper`-based implementation.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,260 @@
...
+
+        public ITagHelperBoundAttributeDescriptorBuilder AddDiagnostic(RazorDiagnostic diagnostic)
+        {
+            EnsureDiagnostics();
+            _diagnostics.Add(diagnostic);
+
+            return this;
+        }
+
+        public IEnumerable<RazorDiagnostic> Validate()
```
rynowak: IIRC from our discussion we never settled exactly how/when validation would occur. To me the most sensible thing to do is to synthesize these errors when `Build()` is called. Then there's no real need for a separate `Validate`.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,260 @@
...
+                _propertyName,
+                _typeName,
+                _indexerNamePrefix,
+                _indexerValueTypeName,
+                _documentation,
+                displayName,
+                _propertyBag,
+                _diagnostics ?? Enumerable.Empty<RazorDiagnostic>());
+
+            return descriptor;
```
rynowak: All of these collections need to make a deep copy

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,260 @@
...
+            var descriptor = new ITagHelperBoundAttributeDescriptor(
+                _isEnum,
+                _name,
+                _propertyName,
+                _typeName,
+                _indexerNamePrefix,
+                _indexerValueTypeName,
+                _documentation,
+                displayName,
+                _propertyBag,
```
rynowak: metadata?

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptorBuilder.cs
```diff
@@ -0,0 +1,244 @@
...
+                    else if (name != TagHelperDescriptorProvider.ElementCatchAllTarget)
+                    {
+                        foreach (var character in name)
+                        {
+                            if (char.IsWhiteSpace(character) || InvalidNonWhitespaceAllowedChildCharacters.Contains(character))
+                            {
+                                var diagnostic = RazorDiagnosticFactory.CreateTagHelper_InvalidRestrictedChild(name, _typeName, character);
+
+                                yield return diagnostic;
+                            }
```
rynowak: This will return one error-per-invalid-character. Is that what you want?

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptor.cs
NTaylorMullen: Don't think so. I can give it a try.

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptor.cs
NTaylorMullen: Figured that organizing the descriptor API around `Kind` might make sense. If a user wants to create their own type of descriptor then `Kind` is the only way to truly tell a new descriptor apart from a built-in one.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: > The nice thing about this as it stands is that only the Create method is what ties it to an ITagHelper-based implementation.

That's what I was going for. With the naming as it is it's obvious that this type of `BoundAttributeDescriptor` belongs to one from an `ITagHelper` implementation.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: Thought about that. The reason why I veered away from that was:
- Re-using a builder becomes impossible because diagnostics accumulate.
- You're forced into validating a `TagHelperDescriptor` even if your runtime system can handle a technically "invalid" one.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: Maybe eventually. Is there any reason to do it now?

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptorBuilder.cs
NTaylorMullen: Yup, that's how it worked in the old world.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: Maybe eventually. Is there any reason to do it now?

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
rynowak: I disagree on both of these points. 
-  That would be a bug. Imagine an implementation where they don't accumulate, that's the right way to code it.
- The rules about what's supported/not-supported are based on HTML and our ability to process it as text. A system that doesn't have the same restrictions would not be Razor.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
rynowak: Yes, because you want to be able to reuse the builder. Sorry, should have said shallow copy.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptorBuilder.cs
rynowak: wierd 😆

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
rynowak: How about we compromise and add this to https://github.com/aspnet/Razor/issues/1006

Not worth blocking on

NTaylorMullen: done

ajaybhargavb:
🆙 📅

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        public BoundAttributeDescriptor Build()
+        {
+            var validationDiagnostics = Validate();
+
+            var diagnostics = new List<RazorDiagnostic>();
+            diagnostics.AddRange(validationDiagnostics);
+
+            if (_diagnostics != null)
```
NTaylorMullen: Could just iterate over each of the validationDiagnostics and call `AddDiagnostic` which will ensure `_diagnostics` is setup appropriately; will be more clear. Same in other builders.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -130,6 +147,18 @@ public BoundAttributeDescriptor Build()
             return descriptor;
         }
 
+        public void Reset()
```
NTaylorMullen: Just to note: It's not a full reset because the `_containingTypeName` will be the same. Only way to get around that fact would be enable a re-initialization of that type.

Just commenting; I think this is fine as is.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
ajaybhargavb: Can't. If someone wants to reuse the builder without resetting it, they will end up with duplicate diagnostics.

NTaylorMullen: Talked offline. `HashSet` to the rescue

ajaybhargavb: Addressed here https://github.com/aspnet/Razor/pull/1055/commits

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,282 @@
...
+        {
+            _name = null;
+            _propertyName = null;
+            _typeName = null;
+            _documentation = null;
+            _isEnum = false;
+            _indexerNamePrefix = null;
+            _indexerValueTypeName = null;
+            _diagnostics?.Clear();
```
rynowak: you need to also clear `_metadata`

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,282 @@
...
+                Name = name;
+                TypeName = typeName;
+                IndexerNamePrefix = dictionaryAttributeNamePrefix;
+                IndexerTypeName = dictionaryValueTypeName;
+                Documentation = documentation;
+                DisplayName = displayName;
+                Diagnostics = diagnostics;
+
+                metadata[PropertyNameKey] = propertyName;
+                Metadata = new Dictionary<string, string>(metadata);
```
rynowak: you should not dirty your parameter (see L277), make a copy first and then modify the copy.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -0,0 +1,282 @@
...
+                IsEnum = isEnum;
+                IsIndexerStringProperty = dictionaryValueTypeName == typeof(string).FullName || dictionaryValueTypeName == "string";
+                IsStringProperty = typeName == typeof(string).FullName || typeName == "string";
+                Name = name;
+                TypeName = typeName;
+                IndexerNamePrefix = dictionaryAttributeNamePrefix;
+                IndexerTypeName = dictionaryValueTypeName;
+                Documentation = documentation;
+                DisplayName = displayName;
+                Diagnostics = diagnostics;
```
rynowak: Why not make a copy of diagnostics as well? The copy of the metadata is at L277. Either make the copy here or make it in the build method, either way be consistent.

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperDescriptorBuilder.cs
```diff
@@ -0,0 +1,265 @@
...
+                DisplayName = displayName;
+                Documentation = documentation;
+                TagOutputHint = tagOutputHint;
+                TagMatchingRules = new List<TagMatchingRule>(tagMatchingRules);
+                BoundAttributes = new List<BoundAttributeDescriptor>(attributeDescriptors);
+                AllowedChildTags = new List<string>(allowedChildTags);
+                Diagnostics = diagnostics;
+
+                metadata[TypeNameKey] = typeName;
+                Metadata = new Dictionary<string, string>(metadata);
```
rynowak: Same comments as the other one.

PR #1054, opened 2017-03-07
Reacted to descriptor changes in product code.

NTaylorMullen:

- Added a new `RazorDiagnosticFactory` abstraction to handle `RazorDiagnostic`s and their corresponding errors/ids etc. This new API should allow for easy addition of new `RazorDiagnostic` errors.
- Updated the `DefaultTagHelperDescriptorFactory` to construct `TagHelperDescriptor`s using the new builder APIs and in the new descriptor format (1 descriptor per type).
- Updated `ViewComponentTagHelperDescriptorFactory` to construct `TagHelperDescriptor`s with the builder API.
- With both factory implementations code was duplicated because the ViewComponent work will be moving outside of Razor once we have the proper hooks.
- Updated `TagHelper` binding bits to capture a binding result in order to query which rules appy to a given tag name.
- Still need to handle deserialization of `TagHelperDescriptor` types within the Razor extension.

#973


src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptorMatchingConventions.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    /*
+     * REVIEWERS: This needs to be public or IVT for tooling to consume until we flesh
+     * out a proper API to decouple tooling from Razors intrinsics. They use these to
+     * determine if their HTML elements match a descriptor or not to provide IntelliSense.
+    */
+    internal static class TagHelperDescriptorMatchingConventions
```
rynowak: and the service will likely call this class.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptorMatchingConventions.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        public static bool CanMatchName(this BoundAttributeDescriptor descriptor, string name) =>
+            IsFullNameMatch(descriptor, name) || IsIndexerNameMatch(descriptor, name);
```
rynowak: arrow function + line break = less readable

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -1,15 +1,15 @@
...
 
+using Microsoft.AspNetCore.Razor.Evolution;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using Microsoft.CodeAnalysis;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq;
 using System.Text.RegularExpressions;
```
rynowak: sort

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -61,87 +61,44 @@ public DefaultTagHelperDescriptorFactory(Compilation compilation, bool designTim
         protected bool DesignTime { get; }
 
         /// <inheritdoc />
-        public virtual IEnumerable<TagHelperDescriptor> CreateDescriptors(
-            INamedTypeSymbol type,
-            ErrorSink errorSink)
+        public virtual TagHelperDescriptor CreateDescriptor(INamedTypeSymbol type)
```
rynowak: yay

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -61,87 +61,44 @@ public DefaultTagHelperDescriptorFactory(Compilation compilation, bool designTim
...
-                }
-
-                string outputElementHint = null;
-                var outputElementHintAttribute = type.GetAttributes().Where(a => a.AttributeClass == _outputElementHintAttributeSymbol).FirstOrDefault();
-                if (outputElementHintAttribute != null)
-                {
-                    outputElementHint = (string)(outputElementHintAttribute.ConstructorArguments[0]).Value;
-                }
+                descriptorBuilder.AddDiagnostic(diagnostic);
+            }
```
rynowak: I commented about this elsewhere, but this shouldn't be necessary (L86-L90). This should just be part of build.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -894,323 +495,299 @@ internal static string ToHtmlCase(string name)
...
+            public static void AddRequiredAttributes(string requiredAttributes, TagMatchingRuleBuilder ruleBuilder)
             {
-                _requiredAttributes = requiredAttributes;
+                var requiredAttributeParser = new DefaultRequiredAttributeParser(requiredAttributes);
+                requiredAttributeParser.AddRequiredAttributes(ruleBuilder);
             }
 
-            private char Current => _requiredAttributes[_index];
+            private class DefaultRequiredAttributeParser
+            {
```
rynowak: I don't get what's going on here.

rynowak: It would be nice to split this out into its out thing.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -894,323 +495,299 @@ internal static string ToHtmlCase(string name)
...
-            {
-                while (!AtEnd && (Current == ' ' || Current == '\t'))
+                private void PassOptionalWhitespace()
                 {
-                    _index++;
+                    while (!AtEnd && (Current == ' ' || Current == '\t'))
+                    {
+                        _index++;
+                    }
                 }
```
rynowak: It's really nice overall to see the changes to this class, this is much cleaner than before.

src/Microsoft.CodeAnalysis.Razor/TagHelpers.cs
```diff
@@ -32,8 +32,8 @@ private static void VisitTagHelpers(Compilation compilation, List<TagHelperDescr
 
             foreach (var type in types)
             {
-                var descriptors = factory.CreateDescriptors(type, errors);
-                results.AddRange(descriptors);
+                var descriptor = factory.CreateDescriptor(type);
+                results.Add(descriptor);
```
rynowak: What if the factory returns null?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -82,42 +83,124 @@ internal class DefaultTagHelperResolver : ITagHelperResolver
...
+            var documentedTagHelpers = new List<TagHelperDescriptor>();
             for (var i = 0; i < tagHelpers.Count; i++)
             {
                 var tagHelper = tagHelpers[i];
-                tagHelper.DesignTimeDescriptor = tagHelper.DesignTimeDescriptor ?? new TagHelperDesignTimeDescriptor();
 
-                var symbol = compilation.GetTypeByMetadataName(tagHelper.TypeName);
+                if (tagHelper.Documentation != null)
+                {
+                    goto AddDocumentedTagHelper;
```
rynowak: No. you haven't earned this.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -50,11 +50,12 @@ internal class DefaultTagHelperResolver : ITagHelperResolver
                                 // Per https://github.com/dotnet/roslyn/issues/12770 - there's currently no support for documentation in the OOP host
                                 // until that's available we add the documentation on the VS side by looking up each symbol again.
                                 var compilation = await project.GetCompilationAsync().ConfigureAwait(false);
-                                AddXmlDocumentation(compilation, result.Descriptors);
-                                return result;
+                                var documentedTagHelpers = GetDocumentedTagHelpers(compilation, result.Descriptors);
+                                var documentedResult = new TagHelperResolutionResult(documentedTagHelpers, result.Diagnostics);
```
rynowak: Why change this data flow? Just because of immutability?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -82,42 +83,124 @@ internal class DefaultTagHelperResolver : ITagHelperResolver
...
+
+                AddDocumentedTagHelper:
+
+                documentedTagHelpers.Add(tagHelper);
+            }
+
+            return documentedTagHelpers;
+        }
+
+        private ITagHelperBoundAttributeDescriptorBuilder ShallowCopy(string tagHelperTypeName, BoundAttributeDescriptor attribute)
```
rynowak: Think about making this an internal method on the abstract class. Then each descriptor type can 'copy' itself to a builder.

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/SetPreallocatedTagHelperPropertyIRNode.cs
```diff
@@ -21,7 +21,9 @@ public class SetPreallocatedTagHelperPropertyIRNode : RazorIRNode
 
         public string PropertyName { get; set; }
 
-        public TagHelperAttributeDescriptor Descriptor { get; set; }
+        public BoundAttributeDescriptor Descriptor { get; set; }
```
rynowak: The IR node I mean

src/Microsoft.AspNetCore.Razor.Evolution/RazorCodeDocumentExtensions.cs
```diff
@@ -9,6 +9,28 @@ namespace Microsoft.AspNetCore.Razor.Evolution
...
+
+        public static void SetTagHelperPrefix(this RazorCodeDocument document, string tagHelperPrefix)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            document.Items[TagHelperPrefixKey] = tagHelperPrefix;
+        }
```
NTaylorMullen: https://github.com/aspnet/Razor/issues/1060

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptorMatchingConventions.cs
NTaylorMullen: But... Buttt.... my pranav points.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: Yup

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: Thought about doing that and then realized this was the only part of the API that'd use it. Decided against it until we need it.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
rynowak: I'd suggest just writing it in a sane, albeit inefficient way.

1. Get the descriptors
2. Create the objects
3. Loop over each one that's missing docs and add them
4. return

Alternatively you could loop over the builders in stage 3.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
rynowak: Sounds good.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
rynowak: NVM I am a sillypants

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -252,32 +254,38 @@ private bool TryRewriteTagHelper(Block tagBlock, ErrorSink errorSink)
...
-                                invalidDescriptor.TypeName,
-                                invalidDescriptor.TagStructure),
-                            tagName.Length);
+                        var boundRules = tagHelperBinding.GetBoundRules(descriptor);
+                        var invalidRule = boundRules.FirstOrDefault(rule => rule.TagStructure == TagStructure.WithoutEndTag);
 
-                        return false;
+                        if (invalidRule != null)
+                        {
+                            var typeName = descriptor.Metadata[ITagHelperDescriptorBuilder.TypeNameKey];
```
ajaybhargavb: https://github.com/aspnet/Razor/issues/1072

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/SetPreallocatedTagHelperPropertyIRNode.cs
ajaybhargavb: Spoke offline. Going to add  a reference to `TagHelperBinding` aka `TagBindingResult`

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperDescriptorMatchingConventions.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    /*
+     * REVIEWERS: This needs to be public or IVT for tooling to consume until we flesh
+     * out a proper API to decouple tooling from Razors intrinsics. They use these to
+     * determine if their HTML elements match a descriptor or not to provide IntelliSense.
+    */
```
rynowak: Don't leave this comment here

src/Microsoft.CodeAnalysis.Razor/RequiredAttributeParser.cs
```diff
@@ -0,0 +1,304 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    // Internal for testing
+    internal static class RequiredAttributeParser
```
rynowak: This whole thing should be independent of the builder... The builder should **call** into this class, not the other way around.

NTaylorMullen: That'd put the logic of CSS based required attributes in the builder and not in the `DefaultTagHelperDescriptorFactory`. That sounds a little backwards compared to how all other declarative information on TagHelper's is consumed.

ajaybhargavb: Filed an issue to track this. https://github.com/aspnet/Razor/issues/1082

PR #1055, opened 2017-03-07
Update tests to react to new `TagHelperDescriptor` API.

NTaylorMullen:

- Finished migrating the `Microsoft.CodeAnalysis.Razor.Test` project. Major work here was getting the various factory tests constructing `TagHelperDescriptor`s correctly in a 1 -> 1 format with types.
- Migrated integration tests for confidence in our product code changes.
- Deleted several tests that raised themselves as being duplicates.
- Updated case sensitive comparers to understand new `TagHelperDescriptor` API.
- Removed design time comparers.
- This does not include the completion of Razor.Evolution test reactions.

#973


test/Microsoft.AspNetCore.Razor.Evolution.Test/CaseSensitiveTagMatchingRuleComparer.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
+using Microsoft.Extensions.Internal;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Evolution
+{
+    internal class CaseSensitiveTagMatchingRuleComparer : TagMatchingRuleComparer
```
rynowak: Maybe you'll have to explain to me a few things, but what's the purpose of this?

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
```diff
@@ -1626,6 +1612,10 @@ private void RunRuntimeTagHelpersTest(IEnumerable<TagHelperDescriptor> descripto
                     builder.AddTagHelpers(descriptors);
                 });
             var document = CreateCodeDocument();
+            if (tagHelperPrefix != null)
+            {
+                document.SetTagHelperPrefix(tagHelperPrefix);
+            }
```
rynowak: If these are integration tests shouldn't the prefix come from the code?

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/InstrumentationPassIntegrationTest.cs
```diff
@@ -66,5 +58,28 @@ public void BasicTest()
             AssertCSharpDocumentMatchesBaseline(csharpDocument);
             Assert.Empty(csharpDocument.Diagnostics);
         }
+
+        private static TagHelperDescriptor CreateTagHelperDescriptor(
+            string tagName,
+            string typeName,
+            string assemblyName,
+            IEnumerable<Action<ITagHelperBoundAttributeDescriptorBuilder>> attributes = null)
```
rynowak: should we put this on the base class?

test/Microsoft.AspNetCore.Razor.Evolution.Test/CaseSensitiveTagMatchingRuleComparer.cs
NTaylorMullen: When comparing `TagHelperDescriptor`s we want to ensure that the tag matching rules are exactly what we expect. In the product code comparer we validate the `TagName` case insensitively; therefore, we add a case insensitive variant of it to ensure our tests exactly match our expectations.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
NTaylorMullen: Agree, we didn't do this in the past because the logic to understand `@tagHelperPrefix` wasn't in the core of Razor. Aka, this is just ported code smell.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/InstrumentationPassIntegrationTest.cs
NTaylorMullen: Nah, this method is just slightly more convenient for creating `TagHelperDescriptor`s. I personally prefer using the builder API's directly when possible.

test/Microsoft.AspNetCore.Razor.Evolution.Test/CaseSensitiveTagMatchingRuleComparer.cs
rynowak: Of what value is this? What are the chances with we'll have a bug?

It seems like we have a lot of infrastructure here to make sure that things we don't do don't happen.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/InstrumentationPassIntegrationTest.cs
rynowak: 👍

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
rynowak: OK. Can you log an item for this?

test/Microsoft.AspNetCore.Razor.Evolution.Test/CaseSensitiveTagMatchingRuleComparer.cs
NTaylorMullen: There's value. Having multiple `HtmlTargetElementAttribute`s with different cases may pass on the product variant of `TagMatchingRuleComparer` but not the case sensitive one depending on if we consume the attributes correctly.

test/Microsoft.AspNetCore.Razor.Evolution.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/1059

test/Microsoft.AspNetCore.Razor.Evolution.Test/CaseSensitiveTagMatchingRuleComparer.cs
rynowak: OK. I'm going to refer to these in the future as **Taylor's comparers** and use it as an example in a talk perhaps. 

If you love someone, you have to accept them as they are.

rynowak: Also my opinion:

> Having multiple HtmlTargetElementAttributes with different cases may pass on the product variant of TagMatchingRuleComparer but not the case sensitive one depending on if we consume the attributes correctly.

Equal means Equal, and equal is defined by the default comparer. Having multiple values that vary by only case is the the same as having multiple values that are literally the same unicode code points. We should either ignore duplicates everywhere or report an error everywhere. Anything else is a bug.

NTaylorMullen: Talked offline. Going to update the base class to reflect the case insensitive variant so all product code is in one place.

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultDirectiveIRPassTest.cs
```diff
@@ -1,171 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
NTaylorMullen: Ya this guy shouldn't have been terminated.
image

ajaybhargavb: I've resurrected this guy here https://github.com/aspnet/Razor/pull/1071/commits

src/Microsoft.AspNetCore.Razor.Evolution/BoundAttributeDescriptorComparer.cs
```diff
@@ -10,10 +10,32 @@ namespace Microsoft.AspNetCore.Razor.Evolution
...
+        /// <summary>
+        /// A default instance of the <see cref="BoundAttributeDescriptorComparer"/>.
+        /// </summary>
         public static readonly BoundAttributeDescriptorComparer Default = new BoundAttributeDescriptorComparer();
 
-        protected BoundAttributeDescriptorComparer()
+        /// <summary>
+        /// A default instance of the <see cref="BoundAttributeDescriptorComparer"/> that does case-sensitive comparison.
+        /// </summary>
+        internal static readonly BoundAttributeDescriptorComparer DefaultCaseSensitive =
```
rynowak: This should just be `CaseSensitive`

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -116,15 +116,7 @@ public ITagHelperBoundAttributeDescriptorBuilder AddDiagnostic(RazorDiagnostic d
...
-            var validationDiagnostics = Validate();
-
-            var diagnostics = new List<RazorDiagnostic>();
-            diagnostics.AddRange(validationDiagnostics);
-
-            if (_diagnostics != null)
-            {
-                diagnostics.AddRange(_diagnostics);
-            }
+            Validate();
```
rynowak: This is wrong. These errors are **computed** based on the state of builder when `Build` is called. Storing them is always wrong.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/RequiredAttributeDescriptorComparer.cs
```diff
@@ -15,16 +15,32 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
...
         /// </summary>
         public static readonly RequiredAttributeDescriptorComparer Default =
             new RequiredAttributeDescriptorComparer();
 
         /// <summary>
-        /// Initializes a new <see cref="RequiredAttributeDescriptor"/> instance.
+        /// A default instance of the <see cref="RequiredAttributeDescriptorComparer"/> that does case-sensitive comparison.
         /// </summary>
-        protected RequiredAttributeDescriptorComparer()
+        internal static readonly RequiredAttributeDescriptorComparer DefaultCaseSensitive =
```
rynowak: CaseSensitive

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TagHelperDescriptorComparer.cs
```diff
@@ -10,10 +10,38 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
...
+        /// <summary>
+        /// A default instance of the <see cref="TagHelperDescriptorComparer"/>.
+        /// </summary>
         public static readonly TagHelperDescriptorComparer Default = new TagHelperDescriptorComparer();
 
-        protected TagHelperDescriptorComparer()
+        /// <summary>
+        /// A default instance of the <see cref="TagHelperDescriptorComparer"/> that does case-sensitive comparison.
+        /// </summary>
+        internal static readonly TagHelperDescriptorComparer DefaultCaseSensitive =
```
rynowak: CaseSensitive

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -471,7 +471,7 @@ private bool HasAllowedChildren()
                 return false;
             }
 
-            return _currentTagHelperTracker.AllowedChildren != null;
+            return _currentTagHelperTracker.AllowedChildren != null && _currentTagHelperTracker.AllowedChildren.Count() > 0;
```
rynowak: Why is this `.Count()`? Aren't all of these things `IReadOnlyList<>`? If they aren't then they should be.

src/Microsoft.AspNetCore.Razor.Evolution/TagMatchingRuleComparer.cs
```diff
@@ -11,10 +11,35 @@ namespace Microsoft.AspNetCore.Razor.Evolution
...
+        /// <summary>
+        /// A default instance of the <see cref="TagMatchingRuleComparer"/>.
+        /// </summary>
         public static readonly TagMatchingRuleComparer Default = new TagMatchingRuleComparer();
 
-        protected TagMatchingRuleComparer()
+        /// <summary>
+        /// A default instance of the <see cref="TagMatchingRuleComparer"/> that does case-sensitive comparison.
+        /// </summary>
+        internal static readonly TagMatchingRuleComparer DefaultCaseSensitive =
```
rynowak: CaseSensitive

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: HashSet helps fight this though. How is it different than storing the `RazorDiagnostic`'s that are manually added and not adding a way to mutate that list?

ajaybhargavb: Spoke offline

PR #1061, opened 2017-03-08
Update versions of VS and Roslyn 2.0.0 dependencies

rynowak:
VS has gone RTM so, updating to the RTM versions of those dependencies.

Roslyn does not publish our shim packages on NuGet.org, so updating those
to a non-ancient version for projects that use 2.0.0. The projects that
use 1.3.x are staying put for now.

The code change is dealing with something that was obsoleted.


rynowak:
FYI only

Issue #1065, opened 2017-03-08
Add TagName to TagHelperContext


jbagga:
Adding `TagName` to the `TagHelperContext` for easy access to the name of HTML element being targeted by the `TagHelper`.

Follow up to https://github.com/aspnet/Mvc/issues/5728#issuecomment-284114612

PR #1066, opened 2017-03-08
Add TagName to TagHelperContext

jbagga:
Addresses #1065 

cc @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
@@ -70,7 +70,7 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         /// <summary>
         /// The parsed HTML tag name of the element.
         /// </summary>
-        public string TagName { get; set; }
+        public string TagName { get; }
 
         /// <summary>
         /// Every attribute associated with the current HTML element.
--- a/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
@@ -70,7 +70,7 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
         /// <summary>
         /// The parsed HTML tag name of the element.
         /// </summary>
-        public string TagName { get; }
+        public string TagName { get; private set; }
 
         /// <summary>
         /// Every attribute associated with the current HTML element.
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -16,6 +16,28 @@ public class TagHelperContext
         /// <summary>
         /// Instantiates a new <see cref="TagHelperContext"/>.
         /// </summary>
+        /// <param name="tagName">The HTML element's tag name.</param>
```
NTaylorMullen: `The parsed HTML tag name of the element`

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -46,6 +68,14 @@ public class TagHelperContext
         }
 
         /// <summary>
+        /// The HTML element's tag name.
+        /// </summary>
+        /// <remarks>
+        /// A whitespace or <c>null</c> value results in no start or end tag being rendered.
```
NTaylorMullen: The only time `TagName` will be null will be if a user constructs `TagHelperContext` in its old format.

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -16,6 +16,28 @@ public class TagHelperContext
         /// <summary>
         /// Instantiates a new <see cref="TagHelperContext"/>.
         /// </summary>
+        /// <param name="tagName">The HTML element's tag name.</param>
+        /// <param name="allAttributes">Every attribute associated with the current HTML element.</param>
+        /// <param name="items">Collection of items used to communicate with other <see cref="ITagHelper"/>s.</param>
+        /// <param name="uniqueId">The unique identifier for the source element this <see cref="TagHelperContext" />
+        /// applies to.</param>
+        public TagHelperContext(
```
NTaylorMullen: Awesome non-breaking change 😄.

@Eilon should we mark the other ctor in this class as deprecated?

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
jbagga: I should remove the remark then?

NTaylorMullen: yup

jbagga:
@NTaylorMullen Updated

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
```diff
@@ -46,6 +68,11 @@ public class TagHelperContext
         }
 
         /// <summary>
+        /// The parsed HTML tag name of the element.
+        /// </summary>
+        public string TagName { get; set; }
```
NTaylorMullen: Oops, missed this part. Remove the setter.

jbagga:
@NTaylorMullen Updated

src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/TagHelperContext.cs
Eilon: Do both ctors work just fine? Do we want to prevent people from calling the old ctor?

If both work fine, and we don't care which people call, then no reason to make a breaking change.

NTaylorMullen: Both work fine. One just results in the `TagName` being `null` so a user can't determine the parsed `TagName`.

jbagga:
@NTaylorMullen Added another Reinitialize method as per our discussion. Also added related test

PR #1067, opened 2017-03-09
Collate diagnositcs from Imports

pranavkm:


test/Microsoft.AspNetCore.Razor.Evolution.Test/DefaultRazorCSharpLoweringPhaseTest.cs
```diff
@@ -75,5 +74,73 @@ public void Execute_ThrowsForMissingDependency_RuntimeTarget()
                 $"The document of kind 'test' does not have a '{nameof(RuntimeTarget)}'. " +
                 $"The document classifier must set a value for '{nameof(DocumentIRNode.Target)}'.");
         }
+
+        [Fact]
+        public void Execute_CollatesSyntaxDiagnosticsFromCodeDocument()
```
rynowak: This isn't 'from the code document' it's from the source document

PR #1071, opened 2017-03-09
Finish porting Razor.Evolution tests

ajaybhargavb:
- Fixed legacy tests
- Added TagHelperDescriptorJsonConverter and added serialization tests

@rynowak @pranavkm @NTaylorMullen


src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+    internal class TagHelperDescriptorJsonConverter : JsonConverter
+    {
+        public static readonly TagHelperDescriptorJsonConverter Instance = new TagHelperDescriptorJsonConverter();
+        private const string RazorDiagnosticMessageKey = "Message";
+        private const string RazorDiagnosticTypeNameKey = "TypeName";
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(TagHelperDescriptor) ||
+                objectType == typeof(DefaultRazorDiagnostic) ||
```
NTaylorMullen: Why are the diagnostic bits in the `CanConvert` method?

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+            if (reader.TokenType != JsonToken.StartObject)
+            {
+                return null;
+            }
+
+            var descriptor = JObject.Load(reader);
+            var descriptorKind = descriptor[nameof(TagHelperDescriptor.Kind)].Value<string>();
+            if (descriptorKind != ITagHelperDescriptorBuilder.DescriptorKind)
+            {
+                throw new NotSupportedException();
```
NTaylorMullen: Should probably provide a nice error message

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            var builder = ITagHelperDescriptorBuilder.Create(typeName, assemblyName);
+
+            builder
+                .Documentation(documentation)
+                .TagOutputHint(tagOutputHint);
+
+            foreach (var tagMatchingRule in tagMatchingRules)
+            {
+                builder.TagMatchingRule(b => ReadTagMatchingRule(b, tagMatchingRule.Value<JObject>(), serializer));
```
NTaylorMullen: Break these things up so they're easy to debug if we need to in the future. Same with the other loops below

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            foreach (var diagnostic in diagnostics)
+            {
+                builder.AddDiagnostic(ReadDiagnostic(diagnostic.Value<JObject>(), serializer));
+            }
+
+            var metadataValue = serializer.Deserialize<Dictionary<string, string>>(metadata.CreateReader());
+            foreach (var item in metadataValue)
+            {
+                if (string.Equals(item.Key, ITagHelperBoundAttributeDescriptorBuilder.PropertyNameKey, StringComparison.Ordinal))
```
NTaylorMullen: Can do this outside of the loop and query the metadata. The builder will ensure that duplicates aren't added to the metadata bag.

src/Microsoft.VisualStudio.LanguageServices.Razor/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="TagHelperDescriptorJsonConverter_UnsupportedRazorDiagnosticType" xml:space="preserve">
+    <value>Deserialization of RazorDiagnostic type '{0}' is not supported.</value>
```
NTaylorMullen: pass in `RazorDiagnostic` in case we refactor/change bits on it. Also don't want it to be localized.

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+            var indexerTypeName = attribute[nameof(BoundAttributeDescriptor.IndexerTypeName)].Value<string>();
+            var documentation = attribute[nameof(BoundAttributeDescriptor.Documentation)].Value<string>();
+            var diagnostics = attribute[nameof(BoundAttributeDescriptor.Diagnostics)].Value<JArray>();
+            var metadata = attribute[nameof(BoundAttributeDescriptor.Metadata)].Value<JObject>();
+
+            builder
+                .Name(name)
+                .TypeName(typeName)
+                .Documentation(documentation)
+                .AsDictionary(indexerNamePrefix, indexerTypeName);
```
NTaylorMullen: What happens if it's not a dictionary?

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                builder.AddMetadata(item.Key, item.Value);
+            }
+        }
+
+        private RazorDiagnostic ReadDiagnostic(JObject diagnostic, JsonSerializer serializer)
+        {
+            var span = diagnostic[nameof(RazorDiagnostic.Span)].Value<JObject>();
+            var absoluteIndex = span[nameof(SourceSpan.AbsoluteIndex)].Value<int>();
+            var lineIndex = span[nameof(SourceSpan.LineIndex)].Value<int>();
```
NTaylorMullen: SourceLocations should be deserializable already.

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
ajaybhargavb: Nothing special really. Only the values change. I just wanted to have another case that reads and writes different types.

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
ajaybhargavb: Not quite. While deserializing `SourceSpan`, the default deserializer tries to use the constructor with `SourceLocation` and just ends up initializing it with default values.

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
ajaybhargavb: Just realized my response probably made no sense to you here. I thought this was the test class 😆

ajaybhargavb:
🆙 📅

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+            {
+                builder.AddMetadata(item.Key, item.Value);
+            }
+
+            return builder.Build();
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
```
NTaylorMullen: Possible to just use the default behavior of the serializer? Seeing that this class is a converter it should techincally be able to handle either serialization or deserialization.

ajaybhargavb: Technically, the default behavior is being used for serialization. `public override bool CanWrite => false;` takes care of that. As `JsonConverter` is an abstract class, I can't call `base.WriteJson` on this and I can't not override `WriteJson` method.

NTaylorMullen: Oh didn't see the CanWrite bit. Sounds awesome 😄

ajaybhargavb:
🆙 📅

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
```diff
@@ -0,0 +1,194 @@
...
+                return null;
+            }
+
+            var descriptor = JObject.Load(reader);
+            var descriptorKind = descriptor[nameof(TagHelperDescriptor.Kind)].Value<string>();
+            if (descriptorKind != ITagHelperDescriptorBuilder.DescriptorKind)
+            {
+                throw new NotSupportedException(
+                    Resources.FormatTagHelperDescriptorJsonConverter_UnsupportedTagHelperDescriptorKind(
+                        typeof(TagHelperDescriptor).Name, descriptorKind));
```
rynowak: This should be supported.

src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs
rynowak: Leave a comment somewhere that explains **why**

ajaybhargavb:
🆙 📅 FYI only. I'll merge these to one big PR.

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperBinderSyntaxTreePass.cs
```diff
@@ -73,6 +73,8 @@ public RazorSyntaxTree Execute(RazorCodeDocument codeDocument, RazorSyntaxTree s
             // Temporary code while we're still using legacy diagnostics in the SyntaxTree.
             errorList.AddRange(errorSink.Errors.Select(error => RazorDiagnostic.Create(error)));
 
+            errorList.AddRange(descriptors.SelectMany(d => d.GetAllDiagnostics()));
```
ajaybhargavb: @NTaylorMullen, I assume this was your feedback. I've added a test for this here https://github.com/aspnet/Razor/pull/1071/files#diff-a56a4864f6790f3cccbd5887c5bc4743R302

NTaylorMullen: Yup, 👍

PR #1078, opened 2017-03-15
Remove net451 as a cross-compile target

pranavkm:


NTaylorMullen:
Will net46 bits work in VS?

rynowak:
Yes. Our VS bits are already net46

PR #1080, opened 2017-03-15
Port the legacy RazorEditorParser

rynowak:
/cc @ntaylormullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TextChange.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TextChange.cs
@@ -9,7 +9,7 @@ using Microsoft.Extensions.Internal;
 
 namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
 {
-    internal struct TextChange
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TextChange.cs
```diff
@@ -165,15 +165,6 @@ public string ApplyChange(string content, int changeOffset)
...
         }
 
-        /// <summary>
-        /// Applies the text change to the content of the span and returns the new content.
-        /// This method doesn't update the span content.
-        /// </summary>
-        public string ApplyChange(Span span)
-        {
-            return ApplyChange(span.Content, span.Start.AbsoluteIndex);
-        }
```
rynowak: I had to remove this to make `TextChange` public

rynowak:
I've updated this with the TagHelperBlock.LocateOwner part, but I still have some tests to investigate.

rynowak:
Fixed by 90b48347a5dfc23defea3759d324d4671c454c0e

PR #1085, opened 2017-03-17
Verify tag name when matching rules for bound descriptors

ajaybhargavb:
@rynowak @NTaylorMullen


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/TagHelperDescriptorProvider.cs
```diff
@@ -102,8 +104,16 @@ public TagHelperDescriptorProvider(string tagHelperPrefix, IEnumerable<TagHelper
             return tagMappingResult;
         }
 
-        private bool MatchesRule(TagMatchingRule rule, IEnumerable<KeyValuePair<string, string>> tagAttributes, string parentTagName)
+        private bool MatchesRule(TagMatchingRule rule, IEnumerable<KeyValuePair<string, string>> tagAttributes, string tagName, string parentTagName)
```
NTaylorMullen: update this `tagName` to be `tagNameWithoutPrefix`

Issue #980, opened 2017-02-08
String token directives should require quoted values.


pranavkm:
Quotes are optional for some string-like tokens in directives (for instance in `@addTagHelper`). However the optionality is enforced at code gen rather than at parse time. We could instead remove quotes from string tokens at parse time.

rynowak:
We should absolutely not do this. The taghelper directives are special and it is a **non goal** that they reuse the extensible directives feature.

pranavkm:
Would we expect to support `@page "some-route"`? Right now, we end up treating quotes as part of the token. Seems reasonable to generalize this to all "string" like values in descriptors.

rynowak:
We should **only** support `@page "some-route"` and the token api we provide should give you access to the both the 'quoted' span/text and the 'unquoted' span/text.

----------------------------------

I'll repeat, you can do things with the taghelper directives like:
```
@addTagHelper Foo, Bar
@addTagHelper "Baz, Bar" 
```

In both of these cases, everything after `@addTagHelper is a single token (`Foo, Bar` and `"Baz, Bar"`). 

This only works because we know what the syntax of `Foo, Bar` is supposed to be. We fully parse it, and then throw away the information so we can do it again later. 

This kind of thing is a trap and should be avoided. We shipped this in 1.0.0 and it has to stay working, but it shouldn't infect the extensible directives feature.

pranavkm:
Cool I'm all for normalizing it (one way or the other).

rynowak:
> Cool I'm all for normalizing it (one way or the other).

No. It has to be with quotes.

pranavkm:
^ See title.

PR #1088, opened 2017-03-17
Generate parser error for unquoted string token directives

ajaybhargavb:
Issue - #980

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
@@ -1553,7 +1553,19 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
                         outputKind = SpanKind.Code;
                         break;
                     case DirectiveTokenKind.String:
-                        AcceptUntil(CSharpSymbolType.WhiteSpace, CSharpSymbolType.NewLine);
+                        if (At(CSharpSymbolType.StringLiteral))
+                        {
+                            AcceptAndMoveNext();
+                        }
+                        else
+                        {
+                            var startLocation = CurrentStart;
+                            AcceptUntil(CSharpSymbolType.WhiteSpace, CSharpSymbolType.NewLine);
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -1553,6 +1553,13 @@ private void HandleDirective(DirectiveDescriptor descriptor)
...
                         break;
                     case DirectiveTokenKind.String:
+                        if (!At(CSharpSymbolType.StringLiteral))
+                        {
+                            Context.ErrorSink.OnError(
+                                CurrentStart,
+                                LegacyResources.FormatDirectiveExpectsQuotedStringLiteral(descriptor.Name),
+                                CurrentSymbol.Content.Length);
+                        }
                         AcceptUntil(CSharpSymbolType.WhiteSpace, CSharpSymbolType.NewLine);
```
NTaylorMullen: Do we need `AcceptUntil` now? Isn't `AcceptAndMoveNext` sufficient?

ajaybhargavb: Don't we still want to try and parse the rest of it as correctly as possible? Otherwise for things like `@custom {blah}` it will only accept `{`

NTaylorMullen: Hmm, if that's what we want to do. Lets be specific and move the `AcceptUntil` into the error case (with a comment) and have an else case with `AcceptAndMoveNext`

pranavkm: Should this AcceptUntil the next double quote?

NTaylorMullen: @pranavkm there isn't another double quote, otherwise it would have been tokenized as a string literal 😄

ajaybhargavb:
🆙 📅

PR #1091, opened 2017-03-19
Optimize memory et performance of HtmlContentIRNode.Content

Yves57:
It changes the `HtmlContentIRNode.Content` property type from `string` to `StringBuilder`.
A test with large pages like [this](https://github.com/aspnet/Performance/blob/dev/testapp/LargePageMvc/Views/Home/Index.cshtml) shows a big performance improvement, because it allows to concatenate content instead of creating new strings in `DefaultRazorIRLoweringPhase.MainSourceVisitor.Combine`.
On my computer, compilation of a page takes now ~1 second when it was taking ~22 seconds (!) before...

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
@@ -54,7 +54,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                         builder.Add(new RazorIRToken()
                         {
                             Kind = RazorIRToken.TokenKind.CSharp,
-                            Content = original.Content,
+                            Content = original.Content.ToString(),
                             Source = original.Source,
                         });
                     }
@@ -85,7 +85,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                                 builder.Add(new RazorIRToken()
                                 {
                                     Kind = RazorIRToken.TokenKind.CSharp,
-                                    Content = html.Content,
+                                    Content = html.Content.ToString(),
                                     Source = html.Source,
                                 });
                             }
--- a/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
@@ -234,7 +234,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
                     var htmlNode = node.Children.First() as HtmlContentIRNode;
                     if (htmlNode != null)
                     {
-                        Context.Writer.WriteStringLiteral(htmlNode.Content);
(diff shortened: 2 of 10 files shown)


Yves57:
AppVeyor failed but it does not seem to be related to my PR.

rynowak:
Hi, sorry for taking so long to look at this.

We're not going to take a change that puts a StringBuilder on the public API here.  We'll be making another change soon to 'break up' `HtmlContent into a series of tokens (like we did for `CSharpStatementIRNode` [here](https://github.com/aspnet/Razor/pull/1115)).

That change will make it very easy to 'build' an `HtmlContentIRNode` because it will be possible to just append another token to the list of tokens. If you're really interested you could take a stab at making the change. It will end up with a massive diff in the baselines :laughing:

Yves57:
So the idea is to remove completly the `HtmlContentIRNode.Content` property, and "replace" it by the children `RazorIRNode`?

Yves57:
I was thinking about your remark on `StringBuilder` in a public API.
I don't know if you will finally do the optimization yourself, but I expose my propositions in case you are okay with one of my propositions...

I have seen that the `Content` property is used in several cases, for example [here](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs#L263), [here](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs#L57) or [here](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs#L88). In the two last cases, the property is even used in a different assembly.

I think that it would sad/difficult to remove the property.

So I propose:

Add extensions methods like `GetContent()` and `GetContentLength()`
------------------------------------------

- Easy to implement.
- Maybe no very performant, because each `GetContent()` will allocate a new string.
- Less clean API (property replaced by an extension method).

Keep an internal StringBuilder and a public read-only Content property
------------------------------------------------------------

In this proposition, the `RazorIRToken` are not used. Of course the Content property would be a wrapper around the StringBuilder (maybe with the string cached to not allocate the string two times if asked two times).

- No `RazorIRToken` to allocate, so less GC pressure.
- Different principle than for `CSharpStatementIRNode`.
- An external customer cannot create itself a `HtmlContentIRNode` instance except if we add a new public constructor with content.

Keep an internal StringBuilder linked to the Children list and a public read-only Content property
------------------------------------------------------------

In this proposition, a wrapper around the `Children` list echoes each children list modification with an update to the internal StringBuilder (see attached file for a proof of concept).

- More memory allocations (because of the `RazorIRToken` instances, and because of the wrapper instances).
- Keep the `Content` property API clean, and keep the new `RazorIRToken` principle clean.

Just remove the Content property
------------------------------------------------------------

- More or less copy/paste of the children visit loop each time the `Content` or `Content.Length` properties were used.
- Loose the `Content` public API property.

Something completly different
----------------------------

- Another idea?

[HtmlContentIRNodeChildList.txt](https://github.com/aspnet/Razor/files/887646/HtmlContentIRNodeChildList.txt)

Yves57:
Replaced by #1198.

Issue #1087, opened 2017-03-17
TagHelperBoundAttributeDescriptor needs a HasIndexer property


ajaybhargavb:
We need a clear way to figure out if a descriptor was built with `AsDictionary()`. Right now we use `IndexerTypeName` (like https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Host/ViewComponentTagHelperPass.cs#L152) which is not very intuitve.

ajaybhargavb:
We also need to add some validation for `TagHelperBoundAttributeDescriptor`. Currently, it can be built with null `Name` with no diagnostics.

ajaybhargavb:
c6a35e7b2647d51b2b5187cd0516b518557340e3

PR #1100, opened 2017-03-21
Added HasIndexer property to BoundAttributeDescriptor

ajaybhargavb:
Issue - https://github.com/aspnet/Razor/issues/1087

@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
@@ -35,6 +35,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
             new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'', '*' });
 
         private bool _isEnum;
+        private bool _hasIndexer;
         private string _indexerValueTypeName;
         private string _name;
         private string _propertyName;
@@ -88,6 +89,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
         {
             _indexerNamePrefix = attributeNamePrefix;
             _indexerValueTypeName = valueTypeName;
+            _hasIndexer = true;
 
             return this;
         }
@@ -136,6 +138,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
                 _typeName,
                 _indexerNamePrefix,
                 _indexerValueTypeName,
+                _hasIndexer,
                 _documentation,
                 displayName,
                 _metadata,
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -165,14 +165,11 @@ private IEnumerable<RazorDiagnostic> Validate()
 
             if (string.IsNullOrWhiteSpace(_name))
             {
-                if (_indexerNamePrefix == null)
```
NTaylorMullen: Name can totally be `null` if we have an indexer. I expect this explodes multiple tests.

ajaybhargavb: This exploded zero tests 😱

ajaybhargavb:
Right now we don't do any null check shenanigans in Razor because we only need to use IsIndexerNameMatch(). The reaction for this will only be in Mvc https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Host/ViewComponentTagHelperPass.cs#L152

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
NTaylorMullen: Talked offline. We definitely have tests for this guy so should investigate why those don't fail.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -168,8 +171,8 @@ private IEnumerable<RazorDiagnostic> Validate()
                 if (_indexerNamePrefix == null)
                 {
                     var diagnostic = RazorDiagnosticFactory.CreateTagHelper_InvalidBoundAttributeNullOrWhitespace(
-                        _containingTypeName,
-                        _propertyName);
```
NTaylorMullen: undo

src/Microsoft.AspNetCore.Razor.Evolution/ITagHelperBoundAttributeDescriptorBuilder.cs
ajaybhargavb: Since we validate as part of Build(), the diagnostics are same for both the expected and actual descriptors which makes the tests pass. I'll file an issue to refactor our tests to test the diagnostics separately.

ajaybhargavb: https://github.com/aspnet/Razor/issues/1101

Issue #852, opened 2016-11-02
Tooling Standalone: Define Razor lanuage service


rynowak:
We want to define a Razor language service that is host independent. This will be used by VS but not tied to VS

NTaylorMullen:
@rynowak I'm envisioning this bit as what I'm currently working on. Let me know if I'm misreading the purpose of this issue 😄

rynowak:
We should come up with smaller items here and close this, it's *TOO BIG TO TRACK*

NTaylorMullen:
Closing this since this is more-or-less a super-meta-issue

PR #1102, opened 2017-03-21
[Design] Add workspace monitoring for Roslyn APIs.

NTaylorMullen:
- Added tracking of TagHelper files and project reference changes. Explored other tracking options and found that these were the minimum bar.
- This does not include build tracking.

#852

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
@@ -1,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class WorkspaceMonitor
+    {
+        private readonly Workspace _workspace;
+
+        public WorkspaceMonitor(Workspace workspace)
+        {
+            if (workspace == null)
+            {
+                throw new ArgumentNullException(nameof(workspace));
+            }
+
+            _workspace = workspace;
+
+            _workspace.WorkspaceChanged += OnWorkspaceChanged;
+        }
+
+        public event Action TagHelperFileChanged;
+
+        public event Action ReferencesChanged;
+
+        private void OnWorkspaceChanged(object sender, WorkspaceChangeEventArgs args)
+        {
+            switch (args.Kind)
+            {
+                case WorkspaceChangeKind.DocumentAdded:
+                case WorkspaceChangeKind.DocumentChanged:
+                    var document = _workspace.CurrentSolution.GetDocument(args.DocumentId);
+                    var filePath = document?.FilePath;
+
+                    if (filePath != null &&
+                        filePath.EndsWith(".cs", StringComparison.OrdinalIgnoreCase) &&
+                        filePath.IndexOf("TagHelper", StringComparison.OrdinalIgnoreCase) >= 0)
+                    {
+                        TagHelperFileChanged?.Invoke();
+                    }
+                    break;
+
+                case WorkspaceChangeKind.ProjectChanged:
+                    var change = args.NewSolution.GetChanges(args.OldSolution);
+                    foreach (var projectChange in change.GetProjectChanges())
+                    {
+                        if (projectChange.GetAddedMetadataReferences().Any() ||
+                            projectChange.GetAddedProjectReferences().Any() ||
+                            projectChange.GetRemovedMetadataReferences().Any() ||
+                            projectChange.GetRemovedProjectReferences().Any())
+                        {
+                            ReferencesChanged?.Invoke();
+                        }
+                    }
+
+                    break;
+            }
+        }
+    }
+}
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitorProvider.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitorProvider.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class WorkspaceMonitorProvider
+    {
+        private readonly List<WorkspaceMonitorEntry> _workspaceMonitorEntries;
+
+        public WorkspaceMonitorProvider()
+        {
+            _workspaceMonitorEntries = new List<WorkspaceMonitorEntry>();
+        }
+
+        public WorkspaceMonitor GetWorkspaceMonitor(Workspace workspace)
+        {
+            var entry = GetEntry(workspace);
+
+            entry.MonitorRequestCount++;
+
+            return entry.Monitor;
+        }
+
+        public WorkspaceMonitor StopMonitoring(Workspace workspace)
+        {
+            var entry = GetEntry(workspace);
+
+            entry.MonitorRequestCount--;
+
+            if (entry.MonitorRequestCount == 0)
+            {
+                _workspaceMonitorEntries.Remove(entry);
+            }
+
+            return entry.Monitor;
+        }
+
+        private WorkspaceMonitorEntry GetEntry(Workspace workspace)
+        {
+            var entry = _workspaceMonitorEntries.FirstOrDefault(e => e.Workspace == workspace);
+            if (entry == null)
+            {
+                entry = new WorkspaceMonitorEntry(workspace);
+                _workspaceMonitorEntries.Add(entry);
+            }
+
+            return entry;
+        }
+
+        private class WorkspaceMonitorEntry
+        {
+            public WorkspaceMonitorEntry(Workspace workspace)
+            {
+                Workspace = workspace;
+                Monitor = new WorkspaceMonitor(workspace);
+            }
+
+            public int MonitorRequestCount { get; set; }
+
+            public Workspace Workspace { get; }
+
+            public WorkspaceMonitor Monitor { get; }
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
```diff
@@ -0,0 +1,63 @@
...
+using System;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class WorkspaceMonitor
+    {
+        private readonly Workspace _workspace;
+
+        public WorkspaceMonitor(Workspace workspace)
```
NTaylorMullen: Didn't think it was worth following the abstract class => static create method pattern due to how this class will be used. Let me know if you think differently.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorkerProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public RazorEditorWorker GetWorker(ITextBuffer documentBuffer)
+        {
+            var roslynBuffer = GetRoslynBuffer(documentBuffer);
+            var editorWorker = new RazorEditorWorker();
+
+            var registrationMonitor = _workspaceProvider.GetRegistrationMonitor(roslynBuffer);
+            registrationMonitor.ConnectedToWorkspace += workspace =>
+            {
```
NTaylorMullen: Is it best practice to have a `ForegroundThreadAffinitizedObject` like approach when dealing with callbacks from external code?

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
rynowak: In general things in 'workspaces' are going to use MEF and also likely the factory pattern. See https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/Workspace/Host/Mef/ILanguageServiceFactory.cs

Assume that everything here **requires** a separation between abstraction and implementation.

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            _workspace = workspace;
+
+            _workspace.WorkspaceChanged += OnWorkspaceChanged;
+        }
+
+        public event Action TagHelperFileChanged;
+
+        public event Action ReferencesChanged;
```
rynowak: These events will probably need an `EventArgs`. The shape of these events implies that you want to know what file has changed, or what reference change as well.

I'd really suggest just consolidating these into one event, and have an `EventArgs` that describes the change. Even if that `EventArgs` class is empty now, we can add members to it in the future.

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        private void OnWorkspaceChanged(object sender, WorkspaceChangeEventArgs args)
+        {
+            switch (args.Kind)
+            {
+                case WorkspaceChangeKind.DocumentAdded:
+                case WorkspaceChangeKind.DocumentChanged:
```
rynowak: What about document removed?

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitorProvider.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class WorkspaceMonitorProvider
```
rynowak: See `ILanguageServiceFactory`

rynowak: Actually this one should be an https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Portable/Workspace/Host/IWorkspaceService.cs

rynowak: I'd suggest to tying the monitor to a single workspace, rather than keeping a global 'workspace tracker'.

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitor.cs
```diff
@@ -0,0 +1,63 @@
...
+        public WorkspaceMonitor(Workspace workspace)
+        {
+            if (workspace == null)
+            {
+                throw new ArgumentNullException(nameof(workspace));
+            }
+
+            _workspace = workspace;
+
+            _workspace.WorkspaceChanged += OnWorkspaceChanged;
```
rynowak: You will need to unsubscribe at some point. If you ever define an event you either need the *source* and *sink* to have the same lifetime, or you need to unsubscription protocol.

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceMonitorProvider.cs
rynowak: Yeah I'd really suggest a `WorkspaceMonitorFactory` that is an `IWorkspaceServiceFactory` that returns a single shared monitor instance per-workspace. Then clients just ask for `WorkspaceMonitor` and the factory transparently handles subscriptions.

rynowak: In this model the monitor has the same lifetime as the workspace so you don't need to unsubscribe.

src/Microsoft.CodeAnalysis.Razor.Workspaces/WorkspaceRegistrationMonitor.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class WorkspaceRegistrationMonitor
```
rynowak: This doesn't abstract anything. It's like, literally the same semantic as `WorkspaceRegistration`.

src/Microsoft.VisualStudio.LanguageServices.Razor/IRazorEditorWorkerProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public interface IRazorEditorWorkerProvider
```
rynowak: Let's make these abstract classes and call em `...Factories`.

In general the VS convention is to use works like `FooServiceFactory` and `FooService`. I think `RazorEditorWorker` is a good name.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorkerProvider.cs
rynowak: Yes, it helps validate your assumptions.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorkerProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+                workspaceMonitor.TagHelperFileChanged += OnTagHelpersChanged;
+                workspaceMonitor.ReferencesChanged += OnTagHelpersChanged;
+            };
+            registrationMonitor.DisconnectedFromWorkspace += workspace =>
+            {
+                var workspaceMonitor = _workspaceMonitorProvider.StopMonitoring(workspace);
+
+                workspaceMonitor.TagHelperFileChanged -= OnTagHelpersChanged;
+                workspaceMonitor.ReferencesChanged -= OnTagHelpersChanged;
+            };
```
rynowak: See you could easily write all of this code with `WorkspaceRegistration` you don't need our type.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorkerProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            void OnTagHelpersChanged()
+            {
+                editorWorker.TagHelpersChanged?.Invoke();
+            }
```
rynowak: I would define everything in this method on the worker itself.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorkerProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+            return editorWorker;
+        }
+
+        private SourceTextContainer GetRoslynBuffer(ITextBuffer documentBuffer)
+        {
+            // Need to design a better approach of extracting the roslyn buffer from the document buffer.
+            var roslynBufferProvider = (Func<SourceTextContainer>)documentBuffer.Properties["RoslynBufferProvider"];
+            var roslynBuffer = roslynBufferProvider();
+
+            return roslynBuffer;
```
rynowak: #YOLO

We can also ask that they give us the `SourceTextContainer`. For instance we should have them create the (*empty*) C# document when the editor is opened and then we don't have to deal with initializing it lazily.

rynowak: I think @ToddGrun said that this approach would be preferable to them long term - the Roslyn types are already available for us to use as exchange types.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorEditorWorker.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public class RazorEditorWorker
```
rynowak: Also consider exposing some additional things here like the current `Workspace/ProjectId/DocumentId`.

NTaylorMullen:
Will revive this when the time is right.

Issue #1069, opened 2017-03-09
Spans created by RazorSyntaxTree.Parse(RazorSourceDocument) do not have FilePath specified on the source.


pranavkm:

ajaybhargavb:
cab6eea66332f911a2f620c89997956840a59592

PR #1103, opened 2017-03-22
Set file path from RazorSourceDocument in syntax tree

ajaybhargavb:
Issue - #1069 

@NTaylorMullen @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SeekableTextReader.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SeekableTextReader.cs
@@ -10,20 +10,22 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
     {
         private readonly LineTrackingStringBuffer _buffer;
         private int _position = 0;
-        private SourceLocation _location = SourceLocation.Zero;
+        private SourceLocation _location;
         private char? _current;
 
-        public SeekableTextReader(string source) : this(source.ToCharArray()) { }
+        public SeekableTextReader(string source, string filePath) : this(source.ToCharArray(), filePath) { }
 
-        public SeekableTextReader(char[] source)
+        public SeekableTextReader(char[] source, string filePath)
         {
             if (source == null)
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
-            _buffer = new LineTrackingStringBuffer(source);
+            _buffer = new LineTrackingStringBuffer(source, filePath);
             UpdateState();
+
+            _location = new SourceLocation(filePath, 0, 0, 0);
         }
 
         public SourceLocation Location => _location;
--- a/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Evolution.Legacy;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Razor.Evolution.Test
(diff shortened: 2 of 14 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SeekableTextReader.cs
```diff
@@ -13,17 +13,19 @@ internal class SeekableTextReader : TextReader, ITextDocument
...
             if (source == null)
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
-            _buffer = new LineTrackingStringBuffer(source);
+            _buffer = new LineTrackingStringBuffer(source, filePath);
             UpdateState();
+
+            _location.FilePath = filePath;
```
NTaylorMullen: This is broken. SourceLocation is a struct. Something amiss here 😄

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
```diff
@@ -20,5 +20,21 @@ public void Parse_CanParseEmptyDocument()
...
+            var filePath = "test.cshtml";
+            var source = TestRazorSourceDocument.Create("@Hello", fileName: filePath);
+
+            // Act
+            var syntaxTree = RazorSyntaxTree.Parse(source);
+
+            // Assert
+            Assert.Empty(syntaxTree.Diagnostics);
+            Assert.NotNull(syntaxTree);
+            Assert.All(syntaxTree.Root.Children, node => Assert.Equal(filePath, node.Start.FilePath));
```
pranavkm: Isn't there like one child in the tree? Could we have a more complex graph /w nested children and verify it's persisted in all of those?

ajaybhargavb: There are 3 children. But I can make this more complex.

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
```diff
@@ -20,5 +22,39 @@ public void Parse_CanParseEmptyDocument()
...
+
+            // Assert
+            Assert.Empty(syntaxTree.Diagnostics);
+            Assert.NotNull(syntaxTree);
+
+            var spans = new List<SyntaxTreeNode>();
+            GetChildren(spans, syntaxTree.Root);
+            Assert.All(spans, node => Assert.Equal(filePath, node.Start.FilePath));
+
+            void GetChildren(List<SyntaxTreeNode> nodes, SyntaxTreeNode node)
```
ajaybhargavb: PranavPoints?

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
pranavkm: Really no good reason to do this. Plus you aren't capturing anything in the closure, so not sure if it helps being a local function.

pranavkm:
Btw, this change will affect https://github.com/aspnet/Mvc/blob/f7fd5114b34cd12ce20ce05fe2111f95b54d2bdb/test/Microsoft.AspNetCore.Mvc.Razor.Test/Internal/RazorCompilerTest.cs#L111.

test/Microsoft.AspNetCore.Razor.Evolution.Test/RazorSyntaxTreeTest.cs
ajaybhargavb: True. I'll make it so that the function directly adds nodes to `spans` instead of passing it as a parameter.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/SeekableTextReader.cs
```diff
@@ -10,20 +10,22 @@ internal class SeekableTextReader : TextReader, ITextDocument
...
             if (source == null)
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
-            _buffer = new LineTrackingStringBuffer(source);
+            _buffer = new LineTrackingStringBuffer(source, filePath);
             UpdateState();
+
+            _location = new SourceLocation(filePath, 0, 0, 0);
```
NTaylorMullen: Did tests fail before you "fixed" this guy? If not, we should add one to catch that scenario.

ajaybhargavb: They didn't. But the new one I added should catch it.

PR #1113, opened 2017-03-22
Don't filter by assembly name if null is provided

rynowak:
This will nullref if null is provided for the assembly name filters. We
want to transition tooling to not pass any assembly names.

/cc @NTaylorMullen


src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -41,7 +39,7 @@ private void VisitTagHelpers(Compilation compilation, IEnumerable<string> assemb
 
             foreach (var type in types)
             {
-                if (assemblyNameFilters.Contains(type.ContainingAssembly.Identity.Name))
+                if (assemblyNameFilters == null ||assemblyNameFilters.Contains(type.ContainingAssembly.Identity.Name))
```
NTaylorMullen: space after `||`

PR #1116, opened 2017-03-24
Simplified if condition

ajaybhargavb:
@rynowak


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRenderingContext.cs
```diff
@@ -62,17 +62,10 @@ public void AddLineMappingFor(RazorIRNode node)
                 return;
             }
 
-            var imports = CodeDocument.GetImportSyntaxTrees();
-            if (imports != null)
+            if (!string.Equals(SourceDocument.FileName, node.Source.Value.FilePath, StringComparison.OrdinalIgnoreCase))
```
rynowak: Should we also check if the `SourceDocument.FileName == null` here? For instance they could both be null, and that shouldn't result in a linemapping.

PR #1123, opened 2017-03-24
Make the parser options configurable

rynowak:
The RazorParserOptions were only configurable via an internal interface
and an extension method on the builder. This isn't suitable for VS because
we need to be able to update the configuration while the editor is open,
without creating a new engine.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
@@ -107,7 +107,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
         internal static void AddDesignTimeDefaults(IRazorEngineBuilder builder)
         {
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/RazorEngine.cs
```diff
@@ -116,17 +116,5 @@ internal static void AddDesignTimeDefaults(IRazorEngineBuilder builder)
         public abstract IReadOnlyList<IRazorEnginePhase> Phases { get; }
 
         public abstract void Process(RazorCodeDocument document);
-
-        internal class ConfigureDesignTimeOptions : IRazorConfigureParserFeature
```
rynowak: Moved to a top-level class and renamed.

rynowak:
565dd309575d3ed22dab545fd454c39cff4da7af

PR #1126, opened 2017-03-24
Add a static 'register' method to MVC's extensions

rynowak:
This will be called by the IDE, and by MVC to register all of the
functionality in Microsoft.AspNet.Mvc.Razor.Extensions.


rynowak:
> Looks good, I assume the Features we're adding which didn't exist in the code below are to handle things that will be calling this function but aren't yet?

I think you probably mean things like the metadata references feature and the taghelpers feature. 

These aren't part of the MVC Razor codegen extensibility, they are part of the environment. Runtime/MVC will provide these in one way, VS will do it in another.

rynowak:
1683018d71361cb5eaea0f878f8649103d942ef5

PR #1130, opened 2017-03-27
Add a service for initializing tooling

rynowak:
This is a new API for tooling to call to get a RazorTemplateEngine that's
initialized correctly. For now this is hardcoded to use MVC's conventions.

This is also a very temporary design to get us past the next wave of
changes.

We'll eventually want to change this to an ILanguageService, but this will
require some new plumbing in the editor, so let's defer that for another
day.

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorTemplateEngineFactoryService.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorTemplateEngineFactoryService.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.AspNetCore.Mvc.Razor.Extensions;
+using Microsoft.AspNetCore.Razor.Evolution;
+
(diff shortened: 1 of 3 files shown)


rynowak:
Currently testing this out.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorTemplateEngineFactoryService.cs
```diff
@@ -0,0 +1,34 @@
...
+            var engine = RazorEngine.CreateDesignTime(b =>
+            {
+                configure?.Invoke(b);
+
+                // For now we're hardcoded to use MVC's extensibility.
+                RazorExtensions.Register(b);
+            });
+
+            var templateEngine = new MvcRazorTemplateEngine(engine, new FileSystemRazorProject(projectPath));
+            templateEngine.Options.ImportsFileName = "_ViewImports.cshtml";
```
NTaylorMullen: Curious: How will `_PageImports.cshtml` fit in here?

rynowak: We need to change the design. Doing conditional work to figure out what filenames we're looking at isn't going to fly.

We've been considering just making both supported for all pages and views.

rynowak:
This was merged a long time ago. bc9b9876e5f1e5295b160f4394d35d49968c4b9a

Issue #1137, opened 2017-03-28
Using TModel in ViewImports causes compile errors at design time


NTaylorMullen:
We generate directive helper statements to provide intellisense for `@inject IHtmlHelper<TModel>` that look like: 

```C#
((System.Action)(() => {
    global::Microsoft.AspNetCore.Mvc.Rendering.IHtmlHelper<TModel> __typeHelper = null;
}
))();
```

The problem is `TModel` isn't a valid C# term. @pranavkm got around this in the past by essentially doing a `using TModel = dynamic` statement at the top of the file.

ajaybhargavb:
0b5113c76ee3391bd2e46f2f62dad7a5081d8c14

PR #1141, opened 2017-03-28
Suppress compilation errors for TModel everywhere

ajaybhargavb:
Issue - #1137 
@NTaylorMullen @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
         public static IRazorEngineBuilder Register(IRazorEngineBuilder builder)
         {
             builder.AddDirective(Directive);
-            builder.Features.Add(new Pass());
+            builder.Features.Add(new Pass(builder.DesignTime));
             return builder;
         }
 
@@ -58,6 +58,13 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 
         internal class Pass : RazorIRPassBase, IRazorDirectiveClassifierPass
         {
+            private readonly bool _designTime;
+
+            public Pass(bool designTime)
+            {
+                _designTime = designTime;
+            }
+
             // Runs after the @inherits directive
             public override int Order => 5;
 
@@ -65,6 +72,19 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             {
                 var visitor = new Visitor();
                 var modelType = GetModelType(irDocument, visitor);
+
+                if (_designTime)
+                {
+                    // Alias the TModel token to a known type ('object' if the model type is unknown).
+                    // This allows design time compilation to succeed for Razor files where the token isn't replaced.
+                    var typeName = modelType == "dynamic" ? $"global::{typeof(object).FullName}" : modelType;
+                    var usingNode = new UsingStatementIRNode()
+                    {
+                        Content = $"TModel = {typeName}"
+                    };
+
+                    visitor.Namespace?.Children.Insert(0, usingNode);
+                }
 
                 var baseType = visitor.Class?.BaseType?.Replace("<TModel>", "<" + modelType + ">");
                 for (var i = visitor.InheritsDirectives.Count - 1; i >= 0; i--)
@@ -85,11 +105,23 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 
         private class Visitor : RazorIRNodeWalker
         {
+            public NamespaceDeclarationIRNode Namespace { get; private set; }
+
             public ClassDeclarationIRNode Class { get; private set; }
 
             public IList<DirectiveIRNode> InheritsDirectives { get; } = new List<DirectiveIRNode>();
 
             public IList<DirectiveIRNode> ModelDirectives { get; } = new List<DirectiveIRNode>();
+
+            public override void VisitNamespace(NamespaceDeclarationIRNode node)
+            {
+                if (Namespace == null)
+                {
+                    Namespace = node;
+                }
+
+                base.VisitNamespace(node);
+            }
 
             public override void VisitClass(ClassDeclarationIRNode node)
             {
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -3,10 +3,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.AspNetCore.Razor.Evolution.Intermediate;
 using Microsoft.AspNetCore.Razor.Evolution.Legacy;
-using System.Diagnostics;
 
 namespace Microsoft.AspNetCore.Razor.Evolution
 {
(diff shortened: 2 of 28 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -38,6 +39,17 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
                 }
             }
 
+            var sourceFileName = Path.GetFileName(codeDocument.Source.FileName);
+            var importsFileName = (string)codeDocument.Items[RazorTemplateEngine.ImportsFileNameKey];
+            if (importsFileName != null &&
```
pranavkm: `string.Equals` doesn't mind null

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.AspNetCore.Razor.Evolution.Intermediate;
 using Microsoft.AspNetCore.Razor.Evolution.Legacy;
 using System.Diagnostics;
+using System.IO;
```
pranavkm: Sort usings

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/RazorEngineTest.cs
```diff
@@ -174,13 +174,14 @@ private static void RunDesignTimeTest(string testName)
             });
 
             var inputContent = ResourceFile.ReadResource(_assembly, inputFile, sourceFile: true);
-            var item = new TestRazorProjectItem(inputFile) { Content = inputContent, };
+            var item = new TestRazorProjectItem("/" + inputFile) { Content = inputContent, };
```
ajaybhargavb: `FindHierarchicalItems` throws if it doesn't start with a `/`.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/RazorEngineTest.cs
```diff
@@ -131,13 +130,14 @@ private static void RunRuntimeTest(string testName)
...
             var project = new TestRazorProject(new List<RazorProjectItem>()
             {
                 item,
             });
 
             var razorTemplateEngine = new MvcRazorTemplateEngine(engine, project);
+            razorTemplateEngine.Options.ImportsFileName = "_ViewImports.cshtml";
```
ajaybhargavb: Although this doesn't matter, I am leaving these here because it is better to test with this set.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -38,6 +38,15 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
                 }
             }
 
+            if (syntaxTree.Options.DesignTimeMode)
+            {
+                // This is needed to supress compilation errors about unknown TModel type.
+                builder.Add(new UsingStatementIRNode()
+                {
+                    Content = $"TModel = {typeof(object).FullName}"
```
NTaylorMullen: Should not be in Razor.Evolution

src/Microsoft.AspNetCore.Razor.Evolution/RazorTemplateEngine.cs
```diff
@@ -151,8 +152,10 @@ public virtual RazorCodeDocument CreateCodeDocument(RazorProjectItem projectItem
 
             var source = RazorSourceDocument.ReadFrom(projectItem);
             var imports = GetImports(projectItem);
+            var codeDocument = RazorCodeDocument.Create(source, imports);
+            codeDocument.Items[ImportsFileNameKey] = Options.ImportsFileName;
```
NTaylorMullen: Don't still need to do this.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -58,6 +58,13 @@ private static string GetModelType(DocumentIRNode document, Visitor visitor)
 
         internal class Pass : RazorIRPassBase, IRazorDirectiveClassifierPass
         {
+            private bool _designTime;
```
NTaylorMullen: readonly

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -66,6 +73,19 @@ public override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode
                 var visitor = new Visitor();
                 var modelType = GetModelType(irDocument, visitor);
 
+                if (_designTime)
+                {
+                    // This is needed to supress compilation errors about unknown TModel type.
+                    var @namespace = visitor.Class.Parent;
```
NTaylorMullen: Lets not assume that the classes parent is the namespace. You can just raise the namespace in the visitor.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -66,6 +73,19 @@ public override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode
...
+                    var classIndex = @namespace.Children.IndexOf(visitor.Class);
```
NTaylorMullen: Same assumption is made here.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -66,6 +73,17 @@ public override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode
                 var visitor = new Visitor();
                 var modelType = GetModelType(irDocument, visitor);
 
+                if (_designTime)
+                {
+                    // This is needed to supress compilation errors about unknown TModel type.
+                    var usingNode = new UsingStatementIRNode()
+                    {
+                        Content = $"TModel = {typeof(object).FullName}"
```
pranavkm: `using TModel = global::{typeof(object).FullName)}`

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -66,6 +73,17 @@ public override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode
                 var visitor = new Visitor();
                 var modelType = GetModelType(irDocument, visitor);
 
+                if (_designTime)
+                {
+                    // This is needed to supress compilation errors about unknown TModel type.
```
pranavkm: > Alias the `TModel` token to a known type (in this case object). This allows design time compilation to succeed for Razor files where the token isn't replaced.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
pranavkm: Wouldn't we use `modelType` here or does that not mater?

NTaylorMullen: > Wouldn't we use modelType here or does that not mater?

Hmm, interesting idea. We didn't in the past, probably a good idea though 👍

ajaybhargavb:
🆙 📅

Issue #1138, opened 2017-03-28
Imported directives have their tokens rendered for IntelliSense


NTaylorMullen:
We currently only verify that source locations do not equal `null` when [rendering directive token nodes](https://github.com/aspnet/Razor/blob/737a9a58ad303b08fa692fe7cb95f80cfda4d8ff/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs#L113). This isn't sufficient, we need to verify that a token node's source location also exists in the primary document.

ajaybhargavb:
What about the default imports aka https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/Output/DesignTime/Basic.cs#L36-L79?

rynowak:
Default imports won't have a source file associated with them

NTaylorMullen:
> What about the default imports aka https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/Output/DesignTime/Basic.cs#L36-L79?

They shouldn't be rendered.

ajaybhargavb:
c1500da2a85eaa88ba9b4a1f32ae4fc5cd69d040

PR #1143, opened 2017-03-29
Don't render directive token from imports in design time

ajaybhargavb:
Issue - #1138 

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
@@ -110,8 +110,13 @@ namespace Microsoft.AspNetCore.Razor.Evolution.CodeGeneration
             const string TypeHelper = "__typeHelper";
 
             var tokenKind = node.Descriptor.Kind;
-            if (node.Source == null)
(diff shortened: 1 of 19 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -110,8 +110,14 @@ public override void VisitDirectiveToken(DirectiveTokenIRNode node)
             const string TypeHelper = "__typeHelper";
 
             var tokenKind = node.Descriptor.Kind;
-            if (node.Source == null)
+            if (!node.Source.HasValue ||
+                (Context.SourceDocument.FileName != null &&
```
rynowak: Why is the null check needed here? Is this a workaround for some test issue?

ajaybhargavb: It is just to handle the unlikely case of both of them being null.

ajaybhargavb: Just realized that doesn't matter. Will remove.

rynowak:
:shipit: after removing silly null check

Issue #1145, opened 2017-03-29
Resolving TagHelpers from an assembly that only has runtime TagHelpers creates an error at design time.


NTaylorMullen:
If an assembly (`Microsoft.AspNetCore.Mvc.Razor`) only contains `TagHelper`s that have editor browsable never then we log an error in our binder syntax tree pass stating that no TagHelpers exist in the assembly.

NTaylorMullen:
I temporarily worked around this issue to unblock tooling folks in [this](3c0d69ddd6d0b8f0941ffc9d5d9d49e05481be35) commit. We never used to validate that an assembly had `TagHelper`s in the old world so my temporary workaround might be an ok fix. However, the error was definitely intentional so it means there was a drive for it. Other fixes could be:

1. Have a property on descriptors, `descriptor.EditorBrowsable`, and then filter them out before returning any to the editor.
2. Leave the error but conditionally add Mvc.Razor tag helper bits based on if design time or not. This would result in the `MvcRazorTemplateEngine` signature changing to include a design time parameter.

Looking for opinions @rynowak / @pranavkm / @ajaybhargavb

ajaybhargavb:
> Have a property on descriptors, descriptor.EditorBrowsable, and then filter them out before returning any to the editor.

Will doing this make it so that we create descriptor only for this error check and the ones with EditorBrowable set to false will be pretty useless otherwise?

NTaylorMullen:
> Will doing this make it so that we create descriptor only for this error check and the ones with EditorBrowable set to false will be pretty useless otherwise?

yar

rynowak:
Close? I like the solution we ended up with.

NTaylorMullen:
https://github.com/aspnet/Razor/commit/dd9eab551d64aafbd9127497bd70a42e12300c1f

PR #1146, opened 2017-03-29
Workaround tooling issue with runtime only TagHelpers.

NTaylorMullen:
This is an FYI PR since it's a potential temporary workaround to unblock tooling.

- Prior to this if all instances of `TagHelper`s in an assembly had editor browsable never we'd log an error saying that no TagHelpers were found for that assembly. Until we can evaluate a better fix for this issue I've removed the logic that logs those errors and its corresponding tests/resources.

#1145


rynowak:
I think this is fine. We can live without this error for a while and see if we need to bring it back.

PR #1151, opened 2017-03-31
Update VSIX metadata

pranavkm:


build/VSIX.targets
```diff
@@ -17,7 +17,7 @@
 
   <Target Name="_LocateMSBuildExe" Condition="Exists('$(MSBuildProgramFiles32)')">
     <ItemGroup>
-      <MSBuild15ExePaths Include="$(MSBuildProgramFiles32)\Microsoft Visual Studio\2017*\**\MSBuild\15.0\Bin\MSBuild.exe" />
+      <MSBuild15ExePaths Include="$(MSBuildProgramFiles32)\Microsoft Visual Studio\**\MSBuild\15.0\Bin\MSBuild.exe" />
```
pranavkm: My VS install is under Microsoft Visual Studio\Preview dir now.

joeloff: TeamCity 10.0.4 I know has support for VS 2017 to figure out all the locations for things like MS BUILD BTW.

joeloff:
:shipit:

build/VSIX.targets
pranavkm: I'd rather not split this off as a separate step though. Doing that requires using the msbuild runner

Issue #1117, opened 2017-03-24
FileSystemRazorProject should be internal


rynowak:
This should be an internal class that you create with a call to `RazorProject.Create(string)`

PR #1152, opened 2017-03-31
Make FileSystemRazorProject internal

Yves57:
See #1117.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorProject.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorProject.cs
@@ -128,7 +128,7 @@ namespace Microsoft.AspNetCore.Razor.Evolution
         /// Create a Razor project based on a physical file system.
         /// </summary>
         /// <param name="rootDirectoryPath">The directory to root the file system at.</param>
-        /// <returns></returns>
+        /// <returns>A <see cref="RazorProject"/></returns>
         public static RazorProject Create(string rootDirectoryPath)
         {
             return new FileSystemRazorProject(rootDirectoryPath);
--- a/src/Microsoft.AspNetCore.Razor.Evolution/RazorProject.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/RazorProject.cs
@@ -123,5 +123,15 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
             return path;
         }
+
+        /// <summary>
+        /// Create a Razor project based on a physical file system.
+        /// </summary>
+        /// <param name="rootDirectoryPath">The directory to root the file system at.</param>
+        /// <returns></returns>
+        public static RazorProject Create(string rootDirectoryPath)
+        {
+            return new FileSystemRazorProject(rootDirectoryPath);
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


pranavkm:
Probably need to make https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Evolution/FileSystemRazorProjectItem.cs#L11 internal too

Yves57:
@pranavkm I agree, but the specific property `FileSystemRazorProjectItem.File` is used in [RazorPageGenerator](https://github.com/aspnet/Razor/blob/dev/src/RazorPageGenerator/Program.cs#L133). Maybe replace this line by `var basePath = Path.GetDirectoryName(_source.PhysicalPath);`?

pranavkm:
Sure, that should suffice.

test/Microsoft.AspNetCore.Razor.Evolution.Test/FileSystemRazorProjectTest.cs
```diff
@@ -68,7 +68,7 @@ public void NormalizeAndEnsureValidPath_NormalizesToForwardSlashes()
         public void EnumerateItems_DiscoversAllCshtmlFiles()
         {
             // Arrange
-            var fileSystemProject = new FileSystemRazorProject(TestFolder);
```
rynowak: It's better to call the constructor here. We want to make sure that we're really testing this class.

src/RazorPageGenerator/Program.cs
```diff
@@ -130,7 +130,7 @@ public override Stream Read()
 
             private string ProcessFileIncludes()
             {
-                var basePath = _source.File.DirectoryName;
+                var basePath = System.IO.Path.GetDirectoryName(_source.PhysicalPath);
```
pranavkm: Does it conflict with the property name?

Yves57: Yes

Yves57:
Updated PR.
Strange thing: I had to change the RazorPageGenerator target framework from `netcoreapp1.0` to `netcoreapp2.0` to be able to run it on my computer.

pranavkm:
> Strange thing: I had to change the RazorPageGenerator target framework from netcoreapp1.0 to netcoreapp2.0 to be able to run it on my computer.

You might not have the 1.1.0 shared runtime installed. Unfortunately the package is used as a tool and there are issues that need to be worked out before it can start targeting netcoreapp2.0.

Yves57:
@pranavkm The 1.1.0 runtime seems to be installed on my computer. Anyway, if it is only on my computer, it is not important 😃. Don't waste your time.
It was just to be sure that there is no bug in the RazorPageGenerator project.

src/Microsoft.AspNetCore.Razor.Evolution/RazorProject.cs
```diff
@@ -123,5 +123,15 @@ protected virtual string NormalizeAndEnsureValidPath(string path)
 
             return path;
         }
+
+        /// <summary>
+        /// Create a Razor project based on a physical file system.
+        /// </summary>
+        /// <param name="rootDirectoryPath">The directory to root the file system at.</param>
+        /// <returns></returns>
```
pranavkm: @Yves57 if you're around, could you fix this to say `The <see cref="RazorProject" />`. I can touch this up prior to merging it if I don't hear from you.

Yves57: Done

pranavkm:
Thanks for the PR!

Issue #1155, opened 2017-04-03
Create LineMappings for 0 length spans to ensure IntelliSense is provided when typing @


NTaylorMullen:
We no longer generate line mappings for 0 length spans but at design time we must. It enables a user typing the at symbol by itself and getting IntelliSense.

NTaylorMullen:
4c0afbad86f969acda6947ef07e464995ac74987

rynowak:
I'm glad that you fixed this to unblock the tooling experience. The way that this works today is largely an artifact of the current design.

I have some ideas about this, we might want to take a harder look at `RazorCSharpDocument` before we ship.

PR #1156, opened 2017-04-03
Generate line mappings for empty expressions.

NTaylorMullen:
- At design time we weren't generating line mappings when a user would type `@` or `@(`. This results in no C# IntelliSense being provided to the user because the editor hasn't mapped any of Razor to the C# buffer.
- Updated the design time renderer and design time writer to account for empty expressions.
- Modified the `DefaultIRLoweringPhase` to set source locations on empty expression nodes.
- Re-generated test files to account for 0 length line mappings on empty expression nodes.

#1155


NTaylorMullen:
Merging this now to unblock tooling folks. Feel free to review after it's been merged, I recommend doing so with `?w=1` in the url to ignore whitespace.

NTaylorMullen:
> Looks good. Did we miss this when we ported the code to Evolution? Or did we never generate this before?

Eh, "cleaned" it up when porting. Didn't realize it was needed.

PR #1167, opened 2017-04-04
Cleanup around ViewImports

rynowak:
Added a default for the imports filename.

Deleted unused test files


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/MvcRazorTemplateEngine.cs
```diff
@@ -23,6 +23,7 @@ public class MvcRazorTemplateEngine : RazorTemplateEngine
             RazorProject project)
             : base(engine, project)
         {
+            Options.ImportsFileName = "_ViewImports.cshtml";
```
rynowak: This is going to be case for both Pages and Views

rynowak: Note that this is in the MVC extensions.

pranavkm: I guess this is going to break Mvc functionals. Could you get the PR for it ready?

rynowak: This should not break anything. The Page loader sets this value explicitly. In fact both places in MVC where this class are used set it.

I'm working on some other changes to MVC to streamline this.

rynowak:
I don't think this is going to break anything because MVC sets this value. Getting this in first is going to help my MVC PR so I'm going to push it.

rynowak:
3ec4e244db1d1398d7b0973b54ac607d1bd91227

Issue #1174, opened 2017-04-04
Identical @using statements from imports or defaults override those from main document


NTaylorMullen:
This results in 0 IntelliSense or coloring for using statements that match ones that exist on the main document. 

The reason this occurs is we add using statement IR nodes as we encounter them and then make sure not to duplicate any. Problem is, we add ViewImports/default usings first which don't get source locations; when we encounter a using statement on the main document the using has already been added so we no-op in an effort to not duplicate an already added using.

NTaylorMullen:
72dfb1d9086872bdab96dbf7171fbf0ff146b222

PR #1175, opened 2017-04-04
Updated the ir lowering phase to lazily add namespaces.

NTaylorMullen:
- The lazy addition of namespaces gives the main document lowering phase an opportunity to add source location information which we then add after the main lowering.
- Re-generated csharp to capture addition of using statements that were previously overridden by defaults/imports.

#1174

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
@@ -22,20 +22,10 @@ namespace Microsoft.AspNetCore.Razor.Evolution
 
             document.Options = syntaxTree.Options;
 
-            var namespaces = new HashSet<string>();
-
-            var i = 0;
-            foreach (var namespaceImport in syntaxTree.Options.NamespaceImports)
+            var namespaces = new Dictionary<string, SourceSpan?>(StringComparer.Ordinal);
+            foreach (var defaultNamespace in syntaxTree.Options.NamespaceImports)
             {
-                if (namespaces.Add(namespaceImport))
-                {
-                    var @using = new UsingStatementIRNode()
-                    {
-                        Content = namespaceImport,
-                    };
-
-                    builder.Insert(i++, @using);
-                }
+                namespaces[defaultNamespace] = null;
             }
 
             var checksum = ChecksumIRNode.Create(codeDocument.Source);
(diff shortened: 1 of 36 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -64,16 +54,31 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
 
             visitor.VisitBlock(syntaxTree.Root);
 
+            // In each lowering piece above, namespaces were tracked. We render them here to ensure every
+            // lowering action has a chance to add a source location to a namespace.
```
rynowak: I would also mention here that the **closest** wins. We want to use the `using` from the main source file if possible.

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -64,16 +54,31 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+                    Source = @namespace.Value,
+                };
+
+                builder.Insert(i++, @using);
+            }
+
             codeDocument.SetIRDocument(document);
         }
 
         private class LoweringVisitor : ParserVisitor
```
rynowak: Unit test?

src/Microsoft.AspNetCore.Razor.Evolution/DefaultRazorIRLoweringPhase.cs
```diff
@@ -64,16 +54,31 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+            // In each lowering piece above, namespaces were tracked. We render them here to ensure every
+            // lowering action has a chance to add a source location to a namespace.
+
+            var i = builder.Current.Children.IndexOf(checksum) + 1;
+            foreach (var @namespace in namespaces)
+            {
+                var @using = new UsingStatementIRNode()
+                {
+                    Content = @namespace.Key,
+                    Source = @namespace.Value,
```
ajaybhargavb: When there are multiple identical usings, any chance one with null source appear later and overwrite one with source?

NTaylorMullen: Not in this phase 😄

Issue #1110, opened 2017-03-22
CodeGeneration should not generate line pragmas from unknown file paths


NTaylorMullen:
In the case of MVC extensions we generate dynamic `_ViewImports.cshtml` file equivalents that don't end up having any actual existence on disk. When one of those files has an `@using Foo` we codegen a line pragma for `Foo` that does not have a filename. We don't want innaccurate line pragma errors due to lack of file name so we shouldn't render a pragma if a filename is not set on a source location.

NTaylorMullen:
fdc8f79fd4ca73b7ceace62e0660c189900a8784

PR #1178, opened 2017-04-04
Do not generate C# line pragmas without a file path.

NTaylorMullen:
- Prior to this change default imports would get line pragmas generated for them because thye'd have a source location but no file path (they were a dynamic document).
- Re-generated C# files to reflect new line pragma changes.

#1110

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeWriter.cs
@@ -509,6 +509,12 @@ namespace Microsoft.AspNetCore.Razor.Evolution.Legacy
 
         public IDisposable BuildLinePragma(SourceSpan documentLocation)
         {
+            if (string.IsNullOrEmpty(documentLocation.FilePath))
+            {
+                // Can't build a valid line pragma without a file path.
+                return NullDisposable.Default;
+            }
+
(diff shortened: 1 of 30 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeWriter.cs
```diff
@@ -509,6 +509,12 @@ public CSharpCodeWritingScope BuildMethodDeclaration(string accessibility, strin
 
         public IDisposable BuildLinePragma(SourceSpan documentLocation)
         {
+            if (string.IsNullOrEmpty(documentLocation.FilePath))
+            {
+                // Can't build a valid line pragma without a file path.
+                return new NoopDisposable();
```
ajaybhargavb: Return a static instance instead of `new`ing it up everytime

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeWriter.cs
```diff
@@ -632,5 +638,12 @@ public void Dispose()
                     .SetIndent(_startIndent);
             }
         }
+
+        private class NoopDisposable : IDisposable
```
ajaybhargavb: Suggest `NullDisposable` or `EmptyDisposable`. Feel free to ignore this comment.

Issue #1051, opened 2017-03-07
Port C# codegen from CSharpRendering to new BasicWriter/TagHelperWriter


rynowak:
See https://github.com/aspnet/Razor/pull/1050

Break this up into chunks and start porting code + adding unit tests in as small of bites as possible.

rynowak:
This is one of the Razor big rocks remaining

ajaybhargavb:
Checklist:

- [ ] CSharpExpressionIRNode
- [ ] CSharpStatementIRNode
- [ ] HtmlAttributeIRNode (HtmlAttributeValueIRNode / CSharpAttributeValueIRNode) 
- [ ] HtmlContentIRNode
- [ ] InitializeTagHelperStructureIRNode
- [ ] SetTagHelperPropertyIRNode and SetPreallocatedTagHelperPropertyIRNode
- [ ] AddTagHelperHtmlAttributeIRNode
- [ ] CreateTagHelperIRNode
- [ ] ExecuteTagHelpersIRNode

ajaybhargavb:
Dup of #846

PR #1182, opened 2017-04-06
Moved ExecuteTagHelpersIRNode from renderer to writer

ajaybhargavb:
Issue - #1051 

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/CSharpRenderingContext.cs
```diff
@@ -151,7 +151,7 @@ public TagHelperWriterScope(CSharpRenderingContext context, BasicWriter writer)
 
             public void Dispose()
             {
-                _context.BasicWriter = _writer;
+                _context.TagHelperWriter = _writer;
```
rynowak: lol oops

rynowak:
I really don't feel like this is too gross.

ajaybhargavb:
I've filed the investigate issue for this https://github.com/aspnet/Razor/issues/1184 and assigned it to myself. This PR doesn't have to be blocked by that.

Issue #1186, opened 2017-04-06
DefaultTagHelperFactService does not handle opt-out TagHelpers correctly.


NTaylorMullen:
If we have a the following tag: `<!a href="...">Hello</!a>`. You can't query DefaultTagHelperFactService for `!a`; you need to query it for just the `a`. If you happen to have a TagHelper that can match `!a` (a catch-all) then it will apply.

NTaylorMullen:
0005cb93ac8f59bde12cb3aff36ae6e9873d59be

PR #1187, opened 2017-04-06
Update TagHelperMatchingConventions to disallow opt-out prefix.

NTaylorMullen:
- Prior to this change the `TagHelper` parsing would strip the opt-out character (`!`) from tag names that got passed to the TagHelper matching services. At design time this proved to be a problem because they have their own understanding of the HTML document and only pass us full tag names (names that include `!`). This changes the matching conventions to immediately return false if a tag name is seen to contain the `TagHelper` opt-out.
- Added two `DefaultTagHelperFactService` tests to verify that tag names with opt-out prefixes are denied `TagHelperDescriptor`s.

#1186

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Evolution/TagHelperMatchingConventions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Evolution/TagHelperMatchingConventions.cs
@@ -12,6 +12,8 @@ namespace Microsoft.AspNetCore.Razor.Evolution
     {
         public const string ElementCatchAllName = "*";
 
+        public const char ElementOptOutCharacter = '!';
+
         public static bool SatisfiesRule(
             string tagNameWithoutPrefix,
             string parentTagName,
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/TagHelperMatchingConventions.cs
```diff
@@ -66,6 +68,12 @@ public static bool SatisfiesTagName(string tagNameWithoutPrefix, TagMatchingRule
                 throw new ArgumentNullException(nameof(rule));
             }
 
+            if (tagNameWithoutPrefix.Length > 0 && tagNameWithoutPrefix[0] == ElementOptOutCharacter)
```
rynowak: Should we just reject anything with `length == 0`?

NTaylorMullen: Sure 👍

src/Microsoft.AspNetCore.Razor.Evolution/TagHelperMatchingConventions.cs
```diff
@@ -66,6 +68,12 @@ public static bool SatisfiesTagName(string tagNameWithoutPrefix, TagMatchingRule
...
+            {
+                // TagHelper's can never satisfy tag names that are prefixed with the opt-out character.
```
ajaybhargavb: TagHelpers

PR #1190, opened 2017-04-07
Moved HtmlContentIRNode from renderer to writer

ajaybhargavb:
Issue - #1051 

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedBasicWriter.cs
```diff
@@ -70,7 +72,39 @@ public override void WriteHtmlAttribute(CSharpRenderingContext context, HtmlAttr
 
         public override void WriteHtmlContent(CSharpRenderingContext context, HtmlContentIRNode node)
         {
-            _previous.WriteHtmlContent(context, node);
+            if (context.Options.DesignTimeMode)
```
NTaylorMullen: why do we need to do anything at design time?

test/Microsoft.AspNetCore.Razor.Evolution.Test/CodeGeneration/RedirectedBasicWriterTest.cs
```diff
@@ -52,7 +52,7 @@ public void WriteCSharpExpression_DesignTime_DoesNormalWrite()
         public void WriteCSharpExpression_Runtime_SkipsLinePragma_WithoutSource()
         {
             // Arrange
-            var writer = new RedirectedBasicWriter(new DesignTimeBasicWriter(), "test_writer")
```
NTaylorMullen: Why'd these guys change?

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedBasicWriter.cs
```diff
@@ -70,7 +72,39 @@ public override void WriteHtmlAttribute(CSharpRenderingContext context, HtmlAttr
...
+            {
+                _previous.WriteHtmlContent(context, node);
+                return;
+            }
+
+            const int MaxStringLiteralLength = 1024;
```
NTaylorMullen: All of these redirected bits are so ugly. I'm torn between leaving these as is vs. consolidating logic into extensions or helpers. If we plan to go down the path of not having redirect writers at all I'd be down to leave as is and nuke later; otherwise isolating would be better.

test/Microsoft.AspNetCore.Razor.Evolution.Test/CodeGeneration/RedirectedBasicWriterTest.cs
ajaybhargavb: These shouldn't really matter here. But I changed them anyway to keep it consistent with what the test is doing.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedBasicWriter.cs
ajaybhargavb: We don't. But that is not up to this writer to decide. We don't know here what type of writer `_previous` is.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedBasicWriter.cs
ajaybhargavb: Either way I'd prefer not doing that as part of this change. In my opinion, that will be a bigger refactor and should be done separately.

rynowak: Please don't move the goalposts

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedBasicWriter.cs
NTaylorMullen: -_- Would it be awful to have a DesignTime redirected writer and a Runtime redirected writer? That would then leave it up to _previous to use the appropriate one.

ajaybhargavb:
🆙 📅 Changed the design a bit.

PR #1193, opened 2017-04-08
Implement @namespace for pages

rynowak:
This change adds support for @namespace for Pages, and introduces a set of
changes that are needed to support @namespace in the parser.

@namespace and @class have always been treated as reserved words by Razor,
with the intent that someday they would be allowed as directives.

This changes makes that possible.

You will still get an error about @namespace being a reserved word if you
don't have the directive.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
@@ -1,0 +1,191 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
+{
+    public static class NamespaceDirective
+    {
+        private static readonly char[] Separators = new char[] { '\\', '/' }; 
+
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptorBuilder.Create("namespace").AddNamespace().Build();
+
+        public static void Register(IRazorEngineBuilder builder)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException();
+            }
+
+            builder.AddDirective(Directive);
+            builder.Features.Add(new Pass());
+        }
+
+        // internal for testing
+        internal class Pass : RazorIRPassBase, IRazorDirectiveClassifierPass
+        {
+            public override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode irDocument)
+            {
+                if (irDocument.DocumentKind != RazorPageDocumentClassifierPass.RazorPageDocumentKind &&
+                    irDocument.DocumentKind != MvcViewDocumentClassifierPass.MvcViewDocumentKind)
+                {
+                    // Not a page. Skip.
+                    return;
+                }
+
+                var visitor = new Visitor();
+                visitor.Visit(irDocument);
+
+                var directive = visitor.LastNamespaceDirective;
+                if (directive == null)
+                {
+                    // No namespace set. Skip.
+                    return;
+                }
+
+                var @namespace = visitor.FirstNamespace;
+                if (@namespace == null)
+                {
+                    // No namespace node. Skip.
+                    return;
+                }
+                
+                if (TryComputeNamespace(codeDocument.Source.FileName, directive, out var computedNamespace))
+                {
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -110,6 +110,17 @@ protected void MapDirectives(Action handler, params string[] directives)
             {
                 _directiveParsers.Add(directive, handler);
                 Keywords.Add(directive);
+
+                // These C# keywords are reserved for use in directives. It's an error to use them outside of
+                // a directive. This code removes the error generation if the directive *is* registered.
+                if (string.Equals(directive, "class", StringComparison.OrdinalIgnoreCase))
```
rynowak: This change was not a straightforward as Taylor said. Remove these as 'reserved words' unconditionally broke about 200 tests.

src/Microsoft.AspNetCore.Razor.Evolution/Legacy/CSharpCodeParser.cs
```diff
@@ -295,8 +305,17 @@ private void AfterTransition()
                         }
                         else if (CurrentSymbol.Type == CSharpSymbolType.Keyword)
                         {
-                            KeywordBlock(topLevel: true);
-                            return;
+                            if (TryGetDirectiveHandler(CurrentSymbol.Content, out var handler))
```
rynowak: These directives will be tokenized as keywords, so you need to handle that case.

rynowak:
FYI this is blocked on #1194

It looks like this would cause a big diff if we fix it, so I want to pull it out separately.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            // Includes the separator
+            return path.Substring(0, lastSeparator + 1);
+        }
+
+        private class NamespaceDirectiveNodeVisitor : RazorIRNodeWalker
+        {
+            // We want the last one, so get them all and then .
+            public DirectiveIRNode LastNamespaceDirective { get; set; }
```
NTaylorMullen: `private set;` here and elsewhere.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -0,0 +1,195 @@
...
+
+            var baseNamespace = directive.Tokens.First().Content;
+            if (string.IsNullOrEmpty(source) || directiveSource == null)
+            {
+                // No sources, can't compute a suffix.
+                @namespace = baseNamespace;
+                return false;
+            }
+
+            if (!source.StartsWith(directiveSource) || source.Length <= directiveSource.Length)
```
NTaylorMullen: ignorecase when comparing file paths.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -0,0 +1,195 @@
...
+            }
+
+            // OK so that this point we know that the 'imports' file containing this directive is in the directory
+            // hierarchy of this soure file. This is the case where we can append a suffix to the baseNamespace.
+            //
+            // Everything so far has just been defensiveness on our part.
+
+            var builder = new StringBuilder(baseNamespace);
+
+            var segments = source.Substring(directiveSource.Length).Split('/', '\\');
```
NTaylorMullen: `.Split(Separators)`

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -0,0 +1,195 @@
...
+            if (lastSeparator == -1)
+            {
+                return null;
+            }
+
+            // Includes the separator
+            return path.Substring(0, lastSeparator + 1);
+        }
+
+        private class NamespaceDirectiveNodeVisitor : RazorIRNodeWalker
```
NTaylorMullen: Any reason to not have 1 visitor that extracts the directives, namespaces and classes in a single pass?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
NTaylorMullen: Will probably need a test for the case sensitivity as well.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
rynowak: Seems fine

rynowak:

Issue #1196, opened 2017-04-10
Performance of IR lowering is bad on large files


rynowak:
Found via benchmarking

![image](https://cloud.githubusercontent.com/assets/1430011/24846464/812a4dbe-1d6f-11e7-8fea-1c8a54ba877d.png)

Generating code for really large files with lots of HTML is really slow. @Yves57 reported this issue as well, and we need to fix this or it will become a blocking issue on new tooling.

See https://github.com/aspnet/Razor/pull/1091 for details.

We wouldn't take the public API change that's part of this issue, we will likely want the 'list of tokens' approach

rynowak:
See https://github.com/aspnet/Razor/pull/1197 for the test

PR #1197, opened 2017-04-10
Benchmarks

rynowak:

Issue #963, opened 2017-01-30
Bound Attributes are lowered to IR as html instead of expressions


rynowak:
Given this document:
```
@addTagHelper TestTagHelper, TestAssembly
<p foo="17">
```

And a matching taghelper where `foo` is bound to an `int`, this will appear in the IR as an HTML node, where we should be able to clearly know that it's C#.

rynowak:
When this is fixed, look to `ModelExpressionPass` in MVC for some cleanup.

ajaybhargavb:
7bd13061b5aee380a0acdbdc1d1a58e8d2da9445

PR #1198, opened 2017-04-10
Perf: Removed Content property from HtmlContentIRNode

ajaybhargavb:
Issue - #1196 
@rynowak @NTaylorMullen 

**Perf Data:**
**Before:**
image

**After:**
Generate separate `WriteLiteral` for each token (first and second commits of this PR) **Just FYI, we don't want this behavior**
image

Combine token contents using `StringBuilder` when generating code (third and fourth commits of this PR)

image

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
 
@@ -37,13 +38,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                         Content = "ModelExpressionProvider.CreateModelExpression(ViewData, __model => ",
                     });
 
-                    if (node.Children.Count == 1 && node.Children[0] is HtmlContentIRNode)
+                    if (node.Children.Count == 1 && node.Children[0] is HtmlContentIRNode original)
                     {
                         // A 'simple' expression will look like __model => __model.Foo
                         //
                         // Note that the fact we're looking for HTML here is based on a bug.
                         // https://github.com/aspnet/Razor/issues/963
-                        var original = ((HtmlContentIRNode)node.Children[0]);
 
                         builder.Add(new RazorIRToken()
                         {
@@ -51,10 +51,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                             Content = "__model."
                         });
 
+                        var content = GetContent(original);
                         builder.Add(new RazorIRToken()
                         {
                             Kind = RazorIRToken.TokenKind.CSharp,
-                            Content = original.Content,
+                            Content = content,
                             Source = original.Source,
                         });
                     }
@@ -62,13 +63,12 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                     {
                         for (var i = 0; i < node.Children.Count; i++)
                         {
-                            var nestedExpression = node.Children[i] as CSharpExpressionIRNode;
-                            if (nestedExpression != null)
+                            if (node.Children[i] is CSharpExpressionIRNode nestedExpression)
                             {
                                 for (var j = 0; j < nestedExpression.Children.Count; j++)
                                 {
-                                    var cSharpToken = nestedExpression.Children[j] as RazorIRToken;
-                                    if (cSharpToken != null && cSharpToken.Kind == RazorIRToken.TokenKind.CSharp)
+                                    if (nestedExpression.Children[j] is RazorIRToken cSharpToken &&
+                                        cSharpToken.IsCSharp)
                                     {
                                         builder.Add(cSharpToken);
                                     }
@@ -79,13 +79,13 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 
                             // Note that the fact we're looking for HTML here is based on a bug.
                             // https://github.com/aspnet/Razor/issues/963
-                            var html = node.Children[i] as HtmlContentIRNode;
-                            if (html != null)
+                            if (node.Children[i] is HtmlContentIRNode html)
                             {
+                                var content = GetContent(html);
                                 builder.Add(new RazorIRToken()
                                 {
                                     Kind = RazorIRToken.TokenKind.CSharp,
-                                    Content = html.Content,
+                                    Content = content,
                                     Source = html.Source,
                                 });
                             }
@@ -103,6 +103,20 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                     node.Children.Add(expression);
                     expression.Parent = node;
                 }
+            }
+
+            private string GetContent(HtmlContentIRNode node)
+            {
+                var builder = new StringBuilder();
+                for (var i = 0; i < node.Children.Count; i++)
+                {
+                    if (node.Children[i] is RazorIRToken token && token.IsHtml)
+                    {
+                        builder.Append(token.Content);
+                    }
+                }
+
+                return builder.ToString();
             }
         }
     }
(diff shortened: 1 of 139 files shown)


src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -260,13 +260,10 @@ public override void VisitSetTagHelperProperty(SetTagHelperPropertyIRNode node)
...
                 Context.Writer.WriteStartAssignment(propertyValueAccessor);
-                if (node.Children.Count == 1 && node.Children.First() is HtmlContentIRNode)
+                if (node.Children.Count == 1 && node.Children.First() is HtmlContentIRNode htmlNode)
                 {
-                    var htmlNode = node.Children.First() as HtmlContentIRNode;
-                    if (htmlNode != null)
-                    {
-                        Context.Writer.WriteStringLiteral(htmlNode.Content);
-                    }
+                    var content = GetContent(htmlNode);
```
ajaybhargavb: This and a few others should go away when https://github.com/aspnet/Razor/issues/963 is fixed.

benaadams:
Not sure am reading benchmarks right; for the curious is that 51 MB for a 314 KB output?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
```diff
@@ -104,6 +104,20 @@ public override void VisitSetTagHelperProperty(SetTagHelperPropertyIRNode node)
                     expression.Parent = node;
                 }
             }
+
+            private string GetContent(HtmlContentIRNode node)
+            {
+                var builder = new StringBuilder();
```
NTaylorMullen: Could re-use this guy if it was a field for more perf awesomeness

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/DesignTimeCSharpRenderer.cs
```diff
@@ -367,5 +355,19 @@ public override void VisitSetTagHelperProperty(SetTagHelperPropertyIRNode node)
                 Context.Diagnostics.Add(RazorDiagnostic.Create(error));
             }
         }
+
+        private string GetContent(HtmlContentIRNode node)
+        {
+            var builder = new StringBuilder();
```
NTaylorMullen: dittos

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RuntimeBasicWriter.cs
```diff
@@ -106,20 +107,31 @@ public override void WriteHtmlContent(CSharpRenderingContext context, HtmlConten
         {
             const int MaxStringLiteralLength = 1024;
 
+            var builder = new StringBuilder();
```
NTaylorMullen: You repeat all this logic a ton. A helper of some sort could be useful.

ajaybhargavb:
@benaadams I used the benchmark test from #1197. @rynowak will have a better idea what those allocation numbers actually represent. Also note, before this change the allocation number is 5.41 GB 😮

benaadams:
> Also note, before this change the allocation number is 5.41 GB

😱

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedRuntimeBasicWriter.cs
```diff
@@ -79,8 +91,8 @@ public override void WriteHtmlContent(CSharpRenderingContext context, HtmlConten
                     .WriteStringLiteral(textToRender)
                     .WriteEndMethodInvocation();
 
-                charactersConsumed += textToRender.Length;
+                    charactersConsumed += textToRender.Length;
```
rynowak: You can rewrite this code to be much more efficient by going direct to `context.Writer` and looking at the lengths of the strings before you write them.

Also `context.Writer` should expose a way to write a portion of a string using index+length

rynowak:
> Not sure am reading benchmarks right; for the curious is that 51 MB for a 314 KB output?

@benaadams - yeah we've got some work to do 👍

test/Microsoft.AspNetCore.Razor.Evolution.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ComplexTagHelpers_DesignTime.mappings.txt
```diff
@@ -119,10 +119,15 @@ Source Location: (711:20,39 [23] TestFiles/IntegrationTests/CodeGenerationIntegr
 Generated Location: (5883:125,38 [23] )
 |DateTimeOffset.Now.Year|
 
-Source Location: (734:20,62 [7] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ComplexTagHelpers.cshtml)
-| - 1970|
-Generated Location: (5906:125,61 [7] )
-| - 1970|
+Source Location: (734:20,62 [2] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ComplexTagHelpers.cshtml)
+| -|
```
NTaylorMullen: Note: Now that we render line mappings for every token in a TagHelper attribute we may get unexpected perf consequences in the editor. Their biggest perf costs are due to projections so it'd be worth following up with them.

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedRuntimeBasicWriter.cs
```diff
@@ -56,20 +57,31 @@ public override void WriteHtmlContent(CSharpRenderingContext context, HtmlConten
...
+            var builder = new StringBuilder();
+            for (var i = 0; i < node.Children.Count; i++)
+            {
+                if (node.Children[i] is RazorIRToken token && token.IsHtml)
+                {
+                    builder.Append(token.Content);
+                }
+            }
+
+            var content = builder.ToString();
```
Yves57: If I can do a suggestion, A solution can be to create only one `new StringBuilder(MaxStringLiteralLength)`, and fill it token by token until the limit is reached. Then a `StringBuilder.Clear()` will reset the StringBuilder without destroying its internal buffer.
It will avoid doing this huge string allocation. But it would make small string allocations necessary.

ajaybhargavb:
🆙 📅 Re-using the StringBuilder. Perf looks better,

image

src/Microsoft.AspNetCore.Razor.Evolution/CodeGeneration/RedirectedRuntimeBasicWriter.cs
ajaybhargavb: Filed a follow up issue https://github.com/aspnet/Razor/issues/1202

src/Microsoft.AspNetCore.Razor.Evolution/Intermediate/HtmlContentIRNodeExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Evolution.Intermediate
+{
+    internal static class HtmlContentIRNodeExtensions
+    {
+        private static StringBuilder _builder;
```
Yves57: Just for curiosity, there is no thread concurrency scenarios?

rynowak: Yeah, listen to this guy, this is not a safe change.

ajaybhargavb: Gonna revert this change and do this optimization as part of https://github.com/aspnet/Razor/issues/1202. I'll rename the issue title to something more appropriate.

benaadams: Can use this code https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/StringBuilderCache.cs

rynowak: We don't need a string builder here because the codewriter already has one.

PR #1204, opened 2017-04-10
Add support for naming a page

pranavkm:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs
```diff
@@ -1,50 +1,87 @@
...
+            }
+
+            pageDirective = new PageDirective(routeTemplate, pageName);
+            return true;
+        }
+
+        private static string TrimQuotes(string content)
+        {
+            Debug.Assert(content.StartsWith("\"", StringComparison.Ordinal));
+            Debug.Assert(content.EndsWith("\"", StringComparison.Ordinal));
```
ryanbrandenburg: I feel like these should be actual checks rather than just `Debug.Assert`s. Is there something up the chain that will guarantee our quotes for us?

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/PageDirectiveTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            var sourceDocument = RazorSourceDocument.Create(content, "file");
+            var codeDocument = RazorCodeDocument.Create(sourceDocument);
+            var engine = CreateEngine();
+            var irDocument = CreateIRDocument(engine, codeDocument);
+
+            // Act
+            var result = PageDirective.TryGetPageDirective(irDocument, out var pageDirective);
+
+            // Assert
+            Assert.False(result);
```
ryanbrandenburg: We should check that `pageDirective` is null too.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/PageDirectiveTest.cs
```diff
@@ -0,0 +1,127 @@
...
+            Assert.True(result);
+            Assert.Equal("some-route-template", pageDirective.RouteTemplate);
+            Assert.Null(pageDirective.PageName);
+        }
+
+        [Fact]
+        public void TryGetPageDirective_ParsesPageName()
+        {
+            // Arrange
+            var content = "@page \"some-route\" \"some name\"";
```
ryanbrandenburg: Can we get a test that `@page "some-route" non-razor garbage` doesn't give you "non-razor garbage" as pageDirective.PageName?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs
pranavkm: The parser requires quotes around string literals.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/PageDirectiveTest.cs
pranavkm: The parser requires quoted string literals. These would be handled by it accordingly.

pranavkm:
🆙 📅

Issue #1181, opened 2017-04-05
Add TagHelper completions API to TagHelperFactService.


NTaylorMullen:
Today we expose several APIs that control the binding of TagHelpers to a given HTML element based on the tag name and presence of attributes; we should own the entire completion API.

rynowak:
The title of this is jankified. This would be a new, completion-focused service.

NTaylorMullen:
https://github.com/aspnet/Razor/commit/a1cfd22a3273ed1b9aca8cb7a3639cb9edc8a32a

PR #1205, opened 2017-04-11
Add element completion API for TagHelpers.

NTaylorMullen:
- Added a new `TagHelperCompletionFactService` which can be queried for information on what completion information should be provided.
- Added tests to validate the completion service's expectations.
- Fixed an issue where `TagHelper`s with output hints that did not exist in a passed in completion would never be highlighted as `TagHelper`'s; even when their primary targeting element was already in the element completion list.

#1181

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
@@ -1,0 +1,151 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Export(typeof(TagHelperCompletionService))]
+    internal class DefaultTagHelperCompletionService : TagHelperCompletionService
+    {
+        private readonly TagHelperFactsService _tagHelperFactsService;
+        private static readonly HashSet<TagHelperDescriptor> _emptyHashSet = new HashSet<TagHelperDescriptor>();
+
+        [ImportingConstructor]
+        public DefaultTagHelperCompletionService(TagHelperFactsService tagHelperFactsService)
+        {
+            _tagHelperFactsService = tagHelperFactsService;
+        }
+
+        public override ElementCompletionResult GetElementCompletions(ElementCompletionContext completionContext)
+        {
+            if (completionContext == null)
+            {
+                throw new ArgumentNullException(nameof(completionContext));
+            }
+
+            var elementCompletions = new Dictionary<string, HashSet<TagHelperDescriptor>>(StringComparer.OrdinalIgnoreCase);
+
+            AddAllowedChildrenCompletions(completionContext, elementCompletions);
+
+            if (elementCompletions.Count > 0)
+            {
+                // If the containing element is already a TagHelper and only allows certain children.
+                var emptyResult = ElementCompletionResult.Create(elementCompletions);
+                return emptyResult;
+            }
+
+            elementCompletions = completionContext.ExistingCompletions.ToDictionary(
+                completion => completion,
+                _ => new HashSet<TagHelperDescriptor>(),
+                StringComparer.OrdinalIgnoreCase);
+
+            var possibleChildDescriptors = _tagHelperFactsService.GetTagHelpersGivenParent(completionContext.DocumentContext, completionContext.ContainingTagName);
+            foreach (var possibleDescriptor in possibleChildDescriptors)
+            {
+                var addRuleCompletions = false;
+                var outputHint = possibleDescriptor.TagOutputHint;
+
+                // Filter out catch-all rules because TagHelpers that target attributes only would light up every child tag otherwise. Force those TagHelpers
+                // to have additional requirements before showing them in the element completion list.
+                var nonCatchAllRules = possibleDescriptor.TagMatchingRules.Where(rule => rule.TagName != TagHelperMatchingConventions.ElementCatchAllName);
+                foreach (var rule in nonCatchAllRules)
+                {
+                    if (elementCompletions.ContainsKey(rule.TagName))
+                    {
+                        addRuleCompletions = true;
+                    }
+                    else if (outputHint != null && elementCompletions.ContainsKey(outputHint))
+                    {
+                        // If the possible descriptors final output tag already exists in our list of completions, we should add every representation
+                        // of that descriptor to the possible element completions.
+                        addRuleCompletions = true;
+                    }
+                    else if (!completionContext.InHTMLSchema(rule.TagName))
+                    {
+                        // If there is an unknown HTML schema tag that doesn't exist in the current completion we should add it. This happens for
+                        // TagHelpers that target non-schema oriented tags.
+                        addRuleCompletions = true;
+                    }
+
+                    if (addRuleCompletions)
+                    {
+                        if (!elementCompletions.TryGetValue(rule.TagName, out var existingRuleDescriptors))
+                        {
+                            existingRuleDescriptors = new HashSet<TagHelperDescriptor>();
+                            elementCompletions[rule.TagName] = existingRuleDescriptors;
+                        }
+
+                        existingRuleDescriptors.Add(possibleDescriptor);
+                    }
+                }
+            }
+
+            var result = ElementCompletionResult.Create(elementCompletions);
+            return result;
+        }
+
+        private void AddAllowedChildrenCompletions(
+            ElementCompletionContext completionContext,
+            Dictionary<string, HashSet<TagHelperDescriptor>> elementCompletions)
+        {
+            if (completionContext.ContainingTagName == null)
+            {
+                // If we're at the root then there's no containing TagHelper to specify allowed children.
+                return;
+            }
+
+            var prefix = completionContext.DocumentContext.Prefix ?? string.Empty;
+            var binding = _tagHelperFactsService.GetTagHelperBinding(
+                completionContext.DocumentContext,
+                completionContext.ContainingTagName,
+                completionContext.Attributes,
+                completionContext.ContainingParentTagName);
+
+            if (binding == null)
+            {
+                // Containing tag is not a TagHelper; therefore, it allows any children.
+                return;
+            }
+
+            foreach (var descriptor in binding.Descriptors)
+            {
+                if (descriptor.AllowedChildTags == null)
+                {
+                    continue;
+                }
+
+                foreach (var childTag in descriptor.AllowedChildTags)
+                {
+                    var prefixedName = string.Concat(prefix, childTag);
+                    var descriptors = _tagHelperFactsService.GetTagHelpersGivenTag(
+                        completionContext.DocumentContext,
+                        prefixedName,
+                        completionContext.ContainingTagName);
+
+                    if (descriptors.Count == 0)
+                    {
+                        if (!elementCompletions.ContainsKey(prefixedName))
+                        {
+                            elementCompletions[prefixedName] = _emptyHashSet;
+                        }
+
+                        continue;
+                    }
+
+                    if (!elementCompletions.TryGetValue(prefixedName, out var existingRuleDescriptors))
+                    {
+                        existingRuleDescriptors = new HashSet<TagHelperDescriptor>();
+                        elementCompletions[prefixedName] = existingRuleDescriptors;
+                    }
+
+                    existingRuleDescriptors.UnionWith(descriptors);
+                }
+            }
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperFactsService.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperFactsService.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNetCore.Razor.Language;
-using Microsoft.AspNetCore.Razor.Language.Legacy;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionContext.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionContext.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public sealed class ElementCompletionContext
+    {
+        public ElementCompletionContext(
+            TagHelperDocumentContext documentContext,
+            IEnumerable<string> existingCompletions,
+            string containingTagName,
+            IEnumerable<KeyValuePair<string, string>> attributes,
+            string containingParentTagName,
+            Func<string, bool> inHTMLSchema)
+        {
+            if (documentContext == null)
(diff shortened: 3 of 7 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionFactsService.cs
```diff
@@ -0,0 +1,155 @@
...
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Export(typeof(TagHelperCompletionFactsService))]
+    internal class DefaultTagHelperCompletionFactsService : TagHelperCompletionFactsService
```
rynowak: `TagHelperCompletionFactsService` I'd suggest just `TagHelperCompletionService`

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionFactsService.cs
```diff
@@ -0,0 +1,155 @@
...
+        [Import(typeof(TagHelperFactsService))]
+        public TagHelperFactsService TagHelperFactsService { get; set; }
+
+        public override ElementCompletionResult GetElementCompletions(
+            TagHelperDocumentContext documentContext,
+            IEnumerable<string> existingCompletions,
+            string containingTagName,
+            IEnumerable<KeyValuePair<string, string>> attributes,
+            string containingParentTagName,
+            Func<string, bool> inHTMLSchema)
```
rynowak: I think that all of this stuff needs to be rolled up into some kind of context object.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionFactsService.cs
```diff
@@ -0,0 +1,155 @@
...
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Export(typeof(TagHelperCompletionFactsService))]
+    internal class DefaultTagHelperCompletionFactsService : TagHelperCompletionFactsService
+    {
+        [Import(typeof(TagHelperFactsService))]
+        public TagHelperFactsService TagHelperFactsService { get; set; }
+
+        public override ElementCompletionResult GetElementCompletions(
```
rynowak: Have you run this by Todd/Alex yet?

rynowak: Would there be another method for attribute completions?

src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionResult.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public abstract class ElementCompletionResult : IEnumerable<KeyValuePair<string, IEnumerable<TagHelperDescriptor>>>
```
rynowak: eww don't do that.

rynowak: Is it possible to use something other `IEnumerable<>` here?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionFactsService.cs
NTaylorMullen: Yup, doing it in two check-ins. One for element completion, one for attribute completion.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionFactsService.cs
NTaylorMullen: Hmm, alright.

src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionResult.cs
NTaylorMullen: Could do IReadOnlyDictionary

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,153 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Export(typeof(TagHelperCompletionService))]
+    internal class DefaultTagHelperCompletionService : TagHelperCompletionService
+    {
+        [Import(typeof(TagHelperFactsService))]
+        public TagHelperFactsService TagHelperFactsService { get; set; }
```
rynowak: Use `ImportingConstructor` here - we don't really need to use property injection, and that will be more natural for tests.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperFactsService.cs
```diff
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.AspNetCore.Razor.Language;
-using Microsoft.AspNetCore.Razor.Language.Legacy;
 using System;
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
```
rynowak: What are we using here from Legacy? - Can we fix that?

src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionContext.cs
```diff
@@ -0,0 +1,64 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public sealed class ElementCompletionContext
+    {
+        private ElementCompletionContext()
```
rynowak: It's ok to make this constructable if it's sealed.

src/Microsoft.VisualStudio.LanguageServices.Razor/ElementCompletionResult.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public abstract class ElementCompletionResult
+    {
+        private ElementCompletionResult()
+        {
+        }
+
+        public abstract IReadOnlyDictionary<string, IEnumerable<TagHelperDescriptor>> Completions { get; }
+
+        public static ElementCompletionResult Create(IReadOnlyDictionary<string, IEnumerable<TagHelperDescriptor>> completions)
```
rynowak: Why does this need to be public?

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Xunit;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public class DefaultTagHelperCompletionFactsServiceTest
+    {
+        [Fact]
+        public void GetElementCompletions_AllowsMultiTargetingTagHelpers()
```
rynowak: I'm checking these tests for sanity, but mostly I'm trusting you.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+        {
+            // Arrange
+            var documentDescriptors = new[]
+            {
+                ITagHelperDescriptorBuilder.Create("BoldTagHelper1", "TestAssembly")
```
rynowak: Ugh, I still **hate** that this is `ITagHelperDescriptorBuilder`

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+                    .TagMatchingRule(rule => rule.RequireTagName("strong"))
+                    .TagMatchingRule(rule => rule.RequireTagName("b"))
+                    .TagMatchingRule(rule => rule.RequireTagName("bold"))
```
rynowak: Should we have an extension method that does this? `TagMatchingRule(string)`

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Xunit;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public class DefaultTagHelperCompletionFactsServiceTest
```
rynowak: Rename this to match

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+                    .Build(),
+            };
+            var documentContext = TagHelperDocumentContext.Create(string.Empty, documentDescriptors);
+            var service = CreateTagHelperCompletionFactsService();
+            var expectedCompletions = ElementCompletionResult.Create(new Dictionary<string, IEnumerable<TagHelperDescriptor>>()
+            {
+                ["strong"] = new[] { documentDescriptors[0], documentDescriptors[1] },
+                ["b"] = new[] { documentDescriptors[0] },
+                ["bold"] = new[] { documentDescriptors[0] },
+            });
```
rynowak: It would be really helpful if you could pull this guy out with some whitespace to make it easier to read. All of this code needs to breathe.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,407 @@
...
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public class DefaultTagHelperCompletionFactsServiceTest
+    {
+        [Fact]
+        public void GetElementCompletions_AllowsMultiTargetingTagHelpers()
+        {
+            // Arrange
+            var documentDescriptors = new[]
```
rynowak: Think about just calling this variable `tagHelpers`

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,153 @@
...
+                    {
+                        addRuleCompletions = true;
+                    }
+                    else if (outputHint != null && elementCompletions.ContainsKey(outputHint))
+                    {
+                        // If the possible descriptors final output tag already exists in our list of completions, we should add every representation
+                        // of that descriptor to the possible element completions.
+                        addRuleCompletions = true;
+                    }
+                    else if (!completionContext.InHTMLSchema(rule.TagName))
```
rynowak: Do we really need to know about the HTML schema? Would it be better if we didn't do this filtering and they applied it after?

I don't have a big problem with this I just wanted to make sure we're not taking too big of a dependency on WTE.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
NTaylorMullen: VS lied to me. It said it refactored! 😄

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperFactsService.cs
NTaylorMullen: The descriptor provider/binding are from legacy. We already have an issue to move the binding out of legacy and just filed one for the descriptor provider: https://github.com/aspnet/Razor/issues/1209

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
NTaylorMullen: Maybe eventually if we see users using these API's more directly.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
NTaylorMullen: https://github.com/aspnet/Razor/issues/1210

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionFactsServiceTest.cs
NTaylorMullen: nah, don't like conflating tag helpers with their descriptors

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
NTaylorMullen: Sadly we do. If we didn't then after-the-fact the editor would have to re-compute each completions descriptors OutputElementHint and cross reference that with its bound rules and the schema table to determine if it should exist in the completions collection.

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,158 @@
...
+                throw new ArgumentNullException(nameof(completionContext));
+            }
+
+            var elementCompletions = new Dictionary<string, IEnumerable<TagHelperDescriptor>>(StringComparer.OrdinalIgnoreCase);
+
+            AddAllowedChildrenCompletions(completionContext, elementCompletions);
+
+            if (elementCompletions.Count > 0)
+            {
+                // If there are allowed children, those are the only possible completions.
```
rynowak: I would say something like "if this element is already a taghelper, and that taghelper only allows certain children....."

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
rynowak: OK, then let's do this for now 👍

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,158 @@
...
+                        // TagHelpers that target non-schema oriented tags.
+                        addRuleCompletions = true;
+                    }
+
+                    if (addRuleCompletions)
+                    {
+                        if (elementCompletions.TryGetValue(rule.TagName, out var existingRuleDescriptors))
+                        {
+                            var combinedDescriptors = new HashSet<TagHelperDescriptor>(existingRuleDescriptors);
+                            combinedDescriptors.Add(possibleDescriptor);
```
rynowak: This is a really expensive way to check `Contains`.

This would be way simpler to build this as `Dictionary<string, List<TagHelperDescriptor>>` or `Dictionary<string, HashSet<TagHelperDescriptor>>`.

PR #1206, opened 2017-04-11
Fix #1194 - DirectiveIRNode doesn't have source

### LINKED ISSUE #1194: DirectiveIRNode doesn't have a source
Right now the source property on directives isn't set during lowering.

This is causing some difficulties writing `@namespace`.

rynowak:
Now with line mappings!

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
@@ -53,7 +53,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                     {
                         TypeName = typeName,
                         MemberName = memberName,
-                        Source = directive.Source,
                         Parent = visitor.Class,
                     };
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/Output/DesignTime/Inject.cs
```diff
@@ -34,8 +34,12 @@ public class _TestFiles_Input_Inject_cshtml : global::Microsoft.AspNetCore.Mvc.R
         {
         }
         #pragma warning restore 1998
-        [global::Microsoft.AspNetCore.Mvc.Razor.Internal.RazorInjectAttribute]
-        public MyApp MyPropertyName { get; private set; }
+#line 2 "/TestFiles/Input/Inject.cshtml"
```
rynowak: These started showing up now that we know where they came from. I think some code we wrote in `@inject` copies these locations over to the new node.

NTaylorMullen: Curious as to what this would do to the editor. Aka there'll be a visible line mapping for the type/property name but not for the get/setter which is what the debugger would try to highlight.

rynowak: Yeah maybe we should remove these from `@inject`

rynowak:
Updated this. We chose to delete the location from `@inject`. Now there's no diff in the baselines.

PR #1217, opened 2017-04-13
Set an order on @inject to run behind @namespace

rynowak:
The problem is that @inject uses the type name, and @namespace can change
it. Setting an explicit order makes sure that these things happen in a
sensible sequence.


rynowak:
c0e3519bc3ddbd58a7226c6f9c14cb58558fb93f

Issue #1224, opened 2017-04-14
TagHelperElementCompletionService does not return TagHelper completions with a prefix.


NTaylorMullen:

NTaylorMullen:
https://github.com/aspnet/Razor/commit/cf44f103c11a7bf01cc37d1f2cd7a1263981dc0b

PR #1226, opened 2017-04-14
Prefix TagHelpers with tag helper prefix in completion list.

NTaylorMullen:
- We don't want tooling applying tag helper prefix to what we deem `TagHelper`s. Ultimately we should consume all the necessary data and do that work.
- Added two tests to validate that we're applying tag helper prefix to `TagHelper` element completions for catch-all's and non-catch-alls.

#1224


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -87,14 +88,16 @@ public override ElementCompletionResult GetElementCompletions(ElementCompletionC
             {
                 foreach (var completionTagNames in elementCompletions.Keys)
                 {
-                    UpdateCompletions(completionTagNames, catchAllDescriptor);
+                    if (completionTagNames.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
```
ajaybhargavb: completionTagNames => completionTagName

Issue #1231, opened 2017-04-14
Bring back ViewComponentTagHelpers for new Razor


NTaylorMullen:
We removed this for the initial OOB VSIX release because we didn't have code gen that supported it. Now that we do, we need to re-add the logic.

NTaylorMullen:
385e2523a9242313e3ab26843d2b39350a495c67

PR #1232, opened 2017-04-14
Re-enable ViewComponentTagHelpers for tooling.

NTaylorMullen:
#1231


rynowak:
Will this cause compile errors on a 1.0.0 site?

NTaylorMullen:
> Will this cause compile errors on a 1.0.0 site?

Shouldn't since we dynamically create the TagHelpers on the page the same way we always have.

PR #1235, opened 2017-04-18
Moved HtmlAttributeIRNode and children from renderer to writer

ajaybhargavb:
@NTaylorMullen @rynowak 

Added an internal abstract `HtmlAttributeWriter`. Will add tests once design is approved.


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/HtmlAttributeWriter.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
+{
+    internal abstract class HtmlAttributeWriter
```
NTaylorMullen: Any reason the `WriteHtmlAttribute` shouldn't also be on here?

ajaybhargavb: This is an internal class. Don't want to expose a way for people to override the HtmlAttribute code generation?

ajaybhargavb: Actually I think we could rename this to `HtmlAttributeValueWriter`.

ajaybhargavb:
:up: 📅 
- Removed RenderingConventions
- Added Redirect writers
- Updated an integration test to verify redirect scenario for Html attributes.
- Cleanup

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/HtmlAttributeWriter.cs
rynowak: All of this should be on BasicWriter

ajaybhargavb:
🆙 📅 No longer in design.

- Addressed feedback
- Added tests

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/TagHelperHtmlAttributeRuntimeBasicWriterTest.cs
```diff
@@ -0,0 +1,135 @@
...
+@"AddHtmlAttributeValue("" "", 27, new Microsoft.AspNetCore.Mvc.Razor.HelperResult(async(__razor_attribute_value_writer) => {
+    Render Children
+}
+), 28, 13, false);
+",
+                csharp,
+                ignoreLineEndingDifferences: true);
+        }
+
+        private static CSharpRenderingContext GetCSharpRenderingContext(BasicWriter writer)
```
NTaylorMullen: Any possibility to make a test base class to hold all these infrastructure bits? It looks like they're duplicated quite a bit

ajaybhargavb: I don't think it's going to add a lot of value. `GetCSharpRenderingContext` does different things for BasicWriter and TagHelperWriter. Only `Lower` can be shared. Is it worth it? I don't know........

NTaylorMullen: Ok fair enough.

Issue #1237, opened 2017-04-18
Generated class for Razor Pages shows up in VS Solution Explorer when the CSHTML file is open in the editor


NTaylorMullen:
_From @DamianEdwards on March 13, 2017 20:28_

![image](https://cloud.githubusercontent.com/assets/249088/23873906/f624e348-07f0-11e7-8a02-a712a8b76692.png)

_Copied from original issue: aspnet/Mvc#5957_

NTaylorMullen:
_From @DamianEdwards on March 14, 2017 4:43_

@rynowak @Eilon can you assign this one pls?

NTaylorMullen:
_From @rynowak on March 21, 2017 16:19_

Should we just slap an editorbrowsable.never on this?

NTaylorMullen:
_From @DamianEdwards on March 21, 2017 16:26_

Worth trying. Is that how generated files are usually suppressed?

NTaylorMullen:
_From @rynowak on April 4, 2017 1:33_

@DamianEdwards does this become a **good thing** if the names this generates aren't horrible?

NTaylorMullen:
_From @Eilon on April 4, 2017 17:13_

Should the user even know that there's this magical class that's generated? I didn't think the concern was how pretty it was.

BTW perhaps https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.compilergeneratedattribute(v=vs.110).aspx would make more sense as the attribute to use? (Assuming it does what we want.)

NTaylorMullen:
_From @JosefPihrt on April 13, 2017 16:42_

When implementing this, please make sure that generated code will be marked so it can be skipped by roslyn analyzers.

Another alternative to `CompilerGeneratedAttribute` would be to create partial class with generated content and mark it with `<auto-generated>` tag.

NTaylorMullen:
_From @rynowak on April 18, 2017 18:22_

Load Balancing to @NTaylorMullen - let's try `[CompilerGenerated]`

NTaylorMullen:
Blocked on: https://github.com/dotnet/roslyn/issues/18881

PR #1241, opened 2017-04-18
Add CompilerGeneratedAttribute to gen'd code.

NTaylorMullen:
- This will stop Visual Studio and other IDEs from manifesting code generated classes when they don't need to be.
- Re-generated test file baselines.

#1237

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
 
 namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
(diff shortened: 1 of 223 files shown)


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -96,6 +97,12 @@ public override void VisitNamespace(NamespaceDeclarationIRNode node)
 
             public override void VisitClass(ClassDeclarationIRNode node)
             {
+                // Mark generated classes with compiler generated attributes so IDEs don't show generaetd classes when unnecessary.
+                Context.Writer
+                    .Write("[global::")
```
ajaybhargavb: nit: IMO `WriteLine($"[global::{typeof(CompilerGeneratedAttribute).FullName}()]")` looks cleaner.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -96,6 +97,12 @@ public override void VisitNamespace(NamespaceDeclarationIRNode node)
...
+                    .Write(typeof(CompilerGeneratedAttribute).FullName)
+                    .WriteLine("()]");
```
ajaybhargavb: Do you need the `()`?

NTaylorMullen: Shouldn't 👍

rynowak:
Does it work?

NTaylorMullen:
> Does it work?

Thought it was investigated as part of the bug. Will do the work to verify all is good in actuality 👍

NTaylorMullen:
Will revive this guy when us and the Roslyn folks come to an agreement on how to hide compiler generated classes from the solution explorer 😄

Issue #1242, opened 2017-04-19
Trying to write addTagHelper with quotes crashes Visual Studio


NTaylorMullen:
If you try and write this out character by character: `@addTagHelper "*, Microsoft.AspNetCore.Mvc.TagHelpers"` you wont get to the end of the line, VS will crash.

Thissss is pretty bad 😄 

/cc @alexgav we should probably add an end-to-end VS test for this.

NTaylorMullen:
This is on the runtime side. We have a faulty check in our TagHelper binder: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/TagHelperBinderSyntaxTreePass.cs#L351

NTaylorMullen:
f31b45d84a23f52fc564b4892327e5e3e981d66a

PR #1246, opened 2017-04-20
Enable TagHelper directives to handle malformed text.

NTaylorMullen:
- Prior to this change we'd try to substring a TagHelper directive with length 1 but our substring call would be for -1 (explosions).
- Added tests to validate that `@tagHelperPrefix`, `@removeTagHelper` and `@addTagHelper` all behave properly when they have malformed quotes.

#1242

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/TagHelperBinderSyntaxTreePass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/TagHelperBinderSyntaxTreePass.cs
@@ -345,7 +345,8 @@ namespace Microsoft.AspNetCore.Razor.Language
                 TagHelperDirectiveType directiveType)
             {
                 directiveText = directiveText.Trim();
-                if (directiveText.StartsWith("\"", StringComparison.Ordinal) &&
+                if (directiveText.Length >= 2 &&
+                    directiveText.StartsWith("\"", StringComparison.Ordinal) &&
                     directiveText.EndsWith("\"", StringComparison.Ordinal))
                 {
                     directiveText = directiveText.Substring(1, directiveText.Length - 2);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/TagHelperBinderSyntaxTreePass.cs
```diff
@@ -345,7 +345,8 @@ public override void VisitTagHelperPrefixDirectiveSpan(TagHelperPrefixDirectiveC
                 TagHelperDirectiveType directiveType)
             {
                 directiveText = directiveText.Trim();
-                if (directiveText.StartsWith("\"", StringComparison.Ordinal) &&
+                if (directiveText.Length >= 2 &&
```
rynowak: Did you take a look at our other directive parsing? It would be good to do a pass anywhere else we look at directive tokens since the parser doesn't validate them super hard.

test/Microsoft.AspNetCore.Razor.Language.Test/TagHelperBinderSyntaxTreePassTest.cs
```diff
@@ -14,6 +14,132 @@ namespace Microsoft.AspNetCore.Razor.Language
     public class TagHelperBinderSyntaxTreePassTest
     {
         [Fact]
+        public void Execute_CanHandleSingleLengthAddTagHelperDirective()
```
ajaybhargavb: Could just have one test for this since they all go through the same path (unless I am missing something). Having three tests feels a bit redundant.

NTaylorMullen: Eh, don't want the tests to be tied directly to the implementation. 3 different features, they just happen to all go through the same path today.

src/Microsoft.AspNetCore.Razor.Language/TagHelperBinderSyntaxTreePass.cs
NTaylorMullen: Will look through our code base to see if anything jumps out. Merging this for now.

NTaylorMullen: Found an issue with `@page` but nothing else stuck out as a potential problem. Lot of code to go through so very good chance I may have missed something. Nothing blatantly obvious though.

Issue #1247, opened 2017-04-20
Trying to write page with quotes sometimes crashes Visual Studio


NTaylorMullen:
If you try and write this out character by character: `@page "foo"` you sometimes wont get to the end of the line, VS will crash. Reason why this is inconsistent is that Roslyn will attempt to auto-complete your quotes so if our Razor parse gets canceled and then re-queued with `""` being in the document thing work; otherwise, if a parse goes through with `@page "` VS crashes because of [this guy](https://github.com/aspnet/Razor/blob/f31b45d84a23f52fc564b4892327e5e3e981d66a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs#L73).

@alexgav can we add an end-to-end VS test for this 👍.

NTaylorMullen:
c25aadf599b4de61879ffc7f40cded23a36bc9e7

PR #1249, opened 2017-04-20
Enable page directive to handle malformed text.

NTaylorMullen:
- Found that our extensible directive string parsing system wasn't consistent with the rest of the extensible directive tokens. Basically, if there were malformed string tokens we'd consume them and pass them along to extensible directive passes. This was a big no-no because it means extensible directive passes weren't able to rely on tokens being passed to them being well-formed.
- Fixed up existing extensible directive tests that relied on output of string tokens.
- Added an integration test to validate malformed page tags can make their way through the entire Razor pipeline and don't destroy the rest of the content.

#1247

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -1614,18 +1614,17 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                         break;
 
                     case DirectiveTokenKind.String:
-                        if (At(CSharpSymbolType.StringLiteral))
+                        if (At(CSharpSymbolType.StringLiteral) && CurrentSymbol.Errors.Count == 0)
                         {
                             AcceptAndMoveNext();
                         }
                         else
                         {
-                            var startLocation = CurrentStart;
-                            AcceptUntil(CSharpSymbolType.WhiteSpace, CSharpSymbolType.NewLine);
                             Context.ErrorSink.OnError(
-                                startLocation,
+                                CurrentStart,
                                 LegacyResources.FormatDirectiveExpectsQuotedStringLiteral(descriptor.Name),
-                                Span.End.AbsoluteIndex - Span.Start.AbsoluteIndex);
+                                CurrentSymbol.Content.Length);
+                            return;
                         }
                         break;
                 }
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1614,18 +1614,17 @@ private void HandleDirective(DirectiveDescriptor descriptor)
                         break;
 
                     case DirectiveTokenKind.String:
-                        if (At(CSharpSymbolType.StringLiteral))
+                        if (At(CSharpSymbolType.StringLiteral) && CurrentSymbol.Errors.Count == 0)
```
NTaylorMullen: The parser would classify incomplete string literals `"aksfj` as string literals with errors. Added this error check to ensure we drop down into our own error case and not consume string literals that aren't valid.

rynowak:
This is more change than I want us to make for preview-1. Can we patch the issue with `@page` and follow up on this separately?

NTaylorMullen:
Got in a targeted variant of this [here](c25aadf599b4de61879ffc7f40cded23a36bc9e7). Will wait till dev is no longer locked down to get this real fix in.

NTaylorMullen:
🆙 📅 this guy to reflect new rebase changes.

NTaylorMullen:
Ping, this guy good now?

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
rynowak: Love it

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/MalformedPageDirective_DesignTime.codegen.cs
```diff
@@ -13,10 +13,6 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_MalformedP
     {
         #pragma warning disable 219
         private void __RazorDirectiveTokenHelpers__() {
-        ((System.Action)(() => {
-global::System.Object __typeHelper = "foo;
```
rynowak: What leads to this being omitted? I agree that it's good, but I want to understand why.

NTaylorMullen: The directive token is no longer valid (we're validating now) so it's not output into the `DirectiveBlock`. Instead the token is treated as markup in the rest of the page. With this approach directive authors never see invalid tokens.

PR #1250, opened 2017-04-20
Enable page directive to handle malformed text.

NTaylorMullen:
- Made a targeted fix to PageDirective that stops it from crashing Visual Studio when its malformed.
- Added unit test to validate PageDirective didn't throw when attempting to determine if something was a page.
- Added an integration test to validate malformed page tags can make their way through the entire Razor pipeline and don't destroy the rest of the content.

#1247


NTaylorMullen:
Verified this worked after patching VS.

PR #1258, opened 2017-04-25
Update our Roslyn dependencies to 15.3

rynowak:


NTaylorMullen:
Does this get us the ServiceHub fixes? Aka, docs and new ctor?

rynowak:
That comes from VS. These binaries are just what we compile against.

PR #1265, opened 2017-04-27
Fail if baseline longer than generated

ryanbrandenburg:
No tests for this since it's in test code, but I tried it out locally by adding lines to our baselines and it caught it.


test/Microsoft.AspNetCore.Razor.Test.Common/Langauge/IntegrationTests/RazorIRNodeVerifier.cs
```diff
@@ -37,6 +38,12 @@ public Walker(string[] baseline)
 
             public TextWriter Writer { get; }
 
+            public void AssertReachedEndOfBaseline()
```
NTaylorMullen: Move this method below the Visit methods

PR #1273, opened 2017-04-29
[Design] Add a new THProvider api

rynowak:
This change adds an API for Tag Helper discovery.

I also got rid of the 'design time' flag for the provider as an
experimental change. We need to think through the consequences of this
before committing to it. Right now I've left those tests failing until we
can make a decision.

This change decouples VCTH discovery a bit more, but we're still not ready
to move that into a the MVC extensions assembly. For that we need the
ability to discover the MVC extensibility.


src/Microsoft.AspNetCore.Razor.Language/ITagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public interface ITagHelperDescriptorProvider : IRazorEngineFeature
```
rynowak: This is the new API. You can have many providers and they are ordered. Because there's a context object with a 'results' collection later providers can modify the results of earlier ones.

rynowak: /cc @sebastienros

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
```diff
@@ -38,6 +50,23 @@ public static TagHelperDescriptorBuilder Create(string typeName, string assembly
...
+        /// <example>
+        /// SomeThing => some-thing
+        /// capsONInside => caps-on-inside
+        /// CAPSOnOUTSIDE => caps-on-outside
+        /// ALLCAPS => allcaps
+        /// One1Two2Three3 => one1-two2-three3
+        /// ONE1TWO2THREE3 => one1two2three3
+        /// First_Second_ThirdHi => first_second_third-hi
+        /// </example>
+        public static string ToHtmlCase(string name)
```
rynowak: This is a better home for this API

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -44,9 +31,8 @@ internal class DefaultTagHelperDescriptorFactory
                 .WithGlobalNamespaceStyle(SymbolDisplayGlobalNamespaceStyle.Omitted)
                 .WithMiscellaneousOptions(SymbolDisplayFormat.FullyQualifiedFormat.MiscellaneousOptions & (~SymbolDisplayMiscellaneousOptions.UseSpecialTypes));
 
-        public DefaultTagHelperDescriptorFactory(Compilation compilation, bool designTime)
+        public DefaultTagHelperDescriptorFactory(Compilation compilation)
```
rynowak: Come at me bro 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -20,68 +19,42 @@ public override TagHelperResolutionResult GetTagHelpers(Compilation compilation,
...
+                new DefaultTagHelperDescriptorProvider(),
+                new ViewComponentTagHelperDescriptorProvider(),
+            };
 
-            var factory = new DefaultTagHelperDescriptorFactory(compilation, DesignTime);
+            var results = new List<TagHelperDescriptor>();
+            var context = new Context(results);
 
-            foreach (var type in types)
+            for (var i = 0; i < providers.Length; i++)
```
NTaylorMullen: Doesn't look like you're ordering the providers based on their `Order`. I get why, cause this is only used by tooling but I think consistency here (with the feature) will limit our exposure to bugs. Might make sense to have a provider executor that's used in both locations.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -20,68 +19,42 @@ public override TagHelperResolutionResult GetTagHelpers(Compilation compilation,
...
-
-            VisitCompilation(visitor, compilation);
+            var providers = new ITagHelperDescriptorProvider[]
+            {
+                new DefaultTagHelperDescriptorProvider(),
+                new ViewComponentTagHelperDescriptorProvider(),
+            };
 
-            var factory = new DefaultTagHelperDescriptorFactory(compilation, DesignTime);
+            var results = new List<TagHelperDescriptor>();
```
NTaylorMullen: Can just use `Context.Results` and not pass this into the ctor

src/Microsoft.CodeAnalysis.Razor/CompilationTagHelperFeature.cs
```diff
@@ -0,0 +1,69 @@
...
+                OnInitialized();
+            }
+        }
+
+        public IReadOnlyList<TagHelperDescriptor> GetDescriptors()
+        {
+            var results = new List<TagHelperDescriptor>();
+
+            var context = new Context(results);
+            var compilation = CSharpCompilation.Create("__TagHelpers", references: _referenceFeature.References);
```
NTaylorMullen: It might make sense to eventually have this as a feature. That way we can create a compilation once and re-use it across features if the need arises. No action now, just a shower thought 😄

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
```diff
@@ -20,68 +19,42 @@ public override TagHelperResolutionResult GetTagHelpers(Compilation compilation,
...
             }
+            
+            var diagnostics = new List<RazorDiagnostic>();
+            var resolutionResult = new TagHelperResolutionResult(results, diagnostics);
+
+            return resolutionResult;
         }
 
-        private void VisitViewComponents(Compilation compilation, IEnumerable<string> assemblyNameFilters, List<TagHelperDescriptor> results)
+        private class Context : TagHelperDescriptorProviderContext
```
NTaylorMullen: Even though this bit is purely for tooling it might make sense to have a `TagHelperDescriptorProvierContext.CreateDefault` since we have a duplication in the TagHelperFeature. I prefer that approach because it limits potential bugs across runtime/tooling experiences.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
NTaylorMullen: Will need to react to this in our layer that returns `TagHelperDescriptor`s to tooling OR will need to make a WTE change. I'd prefer the first.

NTaylorMullen: Also, do all tests pass? I'd think this would break a lot more tests.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
```diff
@@ -429,41 +393,34 @@ private IEnumerable<IPropertySymbol> GetAccessibleProperties(INamedTypeSymbol ty
...
+            var editorBrowsableAttribute = symbol.GetAttributes().Where(a => a.AttributeClass == _editorBrowsableAttributeSymbol).FirstOrDefault();
+            if (editorBrowsableAttribute == null)
+            {
+                return;
+            }
+
+            if (editorBrowsableAttribute.ConstructorArguments.Length > 0 &&
+                (EditorBrowsableState)editorBrowsableAttribute.ConstructorArguments[0].Value == EditorBrowsableState.Never)
+            {
+                builder.AddMetadata("EditorBrowsable", "Never");
```
NTaylorMullen: How do you feel about just adding the visibility outright if EditorBrowsable is specified? Aka, if advanced was specified set this to advanced.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+                    {
+                        visitor.Visit(assembly.GlobalNamespace);
+                    }
+                }
+            }
+
+            var factory = new DefaultTagHelperDescriptorFactory(compilation);
+            for (var i = 0; i < types.Count; i++)
+            {
+                context.Results.Add(factory.CreateDescriptor(types[i]));
```
NTaylorMullen: Can you break apart for debuggability (I debug through this bit a lot)? Aka `var descriptor = factory.CreateDescriptor(...)` on previous line

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+
+        private bool IsTagHelperAssembly(IAssemblySymbol assembly)
```
NTaylorMullen: static

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public sealed class ViewComponentTagHelperDescriptorProvider : ITagHelperDescriptorProvider
+    {
+        private RazorEngine _engine;
+
+        public int Order { get; set; }
+
+        public RazorEngine Engine
```
NTaylorMullen: Making me want a base type for descriptor provider 😉

src/Microsoft.CodeAnalysis.Razor/ViewComponentTypes.cs
```diff
@@ -21,6 +21,8 @@ internal static class ViewComponentTypes
 
         public const string Task = "System.Threading.Tasks.Task";
 
+        public const string IDictionary = "System.Collections.Generic.IDictionary`2";
```
NTaylorMullen: Why put this here and not leave on the `TagHelperTypes`? The `IDictionary` support comes from the TagHelper system. The ViewCOmponentTagHelper system just detects that something is an idictionary.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
NTaylorMullen: Also, it might make sense to add this API to the descriptor builder.

src/Microsoft.AspNetCore.Razor.Language/ITagHelperDescriptorProvider.cs
rynowak: You mean the one in `Legacy`?

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
rynowak: This seems like something that should be public though

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
rynowak: I think I know what I want to do with this now.

I want all of the providers to just do what they do regardless of runtime or design time. I want the tooling-specific API to just remove the ones that we're ignoring.

It's important that we don't break compat with this change 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
rynowak: This *is* `context.Results`

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
rynowak: Yeaht that needs to happen

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
rynowak: yeah I'll just pull this out into its own top level class

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorFactory.cs
rynowak: Yeah I wasn't sure if this should be first class on the API or just a metadata, metadata feels good.

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
rynowak: ok

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
rynowak: why?

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
rynowak: Yeah I kinda want a public base class for features

src/Microsoft.CodeAnalysis.Razor/ViewComponentTypes.cs
rynowak: I originally moved this functionality to the MVC assembly, and I had to do this when I did that. Personally this is one of those 'just do something sensible' cases. The full name of `IDictionary` is not going to change.

src/Microsoft.AspNetCore.Razor.Language/ITagHelperDescriptorProvider.cs
rynowak: Yeah we'll need to delete the one from `.Legacy`. I went ahead and did that. We'll need to sync this change with tooling because of this.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ViewComponentTagHelperPass.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Microsoft.CodeAnalysis.Razor;
```
rynowak: Nope, the VCTH conventions are in this namespace now.

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
NTaylorMullen: `TagHelperDescriptorConventions.ToHtmlCase` as public.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperResolver.cs
rynowak: The ordering is in `OnInitialized`.

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorFactory.cs
```diff
@@ -136,13 +136,17 @@ private ImmutableArray<IParameterSymbol> GetInvokeMethodParameters(INamedTypeSym
...
+                    componentType.ToDisplayString(FullNameTypeDisplayFormat)));
             }
             else if (methods.Length > 1)
             {
-                throw new InvalidOperationException(
-                    ViewComponentResources.FormatViewComponent_AmbiguousMethods(componentType.ToDisplayString(FullNameTypeDisplayFormat), ViewComponentTypes.AsyncMethodName, ViewComponentTypes.SyncMethodName));
+                throw new InvalidOperationException(ViewComponentResources.FormatViewComponent_AmbiguousMethods(
+                    componentType.ToDisplayString(FullNameTypeDisplayFormat),
+                    ViewComponentTypes.AsyncMethodName,
+                    ViewComponentTypes.SyncMethodName));
```
rynowak: @NTaylorMullen - do you know what we throw an exception here? This doesn't seem right. This is currently failing some tests but I wanted to understand why this isn't just a diagnostic.

NTaylorMullen: This guy is here to replicate MVC's ViewComponentFactory discovery since we need to provide somewhat understandable errors at design time. Aka, if you have an invalid ViewComponent you'd want to know why you weren't getting TagHelper intellisense for it.

rynowak: I removed the idea of **global errors** for taghelper discovery. So this exception is now just an exception. It sounds to me like this should be a diagnostic on the TagHelper.

NTaylorMullen: I agree

rynowak:
@NTaylorMullen updated

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
NTaylorMullen: Ping

src/Microsoft.CodeAnalysis.Razor/CompilationTagHelperFeature.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class CompilationTagHelperFeature : ITagHelperFeature
```
NTaylorMullen: RazorEngineFeatureBase

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public sealed class ViewComponentTagHelperDescriptorProvider : RazorEngineFeatureBase, ITagHelperDescriptorProvider
```
NTaylorMullen: This guy can exist in Mvc.Extensions can't he?

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            var compilation = context.GetCompilation();
+            if (compilation == null)
+            {
+                // No compilation, nothing to do.
+                return;
+            }
+
+            var types = new List<INamedTypeSymbol>();
+            var visitor = TagHelperTypeVisitor.Create(compilation, types);
```
NTaylorMullen: Shouldn't this be `ViewComponentTypeVisitor`? Test explosions?

rynowak: Yeah wierd. I wonder if we don't have tests for this...

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
rynowak: No not yet. We need to do multitargeting first. This has to be hardcoded, and I don't want to introduce an assembly dependency that we will undo.

src/Microsoft.CodeAnalysis.Razor/CompilationTagHelperFeature.cs
rynowak: 👍

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
rynowak: I don't want to make `TagHelperDescriptorConventions` public

rynowak: The implementation can live there, but this needs to be a public API somewhere

src/Microsoft.CodeAnalysis.Razor/ViewComponentTagHelperDescriptorProvider.cs
rynowak: SO the answer was that we don't have tests for this. I'm adding one basic test

rynowak:
@NTaylorMullen - updated, I added a test

rynowak:
I'd really not like to cram any more refactoring into this PR given the state of tooling 😑. Doing anything here that's really going to clean things up is going to require the multitargeting feature, which we can get started on now that we've branched.

rynowak:
> Really dislike having ToHtmlCase on the TagHelperDescriptorBuilder but i'll deal.

Let's discuss this tomorrow, maybe we can find a solution we bothlike

PR #1275, opened 2017-05-01
Limit memory allocation when translating 'RazorIRToken' of type HTMLRazorIR to C# code.

Yves57:
Necessary because some files may have large HTML parts.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
@@ -245,7 +245,6 @@ namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 
             var insideWrite = false;
             var writtenLength = 0;
-            CSharpCodeWriter.StringLiteralWriter stringLiteralWriter = default(CSharpCodeWriter.StringLiteralWriter);
 
             for (var i = 0; i < node.Children.Count; i++)
             {
@@ -258,19 +257,19 @@ namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
                         if (!insideWrite)
                         {
                             WriteHtmlContentStartMethodInvocation(context);
-                            stringLiteralWriter = context.Writer.BuildStringLiteralWriter();
+                            context.Writer.Write("@\"");
                             insideWrite = true;
                         }
 
                         var lengthToWrite = Math.Min(remainingLength, MaxStringLiteralLength - writtenLength);
-                        stringLiteralWriter.Write(token.Content, token.Content.Length - remainingLength, lengthToWrite);
+                        WriteStringLiteralPiece(context.Writer, token.Content, token.Content.Length - remainingLength, lengthToWrite);
 
                         remainingLength -= lengthToWrite;
                         writtenLength += lengthToWrite;
 
                         if (writtenLength >= MaxStringLiteralLength)
                         {
-                            stringLiteralWriter.Dispose();
+                            context.Writer.Write("\"");
                             WriteHtmlContentEndMethodInvocation(context);
                             insideWrite = false;
                             writtenLength = 0;
@@ -281,9 +280,29 @@ namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 
             if (insideWrite)
             {
-                stringLiteralWriter.Dispose();
+                context.Writer.Write("\"");
                 WriteHtmlContentEndMethodInvocation(context);
             }
+        }
+
+        private void WriteStringLiteralPiece(CSharpCodeWriter writer, string literalPiece, int startIndex, int length)
+        {
+            // We need to find the index of each '"' (double-quote) to escape it.
+            var start = startIndex;
+            int end;
+            while ((end = literalPiece.IndexOf('\"', start, length - (start - startIndex))) > -1)
+            {
+                writer.Write(literalPiece, start, end - start);
+
+                writer.Write("\"\"");
+
+                start = end + 1;
+            }
+
+            Debug.Assert(end == -1); // We've hit all of the double-quotes.
+
+            // Write the remainder after the last double-quote.
+            writer.Write(literalPiece, start, startIndex + length - start);
         }
 
         protected virtual void WriteHtmlContentStartMethodInvocation(CSharpRenderingContext context)
--- a/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
@@ -7,6 +7,7 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
 
 namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 {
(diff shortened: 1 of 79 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
```diff
@@ -651,5 +656,46 @@ public void Dispose()
             {
             }
         }
+
+        public struct StringLiteralWriter : IDisposable
```
Yves57: Proposition:
- Allows to keep the C# string syntax in CSharpCodeWriter.
- Reusable with the "IDisposable" pattern in the future.

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RedirectedRuntimeBasicWriterTest.cs
```diff
@@ -207,7 +207,7 @@ public void WriteHtmlContent_RendersContentCorrectly()
             // Assert
             var csharp = context.Writer.Builder.ToString();
             Assert.Equal(
-@"WriteLiteralTo(test_writer, ""SomeContent"");
+@"WriteLiteralTo(test_writer, @""SomeContent"");
```
NTaylorMullen: Why's there an `@` now?

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
NTaylorMullen: This could all just be a single method though (no need to maintain state). Don't see the value in it being its own struct.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
```diff
@@ -239,42 +240,60 @@ public override void WriteCSharpAttributeValue(CSharpRenderingContext context, C
...
-                    .WriteStringLiteral(textToRender)
-                    .WriteEndMethodInvocation();
+        protected virtual void WriteHtmlContentStartMethodInvocation(CSharpRenderingContext context, string methodName)
+        {
+            context.Writer.WriteStartMethodInvocation(methodName);
+        }
 
-                charactersConsumed += textToRender.Length;
-            }
+        protected virtual void WriteHtmlContentEndMethodInvocation(CSharpRenderingContext context)
```
NTaylorMullen: Don't see this guy getting a lot of use outside of this class. Any value?

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
Yves57: The quote is closed in the `Dispose`. In fact I tried to create something similar to the private class `LinePragmaWriter` just above.
It allows to create something like
```
using (var literalWriter = writer.BuildStringLiteralWriter())
{
    literalWriter.Write("foo");
    literalWriter.Write("bar");
}
```

NTaylorMullen: Do we do that in Razor though? From the uses I see we render the string entirely every time (no need for the disposable)

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RedirectedRuntimeBasicWriterTest.cs
Yves57: The `RedirectedRuntimeBasicWriter` class is derived from `RuntimeBasicWriter`. And so `WriteLiteral` is replaced by `WriteLiteralTo`, but with the same tokens.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
Yves57: The `WriteHtmlContentStartMethodInvocation` method is virtual. I find logical to make its associated method `WriteHtmlContentEndMethodInvocation` virtual too, even if it is not currently overriden.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeBasicWriter.cs
```diff
@@ -239,42 +240,60 @@ public override void WriteCSharpAttributeValue(CSharpRenderingContext context, C
...
+                stringLiteralWriter.Dispose();
+                WriteHtmlContentEndMethodInvocation(context);
+            }
+        }
 
-                context.Writer
-                    .WriteStartMethodInvocation(WriteHtmlContentMethod)
-                    .WriteStringLiteral(textToRender)
-                    .WriteEndMethodInvocation();
+        protected virtual void WriteHtmlContentStartMethodInvocation(CSharpRenderingContext context, string methodName)
```
Yves57: Old modification, but not necessary anymore. I will change it.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
Yves57: The PR is made to render a list of HTML token into the CSharpCodeWriter, but this list of token is divided into several "string literals", and these "string literals" are composed of the token contents (or even part of token contents).

Yves57:
Sorry to not provide any memory snapshot, I have no tool for that. Just for information, I precise that StringLiteralWriter is a struct to avoid allocating class instance.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
NTaylorMullen: Totally get it. Only thing is you only ever truly use it once in `RuntimeBasicWriter`. Might as well be a method in there that takes a list of the tokens

Yves57: Do you mean something like `CSharpCodeWriter WriteHtmlToken(ICollection<RazorIRNode> nodes)` in the `CSharpCodeWriter` class?
In my opinion `RazorIRNode` should not mixed with the `CSharpCodeWriter` that is at a lower level... But at the end it's up to you 😄

NTaylorMullen: Oh, nonoo. Just make a helper method in `RuntimeBasicWriter`. You wont have to touch the CSharpCodeWriter at all.

Yves57: I was a little bit surprised too 😄 .
I have thought about helper method idea, but it means that this helper method in `RuntimeBasicWriter` have to have the knowledge of the C# syntax.

NTaylorMullen: It already does based on how it writes out C# 😄

ajaybhargavb:
**Perf data**:

Before:
image

After:
image

@Yves57, fyi we have a performance project in Razor. You can use that to generate these benchmarks. https://github.com/aspnet/Razor/tree/dev/test/Microsoft.AspNetCore.Razor.Performance

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
Yves57: So you prefer that I "move" the three `StringLiteralWriter` methods to the `RunTimeBasicWriter`?

Yves57:
Sorry about the benchmark, I was thinking about .NET Memory snapshot or something like that. But don't waste your time. You would have explainged to me, and I would have done the test.
But I'm a little bit surprised by the "Design Time" results: no change in allocations.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
NTaylorMullen: Can be a single `RenderLiteralHtmlTokens` method in `RuntimeBasicWriter`

NTaylorMullen:
> "Design Time" results: no change in allocations.

We don't render `WriteLiteral` methods at design time.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeWriter.cs
Yves57: I don't see how it would be possible in a single method (except with some particular output parameters). Indeed we don't know how many tokens have to written before the `MaxStringLiteralLength`. A token may be even be split in two string literals.
So I think we have to "start the string literal", write the token one by one until we reach the limit, "close the string", restart a new string literal,...

NTaylorMullen: Whenever you cal BuildStringLiteral you'd write out the verbatim string literal start and whenever you disposed it you'd write out the string end. Either way, we can refactor later. 

Don't want you to burn too many cycles on it 😉

NTaylorMullen:
Closing for now. All of this code has significantly changed so we'll need to revisit.

PR #1277, opened 2017-05-03
Introduce SourceChange in place of TextChange

rynowak:
This is the first step on the journey to replacing RazorEditorParser. We
can't just get rid of this because VS is using it today.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/AutoCompleteEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/AutoCompleteEditHandler.cs
@@ -31,11 +31,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
         public string AutoCompleteString { get; set; }
 
-        protected override PartialParseResult CanAcceptChange(Span target, TextChange normalizedChange)
+        protected override PartialParseResult CanAcceptChange(Span target, SourceChange change)
         {
-            if (((AutoCompleteAtEndOfSpan && IsAtEndOfSpan(target, normalizedChange)) || IsAtEndOfFirstLine(target, normalizedChange)) &&
-                normalizedChange.IsInsert &&
-                ParserHelpers.IsNewLine(normalizedChange.NewText) &&
+            if (((AutoCompleteAtEndOfSpan && IsAtEndOfSpan(target, change)) || IsAtEndOfFirstLine(target, change)) &&
+                change.IsInsert &&
+                ParserHelpers.IsNewLine(change.NewText) &&
                 AutoCompleteString != null)
             {
                 return PartialParseResult.Rejected | PartialParseResult.AutoCompleteBlock;
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/BackgroundParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/BackgroundParser.cs
@@ -43,7 +43,9 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             _main.Cancel();
         }
 
+#pragma warning disable CS0612 // Type or member is obsolete
         public void QueueChange(TextChange change)
+#pragma warning restore CS0612 // Type or member is obsolete
         {
             _main.QueueChange(change);
         }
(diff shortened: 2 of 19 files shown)


rynowak:
TODO here:
- [ ] I still need to add some unit tests for the new types
- [ ] I need to figure out a few test failures

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -105,57 +105,57 @@ protected override PartialParseResult CanAcceptChange(Span target, TextChange no
...
         {
             return IsNewDotlessCommitInsertion(target, change) || IsSecondaryDotlessCommitInsertion(target, change);
         }
 
         // Completing 'DateTime' in intellisense with a '.' could result in: '@DateT' -> '@DateT.' -> '@DateTime.' which is accepted.
-        private static bool IsNewDotlessCommitInsertion(Span target, TextChange change)
+        private static bool IsNewDotlessCommitInsertion(Span target, SourceChange change)
         {
             return !IsAtEndOfSpan(target, change) &&
-                   change.NewPosition > 0 &&
```
rynowak: @NTaylorMullen - what's the **meaning** of `NewPosition`?? I just can't get my head around why this code was correct.

rynowak: There are only two uses of `NewPosition` anywhere, both in this file, and I think that's leading to my 4 test failures.

src/Microsoft.AspNetCore.Razor.Language/SourceChange.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            var offset = GetOffset(span);
+            if (offset == span.Content.Length)
+            {
+                return string.Empty;
+            }
+
+            return span.Content.Substring(offset, span.Length);
+        }
+    }
```
rynowak: I need to add a `ToString` and equality here also.

src/Microsoft.AspNetCore.Razor.Language/Legacy/BackgroundParser.cs
```diff
@@ -67,37 +69,43 @@ protected virtual void OnResultsReady(DocumentParseCompleteEventArgs args)
             }
         }
 
-        internal static bool TreesAreDifferent(RazorSyntaxTree leftTree, RazorSyntaxTree rightTree, IEnumerable<TextChange> changes)
-        {
-            return TreesAreDifferent(leftTree, rightTree, changes, CancellationToken.None);
-        }
-
+#pragma warning disable CS0612 // Type or member is obsolete
```
NTaylorMullen: 😢 we could just align an insertion with WTE on the removal of this

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
NTaylorMullen: Absolutely no clue

src/Microsoft.AspNetCore.Razor.Language/Legacy/AutoCompleteEditHandler.cs
```diff
@@ -31,11 +31,12 @@ public AutoCompleteEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer, Acc
...
         public string AutoCompleteString { get; set; }
 
-        protected override PartialParseResult CanAcceptChange(Span target, TextChange normalizedChange)
+        protected override PartialParseResult CanAcceptChange(Span target, SourceChange change)
         {
-            if (((AutoCompleteAtEndOfSpan && IsAtEndOfSpan(target, normalizedChange)) || IsAtEndOfFirstLine(target, normalizedChange)) &&
-                normalizedChange.IsInsert &&
-                ParserHelpers.IsNewLine(normalizedChange.NewText) &&
+            if (((AutoCompleteAtEndOfSpan && IsAtEndOfSpan(target, change)) || IsAtEndOfFirstLine(target, change)) &&
+                change.Span.Length == 0 &&
```
NTaylorMullen: Why not just use IsInsert on the source change?

src/Microsoft.AspNetCore.Razor.Language/Legacy/SpanEditHandler.cs
```diff
@@ -82,42 +82,39 @@ protected virtual SpanBuilder UpdateSpan(Span target, TextChange normalizedChang
             return newSpan;
         }
 
-        protected internal static bool IsAtEndOfFirstLine(Span target, TextChange change)
+        public string ApplyChange(SourceChange change, string content, int changeOffset)
+        {
+            var changeRelativePosition = change.Span.AbsoluteIndex - changeOffset;
+
+            Debug.Assert(changeRelativePosition >= 0);
+            return content.Remove(changeRelativePosition, change.Span.Length).Insert(changeRelativePosition, change.NewText);
```
NTaylorMullen: Could we make this bit more debuggable. Have a feeling we might be debugging through here in the future:
```C#
var appliedChange = content.Remove(changeRelativePosition, change.Span.Length).Insert(changeRelativePosition, change.NewText);
return appliedChange;
```

src/Microsoft.CodeAnalysis.Razor/TextChangeExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    internal static class TextChangeExtensions
+    {
+        public static SourceChange AsSourceChange(this TextChange textChange)
```
NTaylorMullen: Why doesn't this guy need a pragma warn?

src/Microsoft.AspNetCore.Razor.Language/Legacy/BackgroundParser.cs
rynowak: I don't want to make incremental updates to `RazorEditorParser`, I'm trying to replace it completely.

src/Microsoft.AspNetCore.Razor.Language/Legacy/AutoCompleteEditHandler.cs
rynowak: Yeah I will. I think I wrote this before I added `IsInsert`

src/Microsoft.AspNetCore.Razor.Language/Legacy/SpanEditHandler.cs
rynowak: Sure

src/Microsoft.CodeAnalysis.Razor/TextChangeExtensions.cs
rynowak: hmm wierd.

src/Microsoft.AspNetCore.Razor.Language/Legacy/SpanEditHandler.cs
rynowak: Actually going to remove this, this should be part of `SourceChange`

src/Microsoft.CodeAnalysis.Razor/TextChangeExtensions.cs
rynowak: Ok. Because this is the Roslyn type, not the Razor legacy type.

rynowak:
Updated

Issue #1058, opened 2017-03-07
Tag Helpers: enum attribute of nested enum type


johnendev:
#196 allows writing only the enum member as the attribute value.
When using a tag helper property with a nested enum type the generated code is invalid.
```c#
class Parent {
  public enum Foo {
    Bar,
  }
}
```
```html
<mytaghelper someattr="Bar" />
```

The generated code is like this:
```c#
__MyTagHelper.SomeAttr = global::SomeNamespace.Parent+Child.Bar;
```

Microsoft.AspNetCore.Mvc.Razor v1.1.1

Eilon:
@ajaybhargavb we think this might already be fixed. Either way, please make sure there's a test for this case.

Eilon:
3-Done?

ajaybhargavb:
Oops 😄

PR #1280, opened 2017-05-04
Generate expected tag helper descriptors correctly

ajaybhargavb:
Issue - #1058 

@NTaylorMullen @rynowak

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/IntegrationTests/TestTagHelperDescriptors.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/IntegrationTests/TestTagHelperDescriptors.cs
@@ -4,11 +4,15 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.AspNetCore.Razor.Language.IntegrationTests
 {
     public class TestTagHelperDescriptors
     {
+        private static Compilation Compilation { get; } = TestCompilation.Create(typeof(TestTagHelperDescriptors));
+
         public static IEnumerable<TagHelperDescriptor> SimpleTagHelperDescriptors
         {
             get
@@ -158,7 +162,7 @@ namespace Microsoft.AspNetCore.Razor.Language.IntegrationTests
                                 .Name("catch-all")
                                 .PropertyName("CatchAll")
                                 .AsEnum()
-                                .TypeName(typeof(MyEnum).FullName),
+                                .TypeName(GetFullName(typeof(MyEnum))),
                         }),
                     CreateTagHelperDescriptor(
                         tagName: "input",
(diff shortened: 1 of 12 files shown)


test/Microsoft.AspNetCore.Razor.Test.Common/Microsoft.AspNetCore.Razor.Test.Common.csproj
```diff
@@ -5,13 +5,15 @@
     <!-- To generate baselines, run tests with /p:GenerateBaselines=true -->
     <DefineConstants Condition="'$(GenerateBaselines)'=='true'">$(DefineConstants);GENERATE_BASELINES</DefineConstants>
     <DefineConstants>$(DefineConstants);__RemoveThisBitTo__GENERATE_BASELINES</DefineConstants>
-    <TargetFrameworks>netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;net46</TargetFrameworks>
```
ajaybhargavb: `netstandard1.6` was needed because `DependencyContext.Load` is not available in `1.3`. Also needed to cross compile here for compatibility.

test/Microsoft.AspNetCore.Razor.Language.Test/IntegrationTests/TestTagHelperDescriptors.cs
```diff
@@ -158,7 +162,7 @@ public static IEnumerable<TagHelperDescriptor> EnumTagHelperDescriptors
                                 .Name("catch-all")
                                 .PropertyName("CatchAll")
                                 .AsEnum()
-                                .TypeName(typeof(MyEnum).FullName),
+                                .TypeName(GetFullName(typeof(MyEnum))),
```
rynowak: Can we just hardcode the actual string you expect to see? This is too much magic for the test. You copied product code into here to do this.

rynowak: Frankly, this whole approach is wrong. We gain no value from doing a `.Equals()` on the descriptors. It would be much better to have the tests validate their expectations explicitly.

test/Microsoft.CodeAnalysis.Razor.Test/DefaultTagHelperDescriptorFactoryTest.cs
```diff
@@ -1306,7 +1306,7 @@ public class DynamicTestTagHelper : {typeof(AspNetCore.Razor.TagHelpers.TagHelpe
         {{
         }}";
             var syntaxTree = CSharpSyntaxTree.ParseText(text);
-            var compilation = TestCompilation.Create(syntaxTree);
+            var compilation = TestCompilation.Create(typeof(DefaultTagHelperDescriptorFactoryTest), syntaxTree);
```
rynowak: Think about making this into a method, it's not super create just looking at the code of all the tests what this does. It would be better to pull it out into a method and make it clear what it means.

ajaybhargavb:
Spoke offline. Going to hard code the test expected string. Closing this PR as a lot of the changes here are unnecessary.

PR #1281, opened 2017-05-04
Added a test to verify nested enum behavior in taghelper attribute

ajaybhargavb:
Issue - #1058 

@NTaylorMullen @rynowak


test/Microsoft.AspNetCore.Razor.Language.Test/IntegrationTests/TestTagHelperDescriptors.cs
```diff
@@ -158,7 +158,7 @@ public static IEnumerable<TagHelperDescriptor> EnumTagHelperDescriptors
                                 .Name("catch-all")
                                 .PropertyName("CatchAll")
                                 .AsEnum()
-                                .TypeName(typeof(MyEnum).FullName),
+                                .TypeName($"{typeof(TestTagHelperDescriptors).FullName}.{nameof(MyEnum)}"),
```
ajaybhargavb: Could directly hard code the entire string here but felt this was cleaner and consistent with others.

Issue #1230, opened 2017-04-14
Consider not providing catch-all IntelliSense to existing completions.


NTaylorMullen:
Today our `DefaultTagHelperElementCompletionService` will mark every completion with a catch-all `TagHelper` if it exists. Aka you get a massive list of `TagHelper`ified elements:
![image](https://cloud.githubusercontent.com/assets/2008729/25054029/c7404c06-210f-11e7-9f7e-5fc6e0564913.png)

This is consistent with how `TagHelper` IntelliSense has worked in the past but it doesn't seem useful. It might be better to exclude catch-all based `TagHelper`s entirely from IntelliSense completion. This way a catch-all `TagHelper` that has required attributes will only light up when its required attributes have been satisfied. It also means we wont be polluting the completion lists with excess amounts of `TagHelper`s.

rynowak:
@DamianEdwards - do you have an opinion on this behavior?

DamianEdwards:
Agree with @NTaylorMullen. I'm assuming the attributes on said Tag Helpers **will** still show up though, yes?

rynowak:
Ok. I'm going to slot this into preview-2. For preview-1 we're going to have the behavior Dev14 had here.

NTaylorMullen:
> I'm assuming the attributes on said Tag Helpers will still show up though, yes?

Yes.

NTaylorMullen:
e87011f77fb5d287cf7548c64c393115d94bb242

PR #1284, opened 2017-05-05
Reduce catch all element completion usage.

NTaylorMullen:
- This makes it so if you have catch all `TagHelperDescriptor`s their completions don't apply to every existing completion. Instead, they now only apply to already TagHelperified completions and existing completions that are prefixed with a non-empty TagHelperPrefix.
- Updated existing test to have new expectations.
- Added new test to validate non-empty tag helper prefix case.

#1230 

/cc @alexgav This changes the expectations on how catch-all TagHelpers work.

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
@@ -194,8 +194,11 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor
             {
                 foreach (var completionTagName in elementCompletions.Keys)
                 {
-                    if (completionTagName.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                    if (elementCompletions[completionTagName].Count > 0 || 
+                        !string.IsNullOrEmpty(prefix) && completionTagName.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
                     {
+                        // The current completion either has other TagHelper's associated with it or is prefixed with a non-empty
+                        // TagHelper prefix.
                         UpdateCompletions(completionTagName, catchAllDescriptor);
                     }
                 }
--- a/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionServiceTest.cs
+++ b/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionServiceTest.cs
@@ -509,7 +509,7 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor
         }
 
         [Fact]
-        public void GetElementCompletions_CatchAllsApplyToAllCompletions()
+        public void GetElementCompletions_CatchAllsApplyToOnlyTagHelperCompletions()
         {
             // Arrange
             var documentDescriptors = new[]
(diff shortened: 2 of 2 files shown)


rynowak:
We should probably wait until @alexgav is ready to digest these changes on his side since it's going to break their tests. I don't want their team to have to deal with this while he is on vacation.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
```diff
@@ -194,8 +194,11 @@ public override ElementCompletionResult GetElementCompletions(ElementCompletionC
             {
                 foreach (var completionTagName in elementCompletions.Keys)
                 {
-                    if (completionTagName.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+                    if (elementCompletions[completionTagName].Count > 0 ||
```
rynowak: Why do we need to check `count > 0`? I thought we were just saying catch alls wouldn't be offered as completions?

Also this is order dependent. It will only add a single one of the catch alls.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionServiceTest.cs
```diff
@@ -542,6 +542,40 @@ public void GetElementCompletions_CatchAllsApplyToAllCompletions()
...
+        {
+            // Arrange
+            var documentDescriptors = new[]
+            {
+                TagHelperDescriptorBuilder.Create("SuperLiTagHelper", "TestAssembly")
+                    .TagMatchingRule(rule => rule.RequireTagName("superli"))
+                    .Build(),
+                TagHelperDescriptorBuilder.Create("CatchAll", "TestAssembly")
+                    .TagMatchingRule(rule => rule.RequireTagName("*"))
+                    .Build(),
```
rynowak: I think you have a bug if you have multiple catch all tag helpers.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
NTaylorMullen: > Why do we need to check count > 0? I thought we were just saying catch alls wouldn't be offered as completions?

The goal of the issue is to not flood the users completion list with TagHelpers. It made sense to me to still show that catch-alls were possible on already TagHelperified elements.



> Also this is order dependent. It will only add a single one of the catch alls.

I think you're misreading the code. All CatchAlls will be applied.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperCompletionServiceTest.cs
NTaylorMullen: You confuse me. Could you elaborate? 😉

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperCompletionService.cs
rynowak: Yeah Ok my brain is fried from too much multitask.

NTaylorMullen:
> Ok, this change looks fine, but let's wait for @alexgav to return

Totally agree

Issue #1060, opened 2017-03-08
Instead of tracking @tagHelperPrefix at the code document level, find another way


NTaylorMullen:
https://github.com/aspnet/Razor/pull/1054#discussion_r104776376

rynowak:
This is now part of the TagHelperDocumentContext. Let's use that instead of a one-off thing

rynowak:
a570139b080b1f9aef10fa79bf27ad12d0c00f0c

PR #1286, opened 2017-05-08
Fix #1059 and #1060

### LINKED ISSUE #1059: Change ported Razor tests that dynamically add a prefix to use @tagHelperPrefix
We didn't use `@tagHelperPrefix` in test files in the past because the logic to understand `@tagHelperPrefix` wasn't in the core of Razor. Now that it is, we should make sure our tests properly verify that `@tagHelperPrefix` works as a part of a Razor file.

rynowak:


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIRLoweringPhase.cs
```diff
@@ -17,6 +17,9 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
             var syntaxTree = codeDocument.GetSyntaxTree();
             ThrowForMissingDependency(syntaxTree);
 
+            var tagHelperContext = codeDocument.GetTagHelperContext();
```
ajaybhargavb: Can this method be renamed to `GetTagHelperDocumentContext()`? We already have a `TagHelperContext` in `Razor.Runtime` that is more widely known. This could cause some confusion.

rynowak: I'd prefer not to, this is used in tooling. @NTaylorMullen thoughts?

NTaylorMullen: Agree with @ajaybhargavb here. I find that name overlap is incredibly confusing for us and for consumers.

rynowak:
Ok I am going to log a follow up issue for the rename.

rynowak:
a570139b080b1f9aef10fa79bf27ad12d0c00f0c

Issue #1279, opened 2017-05-04
Clean up TagHelperResolver - remove assemblyFilters


rynowak:
We should remove the `assemblyFilters` argument from this class and also consider removing the 'result type'. We've never had 'global' errors from tag helpers, and the editor doesn't use them today.

NTaylorMullen:
cd0fcfb05ede1ef189168c2c4ac63b4b34f71a8a

PR #1288, opened 2017-05-08
Remove assembly name filter from THD API.

NTaylorMullen:
- This parameter was always passed `null` by tooling in the past.

#1279


NTaylorMullen:
> No test changes? (o_O)

We're just that amazing at testing lmao

Issue #1184, opened 2017-04-06
Figure out a cleaner way to deal with redirect writers


ajaybhargavb:
We need to investigate if there is a way to get rid of `WriteTo`, `WriteLiteralTo` etc completely in our runtime code gen which in turn will remove the need to have redirected writers like `RedirectedBasicWriter` which does pretty much what `RuntimeBasicWriter` does except use `WriteTo`.
Also need to investigate the implications of this working with Razor 1.0.

cc @NTaylorMullen @rynowak

ajaybhargavb:
b44d59ca36e63f403d13c37186baef8cc4551158

PR #1291, opened 2017-05-09
Remove redirect writers

ajaybhargavb:
#1184 

@rynowak @NTaylorMullen 

Will send the Mvc PR soon. But this can be reviewed independently.


test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ConditionalAttributes_Runtime.codegen.cs
```diff
@@ -62,13 +62,14 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_Conditiona
...
             WriteAttributeValue("", 256, new Microsoft.AspNetCore.Mvc.Razor.HelperResult(async(__razor_attribute_value_writer) => {
+                PushWriter(__razor_attribute_value_writer);
 #line 10 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ConditionalAttributes.cshtml"
                if(cls != null) { 
 
 #line default
 #line hidden
 #line 10 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ConditionalAttributes.cshtml"
-WriteTo(__razor_attribute_value_writer, cls);
+                            Write(cls);
```
NTaylorMullen: See, this is awesomesauce. Since the prefix is no longer overly long, we can have code gen line up with source as its supposed to be. Debugging win

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/TemplateTargetExtension.cs
```diff
@@ -9,9 +9,15 @@ namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
     internal class TemplateTargetExtension : ITemplateTargetExtension
     {
         public static readonly string DefaultTemplateTypeName = "Microsoft.AspNetCore.Mvc.Razor.HelperResult";
+        public static readonly string DefaultPushWriterMethod = "PushWriter";
+        public static readonly string DefaultPopWriterMethod = "PopWriter";
 
         public string TemplateTypeName { get; set; } = DefaultTemplateTypeName;
 
+        public string PushWriterMethod { get; set; } = DefaultPushWriterMethod;
```
rynowak: Can this just call into the basic writer? Or don't we have access?

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/TemplateTargetExtensionTest.cs
```diff
@@ -32,17 +34,19 @@ public void WriteTemplate_WritesTemplateCode()
             {
                 Assert.Same(node, n);
 
-                Assert.IsType<RedirectedRuntimeBasicWriter>(context.BasicWriter);
+                Assert.IsType<RuntimeBasicWriter>(context.BasicWriter);
```
rynowak: There's no need for this assert anymore.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/TemplateTargetExtension.cs
ajaybhargavb: Not really. Those properties are not static.

rynowak: Why do we need the properties? SHouldn't we just call the methods on basicwriter? or are they not exposed?

ajaybhargavb: Missed this question. Not sure which methods you mean. We do have these as virtual properties on `RuntimeBasicWriter`. Should they just be `static`?

ajaybhargavb: Spoke offline. The logic to scope writer should live in one place. Filed https://github.com/aspnet/Razor/issues/1332 to do this as a follow-up.

PR #1295, opened 2017-05-10
Adds a tracking system for Razor documents

rynowak:
This is a MEF service that can actively or passively track open ITextViews
and give us information about the Razor initialization state and eventing
when it changes.

The purpose of this is to act as a bridge between the VS mef world and the
roslyn world.

For now this doesn't do any passive tracking of Razor documents, it's only
on demand. That means it will only be initialized and used right now when
you are using the Razor developer tools. This is just to reduce our risk,
it's not ideal to ship code in VS that's doing something without anyone
looking at the result.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentTracker.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentTracker.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public abstract class RazorDocumentTracker
+    {
+        public abstract event EventHandler ContextChanged;
+
+        public abstract bool IsSupportedDocument { get; }
+
+        public abstract ProjectId ProjectId { get; }
+
+        public abstract SourceTextContainer TextContainer { get; }
+
+        public abstract Workspace Workspace { get; }
+    }
+}
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTextViewRazorDocumentTracker.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTextViewRazorDocumentTracker.cs
@@ -1,0 +1,267 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.VisualStudio.Text.Projection;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    internal class DefaultTextViewRazorDocumentTracker : RazorDocumentTracker
+    {
+        private readonly DefaultTextViewRazorDocumentTrackerService _service;
+        private readonly ITextView _textView;
+
+        private DocumentId _documentId;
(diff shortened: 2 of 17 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTextViewRazorDocumentTracker.cs
```diff
@@ -0,0 +1,267 @@
...
+
+        public override bool IsSupportedDocument => _isSupportedProject && _textContainer != null;
+
+        public override ProjectId ProjectId => _documentId?.ProjectId;
+
+        public override SourceTextContainer TextContainer => _textContainer;
+
+        public override Workspace Workspace => _workspace;
+
+        private bool Update()
```
rynowak: This code is pretty much impossible to test unfortunately. It depends on statics from Roslyn and I don't think it really makes much sense to decouple it.

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentTracker.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public abstract class RazorDocumentTracker
+    {
+        public abstract event EventHandler ContextChanged;
+
+        public abstract bool IsSupportedDocument { get; }
```
NTaylorMullen: What's this guy supposed to represent? Are we expecting document trackers to be produced for non-valid projects?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTextViewRazorDocumentTrackerService.cs
```diff
@@ -0,0 +1,169 @@
...
+            if (!_documentFactory.TryGetTextDocument(textBuffer, out var textDocument))
+            {
+                return null;
+            }
+
+            RunningDocumentTable.FindDocument(textDocument.FilePath, out var hierarchy, out uint itemId, out uint cookie);
+
+            // We don't currently try to look a Roslyn ProjectId at this point, we just want to know some
+            // basic things.
+            // See https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/Core/SolutionExplorerShim/HierarchyItemToProjectIdMap.cs#L14
```
NTaylorMullen: Should have a specific commit link. Otherwise this comment will go stale.

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(RoslynDevVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.EditorFeatures.Text" Version="$(RoslynDevVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="$(RoslynDevVersion)" />
+    <PackageReference Include="Microsoft.VisualStudio.Editor" Version="15.0.26201" />
```
NTaylorMullen: Should lift this version into our common version file.

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentTracker.cs
rynowak: Yes, the can be. Also an existing document can become invalid and vice-versa

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTextViewRazorDocumentTrackerService.cs
rynowak: sure

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
rynowak: IMO there's very little value to that. 

1. We're only get to reference that version here, the VSIX has to use packages.config
2. VS has like 70 million different versions that are locked and won't change.

PR #1298, opened 2017-05-10
Add annotations

### LINKED ISSUE #990: Add annotations to DocumentIRNode
We want code to be able to attach arbitrary metadata to nodes that isn't consumed by our codegen (think of this like a property bag). 

I'm not sure yet if we want to add this to all IR nodes or just at the document level.

rynowak:
Fixes #990. This should make a lot of IR manipulations very simple.


rynowak:
eb3c47b6cabdc92e7efff9058b215d5281e1c9c6

Issue #1118, opened 2017-03-24
TagHelperBinderSyntaxTreePass should be a phase, not a pass


rynowak:
This is too semantically meaningful to be 'just' a pass. This should be promoted to a phase.

```
Ordering: Parser -> TagHelperBinder -> SyntaxTreePasses
```

ajaybhargavb:
4bc1a76e220926df6326d1de8b94d7248b52be9d

PR #1299, opened 2017-05-10
Moved TagHelperBinder from pass to a phase

ajaybhargavb:
#1118 

@rynowak @NTaylorMullen


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -8,24 +8,23 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    internal class TagHelperBinderSyntaxTreePass : IRazorSyntaxTreePass
+    internal class DefaultRazorTagHelperBinderPhase : RazorEnginePhaseBase, IRazorTagHelperBinderPhase
     {
         private static HashSet<char> InvalidNonWhitespaceNameCharacters = new HashSet<char>(new[]
```
rynowak: Is this used? (It shouldn't be)

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorTagHelperBinderPhaseTest.cs
```diff
@@ -883,11 +908,11 @@ public static TheoryData ProcessTagHelperPrefixData
         {
             // Arrange
             var errorSink = new ErrorSink();
-            var pass = new TagHelperBinderSyntaxTreePass();
+            var phase = new DefaultRazorTagHelperBinderPhase();
             var document = RazorCodeDocument.Create(new DefaultRazorSourceDocument("Test content", encoding: Encoding.UTF8, fileName: "TestFile"));
 
             // Act
-            var prefix = pass.ProcessTagHelperPrefix(((IEnumerable<TagHelperDirectiveDescriptor>)directiveDescriptors).ToList(), document, errorSink);
+            var prefix = phase.ProcessTagHelperPrefix(((IEnumerable<TagHelperDirectiveDescriptor>)directiveDescriptors).ToList(), document, errorSink);
```
rynowak: You can put `IEnumerable<TagHelperDirectiveDescriptor>` in the parameter list and get rid of this cast. This was an old workaround for when THD was internal.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
ajaybhargavb: Yes it is. [Here](https://github.com/aspnet/Razor/blob/ajbaaska/tgbinder-phase/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs#L246)

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -8,24 +8,23 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    internal class TagHelperBinderSyntaxTreePass : IRazorSyntaxTreePass
+    internal class DefaultRazorTagHelperBinderPhase : RazorEnginePhaseBase, IRazorTagHelperBinderPhase
```
NTaylorMullen: From a naming perspective I don't see the value in having `Razor` in this name. Do our other phases follow the inclusion of Razor?

NTaylorMullen: Nm, I see they do. This is good as is 👍

Issue #1253, opened 2017-04-24
Add TagHelperFactService test that validates RequireParentTag with null root.


NTaylorMullen:
We have a test that validates non parent restricted TagHelpers can have a null root [here](https://github.com/aspnet/Razor/blob/d47dc261b814ae2537a5578726f3ba4742c844bd/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/DefaultTagHelperFactsServiceTest.cs#L290-L307). We should also add one that has the paernt tag restriction.

NTaylorMullen:
51fb1b787ab92a95e6e263f7419f80432b6f5f8d

PR #1300, opened 2017-05-10
Add TagHelperFactsServiceTest to validate root with parent expecting …

NTaylorMullen:
…descriptor.

#1253 

/cc @rynowak @ajaybhargavb FYI only.


rynowak:
👍

PR #1302, opened 2017-05-11
Some API cleanup around directives

rynowak:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
```diff
@@ -11,7 +11,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
     public static class InjectDirective
     {
-        public static readonly DirectiveDescriptor Directive = DirectiveDescriptorBuilder.Create("inject").AddType().AddMember().Build();
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "inject",
+            DirectiveDescriptorKind.SingleLine,
+            builder => builder.AddTypeToken().AddMemberToken());
```
rynowak: There are still a few things here that don't feel good. 

I don't like that we put `descriptor` front and center. I want to call this class just `Directive`.

I want to make it easier to build a 'simple' directive.

Something like

`Directive.CreateSimpleDirective("inject", b=> b.AddTypeToken().AddMemberToken());`

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
```diff
@@ -11,7 +11,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
     public static class InjectDirective
     {
-        public static readonly DirectiveDescriptor Directive = DirectiveDescriptorBuilder.Create("inject").AddType().AddMember().Build();
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "inject",
+            DirectiveDescriptorKind.SingleLine,
```
rynowak: This doesn't feel good, also we have `DirectiveDescriptorKind` and `DirectiveTokenKind`. We should be consistent with the use of `Descriptor`.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
NTaylorMullen: Extension methods!!

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs
```diff
@@ -11,12 +11,10 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
...
-        public static readonly DirectiveDescriptor DirectiveDescriptor = DirectiveDescriptorBuilder
-            .Create("page")
-            .BeginOptionals()
-            .AddString() // Route template
-            .AddString() // Page Name
-            .Build();
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "page",
+            DirectiveDescriptorKind.SingleLine,
+            builder => builder.AddOptionalStringToken());
```
NTaylorMullen: What happened to the page name token? Was it just wrong to begin with?

src/Microsoft.AspNetCore.Razor.Language/DirectiveDescriptor.cs
```diff
@@ -1,16 +1,78 @@
...
+
+            public string Name { get; }
+
+            public DirectiveDescriptorKind Kind { get; }
+
+            public IList<DirectiveTokenDescriptor> Tokens { get; }
+
+            public DirectiveDescriptor Build()
+            {
+                return new DefaultDirectiveDescriptor(Name, Kind, Tokens.ToArray());
```
NTaylorMullen: Need to validate that no non-optionals occur after an optional.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InjectDirective.cs
NTaylorMullen: Isn't that going back to the CreateSingleLineDirective/CreateRazorBlockDirective etc. approch?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs
rynowak: Yeah we cut it.

rynowak:
Updated - @ajaybhargavb can you take a quick look since taylor is out

src/Microsoft.AspNetCore.Razor.Language/Resources.resx
```diff
@@ -147,8 +147,8 @@
   <data name="RazorProject_PathMustStartWithForwardSlash" xml:space="preserve">
     <value>Path must begin with a forward slash '/'.</value>
   </data>
-  <data name="DirectiveDescriptor_BeginOptionalsAlreadyInvoked" xml:space="preserve">
-    <value>The method '{0}' has already been invoked.</value>
+  <data name="DirectiveDescriptor_InvalidNonOptionalToken" xml:space="preserve">
+    <value>An non-optional directive token cannot follow an optional directive token.</value>
```
ajaybhargavb: nit: An => A

Issue #1260, opened 2017-04-25
Roslyn-ize all of our VS MEF services


rynowak:
We should make these into `ILanguageService`-based components. Most of the work here will be on the WTE side to flow the workspace through different components.

We will need to stage this:
1. Add new APIs
2. Push this to VS
3. Allow WTE to integrate these
4. Remove old APIs

NTaylorMullen:
aa980fc67c72743694b004bc5dd29869c41281fd

PR #1331, opened 2017-05-11
Add Roslyn language service factories for our exported types.

NTaylorMullen:
- Verified this all worked via our Razor extension:
image

Wasn't going to check in the extension changes due to their lack of usefulness but if you feel different let me know.

#1260

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorSyntaxFactsServiceFactory.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorSyntaxFactsServiceFactory.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(RazorSyntaxFactsService), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultRazorSyntaxFactsServiceFactory : ILanguageServiceFactory
+    {
+        public ILanguageService CreateLanguageService(HostLanguageServices languageServices)
+        {
+            return new DefaultRazorSyntaxFactsService();
+        }
+    }
+}
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Composition;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(ITagHelperResolver), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultTagHelperResolverFactory : ILanguageServiceFactory
+    {
+        private readonly VisualStudioWorkspace _workspace;
+        private readonly SVsServiceProvider _services;
+
+        [ImportingConstructor]
+        public DefaultTagHelperResolverFactory(VisualStudioWorkspace workspace, SVsServiceProvider services)
+        {
+            _workspace = workspace;
+            _services = services;
+        }
+
+        public ILanguageService CreateLanguageService(HostLanguageServices languageServices)
+        {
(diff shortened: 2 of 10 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Composition;
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(ITagHelperResolver), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultTagHelperResolverFactory : ILanguageServiceFactory
```
NTaylorMullen: We should really rename this TagHelper resolver to be a `HostTagHelperResolver` or `VisualStudioTagHelperResolver`. If you agree I can push that in separately.

rynowak: Sure

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultRazorSyntaxFactsServiceFactory.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(RazorSyntaxFactsService), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultRazorSyntaxFactsServiceFactory : ILanguageServiceFactory
```
rynowak: These should live in the Razor.Workspaces assembly

rynowak: Anything in ServiceLayer.Default should live there

NTaylorMullen: Ah, will have to move a lot of codez. Coolio, will do.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
rynowak: This looks like the only one that needs to live in the VS assembly.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private readonly VisualStudioWorkspace _workspace;
+        private readonly SVsServiceProvider _services;
+
+        [ImportingConstructor]
+        public DefaultTagHelperResolverFactory(VisualStudioWorkspace workspace, SVsServiceProvider services)
```
rynowak: For improved do-thingability you can use `[Export(typeof(SVsServiceProvider))] IServiceProvider services`. The `SVsServiceProvider` is just a marker.

src/Microsoft.VisualStudio.LanguageServices.Razor/ITagHelperResolver.cs
```diff
@@ -4,11 +4,12 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Host;
 using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    public interface ITagHelperResolver
+    public interface ITagHelperResolver : ILanguageService
```
rynowak: Can we make a new interface for this in `Microsoft.CodeAnalysis.Razor` namespace (`Microsoft.AspNetCore.Razor.Workspaces` assembly).

I don't want to roslynize this with the wrong API

NTaylorMullen: Sure.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorSyntaxFactsService.cs
```diff
@@ -3,11 +3,12 @@
 
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Host;
 using Microsoft.VisualStudio.Text;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    public abstract class RazorSyntaxFactsService
+    public abstract class RazorSyntaxFactsService : ILanguageService
```
rynowak: My general comment here, I don't really want 'roslynize' the wrong API. For instance this is in the wrong namespace/assembly. I think the simplest thing is to copy these base classes

rynowak:
It would be useful to add something that would display the taghelpers as a POC but I understand if you want me to use my WPF-Fu on that instead

NTaylorMullen:
Closing till we revisit this area.

Issue #976, opened 2017-02-07
Update Razor to use new ServiceHubServiceBase ctor


rynowak:
See: https://github.com/dotnet/roslyn/pull/16943

We will need to update our shim to follow the recommended pattern once these bits are available.

rynowak:
As of today the Roslyn bits have marked the old ctor as obsolete, but the servicehub does not yet work with the new ctor

rynowak:
The work item here is to try this on a new build of VS and see which constructor gets used by servicehub. Then we'll know if we can remove the obsolete one.

NTaylorMullen:
Waiting on VS to take newer Roslyn bits to make this possible.

NTaylorMullen:
870b2640395a9986230cdf131a6be271d74f21f9

PR #1333, opened 2017-05-11
Remove usage of old `ServiceHubServiceBase` ctor.

NTaylorMullen:
- The change wasn't exactly what was expected. The core `ServiceHub` is still what calls into our `Stream`, `IServiceProvider` overload (as intended) which means we needed to call down into the proper basetype constructor to start using Roslyn's non-obsolete one.
- Had to manually start the `StreamJsonRpc` property to work around race conditions in the ServiceHub APIs.

#976 

/cc @heejaechang so once this is in/inserted you'll be able to nuke your obsolete ctor.


heejaechang:
👍

rynowak:
@NTaylorMullen - let's do an insertion into prerel once we're able, and let @heejaechang know when we have done

PR #1334, opened 2017-05-11
[Design] Move a single class over to Workspaces.

NTaylorMullen:
- Had to modify the API a bit to massage the `TagHelperFactsService` into a state that didn't rely on VS.

@rynowak what do you think about this approach for supporting Roslyn's based services + VisualStudio's? I moved one of the more troublesome classes as an experiment.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultRazorSyntaxFactsService.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultRazorSyntaxFactsService.cs
@@ -3,17 +3,13 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
 using System.Linq;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
-using Microsoft.VisualStudio.Text;
 using Span = Microsoft.AspNetCore.Razor.Language.Legacy.Span;
-using ITextBuffer = Microsoft.AspNetCore.Razor.Language.Legacy.ITextBuffer;
 
-namespace Microsoft.VisualStudio.LanguageServices.Razor
+namespace Microsoft.CodeAnalysis.Razor
 {
-    [Export(typeof(RazorSyntaxFactsService))]
     internal class DefaultRazorSyntaxFactsService : RazorSyntaxFactsService
     {
         public override IReadOnlyList<ClassifiedSpan> GetClassifiedSpans(RazorSyntaxTree syntaxTree)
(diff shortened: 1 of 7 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultRazorSyntaxFactsService.cs
```diff
@@ -139,21 +135,21 @@ public override IReadOnlyList<TagHelperSpan> GetTagHelperSpans(RazorSyntaxTree s
             return results;
         }
 
-        public override int? GetDesiredIndentation(RazorSyntaxTree syntaxTree, ITextSnapshot syntaxTreeSnapshot, ITextSnapshotLine line, int indentSize, int tabSize)
+        public override int? GetDesiredIndentation(RazorSyntaxTree syntaxTree, int previousLineEndIndex, Func<int, string> getLineContent, int indentSize, int tabSize)
```
rynowak: This is probably an example of a case where we would just split this particular API out into its own service.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorSyntaxFactsServiceExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public static class RazorSyntaxFactsServiceExtensions
+    {
+        public static int? GetDesiredIndentation(
```
rynowak: ugh. no

src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperSpan.cs
```diff
@@ -6,7 +6,7 @@
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 
-namespace Microsoft.VisualStudio.LanguageServices.Razor
+namespace Microsoft.CodeAnalysis.Razor
```
rynowak: These are breaking changes.

rynowak:
Ah, I think I see what you're doing now. This is a small set of breaking changes.

Some of the details of this are mixed up but the strategy in general is right.

NTaylorMullen:
> Ah, I think I see what you're doing now. This is a small set of breaking changes.
> 
> Some of the details of this are mixed up but the strategy in general is right.

Ya, it's a breaking change as far as namespace go but it doesn't require Alex to change how the underlying code works. Just makes him use a different namespace.

NTaylorMullen:
Closing till we revisit this area.

Issue #1325, opened 2017-05-11
Remove ExtensionIRNodeVisitor<TResult>


rynowak:

rynowak:
ab98c5d837ca3923721476ece55d9ea7b09f8131

Issue #1310, opened 2017-05-11
Remove RazorIRPass class


rynowak:
Move DefaultFeatureOrder to base class for passes

rynowak:
ab98c5d837ca3923721476ece55d9ea7b09f8131

PR #1336, opened 2017-05-12
Fix #1322 #1328 #1310 #1325

### LINKED ISSUE #1322: Seal all public non-abstract IR node types

rynowak:
This are mechanical API Cleanup issues.


rynowak:
FYI only

PR #1341, opened 2017-05-13
Remove default usings

rynowak:
This change removes the default usings for 'System' and
'System.Threading.Tasks' and adds them to the MVC template engine.

This is preparation for removing this feature from the razor options, I
wanted to get all of the intentional diff out of the way.


rynowak:
FYI only

PR #1344, opened 2017-05-13
Fix #1330 - make TagHelpers static class internal

### LINKED ISSUE #1330: Get rid of TagHelpers static class

rynowak:


rynowak:
FYI only

PR #1346, opened 2017-05-15
Set the chunk generator of non-string tag helper attributes to ExpressionChunkGenerator

ajaybhargavb:
https://github.com/aspnet/Razor/issues/963

@NTaylorMullen @rynowak 

When we configure the SpanBuilder from non-string attributes, we were not setting the correct chunk generator. Setting it to `ExpressionChunkGenerator` results in the correct IR being generated.
This is a change from the behavior of old Razor. I am not sure why it was that way @NTaylorMullen?

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
@@ -38,12 +38,9 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                         Content = "ModelExpressionProvider.CreateModelExpression(ViewData, __model => ",
                     });
 
-                    if (node.Children.Count == 1 && node.Children[0] is HtmlContentIRNode original)
+                    if (node.Children.Count == 1 && node.Children[0] is RazorIRToken token && token.IsCSharp)
                     {
                         // A 'simple' expression will look like __model => __model.Foo
-                        //
-                        // Note that the fact we're looking for HTML here is based on a bug.
-                        // https://github.com/aspnet/Razor/issues/963
 
                         builder.Add(new RazorIRToken()
                         {
@@ -51,13 +48,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                             Content = "__model."
                         });
 
-                        var content = GetContent(original);
-                        builder.Add(new RazorIRToken()
-                        {
-                            Kind = RazorIRToken.TokenKind.CSharp,
-                            Content = content,
-                            Source = original.Source,
-                        });
+                        builder.Add(token);
                     }
                     else
                     {
@@ -76,19 +67,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 
                                 continue;
                             }
-
-                            // Note that the fact we're looking for HTML here is based on a bug.
-                            // https://github.com/aspnet/Razor/issues/963
-                            if (node.Children[i] is HtmlContentIRNode html)
-                            {
-                                var content = GetContent(html);
-                                builder.Add(new RazorIRToken()
-                                {
-                                    Kind = RazorIRToken.TokenKind.CSharp,
-                                    Content = content,
-                                    Source = html.Source,
-                                });
-                            }
                         }
                     }
 
@@ -103,20 +81,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                     node.Children.Add(expression);
                     expression.Parent = node;
                 }
-            }
-
-            private string GetContent(HtmlContentIRNode node)
-            {
-                var builder = new StringBuilder();
-                for (var i = 0; i < node.Children.Count; i++)
-                {
-                    if (node.Children[i] is RazorIRToken token && token.IsHtml)
-                    {
-                        builder.Append(token.Content);
-                    }
-                }
-
-                return builder.ToString();
             }
         }
     }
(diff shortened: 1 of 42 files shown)


rynowak:
I'm glad that this ended up being a straightforward change. I'll let @NTaylorMullen review this because only he knows the history.

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/DuplicateAttributeTagHelpers_Runtime.codegen.cs
```diff
@@ -7,12 +7,7 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_DuplicateA
...
         private static readonly global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute __tagHelperAttribute_1 = new global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute("TYPE", new global::Microsoft.AspNetCore.Html.HtmlString("checkbox"), global::Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.DoubleQuotes);
         private static readonly global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute __tagHelperAttribute_2 = new global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute("type", new global::Microsoft.AspNetCore.Html.HtmlString("checkbox"), global::Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.DoubleQuotes);
-        private static readonly global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute __tagHelperAttribute_3 = new global::Microsoft.AspNetCore.Razor.TagHelpers.TagHelperAttribute("checked", new global::Microsoft.AspNetCore.Html.HtmlString("false"), global::Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.DoubleQuotes);
```
NTaylorMullen: What happened here?

ajaybhargavb: We only preallocate tag helper attribute nodes with `HtmlContentIRNode` children. https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/RazorPreallocatedTagHelperAttributeOptimizationPass.cs#L39

rynowak: What does the code look like? Is this a bug?

ajaybhargavb: We are now losing the preallocated optimization for non-string attributes. Also note, this only affects the case when there is a duplicate bound attribute. I don't think it is a bug but I am not sure if we want to leave it this way. I am now checking if I can do some special casing in PreallocatedOptimizationPass for these attributes.

NTaylorMullen: Interesting. So we were pre-allocating dynamic content as strings prior to this? If that's the case then this is definitely fixing a bug.

ajaybhargavb: Not quite. For example (`checked` binds to a boolean type),
`<input checked="true" checked="false" />`
  Here we used to preallocate the second `checked` attribute because the value was `HtmlContentIRNode`. But after my changes the value is a `RazorIRToken` of kind `CSharp` so no preallocation.

`<input checked="true" checked="@myBool" />`
Here we don't preallocate the second `checked` attribute because the value is `CSharpStatementIRNode`.

ajaybhargavb:
🆙 📅  Handled duplicate attributes. Still in design (Some unit tests failing).

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EnumTagHelpers_DesignTime.codegen.cs
```diff
@@ -40,19 +40,19 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_EnumTagHel
             __TestNamespace_InputTagHelper = CreateTagHelper<global::TestNamespace.InputTagHelper>();
             __TestNamespace_CatchAllTagHelper = CreateTagHelper<global::TestNamespace.CatchAllTagHelper>();
 #line 9 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EnumTagHelpers.cshtml"
-__TestNamespace_InputTagHelper.Value = global::Microsoft.AspNetCore.Razor.Language.IntegrationTests.TestTagHelperDescriptors.MyEnum.MyValue;
+__TestNamespace_InputTagHelper.Value = MyValue;
```
NTaylorMullen: Whappen here?

ajaybhargavb: Yup was just investigating this.

ajaybhargavb:
🆙 📅  Fixed enums

rynowak:
> Looks good. We should have follow ups that do similar things as this PR:

We should have a design meeting to review all of the cases and come up with a comprehensive plan. We have an item tracking this

ajaybhargavb:
🆙 📅 No longer in design. Cleaned up ModelExpressionPass. Worth taking another quick look.

PR #1347, opened 2017-05-15
Remove workaround from test csproj files

natemcmaster:
One more cleanup as a result of #1338 . These properties are no longer necessary now that are using VS Test 15.3.


rynowak:
nice :shipit:

Issue #1351, opened 2017-05-16
Razor.Extension line mapping baselines are not being validated.


NTaylorMullen:
The root cause of this is that the design time tests are calling into the wrong baseline validation method. We should rename the `AssertDesignTimeDocumentMatchBaseline` method to `AssertLineMappingsMatchBaseline` and add separate calls for clarity.

Found this when making changes that should have impacted line mappings for the extensions project.

NTaylorMullen:
19f895b9ad918c1465bfd0065ef573408a514db1

PR #1352, opened 2017-05-16
Validate line mappings for integration tests in Razor.Extensions.

NTaylorMullen:
- Split line mapping validation into its own method.
- Prior to this we were calling the wrong method in our design time code generation integration tests within the Razor.Extensions test project.
- Re-generated line mapping baselines to conform with the cshtml files for Razor.extensions code generation integration tests.

#1351

### DIFF
--- a/test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
+++ b/test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
@@ -487,6 +487,7 @@ public class DivTagHelper : {typeof(TagHelper).FullName}
             // Assert
             AssertIRMatchesBaseline(document.GetIRDocument());
             AssertCSharpDocumentMatchesBaseline(document.GetCSharpDocument());
+            AssertLineMappingsMatchBaseline(document);
             AssertDocumentCompiles(document, compilationReferences, expectedErrors);
         }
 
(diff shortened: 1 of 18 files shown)


test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
```diff
@@ -485,6 +485,7 @@ public void ViewWithNamespace_DesignTime()
             engine.Process(document);
 
             // Assert
+            AssertLineMappingsMatchBaseline(document);
```
rynowak: This should come between `AssertCSharpDocumentMatchesBaseline` and `AssertDocumentCompiles`

The line mappings being different is the *least* helpful thing you could use to diagnose a failure, so it should be later

test/Microsoft.AspNetCore.Razor.Test.Common/Langauge/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -184,9 +184,6 @@ protected void AssertDesignTimeDocumentMatchBaseline(RazorCodeDocument document)
             Assert.NotNull(syntaxTree);
             Assert.True(syntaxTree.Options.DesignTimeMode);
```
rynowak: Do we need these other checks here? Why not just pass in the mappings instead of the code document?

NTaylorMullen: Not really. Saw these guys too. Just figured I'd make the smallest changeset possible; can remove.

NTaylorMullen: Can't pass just the mappings. Need the source as well so we can pull from the generated code and from the cshtml file to compare mappings.

NTaylorMullen:
🆙 📅

Issue #1176, opened 2017-04-04
Design Time CodeGen for DirectiveTokenKind.Type doesn't handle value types


rynowak:
This code in `DesignTimeCSharpRenderer` doesn't handle the case where a type used in the directive is a value type:
```
                    case DirectiveTokenKind.Type:

                        Context.AddLineMappingFor(node);
                        Context.Writer
                            .Write(node.Content)
                            .Write(" ")
                            .WriteStartAssignment(TypeHelper)
                            .WriteLine("null;");
                        break;
```

This will generate code like:
```
int __typeHelper = null;
```

We should use `default(node.Content)` here.

rynowak:
Consider this part of the 15.3 release

NTaylorMullen:
b8ea0081518059e3c7ff4292d30216e08c5bd3aa

PR #1353, opened 2017-05-16
Make extensible directive type token rendering allow value types.

NTaylorMullen:
- Instead of rendering a null assigning statement for a type token we now render a `default(TTypeToken)`. With this approach type tokens can be value types without creating a design time error.
- Re-generated baseline files to reflect new directive token code generation.

#1176


NTaylorMullen:
This was already approved. Just changing base branch (from my other branch to rel/15.3) which seems to reset peoples approvals -_-

PR #1357, opened 2017-05-17
Fixes #1245 - Make TemplateCodeExtension public

### LINKED ISSUE #1245: Consider making target extension interfaces public
Today all of our specific target extension interfaces are internal, i.e. `IPreallocatedAttributeTargetExtension`. We should look over all the extensions we have and determine if we think they should stay internal or become public. This way, we can expose a well-defined code generation extension point for Razor.

rynowak:
This removes the hardcoding of an MVC type from Razor.


NTaylorMullen:
I know in the issue you stated that you didn't want to have template on by default but it feels a bit odd to have templates be part of core Razor but not be fully functional. Could you elaborate as to why you think it shouldn't be on by default?

rynowak:
> Could you elaborate as to why you think it shouldn't be on by default?

Let's be real, the default experience is MVC.

NTaylorMullen:
> Let's be real, the default experience is MVC.

lol if that was a driver why not have core Razor attached to MVC?

PR #1358, opened 2017-05-17
Configure xunit for better display

rynowak:


pranavkm:
What does this do?

rynowak:
It makes the names shorter in the test explorer

rynowak:
Let's try this out for a while and see if we like it

Issue #1269, opened 2017-04-28
@page "{routevalue}" doesn't color properly in the editor


NTaylorMullen:

NTaylorMullen:
03f39753603234a3dd59620ad4e4b6a8c78a854f

PR #1359, opened 2017-05-17
Allow directive string tokens to be colorized.

NTaylorMullen:
### Before
image

### After
image

- Without a directive string token having a `SpanKind.Code` it cannot have any sort of C# coloring associated with it.
- Updated tests to reflect new `SpanKind` expectations.

#1269

Positioned this guy to go into **rel/15.3** @rynowak let me know if you feel like it's not worth.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -1559,7 +1559,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                     return;
                 }
 
-                var outputKind = SpanKind.Markup;
                 switch (tokenDescriptor.Kind)
                 {
                     case DirectiveTokenKind.Type:
@@ -1572,8 +1571,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
                             return;
                         }
-
-                        outputKind = SpanKind.Code;
                         break;
 
                     case DirectiveTokenKind.Namespace:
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -120,7 +120,7 @@ public void DirectiveDescriptor_UnderstandsStringTokens()
                     Factory.CodeTransition(),
                     Factory.MetaCode("custom").Accepts(AcceptedCharacters.None),
                     Factory.Span(SpanKind.Markup, " ", markup: false).Accepts(AcceptedCharacters.WhiteSpace),
```
NTaylorMullen: Left the string prefix as markup. This way when attempting to type a string token you don't get C# IntelliSense for things like `this` etc.

rynowak:
Let's rel/15.3 this when it's ready 👍

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1624,7 +1617,7 @@ private void HandleDirective(DirectiveDescriptor descriptor)
                 }
 
                 Span.ChunkGenerator = new DirectiveTokenChunkGenerator(tokenDescriptor);
-                Output(outputKind, AcceptedCharacters.NonWhiteSpace);
+                Output(SpanKind.Code, AcceptedCharacters.NonWhiteSpace);
```
rynowak: So correct me if I'm wrong, the special case here was that strings were markup and all of the others were code? Seems straightforward.

rynowak:
FYI @alexgav

rynowak:
:shipit:

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
NTaylorMullen: Yup

PR #1360, opened 2017-05-18
Split options into ParserOptions and CodeGenerationOptions

rynowak:


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorParserOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+            Directives = directives;
+            DesignTime = designTime;
+            ParseOnlyLeadingDirectives = parseOnlyLeadingDirectives;
+        }
+
+        public override bool DesignTime { get; }
+
+        public override IReadOnlyCollection<DirectiveDescriptor> Directives { get; }
+
+        public override bool ParseOnlyLeadingDirectives { get; }
```
rynowak: I renamed this option because it's really intended to parse ALL leading directives, not just the first line. I'll love an issue

src/Microsoft.AspNetCore.Razor.Language/DocumentClassifierPassBase.cs
```diff
@@ -19,8 +19,9 @@ public abstract class DocumentClassifierPassBase : RazorIRPassBase, IRazorDocume
         protected override void OnInitialized()
         {
             var feature = Engine.Features.OfType<IRazorTargetExtensionFeature>();
-
             TargetExtensions = feature.FirstOrDefault()?.TargetExtensions.ToArray() ?? EmptyExtensionArray;
+
```
NTaylorMullen: 🤒

Issue #1304, opened 2017-05-11
Move public types used by the editor to Microsoft.VisualStudio.LanguageServices.Razor


rynowak:
- [x] AcceptedCharacters
- [x] BlockKind
- [x] LineMapping
- [ ] SpanKind

rynowak:
@NTaylorMullen - linemapping is used pretty heavily in runtime. I don't think we should try to move it.

rynowak:
5a1090f3240aa50bf1930a6f928d755047ef279e

Issue #1259, opened 2017-04-25
Replace RazorEditorParser


rynowak:
We should modernize the API for `RazorEditorParser`

We need to 'Roslyn-ize' this API and get rid of our janky `ITextBuffer` types and things like that.

Strawman:
```
    public abstract class RazorEditorWorker
    {
        public abstract State Current { get; }

        public abstract Task ParseAsync(TextChange change);

        public abstract class State
        {
            public abstract SourceText Text { get; }

            public abstract RazorSyntaxTree SyntaxTree { get; }

            public abstract RazorCSharpDocument CSharpDocument { get; }
        }
    }
```

rynowak:
For preview2 the work item here is to move this and other `.Legacy` types into the `Microsoft.VisualStudio.LanguageServices.Razor` assembly

rynowak:
5a1090f3240aa50bf1930a6f928d755047ef279e removes this component from 2.0.0. Don't worry! It's still there in the rel/15.3 branch. We'll bring it back in the next foundational update.

NTaylorMullen:
f8d43853f89b7ebbc71d6e4d3b65b00126f768e2

PR #1362, opened 2017-05-18
Api Cleanup for language services

rynowak:
Created internal + public versions of
- BlockKind
- SpanKind
- AcceptedCharacters

That way these types are only exposed through the VS apis and not
through the runtime API surface.

Also deleted RazorEditorParser. Yep. It's going to take significant work
to just port it to the language services assembly. Let's reevaluate this
when we get closer to the next foundational update.


rynowak:
Addresses #1304 and part of #1259

rynowak:
FYI Only

Issue #1323, opened 2017-05-11
Reduce the number of IR nodes related to TagHelpers


rynowak:
Get rid of CreateTagHelperIRNode, InitializeTagHelperFields, ExecuteTagHelperIRNode

ajaybhargavb:
1efcdafa450cdae1478de6695d6ee31957909152

ajaybhargavb:
Related follow up https://github.com/aspnet/Razor/issues/1368

PR #1363, opened 2017-05-18
[Design] Get rid of CreateTagHelperIRNode and ExecuteTagHelperIRNode

ajaybhargavb:
Issue - #1323 

This change removes `CreateTagHelperIRNode` and `ExecuteTagHelperIRNode` as discussed.

Alternative idea (just to get some thoughts):
@NTaylorMullen and I discussed and we think we could make the tag helper execution runtime code gen simpler. Especially the part where we do `tagHelperScopeManager.Begin(...)`. Instead we can have a type like `TagHelperExecutionContextBuilder` with which we can do something like,
```c#
__InputTagHelper = CreateTagHelper<global::InputTagHelper>();
ecBuilder.Add(__InputTagHelper);
ecBuilder.AddHtmlAttribute(__tagHelperAttribute_0);
```
This should help reduce the number of IR nodes even further. We haven't hashed out an exact design yet. But just wanted to check if this is even worth looking into as this will involve a larger change. Not sure if we have scope for this in this milestone.


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
```diff
@@ -32,26 +38,23 @@ public override void WriteAddTagHelperHtmlAttribute(CSharpRenderingContext conte
...
-        {
-            // Do nothing
-        }
-
         public override void WriteInitializeTagHelperStructure(CSharpRenderingContext context, InitializeTagHelperStructureIRNode node)
         {
             context.RenderChildren(node);
+
+            // CreateTagHelper
+            foreach (var descriptor in node.TagHelperBinding.Descriptors)
```
ajaybhargavb: @NTaylorMullen, I tried moving `CreateTaghelper` logic over to `TagHelperIRNode` rendering but that ends up making things ugly,
```C#
public override void WriteTagHelper(CSharpRenderingContext context, TagHelperIRNode node)
{
    var bodyNode = context.Children.First() as BodyIRNode;
    Debug.Assert(bodyNode != null);
    WriteBodyIRNode(context, bodyNode);

    // CreateTagHelper logic

    // Render attributes
    context.Children.Remove(bodyNode); // Ugly
    context.RenderChildren(node);

    // ExecuteTagHelpers logic
}
```

rynowak:
> This should help reduce the number of IR nodes even further. We haven't hashed out an exact design yet. But just wanted to check if this is even worth looking into as this will involve a larger change. Not sure if we have scope for this in this milestone.

Open an issue. I'm not sure that reducing the number of nodes beyond what we discussed is a goal.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
```diff
@@ -32,26 +38,23 @@ public override void WriteAddTagHelperHtmlAttribute(CSharpRenderingContext conte
...
-                    "global::" + node.TagHelperTypeName)
-                .WriteEndMethodInvocation();
-        }
-
-        public override void WriteExecuteTagHelpers(CSharpRenderingContext context, ExecuteTagHelpersIRNode node)
-        {
-            // Do nothing
-        }
-
         public override void WriteInitializeTagHelperStructure(CSharpRenderingContext context, InitializeTagHelperStructureIRNode node)
```
rynowak: why is `InitializeTagHelperStructureIRNode` still here?

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeTagHelperWriter.cs
```diff
@@ -303,6 +218,97 @@ public override void WriteInitializeTagHelperStructure(CSharpRenderingContext co
...
+                context.Writer.WriteInstanceMethodInvocation(
+                    ExecutionContextVariableName,
+                    ExecutionContextAddMethodName,
+                    tagHelperVariableName);
+            }
+        }
+
+        public override void WriteAddTagHelperHtmlAttribute(CSharpRenderingContext context, AddTagHelperHtmlAttributeIRNode node)
+        {
+            var attributeValueStyleParameter = $"global::Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.{node.ValueStyle}";
```
rynowak: This should come from a configurable property somewhere

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
ajaybhargavb: This will be renamed to something like `TagHelperBodyIRNode`. But the node will need to stay. I believe this is what we discussed.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
NTaylorMullen: Agree that what you have to do isn't beautiful but it cleans up the IR surface area. Note: You could always just do `bodyNode.Children.Skip(1)` 👍 

By having all the info on `TagHelperIRNode` you can comfortably rename this node from initialize structure to `TagHelperBodyIRNode`

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
rynowak: Ok cool

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
rynowak: You could also just write a loop.

ajaybhargavb: > Agree that what you have to do isn't beautiful but it cleans up the IR surface area.

Not sure how. This is going to have the same number of nodes as what I have now. I guess I don't get why this approach is better.

NTaylorMullen: Having TagHelperIRNode be the source of truth. Aka, what the binding was, what the tag name, tag mode etc. is makes it easy to understand. With how it is right now: Why would a `TagHelperBodyIRNode` need to have knowledge of the binding?

ajaybhargavb: Got it. Thanks 👍

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InstrumentationPass.cs
```diff
@@ -101,14 +101,11 @@ public override void VisitCSharpExpression(CSharpExpressionIRNode node)
                 VisitDefault(node);
             }
 
-            public override void VisitExecuteTagHelpers(ExecuteTagHelpersIRNode node)
+            public override void VisitTagHelper(TagHelperIRNode node)
```
ajaybhargavb: We now instrument around the entire tag helper instead of just the execute block.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ViewComponentTagHelperPass.cs
```diff
@@ -61,24 +61,25 @@ private void GenerateVCTHClass(ClassDeclarationIRNode @class, TagHelperDescripto
...
             NamespaceDeclarationIRNode @namespace,
             ClassDeclarationIRNode @class,
-            CreateTagHelperIRNode node)
+            TagHelperIRNode node)
         {
-            var originalTypeName = node.TagHelperTypeName;
-
-            var newTypeName = GetVCTHFullName(@namespace, @class, node.Descriptor);
-            for (var i = 0; i < node.Parent.Children.Count; i++)
+            foreach (var descriptor in node.TagHelperTypes.Keys)
```
ajaybhargavb: Since we no longer have a node for create, a `Dictionary<TagHelperDescriptor, string>` was needed so that the type name can be overwritten by VCTH pass.

ajaybhargavb: This doesn't look that elegant but every other solution I can think of needs an IR node for create.

ajaybhargavb:
Closing this. Updated PR https://github.com/aspnet/Razor/pull/1373

Issue #1307, opened 2017-05-11
Remove constants from tag helper descriptor builders and hide behind extension methods


rynowak:

NTaylorMullen:
5aababce6e9d009c9f00172b985efc501649dc32

PR #1365, opened 2017-05-18
Remove magic string keys from TH builders.

NTaylorMullen:
- Replaced the magic strings with extension methods that produce the same behavior as the string keys did.
- Added tests to validate new extension methods.

#1307

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorExtensions.cs
@@ -1,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public static class BoundAttributeDescriptorExtensions
+    {
+        public static string GetPropertyName(this BoundAttributeDescriptor descriptor)
+        {
+            descriptor.Metadata.TryGetValue(TagHelperBoundAttributeDescriptorBuilder.PropertyNameKey, out var propertyName);
+
+            return propertyName;
+        }
+
+        public static bool IsDefaultKind(this BoundAttributeDescriptor descriptor)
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class BoundAttributeDescriptorExtensions
+    {
+        public static string GetPropertyName(this BoundAttributeDescriptor descriptor)
+        {
+            descriptor.Metadata.TryGetValue(TagHelperBoundAttributeDescriptorBuilder.PropertyNameKey, out var propertyName);
+
+            return propertyName;
+        }
+
+        public static bool IsDefault(this BoundAttributeDescriptor descriptor)
```
rynowak: `IsDefaultKind`? I don't like this name

NTaylorMullen: Ya I wasn't sold on `IsDefault` either but couldn't think of something better. Will rename.

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
```diff
@@ -9,8 +9,8 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     public sealed class TagHelperDescriptorBuilder
     {
-        public static readonly string DescriptorKind = "ITagHelper";
-        public static readonly string TypeNameKey = "ITagHelper.TypeName";
+        internal const string DescriptorKind = "ITagHelper";
+        internal const string TypeNameKey = "ITagHelper.TypeName";
 
         private static ICollection<char> InvalidNonWhitespaceAllowedChildCharacters { get; } = new HashSet<char>(
             new[] { '@', '!', '<', '/', '?', '[', '>', ']', '=', '"', '\'', '*' });
```
ajaybhargavb: Not related to your changes. But is there somewhere we can put this collection and share it? Right now this is being duplicated in four different locations.

NTaylorMullen: Will do as a follow up.

PR #1366, opened 2017-05-19
Fix aspnet/Mvc#6296 sanitize class and namespace

rynowak:
The new @namespace directive isn't sanitizing class and namespace names
when generating them. This means that a file-system-legal character like
'-' will show up in a class name, and that's not good.

Note that the old code paths (document classifiers) already had tests for
this and did it properly. It was only missing from @namespace.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ClassName.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ClassName.cs
@@ -4,9 +4,9 @@
 using System.Globalization;
 using System.Text;
 
-namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Internal
+namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
-    public static class ClassName
+    internal static class ClassName
     {
         public static string GetClassNameFromPath(string path)
         {
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Internal
                 category == UnicodeCategory.Format; // Cf
         }
 
-        private static string SanitizeClassName(string inputName)
+        public static string SanitizeClassName(string inputName)
         {
             if (!IsIdentifierStart(inputName[0]) && IsIdentifierPart(inputName[0]))
             {
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
@@ -63,13 +63,14 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                 {
                     // Beautify the class name since we're using a hierarchy for namespaces.
                     var @class = visitor.FirstClass;
+                    var prefix = Path.GetFileNameWithoutExtension(codeDocument.Source.FileName);
                     if (@class != null && irDocument.DocumentKind == RazorPageDocumentClassifierPass.RazorPageDocumentKind)
                     {
-                        @class.Name = Path.GetFileNameWithoutExtension(codeDocument.Source.FileName) + "_Page";
+                        @class.Name = ClassName.SanitizeClassName(prefix + "_Page");
                     }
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -63,13 +63,14 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNo
                 {
                     // Beautify the class name since we're using a hierarchy for namespaces.
                     var @class = visitor.FirstClass;
+                    var prefix = Path.GetFileNameWithoutExtension(codeDocument.Source.FileName);
```
pranavkm: `var classNamePrefix = ClassName.SanitizeClassName(Path.GetFileNameWithoutExtension(codeDocument.Source.FileName));`

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ClassName.cs
```diff
@@ -4,9 +4,9 @@
 using System.Globalization;
 using System.Text;
 
-namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Internal
+namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
-    public static class ClassName
+    internal static class ClassName
```
pranavkm: Maybe call this `CSharpIdentifier`?

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/NamespaceDirectiveTest.cs
```diff
@@ -196,6 +196,44 @@ public void Pass_SetsNamespaceAndClassName_ComputedFromSource()
...
+
+            var @namespace = new NamespaceDeclarationIRNode() { Content = "default" };
+            builder.Push(@namespace);
+
+            var @class = new ClassDeclarationIRNode() { Name = "default" };
+            builder.Add(@class);
+
+            document.DocumentKind = RazorPageDocumentClassifierPass.RazorPageDocumentKind;
+
+            var codeDocument = RazorCodeDocument.Create(RazorSourceDocument.Create("ignored", "/Account/Manage-Info/Add+User.cshtml"));
```
pranavkm: Realized, this will run into issues if any of the segments is a keyword. Consider `Registration/Class/Index.cshtml`. Should we mangle the things further if it's a known keyword?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
rynowak: Why is this better?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ClassName.cs
rynowak: Sure

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/NamespaceDirectiveTest.cs
rynowak: I'm going to open a follow up bug for that. This isn't something we do anywhere right now and will be substantially more complex to fix than this.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
pranavkm: Fewer bytes.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/NamespaceDirectiveTest.cs
rynowak: https://github.com/aspnet/Mvc/issues/6298

pranavkm:
🚢 :it:

PR #1367, opened 2017-05-19
Re-enable some skipped tests

rynowak:
The underlying bug in this area has been fixed.


rynowak:
FYI only, will merge once CIs pass

Issue #1201, opened 2017-04-10
Razor should require directives appear at the start of the line


pranavkm:
Right now, an occurence of `@page` (or any other directive for that matter) anywhere in the page causes it to be treated as a directive. e.g. 

```
@{
   var @using = "crowbar";
}

You cannot use an @using indoors.
```

The generated code looks like:

```C#
WriteLiteral("\r\nHello ");
using
```

ajaybhargavb:
Shouldn't this change be in Mvc that uses the api added in https://github.com/aspnet/Razor/issues/1195?

rynowak:
Both, MVC should use the new API, and this should be a parsing error.

pranavkm:
We should check the devolution parser behavior, but I think it required directives to start at the start of a line. Right now, any token that looks like a directive anywhere in the page is treated as one which would be a breaking behavior change.

NTaylorMullen:
0688cd3ef73ea50999e7a95aa53a12e7e03f3e93
3b53f045186b735d28955064cabd21717b0aee68

NTaylorMullen:
@rynowak didn't see you moved this to preview3 until after merge. Let me know if you want me to back it out

ajaybhargavb:
Looks like we missed re-enabling this test https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/PageDirectiveTest.cs#L48

NTaylorMullen:
So one thing that's different about ^ test is that it expects the page directive to not be there. How we decided to fix this issue was to still let the page directive exist BUT file an error. 

Are we ok with `TryGetPageDirective` returning the page directive if it's in an invalid state? That means when a user access' the page they'll get the logged error.

rynowak:
> Are we ok with TryGetPageDirective returning the page directive if it's in an invalid state? That means when a user access' the page they'll get the logged error.

We should decide how we want this to work at a Razor level, and then allow MVC to live in that world.

Reopening this because there's still a skipped test.

NTaylorMullen:
I think we should detect the page directive, log an error and then be on our merry way (how it exists currently). With that approach a user will be able to see that they made a mistake when accessing the page.

Eilon:
@rynowak please do the needful.

rynowak:
@pranavkm - this is all cleaned up now right?

pranavkm:
[...]

PR #1374, opened 2017-05-19
Log errors if directives do not start at beginning of line.

NTaylorMullen:
- Updated tests to validate expectations.
- Added two additional tests to validate extensible directives and built-in directives get start at line verification.

#1201

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -101,7 +101,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
         {
             foreach (var directive in directives)
             {
-                _directiveParsers.Add(directive, handler);
+                _directiveParsers.Add(directive, () =>
+                {
+                    EnsureDirectiveIsAtStartOfLine();
+                    handler();
+                });
                 Keywords.Add(directive);
 
                 // These C# keywords are reserved for use in directives. It's an error to use them outside of
@@ -1517,6 +1521,30 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             MapDirectives(TagHelperPrefixDirective, SyntaxConstants.CSharp.TagHelperPrefixKeyword);
             MapDirectives(AddTagHelperDirective, SyntaxConstants.CSharp.AddTagHelperKeyword);
             MapDirectives(RemoveTagHelperDirective, SyntaxConstants.CSharp.RemoveTagHelperKeyword);
+        }
+
+        private void EnsureDirectiveIsAtStartOfLine()
+        {
+            // 1 is the offset of the @ transition for the directive.
+            if (CurrentStart.CharacterIndex > 1)
+            {
+                var index = CurrentStart.AbsoluteIndex - 1;
+                var lineStart = CurrentStart.AbsoluteIndex - CurrentStart.CharacterIndex;
+                while (--index >= lineStart)
+                {
+                    var @char = Context.SourceDocument[index];
+
+                    if (!char.IsWhiteSpace(@char))
+                    {
+                        var currentDirective = CurrentSymbol.Content;
+                        Context.ErrorSink.OnError(
+                            CurrentStart,
+                            Resources.FormatDirectiveMustAppearAtStartOfLine(currentDirective),
+                            length: currentDirective.Length);
+                        break;
+                    }
+                }
+            }
         }
 
         private void HandleDirective(DirectiveDescriptor descriptor)
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
@@ -11,6 +11,104 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
     public class CSharpDirectivesTest : CsHtmlCodeParserTestBase
     {
         [Fact]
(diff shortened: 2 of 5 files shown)


NTaylorMullen:
Ya we do. That was the initial goal of the fix. That way there's more structure to directives which enable us and users to understand where they can occur. A great example of this is the `@page` diretive; for a while we were hand parsing everything and having the white space in front of it didn't do anyone any favors.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1519,6 +1523,19 @@ private void SetupDirectives(IEnumerable<DirectiveDescriptor> directiveDescripto
...
 
+        private void EnsureDirectiveIsAtStartOfLine()
+        {
+            // 1 is the offset of the @ transition for the directive.
+            if (CurrentStart.CharacterIndex > 1)
+            {
+                var currentDirective = CurrentSymbol.Content;
+                Context.ErrorSink.OnError(
+                    CurrentStart,
+                    Resources.FormatDirectiveMustAppearAtStartOfLine(currentDirective),
```
NTaylorMullen: Figured this was new code so the error could live in `Resources.resx`. Other option would be in `LegacyResources.resx`. Not sure how we're choosing to make those decisions.

ajaybhargavb:
Not sure about this one. If we have some custom directive that makes sense only in a scope, it would suck to not allow it to be indented. Or if there is no such thing as a scoped directive, should we probably restrict all directives to be at the start of the file?

NTaylorMullen:
> Not sure about this one. If we have some custom directive that makes sense only in a scope, it would suck to not allow it to be indented. Or if there is no such thing as a scoped directive, should we probably restrict all directives to be at the start of the file?

@rynowak thoughts?

rynowak:
> Not sure about this one. If we have some custom directive that makes sense only in a scope, it would suck to not allow it to be indented.

We should not be whitespace sensitive

> Or if there is no such thing as a scoped directive, should we probably restrict all directives to be at the start of the file?

Yeah we need a feature like this. I think there are three parts:
- file scoped - one per file, has overriding behavior implemented by lowering (like `@model`)
- multiple - top-level scoped - appears in 'leading' directives, but multiple are allowed (like `@inject`)
- inline - allowed anywhere (like `@section`)

NTaylorMullen:
- [Enable file scoped extensible directives](https://github.com/aspnet/Razor/issues/1376)
- [Enable multiple occurring extensible directives](https://github.com/aspnet/Razor/issues/1377)
- [Enable inline extensible directives](https://github.com/aspnet/Razor/issues/1378)

NTaylorMullen:
🆙 📅  and rebased this guy ontop of nimullen/parsercontext

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -11,6 +11,104 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
     public class CSharpDirectivesTest : CsHtmlCodeParserTestBase
     {
         [Fact]
+        public void ExtensibleDirectiveDoesNotErorrIfNotAtStartOfLineBecauseOfWhitespace()
```
rynowak: Erorr

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -11,6 +11,104 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
...
+                    Factory.Code(Environment.NewLine + "  ").AsStatement(),
+                    Factory.CodeTransition(),
+                    Factory.MetaCode(SyntaxConstants.CSharp.AddTagHelperKeyword + " ")
+                            .Accepts(AcceptedCharactersInternal.None),
+                    Factory.Code("\"*, Foo\"")
+                        .AsAddTagHelper("\"*, Foo\"")));
+        }
+
+        [Fact]
+        public void BuiltInDirectiveErorrsIfNotAtStartOfLine()
```
rynowak: Is this a breaking change? We don't want to just make arbitrary breaking changes without agreement. This will affect users of 1.X as well.

rynowak: I'm not super concerned about this change in particular, but let's discuss with what's his face.

NTaylorMullen: Yes, this is a breaking change that will impact 1.X. If you want to toggle this on/off I could move the error logging into the IR phase to allow for multi-targeting in the future. However, I feel that it'd be quite rare to have directives not on their own line (I can't think of a case where it'd even start to make sense).

NTaylorMullen:
ping.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
rynowak: Please discuss this with @DamianEdwards and see if he's got concerns

NTaylorMullen:
@DamianEdwards signed off on the breaking change. @rynowak / @muratg this ok to get in for preview2?

muratg:
@NTaylorMullen Yup, if @DamianEdwards signed off on this, it's OK to check in. Please make sure to tag the original bug with `breaking-change`.

DamianEdwards:
@NTaylorMullen did you create a breaking change announcement?

NTaylorMullen:
Did not, investigating a VS crash issue then I'll create one.

PR #1382, opened 2017-05-23
Change ParserContext to take a RazorSourceDocument.

NTaylorMullen:
- This moves ParserContext closer to operating on a RazorSourceDocument and exposes it at the parsing layer.
- Was not able to replace the `ITextDocument` property on `ParserContext` due to its current wiring. Our tokenizers rely on a single reader that iterates over the document and take turns tokenizing characters from that reader. The reader that the tokenizers pull from is also highly coupled with the parsers implementations; they end up moving the readers pointer frequently.

I recommend looking over this pr with `?w=1`.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
@@ -9,14 +9,18 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal partial class ParserContext
     {
-        public ParserContext(ITextDocument source, RazorParserOptions options)
+        public ParserContext(RazorSourceDocument source, RazorParserOptions options)
         {
             if (source == null)
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Source = source;
+            SourceDocument = source;
+            var chars = new char[source.Length];
+            source.CopyTo(0, chars, 0, source.Length);
+
+            Source = new SeekableTextReader(chars, source.FileName);
             DesignTimeMode = options.DesignTime;
             ParseOnlyLeadingDirectives = options.ParseOnlyLeadingDirectives;
             Builder = new SyntaxTreeBuilder();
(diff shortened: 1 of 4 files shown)


NTaylorMullen:
This guy enables me to complete #1201.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
```diff
@@ -9,14 +9,18 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
...
             {
                 throw new ArgumentNullException(nameof(source));
             }
 
-            Source = source;
+            SourceDocument = source;
+            var chars = new char[source.Length];
+            source.CopyTo(0, chars, 0, source.Length);
+
+            Source = new SeekableTextReader(chars, source.FileName);
```
ajaybhargavb: I realize that this was already the case. But should we be disposing this reader at some point?

NTaylorMullen: Before I published this PR I removed the TextReader base type from SeekableTextReader because it technically adds nothing today. Reverted those changes because I didn't want to pollute this PR, but you're right, if it's a TextReader this should be disposed.

Thoughts on me nuking the TextReader basetype?

ajaybhargavb: Discussed offline. Leaving it as is for now so it can be cleaned up later.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
```diff
@@ -9,14 +9,18 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal partial class ParserContext
```
rynowak: Does this need to be partial?

NTaylorMullen: It's partial for debug helpers.

Issue #1251, opened 2017-04-20
ViewComponentTagHelper tooltips are ugly.


NTaylorMullen:
When looking at a tooltip for a ViewComponentTagHelper you see the generated TagHelpers name:
![image](https://cloud.githubusercontent.com/assets/2008729/25254059/7ff67258-25d8-11e7-87ac-f375197ddc54.png)

Instead we should come up with something nice for the view component taghelper tooltip.

rynowak:
Is this tooltip coming from WTE or from the C# editor?

NTaylorMullen:
> Is this tooltip coming from WTE or from the C# editor?

WTE consumes our `DisplayName` property on our `TagHelperDescriptor`

rynowak:
Ok cool, this is easy to fix then.

rynowak:
@ryanbrandenburg work with @DamianEdwards to figure out the right 'display name' for a VCTH

NTaylorMullen:
So turns out we'll need editor fixes here. The editor is currently consuming `descriptor.Name` instead of `descriptor.DisplayName` everywhere. I'm going to fix this issue on our side and then file an external issue for the editor side.

NTaylorMullen:
4c98b7f8f3a55648267aca5c727a1c017c7467f8

@alexgav You guys will need to move from `descriptor.Name` to `descriptor.DisplayName` in order to consume these changes.

PR #1384, opened 2017-05-23
Make ViewComponentTagHelper's bound attribute display names nicer.

NTaylorMullen:
- Went from `typeName __Generated__SomeViewComponentTagHelper.PropertyName` to `typeName SomeViewComponentTagHelper.PropertyName`.
- Updated `TagHelperBoundDescriptorBuilder` to allow setting of `DisplayName`.
- Added `TagHelperBoundAttributeDescriptorBuilderTest` class to verify new `DisplayName` additions.
- Updated `ViewComponentTagHelperDescriptorFactoryTest` expectations.

#1251 

Just realized this piece was missed. Right now I have this posed against rel/15.3, let me know if you want to delay the addition of this.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorBuilder.cs
@@ -31,6 +31,7 @@ namespace Microsoft.AspNetCore.Razor.Language
             [typeof(decimal).FullName] = "decimal",
         };
 
+        private string _displayName;
         private bool _isEnum;
         private bool _hasIndexer;
         private string _indexerValueTypeName;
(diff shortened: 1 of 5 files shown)


rynowak:
We don't have approval to put this in 15.3 so just fix it in dev for now. We can consider this for 15.4

rynowak:
Please open another issue to track fixing this for 15.4

src/Microsoft.AspNetCore.Razor.Language/TagHelperBoundAttributeDescriptorBuilder.cs
```diff
@@ -125,12 +138,16 @@ public BoundAttributeDescriptor Build()
...
-            if (!PrimitiveDisplayTypeNameLookups.TryGetValue(_typeName, out var simpleName))
+            if (_displayName == null)
             {
-                simpleName = _typeName;
+                if (!PrimitiveDisplayTypeNameLookups.TryGetValue(_typeName, out var simpleName))
+                {
+                    simpleName = _typeName;
+                }
+
+                _displayName = $"{simpleName} {_containingTypeName}.{_propertyName}";
```
rynowak: This should not dirty state.

Issue #1387, opened 2017-05-25
Can't debug into @functions code


NTaylorMullen:
_From @danroth27 on May 22, 2017 23:23_

I am unable to debug into my `@functions` code in my razor page:

![image](https://cloud.githubusercontent.com/assets/1874516/26332338/1db0a0ea-3f24-11e7-85df-60db817086ec.png)

This is with VS2017 Update3 preview1

_Copied from original issue: aspnet/Mvc#6311_

NTaylorMullen:
Was able to reproduce this as well.

NTaylorMullen:
_From @DamianEdwards on May 22, 2017 23:38_

@NTaylorMullen can you take a look at this one please? Balance as necessary, but we'll need to figure this out of course.

NTaylorMullen:
dcccea300488541adb2f3d165e8e186bb5c81bc3

PR #1388, opened 2017-05-25
Add Checksum computation to RazorSourceDocument.

NTaylorMullen:
This PR enables Razor views to be debuggable. Prior to this checksums were being generated incorrectly because cshtml files were being stringified prior to computing checksums. This resulted in checksum values that did not account for BOMs.

- Renamed many of our `RazorSourceDocument` abstractions to not include the word `Razor`.
- Added a `GetChecksum()` method to `RazorSourceDocument` to allow source documents to compute their own checksums.
- Re-generated codegen tests that did not normalize new lines. Ones that did re-generate newlines converted from stream => string => normalized string and then ran the Razor parser.
- Added tests to validate `GetChecksum` for all source document types.
- Removed unused `LegacySourceDocument`.

#1387


src/Microsoft.AspNetCore.Razor.Language/LargeTextSourceDocument.cs
```diff
@@ -101,6 +103,27 @@ public override void CopyTo(int sourceIndex, char[] destination, int destination
...
+                var bytes = new byte[byteCount];
+                encoder.GetBytes(charBuffer, 0, charBuffer.Length, bytes, 0, flush: true);
+
+                using (var hashAlgorithm = SHA1.Create())
+                {
+                    _checksum = hashAlgorithm.ComputeHash(bytes);
+                }
+            }
+
+            return _checksum;
```
rynowak: `byte[]` is mutable. Either don't cache, or make a copy each time you return it.

src/Microsoft.AspNetCore.Razor.Language/LargeTextSourceDocument.cs
```diff
@@ -17,8 +18,9 @@ internal class LargeTextRazorSourceDocument : RazorSourceDocument
         private readonly RazorSourceLineCollection _lines;
 
         private readonly int _length;
+        private byte[] _checksum;
 
-        public LargeTextRazorSourceDocument(StreamReader reader, int chunkMaxLength, Encoding encoding, string fileName)
+        public LargeTextSourceDocument(StreamReader reader, int chunkMaxLength, Encoding encoding, string fileName)
```
rynowak: Does this actually work if we have a streamreader? Does the streamreader normalize the BOM?

src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
```diff
@@ -52,6 +52,12 @@ public abstract class RazorSourceDocument
         public abstract void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count);
 
         /// <summary>
+        /// Claculates the checksum for the <see cref="RazorSourceDocument"/>.
```
rynowak: calculates

src/Microsoft.AspNetCore.Razor.Language/StreamSourceDocument.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        public StreamSourceDocument(Stream stream, Encoding encoding, string fileName)
+        {
+            if (stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _streamBytes = ReadStreamBytes(stream);
+            _innerSourceDocument = CreateInnerSourceDocument(stream, encoding, fileName);
```
rynowak: Rather than caching the bytes, should we just compute the hash as we read them?

rynowak: It just seems a little silly to keep the bytes around unless we have to.

test/Microsoft.AspNetCore.Razor.Language.Test/LargeTextSourceDocumentTest.cs
```diff
@@ -6,10 +6,27 @@
...
+            var stream = TestRazorSourceDocument.CreateStreamContent(contentString);
+            var reader = new StreamReader(stream, detectEncodingFromByteOrderMarks: true);
+            var document = new LargeTextSourceDocument(reader, 5, Encoding.UTF8, "file.cshtml");
+            var expectedChecksum = new byte[] { 10, 77, 85, 168, 215, 120, 229, 2, 47, 171, 112, 25, 119, 197, 216, 64, 187, 196, 134, 208 };
+
+            // Act
+            var checksum = document.GetChecksum();
+
+            // Assert
+            Assert.Equal(expectedChecksum, checksum);
```
rynowak: Don't we need to test cases with and without BOM? What about other encodings?

src/Microsoft.AspNetCore.Razor.Language/StreamSourceDocument.cs
rynowak: Note btw that Roslyn specifically passes the encoding so that the exact bytes are round-trippable from the chars. https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Text/SourceText.cs#L82

test/Microsoft.AspNetCore.Razor.Language.Test/LargeTextSourceDocumentTest.cs
NTaylorMullen: > Don't we need to test cases with and without BOM?

All of our tests currently verify without BOM. The tests in this PR are verifying with BOM.

> What about other encodings?

You're right, we should be testing that checksum can handle different encodings. Today we have tests that validate the content is created properly based on encodings but not the checksum itself.

src/Microsoft.AspNetCore.Razor.Language/LargeTextSourceDocument.cs
NTaylorMullen: Hm? We computed the stream bytes beforehand which means the stream reader can do whatever it wants. The `StreamSourceDocument` is the one that generates the proper checksum because its the one that needs the BOM bytes, not the `LargeTextSourceDocument`.

I may be misunderstanding your concern.

src/Microsoft.AspNetCore.Razor.Language/StreamSourceDocument.cs
NTaylorMullen: > Rather than caching the bytes, should we just compute the hash as we read them?

Thought about doing this but saw it wouldn't change the `LargeTextSourceDocument` case (it needs to do a pre-pass over the stream to determine bytes). I'll actually code it out and see what it looks like.

> It just seems a little silly to keep the bytes around unless we have to.

Was trying to be lazy about the checksum calculation; with how the world currently works we always evaluate it anyways so I agree with you.

> Note btw that Roslyn specifically passes the encoding so that the exact bytes are round-trippable from the chars.

Hmmm, does getting the character representation of the BOM not result in issues when round tripping?

NTaylorMullen: > I'll actually code it out and see what it looks like.

Ya, the code ends up exploding in complexity. We basically end up having to re-implement what `StreamReader` does for us (buffer limiting, encoding detection, stream resetting).

NTaylorMullen: > Was trying to be lazy about the checksum calculation; with how the world currently works we always evaluate it anyways so I agree with you.

Going to leave the `StringSourceDocument` and `LargeTextSourceDocument` lazy since they're wrapped by `StreamSourceDocument` (their checksums are never calculated).

NTaylorMullen:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/StreamSourceDocument.cs
```diff
@@ -0,0 +1,111 @@
...
+        private readonly byte[] _checksum;
+
+        public StreamSourceDocument(Stream stream, Encoding encoding, string fileName)
+        {
+            if (stream == null)
+            {
+                throw new ArgumentNullException(nameof(stream));
+            }
+
+            _checksum = ComputeChecksum(stream);
```
ajaybhargavb: Why is this being computed in the constructor instead of in `GetChecksum()`?

NTaylorMullen: So we don't have to hang onto the underlying bytes of the stream. There's a lot of them 😉

src/Microsoft.AspNetCore.Razor.Language/LargeTextSourceDocument.cs
rynowak: I see, so this will never be exposed.

Issue #1393, opened 2017-05-30
@namespace directive crashes VS if at end of file with an invalid namespace name.


NTaylorMullen:
The invalid type name that gets picked up here is `SomeName.` (note the dot at the end)

```
Application: devenv.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException
   at Microsoft.AspNetCore.Razor.Language.Legacy.CSharpCodeParser.HandleDirective(Microsoft.AspNetCore.Razor.Language.DirectiveDescriptor)
   at Microsoft.AspNetCore.Razor.Language.Legacy.CSharpCodeParser+<>c__DisplayClass93_0.<SetupDirectives>b__0()
   at Microsoft.AspNetCore.Razor.Language.Legacy.CSharpCodeParser.AfterTransition()
   at Microsoft.AspNetCore.Razor.Language.Legacy.CSharpCodeParser.AtTransition(Microsoft.AspNetCore.Razor.Language.Legacy.CSharpSymbol)
   at Microsoft.AspNetCore.Razor.Language.Legacy.CSharpCodeParser.ParseBlock()
   at Microsoft.AspNetCore.Razor.Language.Legacy.HtmlMarkupParser.OtherParserBlock()
   at Microsoft.AspNetCore.Razor.Language.Legacy.HtmlMarkupParser.SkipToAndParseCode(System.Func`2<Microsoft.AspNetCore.Razor.Language.Legacy.HtmlSymbol,Boolean>)
   at Microsoft.AspNetCore.Razor.Language.Legacy.HtmlMarkupParser.SkipToAndParseCode(Microsoft.AspNetCore.Razor.Language.Legacy.HtmlSymbolType)
   at Microsoft.AspNetCore.Razor.Language.Legacy.HtmlMarkupParser.ParseDocument()
   at Microsoft.AspNetCore.Razor.Language.Legacy.RazorParser.Parse(Microsoft.AspNetCore.Razor.Language.RazorSourceDocument)
   at Microsoft.AspNetCore.Razor.Language.RazorSyntaxTree.Parse(Microsoft.AspNetCore.Razor.Language.RazorSourceDocument, Microsoft.AspNetCore.Razor.Language.RazorParserOptions)
   at Microsoft.AspNetCore.Razor.Language.DefaultRazorParsingPhase.ExecuteCore(Microsoft.AspNetCore.Razor.Language.RazorCodeDocument)
[...]

PR #1396, opened 2017-06-01
Make namespace tokens tolerant to EOF and invalid states.

NTaylorMullen:
- Updated our `QualifiedIdentifier` to properly parse invalid namespaces. This is more consistent with how the rest of the system works; we consume tokens until we determine if we're in an invalid or valid state. If invalid, we log an error and put back all invalid tokens for the parser to treat as non-directive tokens.
- Added unit tests to validate our extensible directive system can handle malformed namespace tokens at EOF and inline.
- Added a code gen test for Razor.Extensions to prove we've fixed the underlying issue for our `@namespace` directive that crashed VS.

Make invalid namespace tokens support IntelliSense. - This was needed because invalid namespace tokens end up being de-classified as code.
- Added a `DirectiveTokenEditHandler` to enable IntelliSense for invalid namespace tokens.
- Added tests to verify new `DirectiveTokenEditHandler`.
- Updated test expectations for `DirectiveTokenEditHandler`.

#1393 

Verified this by running our RazorExtension bits and opening up Razor files.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DirectiveTokenEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DirectiveTokenEditHandler.cs
@@ -1,0 +1,49 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class DirectiveTokenEditHandler : SpanEditHandler
+    {
+        public DirectiveTokenEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer) : base(tokenizer)
+        {
+        }
+
+        protected override PartialParseResult CanAcceptChange(Span target, SourceChange change)
+        {
+            if (AcceptedCharacters == AcceptedCharacters.NonWhiteSpace)
+            {
+                var originalText = change.GetOriginalText(target);
+                var editedContent = change.GetEditedContent(target);
+
+                if (!ContainsWhitespace(originalText) && !ContainsWhitespace(editedContent))
+                {
+                    // Did not modify whitespace, directive format should be the same.
+                    // Return provisional so extensible IR/code gen pieces can see the full directive text
+                    // once the user stops editing the document.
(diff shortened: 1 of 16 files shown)


src/Microsoft.AspNetCore.Razor.Language/DirectiveTokenEditHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        public DirectiveTokenEditHandler(Func<string, IEnumerable<ISymbol>> tokenizer) : base(tokenizer)
+        {
+        }
+
+        protected override PartialParseResult CanAcceptChange(Span target, SourceChange change)
+        {
+            if (AcceptedCharacters == AcceptedCharacters.NonWhiteSpace)
+            {
+                var currentWhitespace = CountWhitespace(target.Content);
```
rynowak: Would it be better to check if there are **any** whitespace characters in the edited region before and after?

This counting business seems pretty fuzzy

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1574,12 +1606,12 @@ private void HandleDirective(DirectiveDescriptor descriptor)
...
 
                     case DirectiveTokenKind.Namespace:
-                        if (!QualifiedIdentifier())
+                        if (!QualifiedIdentifier(out var identifierLength))
                         {
                             Context.ErrorSink.OnError(
                                 CurrentStart,
                                 LegacyResources.FormatDirectiveExpectsNamespace(descriptor.Name),
-                                CurrentSymbol.Content.Length);
+                                identifierLength);
```
rynowak: This is a really strange way to do things, but OK. We'll revisit when we do error recovery.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
```diff
@@ -236,6 +246,17 @@ public void ViewWithNamespace_Runtime()
...
         #region DesignTime
         [Fact]
+        public void InvalidNamespaceAtEOF_DesignTime()
+        {
+            var references = CreateCompilationReferences(CurrentMvcShim);
+            RunDesignTimeTest(references,
+            expectedErrors: new[]
+            {
+                "Identifier expected"
+            });
```
rynowak: formatting

Issue #1394, opened 2017-05-31
Razor formatting issues


anurse:
Repro steps:

1. Start with the following Razor Page in Visual Studio:

```html
@page "{groupName}"
@model IndexModel
@{
    ViewData["PageTitlePrefix"] = Model.GroupName;
    ViewData["CurrentGroup"] = Model.GroupName;
}

<div class="row">
    <div class="col-md-8">
        <ul class="nav nav-pills" id="app-tabs">
            <li class="active"><a href="#group-by-assignee">Group by assignee</a></li>
            <li class=""><a href="#group-by-milestone">Group by milestone</a></li>
            <li class=""><a href="#group-by-repo">Group by repo</a></li>
        </ul>
    </div>
</div>

<div class="row">
    <div class="col-md-10 col-md-pull-2">
        <div class="tab-content">
            <div class="tab-pane active" id="group-by-assignee">
                @foreach(var person in Model.People)
                {
                }
            </div>
        </div>
    </div>
</div>
```

2. Press Ctrl-K Ctrl-D

Expected Results: Reformatted according to my indentation rules (which should mean it is not modified, as this document is already formatted correctly)

Actual Results:

```html
@page "{groupName}"
    @model IndexModel
        @{
            ViewData["PageTitlePrefix"] = Model.GroupName;
            ViewData["CurrentGroup"] = Model.GroupName;
        }

        <div class="row">
            <div class="col-md-8">
                <ul class="nav nav-pills" id="app-tabs">
                    <li class="active"><a href="#group-by-assignee">Group by assignee</a></li>
                    <li class=""><a href="#group-by-milestone">Group by milestone</a></li>
                    <li class=""><a href="#group-by-repo">Group by repo</a></li>
                </ul>
            </div>
        </div>

        <div class="row">
            <div class="col-md-10 col-md-pull-2">
                <div class="tab-content">
                    <div class="tab-pane active" id="group-by-assignee">
                        @foreach (var person in Model.People)
                        {
                        }
                    </div>
                </div>
            </div>
        </div>
```

:(

/cc @NTaylorMullen

anurse:
Note, removing the `@page` **and** `@model` directives at the top "fixes" the problem and the indentation goes back to normal

NTaylorMullen:
Going to have @ajaybhargavb look at this for Friday so I can investigate desktop issues with MVC/MvcPrecompilation.

anurse:
[...]

PR #1398, opened 2017-06-01
Fix unexpected indentation when autoformatting in VS

ajaybhargavb:
Issue - #1394 

@NTaylorMullen @rynowak @anurse 

We were incorrectly outputting the whitespace and newline at the end of a single line directive as markup which caused VS to indent everything below the directive. We should instead output it as MetaCode.


ajaybhargavb:
I'll merge this once @rynowak looks at this.

rynowak:
👏  :shipit:

PR #1403, opened 2017-06-04
Fixes #1296 - ItemCollection should be ICollection

### LINKED ISSUE #1296: ItemCollection should inherit ICollection<KeyValuePair<object, object>>

rynowak:
A little bit more kindness for consumers.

I didn't add any tests because these are all just trivial forwards.


rynowak:
Ah right

rynowak:
@ajaybhargavb updated

PR #1406, opened 2017-06-06
Add DisplayName and Description for directives

rynowak:
Renamed Name -> Directive so that it doesn't overlap with DisplayName.


src/Microsoft.AspNetCore.Razor.Language/DirectiveDescriptor.cs
```diff
@@ -7,124 +7,202 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
+    /// <summary>
+    /// A descriptor type for a directive that can be parsed by the Razor parser.
+    /// </summary>
     public abstract class DirectiveDescriptor
```
rynowak: yeah docs

PR #1409, opened 2017-06-07
Fix #1407 null refs in VCTH tests

### LINKED ISSUE #1407: Flakey test CreateDescriptor_AddsDiagnostic_ForViewComponentWithNoInvokeMethod
This seems to be occasionally null refing on the CI. From a brief look I don't know how that's possible; worth spending more time to investigate.

rynowak:
Dictionary isn't thread safe YO.


rynowak:
FYI only

PR #1410, opened 2017-06-07
React to downstream upgrades to .NET Standard 2.0

natemcmaster:
No changes to src/, just one test project that uses a package that will be lifted to .NET Standard 2.0

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Runtime.Test/Microsoft.AspNetCore.Razor.Runtime.Test.csproj
+++ b/test/Microsoft.AspNetCore.Razor.Runtime.Test/Microsoft.AspNetCore.Razor.Runtime.Test.csproj
@@ -3,7 +3,7 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Razor.Runtime.Test/Microsoft.AspNetCore.Razor.Runtime.Test.csproj
```diff
@@ -29,4 +29,8 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitVersion)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)'=='net461'">
```
JunTaoLuo: Should this be in common.props? It's only one test project so it's fine for now but maybe there will be more in the future?

natemcmaster: Was trying to keep the change as scoped as possible. We're still discussing internally if Razor should move to ns2.0. If it does, then this will move to common.props.

PR #1415, opened 2017-06-08
Target .NET Standard 2.0

natemcmaster:
Changes

 - lifts from netstandard1.3 to netstandard2.0
 - added net46 where required to ensure we don't need netstandard.dll in the VSIX
 - updated corefx version (only affects one package in test, System.Diagonstics.DiagnosticsSource)
 - float NS.Library correctly to avoid package downgrade warnings


build/common.props
```diff
@@ -18,7 +18,11 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
-    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
+    <PackageReference Include="NETStandard.Library" Version="$(NETStandardImplicitPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" PrivateAssets="All" />
```
rynowak: Is this still going to add the `netstandard.dll` to projects like `Microsoft.AspNetCore.Razor.Language`? I might be confused, can you explain?

test/Microsoft.AspNetCore.Razor.Test.MvcShim/Microsoft.AspNetCore.Razor.Test.MvcShim.csproj
```diff
@@ -3,13 +3,13 @@
   <Import Project="..\..\build\common.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net46;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
```
rynowak: Why the change here? I'd prefer to keep these in sync with the src projects, but it's not a huge deal if there's a good reason why we need to lift it.

rynowak:
See https://github.com/aspnet/Razor/blob/dev/tooling/Microsoft.VisualStudio.RazorExtension/Microsoft.VisualStudio.RazorExtension.csproj#L252

I think you're probably ok since all of these still build as net46.

rynowak:
The other thing to check is, if you build the VSIX and then unzip it, are they any extraneous assemblies in there. This is really easy to regress.

build/common.props
natemcmaster: Technically it shouldn't, but to be safe, I'll move this package ref out of common and add it only to csproj's were we are sure it's ok.

test/Microsoft.AspNetCore.Razor.Test.MvcShim/Microsoft.AspNetCore.Razor.Test.MvcShim.csproj
natemcmaster: This test project depends on Razor.Runtime, which is ns2.0 only so I had to lift to net461 to make it compatible.

natemcmaster:
VSIX contents look ok. Plus, I verified manually that none of these assemblies were compiled with a reference to netstandard.dll.

image

test/Microsoft.AspNetCore.Razor.Test.MvcShim/Microsoft.AspNetCore.Razor.Test.MvcShim.csproj
rynowak: Ah right! 👍

PR #1418, opened 2017-06-08
Fix #1355 Make extension nodes check for their extensions

### LINKED ISSUE #1355: Add error checking for target extensions
Right now we don't check for the absence of target extensions when we render an extension node. This means that we could null-ref instead of reporting an error.

rynowak:


rynowak:
@NTaylorMullen

Issue #900, opened 2016-12-15
ValueStyle of tag helper bound attributes with literal values is not set correctly


ajaybhargavb:
Example `<input bound=foo />`
When this block is rewritten by `TagHelperParseTreeRewriter`, it sets the ValueStyle of `bound` attribute to `DoubleQuotes`. Since this is just a literal value it should be set to `NoQuotes`.

ajaybhargavb:
27b73d737bff33b0d62d1612862a8d1fccecc9c5

PR #1419, opened 2017-06-08
Set correct value style for literal tag helper bound attributes

ajaybhargavb:
Issue - #900 

@NTaylorMullen @rynowak 

We were treating all `NoQuotes` as `DoubleQuotes` to make dynamic values work. We shouldn't be doing that for simple literal values.


rynowak:
:shipit:

PR #1420, opened 2017-06-09
Add collection types for IR children

rynowak:
/cc @NTaylorMullen


src/Microsoft.AspNetCore.Razor.Language/Intermediate/ReadOnlyIRNodeCollection.cs
```diff
@@ -0,0 +1,133 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language.Intermediate
+{
+    public sealed class ReadOnlyIRNodeCollection : RazorIRNodeCollection
```
NTaylorMullen: ReadOnly is kind of misleading here. It's more like `EmptyIRNodeCollection`

src/Microsoft.AspNetCore.Razor.Language/Intermediate/DefaultIRNodeCollection.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public override bool Contains(RazorIRNode item)
+        {
+            return _inner.Contains(item);
+        }
+
+        public override void CopyTo(RazorIRNode[] array, int arrayIndex)
+        {
+            if (array == null)
```
NTaylorMullen: Does the inner not do all these checks?

src/Microsoft.AspNetCore.Razor.Language/Intermediate/DefaultIRNodeCollection.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+
+            return _inner.IndexOf(item);
+        }
+
+        public override void Insert(int index, RazorIRNode item)
+        {
+            if (index < 0 || index > Count)
```
NTaylorMullen: ditto and elsewhere.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/DefaultIRNodeCollection.cs
rynowak: I'll get the opposite feedback from the next person to look at it.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/ReadOnlyIRNodeCollection.cs
rynowak: I disagree. It's whole purpose is to be read only. It happens to also be empty.

NTaylorMullen:
This be good.

PR #1421, opened 2017-06-09
Remove Parent from IR nodes

rynowak:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InstrumentationPass.cs
```diff
@@ -45,33 +42,33 @@ private static void AddInstrumentation(InstrumentationItem item)
...
-            item.Node.Parent.Children.Insert(nodeIndex + 2, endNode);
+            var nodeIndex = item.Parent.Children.IndexOf(item.Node);
+            item.Parent.Children.Insert(nodeIndex, beginNode);
+            item.Parent.Children.Insert(nodeIndex + 2, endNode);
         }
 
         private struct InstrumentationItem
         {
-            public InstrumentationItem(RazorIRNode node, bool isLiteral, SourceSpan source)
+            public InstrumentationItem(RazorIRNode node, RazorIRNode parent, bool isLiteral, SourceSpan source)
```
rynowak: This is the new basic pattern for rewriters - grab the parent and the node during the visiting phase, and then act on it later.

src/Microsoft.AspNetCore.Razor.Language/DefaultDirectiveIRPass.cs
```diff
@@ -20,20 +20,19 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNo
             walker.VisitDocument(irDocument);
 
             var classNode = walker.ClassNode;
-            foreach (var node in walker.FunctionsDirectiveNodes)
+            foreach (var (node, parent) in walker.FunctionsDirectiveNodes)
```
rynowak: This class shows that same basic pattern but with valuetuples

rynowak: I've actually got a super cool idea about this, I'll send another PR

NTaylorMullen: Generic types instead of tuples?

rynowak: See my other PR

rynowak:
> Only concern I have is of the repeated value tuple patterns everywhere. Leads me to think there's a better alternative.

See my next PR in this series

rynowak:
I'm going to go ahead and merge this, but leave https://github.com/aspnet/Razor/pull/1420 in case we decide to address your crazy feedback from that one.

PR #1427, opened 2017-06-12
Remove Roslyn from NuGet.config.

pranavkm:
Update StreamRpcJson to 1.0.36. This package has an assembly version that matches the version shipped in VS 2017 RTM.


rynowak:
:shipit:

PR #1429, opened 2017-06-12
Split the functions/section/inherits directives

rynowak:


src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -47,15 +47,13 @@ public override bool Equals(object obj)
         {
             var other = obj as ImplicitExpressionEditHandler;
             return base.Equals(other) &&
-                _keywords.SetEquals(other._keywords) &&
```
rynowak: @NTaylorMullen lets talk about this. I think there are some lurking issues

src/Microsoft.AspNetCore.Razor.Language/Extensions/FunctionsDirectivePass.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    public class FunctionsDirectivePass : RazorIRPassBase, IRazorDirectiveClassifierPass
```
NTaylorMullen: We should pick an approach of static XYZDirective class with a static register method and a nested private class (that's what `InjectDirective` does as an example) or what you have in this PR, static class, register, the directive class and a public pass. 

I worry about exposing the Functions pass here. Is there value in letting users see this? Same questions for inject.

src/Microsoft.AspNetCore.Razor.Language/Extensions/InheritsDirectivePass.cs
```diff
@@ -0,0 +1,34 @@
...
+            }
+
+            foreach (var inherits in irDocument.FindDirectiveReferences(InheritsDirective.Directive))
+            {
+                inherits.Remove();
+
+                var token = ((DirectiveIRNode)inherits.Node).Tokens.FirstOrDefault();
+                if (token != null)
+                {
+
```
NTaylorMullen: newline

src/Microsoft.AspNetCore.Razor.Language/Extensions/SectionDirectivePass.cs
```diff
@@ -1,53 +1,25 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Collections.Generic;
+
```
NTaylorMullen: newline

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -42,11 +28,8 @@ internal class CSharpCodeParser : TokenizerBackedParser<CSharpTokenizer, CSharpS
             DirectiveKind.SingleLine,
             builder => builder.AddStringToken());
 
-        internal static readonly IEnumerable<DirectiveDescriptor> DefaultDirectiveDescriptors = new[]
+        internal static readonly IEnumerable<DirectiveDescriptor> DefaultDirectiveDescriptors = new DirectiveDescriptor[]
```
NTaylorMullen: Could nuke this guy now 👍

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
NTaylorMullen: I think this equality bits was here only for tests

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
```diff
@@ -239,6 +240,9 @@ private RazorEngine CreateEngine()
             {
                 // Notice we're not registering the ModelDirective.Pass here so we can run it on demand.
                 b.AddDirective(ModelDirective.Directive);
+
+                // There's some special interaction with the inherits directive
```
NTaylorMullen: lol wth, nice comment bro

test/Microsoft.AspNetCore.Razor.Language.Test/Extensions/TemplateTargetExtensionTest.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Xunit;
+using static Microsoft.AspNetCore.Razor.Language.Intermediate.RazorIRAssert;
```
NTaylorMullen: Needed?

src/Microsoft.AspNetCore.Razor.Language/Extensions/FunctionsDirectivePass.cs
rynowak: > We should pick an approach of static XYZDirective class with a static register method and a nested private class (that's what InjectDirective does as an example) or what you have in this PR, static class, register, the directive class and a public pass.

Meh, every case is case-by-case.

In these cases, it might be useful to reuse the directive definition, but implement it another way. So you'd want the directive descriptor but not the pass, which means the pass has to be public so you can remove it.

NTaylorMullen: Ah, for removal. Like it 👍

Issue #1173, opened 2017-04-04
Implement error recovery for directive tokens


rynowak:
We're not strict enough with error detection for directive tokens - for example
```
@custom global::Some_Namespace
```

If the `custom` directive expects a namespace, it will parse the `global` token, and then fail because the `::`Some_Namespace` token remains. If instead, another token was part of the directive that could consume this text, it was parse correctly.

So, that means that tokens are allowed to run together, and we don't prevent it today.

We also need to implement error recovery for tokens universally.

NTaylorMullen:
This is blocked on https://github.com/aspnet/Razor/issues/1412 being completed.

NTaylorMullen:
22060a15dc7f6eb37df3ee0b700b112b98b36fbe

PR #1430, opened 2017-06-12
Implement consistent error story for directives.

NTaylorMullen:
## Basics of what's in this PR
- Enforce whitespace inbetween directive tokens
- Re-write malformed directives into `MalformedDirectiveIRNode`s via a `MalformedDirectiveIRPass`
- Raise directive errors to the IR layer.

## Goals for extensible directive error handling
Tried to encompass 3 primary goals when coming up with a design for our directive error story.

### Consistency
When we  encounter a malformed token when parsing an extensible directive we don't consume the token; instead, we bail out and log an appropriate error that's raised to the `DirectiveIRNode`. Therefore, a user will never have to deal with a malformed directive token state unless they're purposefully consuming `MalformedDirectiveIRNode`s.

### Default simple approach for malformed directives
If a user writes a malformed directive an extensible directive author will not see it represented as a `DirectiveIRNode` as long as their pass operates after the `MalformedDirectiveIRPass`. This means for malformed directives Razor will provide IntelliSense until they have fully satisfied all tokens of a directive, at which point the directive will be valid and will make its way over to extensible directive IR pass code. 

Since a user never sees malformed directives, in the case of a block directive we flow the body of the malformed block directive to the execute method so that all contents continue to get IntelliSense/coloring.

Basically, I made it so extensible directive code is not called until its structure is 100% while maintaining the current editing experience; this should help us avoid VS explosions and make extensible directive IR consumption code simple.

### Opt-in customizable approach for malformed directive
If a user does not like how we handle malformed directives they can write an IR pass that consumes `MalformedDirectiveIRNode`s. They can have it log additional errors, modify code gen... basically whatever they want. A great example of when you'd want to do this is if you wish your malformed directive to still impact code gen to provide more verbose/specific errors (think `@model` where TModel was restricted with a `where` clause).

-----------------------------------------------

#1173 

@ajaybhargavb @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIRLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIRLoweringPhase.cs
@@ -131,12 +131,32 @@ namespace Microsoft.AspNetCore.Razor.Language
 
             public override void VisitAddTagHelperSpan(AddTagHelperChunkGenerator chunkGenerator, Span span)
             {
-                _builder.Push(new DirectiveIRNode()
+                RazorIRNode directiveNode;
+                if (IsMalformed(chunkGenerator.Diagnostics))
                 {
-                    Name = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Directive,
-                    Descriptor = CSharpCodeParser.AddTagHelperDirectiveDescriptor,
-                    Source = BuildSourceSpanFromNode(span),
-                });
+                    directiveNode = new MalformedDirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.AddTagHelperDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+                else
+                {
+                    directiveNode = new DirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.AddTagHelperDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+
+                for (var i = 0; i < chunkGenerator.Diagnostics.Count; i++)
+                {
+                    directiveNode.Diagnostics.Add(chunkGenerator.Diagnostics[i]);
+                }
+
+                _builder.Push(directiveNode);
 
                 _builder.Add(new DirectiveTokenIRNode()
                 {
@@ -150,12 +170,32 @@ namespace Microsoft.AspNetCore.Razor.Language
 
             public override void VisitRemoveTagHelperSpan(RemoveTagHelperChunkGenerator chunkGenerator, Span span)
             {
-                _builder.Push(new DirectiveIRNode()
+                RazorIRNode directiveNode;
+                if (IsMalformed(chunkGenerator.Diagnostics))
                 {
-                    Name = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Directive,
-                    Descriptor = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor,
-                    Source = BuildSourceSpanFromNode(span),
-                });
+                    directiveNode = new MalformedDirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+                else
+                {
+                    directiveNode = new DirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.RemoveTagHelperDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+
+                for (var i = 0; i < chunkGenerator.Diagnostics.Count; i++)
+                {
+                    directiveNode.Diagnostics.Add(chunkGenerator.Diagnostics[i]);
+                }
+
+                _builder.Push(directiveNode);
 
                 _builder.Add(new DirectiveTokenIRNode()
                 {
@@ -169,12 +209,32 @@ namespace Microsoft.AspNetCore.Razor.Language
 
             public override void VisitTagHelperPrefixDirectiveSpan(TagHelperPrefixDirectiveChunkGenerator chunkGenerator, Span span)
             {
-                _builder.Push(new DirectiveIRNode()
+                RazorIRNode directiveNode;
+                if (IsMalformed(chunkGenerator.Diagnostics))
                 {
-                    Name = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor.Directive,
-                    Descriptor = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
-                    Source = BuildSourceSpanFromNode(span),
-                });
+                    directiveNode = new MalformedDirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+                else
+                {
+                    directiveNode = new DirectiveIRNode()
+                    {
+                        Name = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor.Directive,
+                        Descriptor = CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
+                        Source = BuildSourceSpanFromNode(span),
+                    };
+                }
+
+                for (var i = 0; i < chunkGenerator.Diagnostics.Count; i++)
+                {
+                    directiveNode.Diagnostics.Add(chunkGenerator.Diagnostics[i]);
+                }
+
+                _builder.Push(directiveNode);
 
                 _builder.Add(new DirectiveTokenIRNode()
                 {
@@ -235,12 +295,32 @@ namespace Microsoft.AspNetCore.Razor.Language
                 {
                     _insideLineDirective = true;
 
-                    _builder.Push(new DirectiveIRNode()
+                    RazorIRNode directiveNode;
+                    if (IsMalformed(chunkGenerator.Diagnostics))
                     {
-                        Name = chunkGenerator.Descriptor.Directive,
-                        Descriptor = chunkGenerator.Descriptor,
-                        Source = BuildSourceSpanFromNode(block),
-                    });
+                        directiveNode = new MalformedDirectiveIRNode()
+                        {
+                            Name = chunkGenerator.Descriptor.Directive,
+                            Descriptor = chunkGenerator.Descriptor,
+                            Source = BuildSourceSpanFromNode(block),
+                        };
+                    }
+                    else
+                    {
+                        directiveNode = new DirectiveIRNode()
+                        {
+                            Name = chunkGenerator.Descriptor.Directive,
+                            Descriptor = chunkGenerator.Descriptor,
+                            Source = BuildSourceSpanFromNode(block),
+                        };
+                    }
+
+                    for (var i = 0; i < chunkGenerator.Diagnostics.Count; i++)
+                    {
+                        directiveNode.Diagnostics.Add(chunkGenerator.Diagnostics[i]);
+                    }
+
+                    _builder.Push(directiveNode);
 
                     base.VisitDirectiveBlock(chunkGenerator, block);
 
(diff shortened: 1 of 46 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
```diff
@@ -29,7 +29,7 @@ public ParserContext(RazorSourceDocument source, RazorParserOptions options)
 
         public SyntaxTreeBuilder Builder { get; }
 
-        public ErrorSink ErrorSink { get; }
+        public ErrorSink ErrorSink { get; set; }
```
NTaylorMullen: This is super hackery. Essentially we need to swap out the underlying error sink when parsing directive tokens because the tokenizer/other parsing pieces can log their own errors which we still need to capture.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1587,165 +1587,191 @@ private void EnsureDirectiveIsAtStartOfLine()
         private void HandleDirective(DirectiveDescriptor descriptor)
```
NTaylorMullen: Worth looking at this file with `w=1` in the url.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1587,165 +1587,191 @@ private void EnsureDirectiveIsAtStartOfLine()
...
+                        });
+                        break;
+                }
+            }
+            finally
+            {
+                if (directiveErrorSink.Errors.Count > 0)
+                {
+                    var directiveDiagnostics = directiveErrorSink.Errors.Select(error => RazorDiagnostic.Create(error));
+                    directiveChunkGenerator.Diagnostics.AddRange(directiveDiagnostics);
```
NTaylorMullen: Capturing the diagnostics on the parent directive and not on the actual tokens because we can't properly parse the tokens (they're errored) without risking throwing the entire parsing system into disarray.

rynowak:
I like the approach here, this is very thoughtful and has clear goals and principles in mind.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIRLoweringPhase.cs
```diff
@@ -134,12 +134,19 @@ public override void VisitImportSpan(AddImportChunkGenerator chunkGenerator, Spa
...
                     Source = BuildSourceSpanFromNode(span),
-                });
+                };
+
+                for (var i = 0; i < chunkGenerator.Diagnostics.Count; i++)
+                {
+                    directiveNode.Diagnostics.Add(chunkGenerator.Diagnostics[i]);
+                }
+
+                _builder.Push(directiveNode);
```
rynowak: Why not just create the error node directly here? I'd rather have the fork in lowering than a separate pass

src/Microsoft.AspNetCore.Razor.Language/Intermediate/MalformedDirectiveIRNode.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language.Intermediate
+{
+    public sealed class MalformedDirectiveIRNode : RazorIRNode
```
rynowak: This should still probably have a no-arg constructor

src/Microsoft.AspNetCore.Razor.Language/Intermediate/MalformedDirectiveIRNode.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            for (var i = 0; i < directive.Children.Count; i++)
+            {
+                Children.Add(directive.Children[i]);
+            }
+
+            Source = directive.Source;
+            Name = directive.Name;
+            Descriptor = directive.Descriptor;
```
rynowak: If you just created these during lowering instead of having a pass do it, none of this silliness would be required

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
rynowak: #YOLO

src/Microsoft.AspNetCore.Razor.Language/MalformedDirectiveIRPass.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class MalformedDirectiveIRPass : RazorIRPassBase, IRazorDirectiveClassifierPass
+    {
+        // -------------------------------------- REVIEWERS --------------------------------------------------
+        // Should this guy override order to ensure it runs before all of our features and any that users add?
+        // I'm thinking this should be -50 before default feature order (1000).
```
rynowak: Or just make it part of the lowering phase since it's an intrinsic node type

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIRLoweringPhase.cs
NTaylorMullen: My thought was that errored directive nodes can happen in many different ways:

- Malformed extensible directives. 1 lowering point.
- Malformed built-in directives. These each have their own lowering points.
- After-lowering "malformed" extensible directives. Basically someone has their own logic that determines if a directive is valid/invalid and adds diagnostics to a `DirectiveIRNode` that previously wasn't considered "malformed"

Since there's multiple ways the `DirectiveIRNode`s can flow into the system as malformed I thought it made the most sense to have a single deciding point on if a directive was malformed. 

Figured the only negative of having the pass was in the opt-in customization scenario where a users pass would need to pick an appropriate order in relation to the `MalformedDirectiveIRPass`. That negative also enables users to invalidate directives on their own terms though. I can see it going either way 😄

rynowak: >Malformed extensible directives. 1 lowering point.....Malformed built-in directives. These each have their own lowering points.

None of these are arguments about why this is the best design, they are arguments for why this is the least code required to implement the design.

NTaylorMullen: Sure. Thoughts on third point?

rynowak: The whole purpose of 'malformed' is that you don't process them by default. In the case you're suggesting, there's no value to be gained from 'malforming' since user code is doing it - (other than consistency).

Still if the user wanted to do it, it's trivial.

NTaylorMullen: Done.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -91,11 +91,7 @@ protected void MapDirectives(Action handler, params string[] directives)
         {
             foreach (var directive in directives)
             {
-                _directiveParsers.Add(directive, () =>
-                {
-                    EnsureDirectiveIsAtStartOfLine();
```
NTaylorMullen: Had to move this into each of the directive parsing bits so that the start of line errors made their way onto the `DirectiveIRNode`s and not the document.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -91,11 +91,7 @@ protected void MapDirectives(Action handler, params string[] directives)
         {
```
NTaylorMullen: Look at this file with `?w=1`

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
```diff
@@ -28,10 +28,7 @@ public class CodeGenerationIntegrationTest : IntegrationTestBase
         public void InvalidNamespaceAtEOF_Runtime()
         {
             var references = CreateCompilationReferences(CurrentMvcShim);
-            RunRuntimeTest(references, expectedErrors: new[]
-            {
-                "Identifier expected"
```
NTaylorMullen: Since we don't call into user code if a directive is invalid we no longer generate an empty namespace declaration (hence no error).

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/PageDirectiveTest.cs
```diff
@@ -10,24 +11,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
     public class PageDirectiveTest
     {
         [Fact]
-        public void TryGetPageDirective_CanHandleMalformedPageDirectives()
```
NTaylorMullen: MalformedDirectives = no page directive (doesn't make its way to user code).

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives_DesignTime.codegen.cs
```diff
@@ -1,4 +1,5 @@
-namespace 
+[assembly:global::Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.RazorPageAttribute(null, typeof(AspNetCore.TestFiles_IntegrationTests_CodeGenerationIntegrationTest_IncompleteDirectives_cshtml), null)]
```
NTaylorMullen: More examples of where not calling into user code for malformed directives saves the day.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1576,166 +1572,198 @@ private void EnsureDirectiveIsAtStartOfLine()
...
-            Context.Builder.CurrentBlock.Type = BlockKindInternal.Directive;
-            Context.Builder.CurrentBlock.ChunkGenerator = new DirectiveChunkGenerator(descriptor);
             AssertDirective(descriptor.Directive);
 
-            AcceptAndMoveNext();
-            Output(SpanKindInternal.MetaCode, AcceptedCharactersInternal.None);
+            var directiveErrorSink = new ErrorSink();
+            var savedErrorSink = Context.ErrorSink;
+            Context.ErrorSink = directiveErrorSink;
```
ajaybhargavb: Is there any significance of switching the error sink here? Looks like we are directly consuming it from `directiveErrorSink` at the end.

src/Microsoft.AspNetCore.Razor.Language/Legacy/DirectiveChunkGenerator.cs
```diff
@@ -17,6 +21,19 @@ public DirectiveChunkGenerator(DirectiveDescriptor descriptor)
 
         public DirectiveDescriptor Descriptor { get; }
 
+        public List<RazorDiagnostic> Diagnostics
```
ajaybhargavb: Why not `RazorDiagnosticCollection`?

src/Microsoft.AspNetCore.Razor.Language/Legacy/DirectiveChunkGenerator.cs
```diff
@@ -37,5 +55,24 @@ public override int GetHashCode()
 
             return combiner.CombinedHash;
         }
+
+        public override string ToString()
+        {
+            // This is used primarily at test time to show an identifiable representation of the chunk generator.
+
+            var toStringBuilder = new StringBuilder("Directive {");
```
ajaybhargavb: nit: Just `builder` would look cleaner.

src/Microsoft.AspNetCore.Razor.Language/Legacy/DirectiveChunkGenerator.cs
NTaylorMullen: This is an internal API (the chunk generator) which doesn't need to have the extra pieces of the diagnostic collection. We only use the `RazorDiagnosticCollection` to have control over the public API surface of our IR.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/1430#discussion_r121548559

NTaylorMullen:
@rynowak this good?

src/Microsoft.AspNetCore.Razor.Language/Legacy/DirectiveChunkGenerator.cs
rynowak: yeah lol

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
rynowak: nice

rynowak: but should we lol? Let's talk tomorrow, don't want to block this PR.

PR #1431, opened 2017-06-12
Make usage of FileName and FilePath consistent

rynowak:
FileName => "Foo.cshtml"
FilePath => "/Bar/Baz/Foo.cshtml"


rynowak:
FYI only

Issue #1424, opened 2017-06-09
Move CSharpCodeWriter and CodeWriter out of legacy


rynowak:

ajaybhargavb:
4811807329b9033950039e430a95562b2c1c40cf

PR #1433, opened 2017-06-13
Moved CSharpCodeWriter out of legacy

ajaybhargavb:
Issue - #1424 

- Moved `CSharpCodeWriter` and `CSharpCodeWritingScope` out of legacy.
- Removed `CodeWriter` base class and merged contents with `CSharpCodeWriter`.
- Modified `DefaultDocumentWriter` to use methods from `CSharpCodeWriter`.
- Removed `CSharpDisableWarningScope`
- Removed some more unused overloads.
- Moved corresponding tests.

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -6,11 +6,11 @@
...
 using System.Globalization;
 using System.Linq;
-using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
+using System.Text;
 
-namespace Microsoft.AspNetCore.Razor.Language.Legacy
+namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 {
-    public class CSharpCodeWriter : CodeWriter
+    public class CSharpCodeWriter : IDisposable
```
ajaybhargavb: Should this be sealed?

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -26,44 +26,183 @@ public class CSharpCodeWriter : CodeWriter
...
+        public CSharpCodeWriter SetIndent(int size)
         {
-            return (CSharpCodeWriter)base.SetIndent(size);
+            CurrentIndent = size;
+
+            return this;
         }
 
-        public new CSharpCodeWriter IncreaseIndent(int size)
+        public CSharpCodeWriter Indent(int size)
```
ajaybhargavb: A few methods like this one are unused outside this class. But I've left them as public it makes sense to have this functionality. Let me know if you feel otherwise.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -6,11 +6,11 @@
 using System.Diagnostics;
```
NTaylorMullen: Fix up the formatting in this class. Newlines, whitespace and such.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
rynowak: abstract or sealed

rynowak: This should **not** be disposable

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -26,44 +26,183 @@ public class CSharpCodeWriter : CodeWriter
...
+            if (_dirty)
+            {
+                _cache = Builder.ToString();
+                _dirty = false;
+            }
+
+            return _cache;
+        }
+
+        public SourceLocation GetCurrentSourceLocation()
```
rynowak: What about a property `Location`

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
```diff
@@ -2,22 +2,21 @@
...
+        private CSharpCodeWriter _writer;
         private bool _autoSpace;
         private int _tabSize;
         private int _startIndent;
 
-        public CSharpCodeWritingScope(CodeWriter writer) : this(writer, true) { }
-        public CSharpCodeWritingScope(CodeWriter writer, int tabSize) : this(writer, tabSize, true) { }
+        public CSharpCodeWritingScope(CSharpCodeWriter writer) : this(writer, true) { }
+        public CSharpCodeWritingScope(CSharpCodeWriter writer, int tabSize) : this(writer, tabSize, true) { }
         // TODO: Make indents (tabs) environment specific
```
rynowak: lol

ajaybhargavb:
🆙 📅  Will make the rest of the changes as a separate PR.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -302,9 +362,10 @@ public CSharpCodeWriter WriteEndMethodInvocation(bool endLine)
         }
```
NTaylorMullen: Is the WriteStartMethodInvocation with generic parameters used a lot? It's a rough API and thinking it'd make more sense to not have it.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -542,42 +598,50 @@ public CSharpCodeWritingScope BuildNamespace(string name)
...
-        {
-            return BuildConstructor("public", name);
-        }
-
-        public CSharpCodeWritingScope BuildConstructor(string accessibility, string name)
-        {
-            return BuildConstructor(accessibility, name, Enumerable.Empty<KeyValuePair<string, string>>());
-        }
-
         public CSharpCodeWritingScope BuildConstructor(
```
NTaylorMullen: Hmmmm, this is pretty odd to have on here at all. Lets inline this functionality where it's used. I believe its only used by our ViewComponent bits.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
```diff
@@ -2,22 +2,21 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 
-namespace Microsoft.AspNetCore.Razor.Language.Legacy
+namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 {
     public struct CSharpCodeWritingScope : IDisposable
```
NTaylorMullen: So the only reason why this is public today is to prevent boxing. Thinking we should nest this publicly in the `CSharpCodeWriter` since that's the only thing that uses it.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
NTaylorMullen: Also Fix up the OnClose field below. It be bad.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
```diff
@@ -2,22 +2,21 @@
...
     {
-        private CodeWriter _writer;
+        private CSharpCodeWriter _writer;
         private bool _autoSpace;
         private int _tabSize;
         private int _startIndent;
 
-        public CSharpCodeWritingScope(CodeWriter writer) : this(writer, true) { }
-        public CSharpCodeWritingScope(CodeWriter writer, int tabSize) : this(writer, tabSize, true) { }
+        public CSharpCodeWritingScope(CSharpCodeWriter writer) : this(writer, true) { }
```
NTaylorMullen: Too many ctors. Do we really use all these?

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
```diff
@@ -26,44 +26,155 @@ public class CSharpCodeWriter : CodeWriter
...
-            return (CSharpCodeWriter)base.Indent(size);
-        }
+        private string _cache = string.Empty;
+        private bool _dirty;
+
+        private int _absoluteIndex;
+        private int _currentLineIndex;
+        private int _currentLineCharacterIndex;
+
+        public StringBuilder Builder { get; } = new StringBuilder();
```
NTaylorMullen: Shouldn't be public. It's used by `CSharpCodeWritingScope` but this still shouldn't be public. If a user modifies this directly it throws off the internal state of this writer entirely.

ajaybhargavb: Hmm if they can't see the builder, how will they know what got written? Eg: When I am debugging I do `Builder.ToString()` to check what got written. Is this something we care about?

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
ajaybhargavb: Yes. This and `WriteEndMethodInvocation` combined is used in about 50 places. Might be useful. But I don't feel strongly either way.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
ajaybhargavb: Nah. Seems unnecessary I'll clean this up.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
NTaylorMullen: That's equivalent to calling `GenerateCode()`.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWritingScope.cs
rynowak: I would prefer to just get rid of this pattern.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CSharpCodeWriter.cs
NTaylorMullen: Ya lets fix this in a follow up.

Issue #1329, opened 2017-05-11
Review Properties on all public Tag Helper IR Nodes


rynowak:
My notes from review of all of our IR:

**General**

- [x] Want to consider a rename from FooIRNode -> FooIntermediateNode
- [x] Can we get rid of `AccessModifier` where it appears?

**Preallocated Nodes** 

- [x] Move to .Extensions namespace (still internal)

**Using**
- [x] UsingStatement -> UsingDirective (a using statement is something else altogether)

**Tag Helper**
- [x] Should have tag helper descriptors

**Tag Helper Body**
- [x] Should have tag helper descriptors

**Set Tag Helper Property**
- [x] Needs lots of love
- [x] Should not have a type name
- [x] Can we get rid of property name (redundant)?
- [x] Can we get rid of attribute name?

**DeclareTagHelperFieldsIRNode**
- [x] Can we delete it?

**Checksum**
- [x] Can we delete it?

**AddTagHelperHtmlAttributeIRNode**
- [x] Better name?

rynowak:
My notes from review of all of our IR:

**General**

Want to consider a rename from FooIRNode -> FooIntermediateNode
What things need a source besides tokens? where is it used?

Can we get rid of `AccessModifier` where it appears?

**Preallocated Nodes** 

Move to .Extensions namespace (still internal)

**Using**
- UsingStatement -> UsingDirective (a using statement is something else altogether)

**Tag Helper**
- Should have tag helper descriptors

**Tag Helper Body**
- Should have tag helper descriptors

**Set Tag Helper Property**
- Needs lots of love
- Should not have a type name
- Can we get rid of property name (redundant)?
- Can we get rid of attribute name?

**DirectiveIRToken**
- Can we get rid of it and use regular token?

**DeclareTagHelperFieldsIRNode**
- Can we delete it?

**Checksum**
- Can we delete it?

**AddTagHelperHtmlAttributeIRNode**
- Better name?

rynowak:
@ajaybhargavb - let's divide and conquer to make these a reality.

ajaybhargavb:
> What things need a source besides tokens? where is it used?

From a preliminary investigation, here is what I found. Note: this is not an exhaustive list.

Nodes | Reason
------|------
[...]

PR #1434, opened 2017-06-14
Removed DeclareTagHelperFieldsIRNode

ajaybhargavb:
Part of #1329 

- Removed `DeclareTagHelperFieldsIRNode`
- Using `FieldDeclarationIRNode` combined with annotations to keep track of the tag helper fields
- Added two entries to `CommonAnnotations`
- Added `ContainsKey()` to `ItemCollection`
- Updated tests

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ViewComponentTagHelperPass.cs
```diff
@@ -265,14 +272,11 @@ public override void VisitClassDeclaration(ClassDeclarationIRNode node)
...
-            public override void VisitDeclareTagHelperFields(DeclareTagHelperFieldsIRNode node)
+            public override void VisitFieldDeclaration(FieldDeclarationIRNode node)
             {
-                if (Fields == null)
-                {
-                    Fields = node;
-                }
+                TagHelperFields.Add(node);
 
-                base.VisitDeclareTagHelperFields(node);
```
rynowak: Should we be looking for an annotation here?

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -176,7 +176,14 @@ public override void VisitMethodDeclaration(MethodDeclarationIRNode node)
...
             {
-                Context.Writer.WriteField(node.AccessModifier, node.Modifiers, node.Type, node.Name);
+                if (node.Annotations.ContainsKey(CommonAnnotations.InitializeTagHelperVariables))
+                {
+                    Context.TagHelperWriter.WriteDeclareTagHelperFields(Context);
+                }
+                else
+                {
+                    Context.Writer.WriteField(node.AccessModifier, node.Modifiers, node.Type, node.Name);
+                }
```
rynowak: This isn't right at all.

This doesn't semantically change anything, it's just using the file declaration as a proxy.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
```diff
@@ -14,18 +14,9 @@ public class DesignTimeTagHelperWriter : TagHelperWriter
     {
         public string CreateTagHelperMethodName { get; set; } = "CreateTagHelper";
 
-        public override void WriteDeclareTagHelperFields(CSharpRenderingContext context, DeclareTagHelperFieldsIRNode node)
+        public override void WriteDeclareTagHelperFields(CSharpRenderingContext context)
```
rynowak: This method shouldn't exist anymore

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ViewComponentTagHelperPass.cs
ajaybhargavb: Yes. I meant to do that but missed it.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
ajaybhargavb: Still need this to write out the declaration and initialization of all the tag helper variables like `TagHelperExecutionContext`, `TagHelperRunner` etc. Should this be internal?
We can't make them be regular `FieldDeclarationIRNode`s because the type names used aren't known at IR time.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
ajaybhargavb: I agree this is the ugliest part of this PR. As mentioned here https://github.com/aspnet/Razor/pull/1434/files#r121839784 those field declarations are special and they need some node to track their location in the IR tree and we can't split them into other basic nodes because all the information is not available at IR phase.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeTagHelperWriter.cs
rynowak: > We can't make them be regular FieldDeclarationIRNodes because the type names used aren't known at IR time.

Why not?

ajaybhargavb: Technically we can make it work. But a lot of the properties here https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/RuntimeTagHelperWriter.cs#L21 will need to move out of here into IR phase. We'll need to use basic nodes to write out this entire block https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/AttributeTargetingTagHelpers_Runtime.codegen.cs#L11-L27

ajaybhargavb:
Closing. We are going with a different approach.

PR #1435, opened 2017-06-14
Make section an extension node

rynowak:


test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/DesignTime_DesignTime.codegen.cs
```diff
@@ -49,13 +49,14 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_DesignTime
...
-            DefineSection("Footer", async (__razor_section_writer) => {
+            DefineSection("Footer", async(__razor_section_writer) => {
 #line 14 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/DesignTime.cshtml"
 __o = bar;
 
 #line default
 #line hidden
-            });
+            }
+            );
```
rynowak: Offhand remark.

There's just sooooo much policy in some of these csharpcodewriter methods. We should think about making them internal or deleting them.

The fact that I can't use `BuildAsyncLambda` without the newline here means that I really don't want to use it.

src/Microsoft.AspNetCore.Razor.Language/Extensions/SectionTargetExtension.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    public sealed class SectionTargetExtension : ISectionTargetExtension
+    {
+        private static readonly string DefaultWriterName = "__razor_section_writer";
```
rynowak: No. This is turd that will go away when we do multitargeting. I don't want it on the public API

src/Microsoft.AspNetCore.Razor.Language/Extensions/SectionDirectivePass.cs
```diff
@@ -17,30 +17,30 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNo
...
-                });
-                section.InsertBefore(builder.Build());
+                    Name = sectionName,
+                };
 
-                section.InsertBefore(section.Node.Children.Except(((DirectiveIRNode)section.Node).Tokens));
+                var i = 0;
+                for (; i < directive.Node.Children.Count; i++)
+                {
+                    if (!(directive.Node.Children[i] is DirectiveTokenIRNode))
```
rynowak: Cool beans

Issue #1321, opened 2017-05-11
Get rid of checksum ir node


rynowak:
This should just be an artifact of the Runtime codegen for a document

ajaybhargavb:
3892a6fedead6a8db151a21122cbcd0af9c24bb5

PR #1442, opened 2017-06-15
Removed ChecksumIRNode

ajaybhargavb:
Issue - #1321 

- Removed `ChecksumIRNode` and all visitors
- Added a switch in `RazorCodeGenerationOptions` to disable checksum generation
- Checksum generation now happens as part of `VisitDocument` at code gen phase
- Regenerated baslines and updated tests

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorCodeGenerationOptionsBuilder.cs
```diff
@@ -11,9 +11,11 @@ internal class DefaultRazorCodeGenerationOptionsBuilder : RazorCodeGenerationOpt
 
         public override bool IndentWithTabs { get; set; }
 
+        public override bool GenerateChecksum { get; set; } = true;
```
ajaybhargavb: Defaulting this to `true` here. Should we instead have a `RuntimeOptionsFeature` like we do for design time? https://github.com/aspnet/Razor/blob/ajbaaska/nuke-checksum/src/Microsoft.AspNetCore.Razor.Language/DesignTimeOptionsFeature.cs#L30

NTaylorMullen: Thinking a `RuntimeOptionsFeature` makes more sense. It's consistent that way.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -69,12 +70,36 @@ public void RenderChildren(RazorIRNode node)
 
             public override void VisitDocument(DocumentIRNode node)
             {
-                RenderChildren(node);
-            }
+                if (Context.Options.GenerateChecksum)
```
rynowak: This option should be `false` by default. Call it `SuppressChecksum`

PR #1450, opened 2017-06-21
Add Tag Helper descriptors to the Tag Helper node

rynowak:


src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperOptimizationPass.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    internal class DefaultTagHelperOptimizationPass : RazorIRPassBase, IRazorIROptimizationPass
+    {
+        // Run later than default order so other passes have a chance to modify the
+        // tag helper nodes.
+        public override int Order => 1000;
```
NTaylorMullen: Hmm? DefaultFeatureOrder is 1k

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperOptimizationPass.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentIRNode irDocument)
+        {
```
NTaylorMullen: This class not needed?

src/Microsoft.AspNetCore.Razor.Language/Intermediate/ExtensionIRNode.cs
```diff
@@ -36,6 +36,8 @@ public override RazorDiagnosticCollection Diagnostics
             }
         }
 
+        public override SourceSpan? Source { get; set; }
```
NTaylorMullen: 👍

src/Microsoft.AspNetCore.Razor.Language/Intermediate/TagHelperIRNode.cs
```diff
@@ -46,6 +47,8 @@ public override RazorDiagnosticCollection Diagnostics
 
         public TagMode TagMode { get; set; }
 
+        public ICollection<TagHelperDescriptor> TagHelpers { get; set; }
```
NTaylorMullen: Thinking it's probably better to treat this guy like Children, Diagnostics, Annotations etc. and have a getter only.

rynowak: Sure

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperOptimizationPass.cs
rynowak: oh lol, that's not supposed to be in this commit

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperOptimizationPass.cs
rynowak: does it be?

rynowak: yeah I should be using the constant here. The default for user code is 0. The default for our code is 1000 where we want to run **after** user code. This is one of those cases

NTaylorMullen:
Looks like you're failing in the travis/appveyor builds because of compile errors. :shipit: once everything be good.

test/Microsoft.AspNetCore.Razor.Test.Common/Langauge/Intermediate/RazorIRAssert.cs
```diff
@@ -342,14 +342,16 @@ internal static void TagHelperFieldDeclaration(RazorIRNode node, params string[]
...
+        internal static void TagHelper(string tagName, TagMode tagMode, IEnumerable<TagHelperDescriptor> tagHelpers, RazorIRNode node, params Action<RazorIRNode>[] childValidators)
         {
             var tagHelperNode = Assert.IsType<TagHelperIRNode>(node);
 
             try
             {
                 Assert.Equal(tagName, tagHelperNode.TagName);
                 Assert.Equal(tagMode, tagHelperNode.TagMode);
+
+                Assert.Equal(tagHelpers, tagHelperNode.TagHelpers);
```
ajaybhargavb: @NTaylorMullen, should this use the case sensitive comparer?

NTaylorMullen: You totally be right 👍

PR #1452, opened 2017-06-21
Clean up .Legacy

rynowak:
Also sealed two types that didn't get the treatment.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/HtmlAttributeValueStyle.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/HtmlAttributeValueStyle.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace Microsoft.AspNetCore.Razor.Language.Legacy
+namespace Microsoft.AspNetCore.Razor.Language
 {
+    // This is the design time equivalent of Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.
+    // They should be kept in sync.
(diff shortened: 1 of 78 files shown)


src/Microsoft.AspNetCore.Razor.Language/HtmlAttributeValueStyle.cs
```diff
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace Microsoft.AspNetCore.Razor.Language.Legacy
+namespace Microsoft.AspNetCore.Razor.Language
 {
+    // This is the design time equivalent of Microsoft.AspNetCore.Razor.TagHelpers.HtmlAttributeValueStyle.
+    // They should be kept in sync.
     public enum HtmlAttributeValueStyle
```
rynowak: @NTaylorMullen - I want to find a better name for this if possible.

rynowak: Idea that I had was `AttributeStructure` to go along with `TagStructure`

NTaylorMullen: Do it.

src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
-    public enum BlockKindInternal
+    internal enum BlockKindInternal
```
NTaylorMullen: 😆

PR #1453, opened 2017-06-21
Rename RazorIRNode

rynowak:
Get rid of references to 'IR'


rynowak:
This is mostly and FYI, as it's too large to really review and most changes are mechanical.

Will also send reaction PRs for other repos as necessary. Planning on merging this later tonight.

PR #1454, opened 2017-06-21
Get rid of AccessModifiers

rynowak:


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -122,9 +122,14 @@ public override void VisitNamespaceDeclaration(NamespaceDeclarationIntermediateN
 
             public override void VisitClassDeclaration(ClassDeclarationIntermediateNode node)
             {
+                for (var i = 0; i < node.Modifiers.Count; i++)
+                {
+                    Context.Writer.Write(node.Modifiers[i]);
+                    Context.Writer.Write(" ");
```
NTaylorMullen: chain writes here and below for consistency.

rynowak: I hate that crap so much

Issue #1432, opened 2017-06-12
Review RazorIRBuilder


rynowak:
RazorIRReference does (better) a lot of the things the builder did. Should review the builder and make it more relevant.

rynowak:
What I'd like to do here is to make `RazorIRBuilder` internal. All of the users outside of our main lowering code are pretty trivial.

@ajaybhargavb @NTaylorMullen any objections?

NTaylorMullen:
> @ajaybhargavb @NTaylorMullen any objections?

Sounds good to me.

ajaybhargavb:
5e57e9f2359149ec126ab595027b93de1ceaa902

PR #1455, opened 2017-06-21
Make IntermediateNodeBuilder internal

ajaybhargavb:
#1432 

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeBuilder.cs
@@ -5,7 +5,7 @@ using System;
 
 namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
-    public abstract class IntermediateNodeBuilder
+    internal abstract class IntermediateNodeBuilder
     {
         public static IntermediateNodeBuilder Create(IntermediateNode root)
         {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeBuilder.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
-    public abstract class IntermediateNodeBuilder
+    internal abstract class IntermediateNodeBuilder
```
NTaylorMullen: @rynowak do you envision us ever wanting to expose this? I'm thinking no.

If so, no need for the abstractness and a default implementation (only one implementation); far simpler to just have all the logic live here.

rynowak: I don't want to make a bunch of change for change sake for something that's already working, tested and non-public.

@ajaybhargavb if you want to simplify this code go ahead, but it's not a priority

ajaybhargavb: @NTaylorMullen, I'm thinking I'll keep this how it is right now unless you feel strongly otherwise.

NTaylorMullen: Nope, all good.

Issue #1376, opened 2017-05-22
Enable file scoped extensible directives


NTaylorMullen:
These directives would be top-level scoped, one per file, would have overriding behavior implemented by lowering (like `@model`)

NTaylorMullen:
WIP
1877a29bb8151fc3a8988ee31370fe0ee661715c
1dd4feb1f0ba00697c0c4c1ec0fcd720c8dfb520

PR #1457, opened 2017-06-22
Add file scoped extensible directives.

NTaylorMullen:
### What's in the PR
Introduced a new `DirectiveUsage` `enum` that enables directive authors to specify if they want their directives be have a specific usage. The default (what exists today) is `DirectiveUsage.Unrestricted`. The new one that was introduced in this PR (file scoped) is labeled as `SinglePreContent`. It's a semi-verbose name but I think it's important that when a user sees the enum they have a clear indication as to what it entails; this of course is further elaborated in doc comments.

For a `SinglePreContent` directive there is the following restrictions:
1. Must exist prior to any statements or expressions
1. Must exist prior to any HTML (including TagHelpers)
1. Must not have any duplicates

With these restrictions in mind these are the high level pieces that this enables
- Directives sitting next to other directives
- Directives sitting indiscriminately next to any number of whitespace
- Directives being decorated with comments

### Inheriting Directives
This PR does not contain the code to automatically inherit directives that are labeled as `SinglePreContent`. I thought about making this automatic but felt that it could go in a follow up if deemed important. The primary reason for it not being in this PR is due to it requiring a change in design on how we inherit directives. Today we append imported IR to the bottom of our IR document; this isn't sufficient if you have a default importing story. If we start specifying how directives are inherited we then need to start marking `DirectiveIntermediateNode`s (in a way that stops overridden directives from flowing into extensibility) as inherited or not inherited and enable ways on how to change that.

--------------------------------------------------------------

- Added `DirectiveUsage` to enable extensible directive authors to indicate how their directives should be used. Currently support `Unrestricted` (how section directives have always worked and a single pre-content based directive.
- Added directive parsing tests.
- Removed no longer used `BlockKindInternal` items.
- Update built-in directives to use DirectiveUsage.
- Updated integration code gen and IR bits to reflect new directive usage.
- Updated existing unit tests that happened to test directives in code blocks to now test what happens when they exist at the document level. Being inside of a code block is now invalid and we have separate tests for that scenario.

#1376 

/cc @rynowak @ajaybhargavb

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
@@ -14,7 +14,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
         public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
             "model",
             DirectiveKind.SingleLine,
-            builder => builder.AddTypeToken());
+            builder =>
+            {
+                builder.AddTypeToken();
+                builder.Usage = DirectiveUsage.FileScopedSinglyOccurring;
+            });
 
         public static IRazorEngineBuilder Register(IRazorEngineBuilder builder)
         {
@@ -55,7 +59,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             }
             else
             {
-                return  "dynamic";
+                return "dynamic";
             }
         }
 
--- a/src/Microsoft.AspNetCore.Razor.Language/IDirectiveDescriptorBuilder.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/IDirectiveDescriptorBuilder.cs
@@ -31,6 +31,11 @@ namespace Microsoft.AspNetCore.Razor.Language
         DirectiveKind Kind { get; }
 
         /// <summary>
+        /// Gets or sets the directive usage. The usage determines how many, and where directives can exist per document.
+        /// </summary>
+        DirectiveUsage Usage { get; set; }
+
+        /// <summary>
         /// Gets a list of the directive tokens.
         /// </summary>
         IList<DirectiveTokenDescriptor> Tokens { get; }
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -54,6 +54,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
         private Dictionary<string, Action> _directiveParsers = new Dictionary<string, Action>(StringComparer.Ordinal);
         private Dictionary<CSharpKeyword, Action<bool>> _keywordParsers = new Dictionary<CSharpKeyword, Action<bool>>();
+        private HashSet<string> _seenDirectives = new HashSet<string>(StringComparer.Ordinal);
 
         public CSharpCodeParser(ParserContext context)
             : this(directives: Enumerable.Empty<DirectiveDescriptor>(), context: context)
@@ -91,7 +92,12 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
         {
             foreach (var directive in directives)
             {
-                _directiveParsers.Add(directive, handler);
+                _directiveParsers.Add(directive, () =>
+                {
+                    handler();
+                    _seenDirectives.Add(directive);
+                });
+
                 Keywords.Add(directive);
 
                 // These C# keywords are reserved for use in directives. It's an error to use them outside of
@@ -1589,10 +1595,13 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                 AcceptAndMoveNext();
                 Output(SpanKindInternal.MetaCode, AcceptedCharactersInternal.None);
 
+                // Even if an error was logged do not bail out early. If a directive was used incorrectly it doesn't mean it can't be parsed.
+                ValidateDirectiveUsage(descriptor);
+
                 for (var i = 0; i < descriptor.Tokens.Count; i++)
                 {
                     if (!At(CSharpSymbolType.WhiteSpace) &&
-                        !At(CSharpSymbolType.NewLine) && 
+                        !At(CSharpSymbolType.NewLine) &&
                         !EndOfFile)
                     {
                         Context.ErrorSink.OnError(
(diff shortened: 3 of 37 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1767,6 +1776,56 @@ private void HandleDirective(DirectiveDescriptor descriptor)
...
+        {
+            if (descriptor.Usage == DirectiveUsage.SinglePreContent)
+            {
+                if (_seenDirectives.Contains(descriptor.Directive))
+                {
+                    UsageError(Resources.FormatDuplicateDirective(descriptor.Directive));
+                    return;
+                }
+
+                var root = Context.Builder.ActiveBlocks.Last();
```
NTaylorMullen: @ajaybhargavb this is the magic that determines if a directive exists prior to code, html etc.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
     public class ModelDirectiveTest
     {
         [Fact]
-        public void ModelDirective_GetModelType_GetsTypeFromLastWellFormedDirective()
+        public void ModelDirective_GetModelType_GetsTypeFromFirstWellFormedDirective()
```
NTaylorMullen: This means that `@model` existing multiple times in a single document results in the first valid `@model` being used. If you have a valid model in an import file that carries over and properly overrides proper `@model` directives due to how our `ModelDirective` code passes over the IR.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/MultipleModels_DesignTime.codegen.cs
```diff
@@ -39,6 +39,6 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_MultipleMo
         [global::Microsoft.AspNetCore.Mvc.Razor.Internal.RazorInjectAttribute]
         public global::Microsoft.AspNetCore.Mvc.Rendering.IJsonHelper Json { get; private set; }
         [global::Microsoft.AspNetCore.Mvc.Razor.Internal.RazorInjectAttribute]
-        public global::Microsoft.AspNetCore.Mvc.Rendering.IHtmlHelper<System.Collections.IEnumerable> Html { get; private set; }
+        public global::Microsoft.AspNetCore.Mvc.Rendering.IHtmlHelper<ThisShouldBeGenerated> Html { get; private set; }
```
NTaylorMullen: Should have been this way to start with.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -1062,57 +1351,53 @@ public void AddTagHelperDirective_EndQuoteRequiresDoubleQuotesAroundValue()
         }
 
         [Fact]
-        public void ParseBlock_InheritsDirective()
```
NTaylorMullen: No longer valid. Have other tests.

rynowak:
I like the naming `File Scoped` rather than pre-content. I call it file scoped, because the unit of operation is the file, and it is only allowed at the top level of scoping. 

I think we have the following categories:

File Scoped - Singly Occurring
File Scoped - Multiple Occurring
Block Scoped

I defined `File Scoped` in the same way that you define `PreContent` as far as lexically how they are in Razor code. I think that part is perfect in your description.

------

Additionally they should have the following semantics:

**File Scoped - Single**

Imported directives can be overridden - from the furthest to the closest. An overridden directive **does not appear in the IR**. It is invisible to user code. 

If you don't want this, don't do this. File Scoped - Multiple will allow you to see multiple uses of the directive. 

Using a File Scoped - Single directive multiple times in any single file is an error.

Using a File Scoped - Single directive outside of the 'top' of the file is an error.

Does not support blocks. 

Examples: model, inherits, page

**MultipleFileScoped**

Imported directives are all added to the IR document - from the furthest to the closest.

Using a File Scoped - Multiple directive outside of the 'top' of the file is an error.

Does not support blocks. 

Examples: inject, using

**Block**

Imported directives are all added to the IR document - from the furthest to the closest.

Can appear anywhere in the document.

rynowak:
For what I'm calling a Block directive, I'm not sure if we should make it possible to make a directive importable/not-importable.

rynowak:
My guess btw is that the code for this is mostly right, I feel like my concept of this is about 10% off from yours

rynowak:
Also - notice how 'Block' is pretty much how directives behave today. The File Scoped variations just remove the need to extensiblity authors to do error checking.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -14,7 +14,11 @@ public static class ModelDirective
         public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
             "model",
             DirectiveKind.SingleLine,
-            builder => builder.AddTypeToken());
+            builder =>
+            {
+                builder.AddTypeToken();
+                builder.Usage = DirectiveUsage.SinglePreContent;
```
rynowak: Should usage just be a required argument to `CreateDirective`? I don't have a major grip with this, I just want to understand how you see the tradeoffs.

src/Microsoft.AspNetCore.Razor.Language/IDirectiveDescriptorBuilder.cs
```diff
@@ -31,6 +31,11 @@ public interface IDirectiveDescriptorBuilder
         DirectiveKind Kind { get; }
 
         /// <summary>
+        /// Gets the directive usage. The usage determines how many, and where directives can exist per document.
```
rynowak: gets/sets

src/Microsoft.AspNetCore.Razor.Language/Properties/Resources.Designer.cs
```diff
@@ -514,6 +514,34 @@ internal static string IntermediateNodeReference_CollectionIsReadOnly
         internal static string FormatIntermediateNodeReference_CollectionIsReadOnly(object p0)
             => string.Format(CultureInfo.CurrentCulture, GetString("IntermediateNodeReference_CollectionIsReadOnly"), p0);
 
+        /// <summary>
+        /// The '{0}' directive may only occurr once per document.
```
rynowak: occurr

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
rynowak: I don't understand what you're saying

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
```diff
@@ -31,7 +31,7 @@ @model Type2
             var result = ModelDirective.GetModelType(irDocument);
 
             // Assert
-            Assert.Equal("Type2", result);
```
rynowak: I think this is fine.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives_DesignTime.diagnostics.txt
```diff
@@ -1,8 +1,14 @@
...
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(8,8): Error RZ9999: The 'model' directive expects a type name.
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(10,8): Error RZ9999: The 'inject' directive expects a type name.
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(11,9): Error RZ9999: The 'inject' directive expects a type name.
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(12,26): Error RZ9999: The 'inject' directive expects an identifier.
+TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(14,1): Error RZ9999: The 'namespace' directive must exist prior to markup or code.
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(14,11): Error RZ9999: The 'namespace' directive expects a namespace name.
+TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(15,1): Error RZ9999: The 'namespace' directive may only occurr once per document.
```
rynowak: occurrrrrrr

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
NTaylorMullen: Thought about that approach but didn't feel like it deserved its own argument since having an unrestricted directive was consistent with what we've always had in the past. Aka, users previous expectations of directives is the default experience.

rynowak: OK cool, I think that's fine.

NTaylorMullen:
Running an MVC build with these bits to ensure I didn't break anything and then will get this in (assuming all is well). I'll then do a follow up PR to implement _ViewImports support by default for `FileScopedSinglyOccurring` directive usages.

PR #1459, opened 2017-06-22
Move some stuff to .Extensions

rynowak:
Moved design time directives and tag helper attribute preallocation to
.Extensions.

Just mechanical namespace changes here.


rynowak:
FYI only, this is just mechanical renames. Will merge when CI is green

Issue #1449, opened 2017-06-20
razor generated code should have the ubiquitous auto-generated in the comments


rynowak:
_From @jimmymain on June 20, 2017 12:20_

The omitted `<auto-generated/>` in the comments for code files generated by the razor engine are causing [multiple issues](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2373#issuecomment-309723230) to be raised in [other projects](https://github.com/aspnet/Tooling/issues/498).

Please could we have `<auto-generated />` added as the first line of the comments in razor generated code files. 

This will **alleviate** the **huge volume of noise** being generated in code analysis when opening *.cshtml* files.

https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2373
https://github.com/aspnet/Tooling/issues/498

Perhaps also consider including `#pragma warning suppress 1591` as well. Either that, or ensure that each generated comment is XML compliant.

https://github.com/dotnet/roslyn/issues/12702

_Copied from original issue: aspnet/Mvc#6414_

rynowak:
/cc @sharwell

So I'm hearing the feedback loud and clear about `CS1591` and XML doc related warnings.

Is there anything we can do that will get stylecop to ignore a file?

sharwell:
You should do two things:

1. Make sure the top of the file contains a comment which starts with `<auto-generated/>`. If you want to generate more detail, e.g. the version of the tool that generated the file or a message explaining that the code will be regenerated, I can show examples of standard cases.

2. Add a `#pragma` line to disable CS1591 with a comment explaining the meaning:

    ```csharp
    // Missing XML comment for publicly visible type or member '...'
    #pragma warning disable 1591
    ```

rynowak:
@ajaybhargavb putting on your plate for the 2.0.0 runtime release. 

I will report back about whether or not we can fix for VS 15.3.

rynowak:
Tracking this on TFS as well with 454556

ajaybhargavb:
rel/15.3: 582cd877bb47862112610a43c59c536c423fc3f1
dev: d73ecf8b36ec865051b5f835026d00763d7b1865

PR #1463, opened 2017-06-23
Suppress warnings for auto generated code

ajaybhargavb:
#1449 

- Added `// <auto-generated/>` to suppress the StyleCop Analyzer warnings
- Added `#pragma warning disable 1591` to suppress the XML doc comments missing warnings. Added a corresponding restore at the end of file (Not necessary but for consistency)
- Verified this works correctly during design time ~~and runtime~~(runtime changes are only needed in dev branch) and the debugging experience is not affected.
- Added a test and updated baselines.

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -69,7 +69,12 @@ public void RenderChildren(RazorIRNode node)
 
             public override void VisitDocument(DocumentIRNode node)
             {
+                Context.Writer.WriteLine("// <auto-generated/>");
```
ajaybhargavb: I can't think of a reason to have this not enabled. I haven't added any option to disable this behavior.

global.json
```diff
@@ -0,0 +1,5 @@
+{
+    "sdk": {
+        "version": "2.0.0-preview2-006215"
```
ajaybhargavb: Pinning the cli and packages version to have this build successfully for rel/15.3 branch.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/PageWithNamespace_DesignTime.codegen.cs
```diff
@@ -1,3 +1,5 @@
+// <auto-generated/>
```
pranavkm: Does this appear after the checksum pragma? That has to be the first line in the generated code.

.gitignore
```diff
@@ -31,6 +31,5 @@ project.lock.json
 .build/
 .vs/
 launchSettings.json
-global.json
```
pranavkm: ?

build/dependencies.props
```diff
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <AspNetCoreVersion>2.0.0-preview2-25661</AspNetCoreVersion>
```
pranavkm: ?

global.json
pranavkm: `git add -f global.json` and leave the .gitignore alone.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/PageWithNamespace_DesignTime.codegen.cs
ajaybhargavb: It doesn't. I messed it up when I cherry picked from dev. Checksum generation behavior is different in dev and rel/15.3. I'll update this.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -70,11 +70,16 @@ public void RenderChildren(RazorIRNode node)
...
                 RenderChildren(node);
+
+                Context.Writer.WriteLine("#pragma warning restore 1591");
             }
 
             public override void VisitChecksum(ChecksumIRNode node)
             {
                 Context.BasicWriter.WriteChecksum(Context, node);
+
+                Context.Writer.WriteLine("// <auto-generated/>");
```
ajaybhargavb: Having this here is kind of ugly. Spoke to @NTaylorMullen, This is temporary code and is only going to live inside VS and we don't need to care about customization.

ajaybhargavb: Actually I am going to revert this because we only need this for design time in this branch.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/PageWithNamespace_DesignTime.codegen.cs
ajaybhargavb: I've updated this to only do this in design time. We don't add pragma checksum in design time.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -69,7 +69,18 @@ public void RenderChildren(RazorIRNode node)
 
             public override void VisitDocument(DocumentIRNode node)
             {
+                if (Context.Options.DesignTimeMode)
+                {
+                    Context.Writer.WriteLine("// <auto-generated/>");
```
pranavkm: Does not having this cause issues at runtime for instance if you have `<GenerateDocumentationFile>true</GenerateDocumentationFile>` on and `WarningsAsErrors` enabled? If so we should just generate these all the time. For comparison, here's the blob of text that System.Web generates at the top of generated files:

```C#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     Runtime Version:4.0.30319.42000
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

[assembly: System.CodeDom.Compiler.GeneratedCodeAttribute("ASP.NET", "4.0.30319.42000")]
[assembly: System.Security.SecurityRulesAttribute(System.Security.SecurityRuleSet.Level2)]
[assembly: System.Runtime.Versioning.TargetFrameworkAttribute(".NETFramework,Version=v4.6.1")]

```

NTaylorMullen: We will in the dev branch. This branch will not make it into a runtime release.

PR #1465, opened 2017-06-24
Make single line single file scoped directives automatically import.

NTaylorMullen:
### Gist of the PR
- Single line directives in imports always flow to the main document. This is to enable directives like `@inject` (a non-file scoped singly occurring directive) to provide its own import behavior.
- Single line directives that are `DirectiveUsage.FileScopedSinglyOccurring` only import if they're the first of their kind. Basically implementing overriding behavior (like what we have with `@model`) today.
- Block directives that are `DirectiveUsage.FileScopedSinglyOccurring` and exist in an import do not import AND have an error logged. 

*The third bullet point I'm unsure about, it seems too specific. Despite what's in this PR I think i'm leaning towards not having an error at all and just not flowing them* @rynowak / @ajaybhargavb please share your thoughts on this.

---------------------------------------
- Added an inner pass inside of the intermediate lowering phase to determine which directives get flowed to the final document. There were many ways to accomplish this but in order to keep the last wins mechanic for non-auto imported directives I had to let the directives get created and then removed based on if they were inherited.
- Added error case if a user attempts to import a block directive with a `FileScopedSinglyOccurring` directive usage.
- Added test cases that validate directives are properly inherietd at the intermediate lowering phase.
- Updated a few tests that had incorrect assumptions.
- Left the default directive passes alone in regards to determining the "imported" directive to enable users to add their own model, inherits, etc. directives that take precedence.
- Normalized the passes in the intermediate lowering phase to handle directives identically (we don't conditionally lower anymore).

#1376 

/cc @rynowak @ajaybhargavb

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
@@ -73,6 +73,8 @@ namespace Microsoft.AspNetCore.Razor.Language
                 builder.Insert(i++, @using);
             }
 
+            ImportDirectives(document);
+
             // The document should contain all errors that currently exist in the system. This involves
             // adding the errors from the primary and imported syntax trees.
 
@@ -94,6 +96,56 @@ namespace Microsoft.AspNetCore.Razor.Language
             }
 
             codeDocument.SetDocumentIntermediateNode(document);
+        }
+
+        private void ImportDirectives(DocumentIntermediateNode document)
+        {
+            var visitor = new DirectiveVisitor();
+            visitor.VisitDocument(document);
+
+            var seenDirectives = new HashSet<DirectiveDescriptor>();
+            for (var i = visitor.Directives.Count - 1; i >= 0; i--)
+            {
+                var reference = visitor.Directives[i];
+                var directive = (DirectiveIntermediateNode)reference.Node;
+                var descriptor = directive.Descriptor;
+                var seenDirective = !seenDirectives.Add(descriptor);
+                var imported = ReferenceEquals(directive.Annotations[CommonAnnotations.Imported], CommonAnnotations.Imported);
+
+                if (!imported)
+                {
+                    continue;
+                }
+
+                switch (descriptor.Kind)
+                {
+                    case DirectiveKind.SingleLine:
+                        if (seenDirective && descriptor.Usage == DirectiveUsage.FileScopedSinglyOccurring)
+                        {
+                            // This directive has been overridden, it should be removed from the document.
+
+                            break;
+                        }
+
+                        continue;
+                    case DirectiveKind.RazorBlock:
+                    case DirectiveKind.CodeBlock:
+                        if (descriptor.Usage == DirectiveUsage.FileScopedSinglyOccurring)
+                        {
+                            // A block directive cannot be imported.
+
+                            document.Diagnostics.Add(
+                                RazorDiagnosticFactory.CreateDirective_BlockDirectiveCannotBeImported(descriptor.Directive));
+                        }
+                        break;
+                    default:
+                        throw new InvalidOperationException(Resources.FormatUnexpectedDirectiveKind(typeof(DirectiveKind).FullName));
+                }
+
+                // Overridden and invalid imported directives make it to here. They should be removed from the document.
+
+                reference.Remove();
+            }
         }
 
         private RazorCodeGenerationOptions CreateCodeGenerationOptions()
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -73,6 +73,8 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
                 builder.Insert(i++, @using);
             }
 
+            InheritImportedDirectives(document, codeDocument.Source.FilePath);
```
NTaylorMullen: So I experimented with reversing the lowering order so I could conditionally not add a directive instead of removing it later. Problem with this approach is it totally messes with other directives like `@inject` who implement their own import behavior (it makes the most relevant directive first instead of last).

Therefore, went with an inner-pass approach where we remove directives that have been added if we feel that they should not have been imported.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -96,6 +98,54 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+                        }
+
+                        continue;
+                    case DirectiveKind.RazorBlock:
+                    case DirectiveKind.CodeBlock:
+                        if (descriptor.Usage == DirectiveUsage.FileScopedSinglyOccurring)
+                        {
+                            // A block directive cannot be imported.
+
+                            document.Diagnostics.Add(
```
NTaylorMullen: See PR details. Not sure I like what I did here.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -122,6 +172,50 @@ public LoweringVisitor(DocumentIntermediateNode document, IntermediateNodeBuilde
 
             public string FilePath { get; set; }
 
+            public override void VisitDirectiveToken(DirectiveTokenChunkGenerator chunkGenerator, Span span)
```
NTaylorMullen: These visit methods are a direct copy from the `MainSourceVisitor` (this class is used for visiting the primary document and the import documents). There was no need to have a separate import pass anymore because we need to remove overridden directives; to do that I added all directives to the document and then decisively remove them / log diagnostics before we finish the phase.

rynowak:
> Block directives that are DirectiveUsage.FileScopedSinglyOccurring and exist in an import do not import AND have an error logged.
> The third bullet point I'm unsure about, it seems too specific. 

I want the error. We have no reason to be tolerant of this.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
rynowak: Preserving the source ordering is **critical** 👍 If it makes our jobs harder, too bad, that's what we signed up for.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -96,6 +98,54 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
             codeDocument.SetDocumentIntermediateNode(document);
         }
 
+        private void InheritImportedDirectives(DocumentIntermediateNode document, string mainDocumentFilePath)
```
rynowak: I don't like the term *inherit* showing up all of the sudden. We don't use this anywhere else and it's overloaded with other concepts in c#. I'd really prefer this not show up anywhere (comments, code, docs).

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -96,6 +98,54 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+            visitor.VisitDocument(document);
+
+            var seenDirectives = new HashSet<DirectiveDescriptor>();
+            for (var i = visitor.Directives.Count - 1; i >= 0; i--)
+            {
+                var reference = visitor.Directives[i];
+                var directive = (DirectiveIntermediateNode)reference.Node;
+                var descriptor = directive.Descriptor;
+                var seenDirective = !seenDirectives.Add(descriptor);
+                var imported = directive.Source.HasValue && directive.Source.Value.FilePath != mainDocumentFilePath;
```
rynowak: We don't need to use file paths here. We have the code document and we know what are imports and what aren't.

rynowak: BTW we **will** have bugs if we rely on file paths being unequal. It's possible for the main document to not have a location

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -96,6 +98,54 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+                    case DirectiveKind.RazorBlock:
+                    case DirectiveKind.CodeBlock:
+                        if (descriptor.Usage == DirectiveUsage.FileScopedSinglyOccurring)
+                        {
+                            // A block directive cannot be imported.
+
+                            document.Diagnostics.Add(
+                                RazorDiagnosticFactory.CreateDirective_BlockDirectiveCannotBeImported(descriptor.Directive));
+                        }
+                        break;
```
rynowak: Make sure to have a default case here, even if it just throws and says unreachable. Then if we add more kinds then we know this code has to be touched

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
rynowak: ok

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorIntermediateNodeLoweringPhaseIntegrationTest.cs
```diff
@@ -361,8 +361,8 @@ public void Lower_WithImports_Using()
 <p>Hi!</p>");
             var imports = new[]
             {
-                TestRazorSourceDocument.Create("@using System.Globalization"),
-                TestRazorSourceDocument.Create("@using System.Text"),
+                TestRazorSourceDocument.Create("@using System.Globalization", fileName: "import1.cshtml"),
+                TestRazorSourceDocument.Create("@using System.Text", fileName: "import2.cshtml"),
```
rynowak: This is why we shouldn't rely on filenames.

NTaylorMullen:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
NTaylorMullen: Done.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -96,6 +98,57 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
+                    default:
+                        Debug.Fail($"Unreachable code. This can happen when a new {typeof(DirectiveKind).FullName} is introduced.");
+                        break;
```
rynowak: This should throw

PR #1469, opened 2017-06-26
Suppress warnings for auto generated code

ajaybhargavb:
#1449 

Port of #1463 to dev branch. Note that we now mark both runtime and design time code as auto generated.

@NTaylorMullen @rynowak @pranavkm


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -99,7 +99,12 @@ public override void VisitDocument(DocumentIntermediateNode node)
                     }
                 }
 
+                Context.Writer.WriteLine("// <auto-generated/>");
+                Context.Writer.WriteLine("#pragma warning disable 1591");
```
NTaylorMullen: Chain writelines. Also, indent the Write's in the `IsNullOrEmpty` if above this.

PR #1471, opened 2017-06-27
Restore the VSIX project with MSBuild instead of NuGet.exe

natemcmaster:
Preparation for changes coming soon to a KoreBuild near you (https://github.com/aspnet/KoreBuild/issues/258).

This replaces the `nuget.exe restore` step with `msbuild /t:Restore`.

### DIFF
--- a/build/VSIX.targets
+++ b/build/VSIX.targets
@@ -1,14 +1,9 @@
 <Project>
   <PropertyGroup>
     <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);GenerateVSIX</PackageDependsOn>
-    <NuGetCommandLineVersion>4.1.0</NuGetCommandLineVersion>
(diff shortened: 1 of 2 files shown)


natemcmaster:
cc @ryanbrandenburg - this will unblock the build

build/VSIX.targets
```diff
@@ -44,7 +32,8 @@
       <VSIXOutputPath>$(BuildDir)$(VSIXName).vsix</VSIXOutputPath>
     </PropertyGroup>
 
-    <CallTarget Targets="_RestoreProject" />
+    <Exec Command="&quot;$(MSBuildExePath)&quot; &quot;$(VSIXProject)&quot; /nologo /t:Restore /v:m" />
```
pranavkm: Neat, didn't know this target was a thing.

natemcmaster: Yup. PackageReference in old-school csproj should work too once you have the latest public 15.3 preview. I didn't change to that yet, though, as this caused issues for the VSIX targets.

Issue #1361, opened 2017-05-18
Parser options should have the ability to parse all leading directives


rynowak:
Currently we parse just the first leading directive, we should parse all *simple* directives before the first markup or code block.

rynowak:
For 2.0.0 RTM let's document the quirk that this only reads the first line, and then address this in 2.1.

rynowak:
Added docs for 2.0.0 in e70ce98

rynowak:
Backlogging this until we have a need for it.

PR #1476, opened 2017-06-28
Add docs about limitation of this option

rynowak:
This is all the work we're planning to do for #1361 for 2.0.0. Will
revisit this functionality in 2.1.0.


rynowak:
FYI only docs update

Issue #1440, opened 2017-06-14
Improve DesignTimeDirectivePass


rynowak:
See https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/RazorDesignTimeIRPass.cs#L13

We want this pass to run later than it does today, so that passes have a chance to implement their own design time support by removing their directive.

If you leave your directive in, then it will have auto-generated design time support.

We should also merge this with the removal pass

ajaybhargavb:
ff433f72b88e6bab61ec530b828bd70f459da79a

PR #1481, opened 2017-06-28
Run DesignTimeDirectivePass later in the phase

ajaybhargavb:
#1440 

- Running DesignTimeDirectivePass after all other DirectiveClassifier passes.
- Modified `Section`, `Functions` and `Inherits` directive passes to not remove themselves so that the DesignTimeDirectivePass has a chance to add design time support. They are then removed by the `DirectiveRemovalOptimizationPass`.
- Updated some unit tests to reflect the changes.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
@@ -94,18 +94,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                 }
 
                 var baseType = visitor.Class?.BaseType?.Replace("<TModel>", "<" + modelType + ">");
-                for (var i = visitor.InheritsDirectives.Count - 1; i >= 0; i--)
-                {
-                    var directive = visitor.InheritsDirectives[i];
-                    var tokens = directive.Tokens.ToArray();
-                    if (tokens.Length >= 1)
-                    {
-                        baseType = tokens[0].Content.Replace("<TModel>", "<" + modelType + ">");
-                        tokens[0].Content = baseType;
-                        break;
-                    }
-                }
-
                 visitor.Class.BaseType = baseType;
             }
         }
@@ -115,8 +103,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             public NamespaceDeclarationIntermediateNode Namespace { get; private set; }
 
             public ClassDeclarationIntermediateNode Class { get; private set; }
-
-            public IList<DirectiveIntermediateNode> InheritsDirectives { get; } = new List<DirectiveIntermediateNode>();
 
             public IList<DirectiveIntermediateNode> ModelDirectives { get; } = new List<DirectiveIntermediateNode>();
 
@@ -145,10 +131,6 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                 if (node.Descriptor == Directive)
                 {
                     ModelDirectives.Add(node);
-                }
-                else if (node.Descriptor.Directive == "inherits")
-                {
-                    InheritsDirectives.Add(node);
                 }
             }
         }
--- a/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectivePass.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectivePass.cs
@@ -9,8 +9,9 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
     {
         internal const string DesignTimeVariable = "__o";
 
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectivePass.cs
```diff
@@ -9,8 +9,9 @@ internal class DesignTimeDirectivePass : IntermediateNodePassBase, IRazorDirecti
     {
         internal const string DesignTimeVariable = "__o";
 
-        // This needs to run before other directive classifiers.
-        public override int Order => -10;
+        // This needs to run after other directive classifiers. Any DirectiveToken that is not removed 
+        // by the previous classifiers will have auto-generated design time support.
+        public override int Order => 100;
```
ajaybhargavb: Is this too high?

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
```diff
@@ -259,6 +259,13 @@ private DocumentIntermediateNode CreateIRDocument(RazorEngine engine, RazorCodeD
                 }
             }
 
+            // InheritsDirectivePass needs to run before ModelDirective.
+            var pass = new InheritsDirectivePass()
```
ajaybhargavb: These unit tests were verifying the right code. Without the `InheritsDirectivePass` running, the class won't have a basetype set. Removing the unnecessary code in ModelDirective uncovered this problem.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -94,17 +94,6 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
                 }
 
                 var baseType = visitor.Class?.BaseType?.Replace("<TModel>", "<" + modelType + ">");
-                for (var i = visitor.InheritsDirectives.Count - 1; i >= 0; i--)
```
ajaybhargavb: This is unnecessary code. This was not getting executed because they will be no Inherits directives left at this point before this change. The unit tests were incorrectly verifying this code instead of the line above.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectivePass.cs
NTaylorMullen: Thinking too low. Hmm, since these are only ordered against other classifier passes I'm thinking this might make sense to be 1000 (`DefaultFeatureOrder`)

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
NTaylorMullen: So even if you remove this code the tests pass? If that's the case we need to update how the tests validate their expectations.

test/Microsoft.AspNetCore.Razor.Language.Test/Extensions/FunctionsDirectivePassTest.cs
```diff
@@ -69,7 +69,7 @@ public void Execute_MovesStatementsToClassLevel()
                 node => CSharpCode(" var value = true; ", node));
 
             var method = (MethodDeclarationIntermediateNode)@class.Children[0];
-            Assert.Empty(method.Children);
+            SingleChild<DirectiveIntermediateNode>(method);
```
NTaylorMullen: Validate this like you do the others bits in this test. It's now an add test, not a move.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -94,17 +94,6 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
...
-                {
-                    var directive = visitor.InheritsDirectives[i];
-                    var tokens = directive.Tokens.ToArray();
-                    if (tokens.Length >= 1)
-                    {
-                        baseType = tokens[0].Content.Replace("<TModel>", "<" + modelType + ">");
-                        tokens[0].Content = baseType;
-                        break;
-                    }
-                }
```
NTaylorMullen: Can remove this line.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/ModelDirectiveTest.cs
ajaybhargavb: No they don't.

ajaybhargavb:
🆙 📅

PR #1483, opened 2017-06-29
Added convenience properties to TagHelperIntermediateNode

ajaybhargavb:
#1329 

@NTaylorMullen @rynowak 

The reason behind doing this is to have all the parts of a tag helper tied to a single intermediate node. I have not made any changes to the code generation to use these properties. I will do that (if needed) after @rynowak's changes are in.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Intermediate/TagHelperIntermediateNode.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Intermediate/TagHelperIntermediateNode.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Razor.Language/Intermediate/TagHelperIntermediateNode.cs
```diff
@@ -49,6 +50,30 @@ public override RazorDiagnosticCollection Diagnostics
...
 
+        public TagHelperBodyIntermediateNode Body
+        {
+            get
+            {
+                return Children.OfType<TagHelperBodyIntermediateNode>().SingleOrDefault();
+            }
+        }
+
+        public IReadOnlyCollection<SetTagHelperPropertyIntermediateNode> SetTagHelperProperties
```
ajaybhargavb: Since we decided to rename these nodes, I did not put a lot of thought in to the property name.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/TagHelperIntermediateNode.cs
```diff
@@ -49,6 +50,30 @@ public override RazorDiagnosticCollection Diagnostics
...
+        {
+            get
+            {
+                return Children.OfType<SetTagHelperPropertyIntermediateNode>().ToList();
+            }
+        }
+
+        public IReadOnlyCollection<AddTagHelperHtmlAttributeIntermediateNode> AddTagHelperHtmlAttributes
```
NTaylorMullen: These should be IEnumerable.

Issue #1473, opened 2017-06-27
Update to RTM 2.3.0 of Roslyn


rynowak:

pranavkm:
We need to mirror the 2.6.0-* builds of RoslynDev first to avoid downgrade warnings.

pranavkm:
This is done. https://github.com/aspnet/Razor/commit/c812f9d8852856219619a29c5d58eedde9c37ee0

PR #1484, opened 2017-06-29
Update to StreamRpc 1.1.92

pranavkm:


rynowak:
:shipit:

pranavkm:
We're down to one warning in Razor. That should be fixed once we update to Roslyn 2.3.0 RTM (https://github.com/aspnet/Razor/issues/1473). In fact, we could try getting newer builds of Roslyn package (https://github.com/aspnet/CoreCLR/blob/dev/tools/MirrorLatestCoreCLR/restore-projects/Roslyn.proj#L5) and verify this goes away. 

Relevant Roslyn bug https://github.com/dotnet/roslyn/issues/16040 that says updating to `2.3.0-beta3` should address this.

mikeharder:
I chatted with @rynowak, and he agrees we should update to 2.3.0-beta3.  @pranavkm, can you make this chagne today?  Otherwise I can work on it.  I assume it requires changes to both the mirror and Razor, and the version should be set to `2.3.0-beta3-*`.

pranavkm:
I can get it updated.

Issue #1477, opened 2017-06-28
Enhancements to TagHelperDescriptorBuilders


rynowak:
- [ ] Convert lots of stuff to get/set properties 
- [x] Move TypeName and PropertyName to extension methods

ajaybhargavb:
7c7bb627b96b8c06e84baeecae81de9606b47aad

ajaybhargavb:
9c031e6fb571a616a69d9ddccd6abfa46260b5c0

PR #1488, opened 2017-06-30
Taskhancements to TagHelperDescriptor builders - part 2

ajaybhargavb:
#1477 

- Moved all TagHelperDescriptor related builders to have getter setter instead of methods except the methods that are needed to configure other builders (eg. `BindAttribute(Action<BoundAttributeDescriptorBuilder> configure)`)
- The methods in these builders are now not chainable
- Added necessary extensions methods in Test.Common so that the tests don't break
- **Pay extra attention** to the names of the new properties since I had to change some of them because they looked like method names

@NTaylorMullen @rynowak


src/Microsoft.AspNetCore.Razor.Language/DefaultTagMatchingRuleDescriptorBuilder.cs
```diff
@@ -10,38 +10,20 @@ namespace Microsoft.AspNetCore.Razor.Language
...
+        public override string RequireTagName { get; set; }
 
-            return this;
-        }
+        public override string RequireParentTag { get; set; }
 
-        public override TagMatchingRuleDescriptorBuilder RequireTagStructure(TagStructure tagStructure)
-        {
-            _tagStructure = tagStructure;
+        public override TagStructure RequireTagStructure { get; set; }
```
ajaybhargavb: Should we remove the `Require**` prefix on these properties?

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
```diff
@@ -32,9 +29,41 @@ public DefaultTagHelperDescriptorBuilder(string kind, string name, string assemb
...
+                if (_diagnostics == null)
+                {
+                    _diagnostics = new HashSet<RazorDiagnostic>();
+                }
+
+                return _diagnostics;
+            }
+        }
+
+        public override void BindAttribute(Action<BoundAttributeDescriptorBuilder> configure)
```
ajaybhargavb: @NTaylorMullen, I know we discussed exposing the underlying list as a readonly collection, but looking at how this turned out I felt like there was not much value in having that here. Plus there was nothing exposing it prior to this change. Am I missing something?

test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestTagHelperDescriptorBuilderExtensions.cs
```diff
@@ -14,7 +14,109 @@ public static TagHelperDescriptorBuilder TypeName(this TagHelperDescriptorBuilde
...
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder.Diagnostics.Add(diagnostic);
+
+            return builder;
+        }
+
+        public static TagHelperDescriptorBuilder BoundAttributeDescriptor(
```
ajaybhargavb: Note: I had to rename this and similar methods to something other that what we have in the product code to avoid collision. The compiler cannot handle overloads that are of the same name, arguments but different return type.

ajaybhargavb: This is the reason we have a lot of test changes.

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
```diff
@@ -42,23 +42,21 @@ public static TagHelperDescriptorBuilder Create(string kind, string name, string
...
+        public abstract string DisplayName { get; set; }
 
-        public abstract TagHelperDescriptorBuilder TagMatchingRule(Action<TagMatchingRuleDescriptorBuilder> configure);
+        public abstract string TagOutputHint { get; set; }
 
-        public abstract TagHelperDescriptorBuilder AllowChildTag(string allowedChild);
+        public abstract string Documentation { get; set; }
 
-        public abstract TagHelperDescriptorBuilder TagOutputHint(string hint);
+        public abstract ICollection<string> AllowedChildTags { get; }
```
NTaylorMullen: Thinking we might want to make this a custom collection to protect ourselves. @rynowak thoughts?

test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestBoundAttributeDescriptorBuilderExtensions.cs
```diff
@@ -0,0 +1,123 @@
...
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public static class TestBoundAttributeDescriptorBuilderExtensions
+    {
+        public static BoundAttributeDescriptorBuilder Name(this BoundAttributeDescriptorBuilder builder, string name)
+        {
+            if (builder == null)
```
NTaylorMullen: No need to null check our test helpers 😄

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorBuilder.cs
```diff
@@ -7,23 +7,24 @@ namespace Microsoft.AspNetCore.Razor.Language
...
+        public abstract string Name { get; set; }
 
-        public abstract BoundAttributeDescriptorBuilder Name(string name);
+        public abstract string TypeName { get; set; }
 
-        public abstract BoundAttributeDescriptorBuilder TypeName(string typeName);
+        public abstract bool IsEnum { get; set; }
 
-        public abstract BoundAttributeDescriptorBuilder AsEnum();
+        public abstract bool IsDictionary { get; set; }
```
rynowak: We can still have extension methods dog

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
ajaybhargavb: > I still think it'd be valuable to expose the builders as their public non-default type.

As read-only?

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorBuilder.cs
ajaybhargavb: I'll create an `AsDictionary()` extension method.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestBoundAttributeDescriptorBuilderExtensions.cs
ajaybhargavb: It was just copy paste. No time wasted 😄

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
```diff
@@ -15,9 +15,6 @@ internal class DefaultTagHelperDescriptorBuilder : TagHelperDescriptorBuilder
         private readonly string _assemblyName;
```
ajaybhargavb: Done

src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorBuilder.cs
ajaybhargavb: Filed https://github.com/aspnet/Razor/issues/1493

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
```diff
@@ -32,86 +29,92 @@ public DefaultTagHelperDescriptorBuilder(string kind, string name, string assemb
             _metadata = new Dictionary<string, string>(StringComparer.Ordinal);
         }
 
-        public override IDictionary<string, string> Metadata => _metadata;
+        public override string Name => _name;
```
NTaylorMullen: `{ get; }` and for kind/assemblyname (don't need the backing field)

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
```diff
@@ -32,86 +29,92 @@ public DefaultTagHelperDescriptorBuilder(string kind, string name, string assemb
...
-        public override TagHelperDescriptorBuilder TagMatchingRule(Action<TagMatchingRuleDescriptorBuilder> configure)
+        public override ICollection<string> AllowedChildTags
         {
-            if (configure == null)
+            get
             {
-                throw new ArgumentNullException(nameof(configure));
+                if (_allowedChildTags == null)
+                {
+                    _allowedChildTags = new HashSet<string>();
```
NTaylorMullen: ignorecase

src/Microsoft.AspNetCore.Razor.Language/DefaultTagHelperDescriptorBuilder.cs
```diff
@@ -32,86 +29,92 @@ public DefaultTagHelperDescriptorBuilder(string kind, string name, string assemb
...
+                    _diagnostics = new DefaultRazorDiagnosticCollection();
+                }
 
-            return this;
+                return _diagnostics;
+            }
         }
 
-        public override TagHelperDescriptorBuilder TagOutputHint(string hint)
+        public override IEnumerable<BoundAttributeDescriptorBuilder> BoundAttributes
```
NTaylorMullen: `IReadOnlyList` for builders

PR #1498, opened 2017-07-01
Check for warnrings in tests taht compile

rynowak:
This fixes and issue that snuck through and broke MVC as well as
preventing this from happening in the future.


rynowak:
FYI only, merging to fix mvc build

PR #1502, opened 2017-07-03
DiagnosticSource to 4.4.1

ryanbrandenburg:


NTaylorMullen:
This is good as long as this change is also making its way into MVC

NTaylorMullen:
Btw, looks like this package isn't available on linux? (travis failing)

ryanbrandenburg:
@NTaylorMullen all of the PR's like this are part of getting the Mirror passing because of a change in versioning made by the CLR team. It's failing because the mirror hasn't copied these packages yet because we lack these changes. I have a PR out for MVC too.

Issue #1503, opened 2017-07-03
Imported @page directive should log a diagnostic.


NTaylorMullen:
Today if you import an `@page` directive you end up with Razor thinking a view is a "page" but then MVC doesn't detect that. We should log a diagnostic indicating that imported `@page` directives are invalid.

pranavkm:
Couldn't we make it so that `@page` does not get imported?

NTaylorMullen:
@pranavkm You could but then you'd have silent errors in the system. I'd rather let it be imported and indicate that it's an error when a user attempts to use that imported view.

NTaylorMullen:
d36838ed883a84d9f85c8637c1f994c4e835086b

PR #1504, opened 2017-07-03
Handle imported @page directives.

NTaylorMullen:
- Prior to this imported `@page` directives would flow through the Razor system without error. This resulted in inconsistent behavior between MVC and Razor. Now, imported `@page` directives result in diagnostics on the page directive node.
- Added two tests to verify that we still treat views with imported page directives as Razor pages BUT we also log a diagnostic on the page directive node.
- Renamed the `ViewComponentDiagnosticFactory` class to `RazorExtensionsDiagnosticFactory` so it can be used for more than just view component diagnostics. This way we can ensure that our diagnostics don't end up overlapping.

#1503

/cc @pranavkm @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensionsDiagnosticFactory.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensionsDiagnosticFactory.cs
@@ -1,12 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.IO;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.Language;
 
 namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
-    internal class ViewComponentDiagnosticFactory
+    internal class RazorExtensionsDiagnosticFactory
     {
         private const string DiagnosticPrefix = "RZ";
 
@@ -95,6 +96,20 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
                 ViewComponentTypes.SyncMethodName,
                 tagHelperType,
                 nameof(Task));
+
+            return diagnostic;
+        }
+
+        public static readonly RazorDiagnosticDescriptor PageDirective_CannotBeImported =
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}3905",
+                () => Resources.PageDirectiveCannotBeImported,
+                RazorDiagnosticSeverity.Error);
+
+        public static RazorDiagnostic CreatePageDirective_CannotBeImported(SourceSpan source)
+        {
+            var fileName = Path.GetFileName(source.FilePath);
+            var diagnostic = RazorDiagnostic.Create(PageDirective_CannotBeImported, source, PageDirective.Directive.Directive, fileName);
 
             return diagnostic;
         }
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensionsDiagnosticFactory.cs
```diff
@@ -98,5 +98,17 @@ public static RazorDiagnostic CreateViewComponent_SyncMethod_CannotReturnTask(st
 
             return diagnostic;
         }
+
+        public static readonly RazorDiagnosticDescriptor PageDirective_CannotBeImported =
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}3905",
+                () => Resources.PageDirectiveCannotBeImported,
+                RazorDiagnosticSeverity.Error);
+        public static RazorDiagnostic CreatePageDirective_CannotBeImported(SourceSpan source)
```
pranavkm: New line before this.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensionsDiagnosticFactory.cs
```diff
@@ -98,5 +98,17 @@ public static RazorDiagnostic CreateViewComponent_SyncMethod_CannotReturnTask(st
 
             return diagnostic;
         }
+
+        public static readonly RazorDiagnosticDescriptor PageDirective_CannotBeImported =
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}3905",
+                () => Resources.PageDirectiveCannotBeImported,
```
pranavkm: wtf, why is the error message a Func?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Resources.resx
```diff
@@ -135,4 +135,7 @@
   <data name="MvcRazorParser_InvalidPropertyType" xml:space="preserve">
     <value>Invalid tag helper property '{0}.{1}'. Dictionary values must not be of type '{2}'.</value>
   </data>
+  <data name="PageDirectiveCannotBeImported" xml:space="preserve">
+    <value>Page directive cannot be imported.</value>
```
pranavkm: Better error message?

> `The `@page` directive specified in {0} file will not be imported. The directive must appear at the top of each Razor Page cshtml file.

or something like that

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensionsDiagnosticFactory.cs
NTaylorMullen: That's the signature of the diagnostic descriptor. Essentially it enables errors to have scoped values intertwined (taken from Roslyn).

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Resources.resx
NTaylorMullen: Will do all but the file part (that information is included in the source part of the error)

pranavkm: I was hoping we could include the {ImportsFileName}. Usually the error has the file the error originated in.

NTaylorMullen: Ok will do.

Issue #1448, opened 2017-06-20
@page should look at malformed directives and attempt to classify a document as a page.


NTaylorMullen:
With the new Razor error story we transform malformed `@page` directives into `MalformedDirectiveIRNode`s which in turn result in pages not being classified as Razor pages. This impacts the way Mvc routes to said pages; we should be smarter and also inspect malformed directives to determine our "pageness"

rynowak:
@NTaylorMullen - done?

NTaylorMullen:
@rynowak nope, waiting for your review on the PR.

NTaylorMullen:
a78202e9379728494c409c5eff378ec966dbce5e

PR #1506, opened 2017-07-04
Have `@page` understand malformed directives.

NTaylorMullen:
- The `@page` directive will now look for malformed directives and treat a view as a Razor page based on the existence of the malformed directive.
- Updated the `PageDirective` type to contain a reference to its responsible directive node.
- Added unit tests to validate malformed directives were picked up accordingly.
- Updated code generation test baselines to reflect the new malformed directive understanding.

#1448 

/cc @rynowak @pranavkm


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/PageDirective.cs
```diff
@@ -69,13 +75,25 @@ private static string TrimQuotes(string content)
 
         private class Visitor : IntermediateNodeWalker
         {
-            public DirectiveIntermediateNode DirectiveNode { get; private set; }
+            public IntermediateNode DirectiveNode { get; private set; }
+
+            public IEnumerable<DirectiveTokenIntermediateNode> DirectiveTokens { get; private set; }
```
pranavkm: Why set this? As opposed to just doing `node.DirectiveTokens` at the callsite?

NTaylorMullen: The node is an `IntermediateNode` because it can be a malformed node or a non-malformed node.

NTaylorMullen:
> Can you have a PR for https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageDirectiveFeature.cs#L13 ready since this would mess up the logging here (https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageDirectiveFeature.cs#L51)?

Will do.

PR #1507, opened 2017-07-05
Streamline collections

rynowak:
Makes our collections sealed instead of abstract.

Only the IntermediateNodeCollection needs to have a read only variant.


src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeCollection.cs
```diff
@@ -1,40 +1,239 @@
...
 using System.Collections;
 using System.Collections.Generic;
 
 namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
-    public abstract class IntermediateNodeCollection : IList<IntermediateNode>
+    public sealed class IntermediateNodeCollection : IList<IntermediateNode>
     {
-        public abstract IntermediateNode this[int index] { get; set; }
+        public static readonly IntermediateNodeCollection ReadOnly = new IntermediateNodeCollection(null);
```
NTaylorMullen: use a readonly collection instead of passing around null

src/Microsoft.AspNetCore.Razor.Language/ReadOnlyItemCollection.cs
```diff
@@ -1,93 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Generic;
-using System.Linq;
-
-namespace Microsoft.AspNetCore.Razor.Language
-{
-    internal class ReadOnlyItemCollection : ItemCollection
```
NTaylorMullen: lol interesting, shocked we didn't use this anywhere.

PR #1509, opened 2017-07-05
Remove explicit scope managment

rynowak:


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CodeRenderingContext.cs
```diff
@@ -30,36 +31,14 @@ public abstract class CodeRenderingContext
...
         public abstract RazorSourceDocument SourceDocument { get; }
 
-        public abstract Scope CreateScope();
-
-        public abstract Scope CreateScope(IntermediateNodeWriter writer);
-
-        public abstract void EndScope();
+        public abstract void AddLineMappingFor(IntermediateNode node);
 
         public abstract void RenderNode(IntermediateNode node);
```
NTaylorMullen: Could do without this now (and the RenderChildren variant).

Issue #1510, opened 2017-07-06
Razor.Language API polish


NTaylorMullen:
#### General
- [x] Make `DocumentClassifierPassBase.TargetExtensions` private. - @NTaylorMullen
- [x] `IRazorEngineBuilder.DesignTime` should be getter only. - @NTaylorMullen
- [x] ~`RazorCSharpDocument.GeneratedCode` should be renamed to `Code`.~ - @NTaylorMullen - Unimportant/too cross-cutting for repos.
- [x] Rename `RazorEngineFeatureBase.ThrowForMissingEngineDependency` to `ThrowForMissingFeatureDependency` - @NTaylorMullen
- [x] Rename `RazorEnginePhaseBase.ThrowForMissingEngineDependency` to `ThrowForMissingFeatureDependency` - @NTaylorMullen
- [x] Find all instances of base / path and rename them to FilePath / BasePath - @ajaybhargavb
- [x] Abstractify `ICodeTargetBuilder`. - @ajaybhargavb
- [x] `IntermediateNodeWalker` should be indexable - @ajaybhargavb
- [x] Do a pass on all `IntermediateNode`s and cross-check the verbosity of their properties. They should be consistent which may mean having more verbose properties. - @ajaybhargavb
- [x] Unnest `IntermediateToken.TokenKind` - @ajaybhargavb
- [x] Make `TagHelperIntermediateNode.TagHelpers` an IList - @ajaybhargavb

#### HtmlConventions.InvalidNonWhitespaceHtmlCharacters - @ajaybhargavb
- [x] Make internal.
- [x] Make non-hashset.

#### LineMapping - @NTaylorMullen
- [x] Rename to `SourceMapping`.
- [x] Remove `==` and `!=` operators.

#### RazorParserOptions - @NTaylorMullen
- [x] Add `CreateDesignTime` method 
- [x] Rename `ParseOnlyLeadingDirectives` => `ParseLeadingDirectives`

#### RazorParserOptionsBuilder - @NTaylorMullen
- [x] `DesignTime` should be getter only.
- [x] Rename `ParseOnlyLeadingDirectives` => `ParseLeadingDirectives`

#### TagHelperBinding - @NTaylorMullen
- [x] Attributes should be IReadOnlyList if it makes sense.
- [x] GetBoundRules should return IReadOnlyList if it makes sense.

#### TagHelper descriptors in general - @NTaylorMullen
[...]

PR #1514, opened 2017-07-06
Make `RazorParserOptions.DesignTime` getter only.

NTaylorMullen:
- Renamed `IRazorParserOptionsFeature` to `IConfigureRazorParserOptionsFeature`, the original interface was re-purposed to get the options rather than configure them.
- This involved re-designing how we set design time on the `RazorParserOptions` object. The indicator of `DesignTime` is now configured at the RazorEngine level via a parser options provider feature.
- Moved options construction from the phase into an `IRazorParserOptionsFeature` type.

#1510 

This required a design change so looking for feedback.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DesignTimeOptionsFeature.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DesignTimeOptionsFeature.cs
@@ -5,19 +5,9 @@ using System;
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    internal class DesignTimeOptionsFeature : RazorEngineFeatureBase, IRazorParserOptionsFeature, IRazorCodeGenerationOptionsFeature
+    internal class DesignTimeOptionsFeature : RazorEngineFeatureBase, IRazorCodeGenerationOptionsFeature
     {
         public int Order { get; set; }
-
-        public void Configure(RazorParserOptionsBuilder options)
-        {
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.Razor.Language/DesignTimeOptionsFeature.cs
```diff
@@ -5,20 +5,10 @@
...
-        {
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
-            options.DesignTime = true;
-        }
-
         public void Configure(RazorCodeGenerationOptionsBuilder options)
```
NTaylorMullen: Will probably end up moving the code generation options to follow the same pattern as the parser options which would result in this class going away.

src/Microsoft.AspNetCore.Razor.Language/IRazorParserOptionsFeature.cs
```diff
@@ -5,8 +5,6 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     public interface IRazorParserOptionsFeature : IRazorEngineFeature
     {
-        int Order { get; }
-
-        void Configure(RazorParserOptionsBuilder options);
+        RazorParserOptions GetOptions();
```
NTaylorMullen: This is the same design as our TagHelper feature bits.

NTaylorMullen:
Will get an MVC reaction done in parallel

NTaylorMullen:
Getting this in to unblock other PRs.

PR #1523, opened 2017-07-07
Adjusted verbosity of some property names and made

ajaybhargavb:
TagHelperIntermediateNode.TagHelpers and IList

#1510 

@NTaylorMullen @rynowak, I think we already agreed on these renames. So I'll merge this now. But do have a quick look to see if I missed something.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
@@ -55,7 +55,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 
             if (document.DocumentKind == RazorPageDocumentClassifierPass.RazorPageDocumentKind)
             {
-                return visitor.Class.Name;
+                return visitor.Class.ClassName;
             }
             else
             {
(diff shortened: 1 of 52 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -128,7 +128,7 @@ public override void VisitClassDeclaration(ClassDeclarationIntermediateNode node
 
             public override void VisitDirective(DirectiveIntermediateNode node)
             {
-                if (node.Descriptor == Directive)
```
rynowak: Agree with this, we shouldn't use `descriptor` as an identifier

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelDirective.cs
```diff
@@ -55,7 +55,7 @@ private static string GetModelType(DocumentIntermediateNode document, Visitor vi
 
             if (document.DocumentKind == RazorPageDocumentClassifierPass.RazorPageDocumentKind)
             {
-                return visitor.Class.Name;
```
rynowak: Agree with this, using `ClassName` means it can be consistent everywhere

PR #1526, opened 2017-07-07
Made IntermediateNodeWalker.Ancestors indexable

ajaybhargavb:
#1510 

@rynowak @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeWalker.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeWalker.cs
@@ -7,11 +7,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
     public abstract class IntermediateNodeWalker : IntermediateNodeVisitor
     {
-        private readonly Stack<IntermediateNode> _ancestors = new Stack<IntermediateNode>();
+        private readonly List<IntermediateNode> _ancestors = new List<IntermediateNode>();
 
-        protected IEnumerable<IntermediateNode> Ancestors => _ancestors;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/Intermediate/IntermediateNodeWalker.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Intermediate
 {
     public abstract class IntermediateNodeWalker : IntermediateNodeVisitor
     {
-        private readonly Stack<IntermediateNode> _ancestors = new Stack<IntermediateNode>();
+        private readonly List<IntermediateNode> _ancestors = new List<IntermediateNode>();
 
-        protected IEnumerable<IntermediateNode> Ancestors => _ancestors;
```
rynowak: Unit test?

ajaybhargavb: There is already a unit test that verifies if the ancestors are set correctly. Did you mean add one that tests indexability?

rynowak: Does it verify the iteration order? If it does then consider me signed off.

ajaybhargavb: Yes it does 👍 https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Razor.Language.Test/Intermediate/IntermediateNodeWalkerTest.cs#L68.

PR #1527, opened 2017-07-07
A different design for extensibility of TagHelpers

rynowak:


test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ModelExpressionTagHelper_DesignTime.ir.txt
```diff
@@ -43,7 +43,7 @@ Document -
                 TagHelper - (66:4,0 [25] ModelExpressionTagHelper.cshtml) - input-test - TagMode.SelfClosing
                     DefaultTagHelperBody - 
                     DefaultTagHelperCreate -  - InputTestTagHelper
-                    DefaultTagHelperProperty - (83:4,17 [4] ModelExpressionTagHelper.cshtml) - for - Microsoft.AspNetCore.Mvc.ViewFeatures.ModelExpression InputTestTagHelper.For - HtmlAttributeValueStyle.DoubleQuotes
+                    DefaultTagHelperProperty - (83:4,17 [4] ModelExpressionTagHelper.cshtml) - For - __InputTestTagHelper - for - Microsoft.AspNetCore.Mvc.ViewFeatures.ModelExpression InputTestTagHelper.For - HtmlAttributeValueStyle.DoubleQuotes
```
rynowak: The changes to IR are because I added new information that wasn't getting into the baseline before.

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/PrefixedAttributeTagHelpers_Runtime.codegen.cs
```diff
@@ -181,10 +181,6 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_PrefixedAt
             __tagHelperExecutionContext.AddTagHelperAttribute(__tagHelperAttribute_5);
             __TestNamespace_InputTagHelper2.StringDictionaryProperty["paprika"] = (string)__tagHelperAttribute_5.Value;
             __tagHelperExecutionContext.AddTagHelperAttribute(__tagHelperAttribute_5);
-            if (__TestNamespace_InputTagHelper1.StringDictionaryProperty == null)
```
rynowak: This is the only change to codegen, it removed a redundant check on this dictionary

rynowak: See line 176

NTaylorMullen:
Not sure I like this approach. The more I look at `ExtensionIntermediateNode` the more I think it makes sense as an interface. Today `ExtensionIntermediateNode` type only exposes two things:

- abstract WriteNode
- protected ReportMissingCodeTargetExtension

It also has a static AcceptExtensionNode:
```C#
protected static void AcceptExtensionNode<TNode>(TNode node, IntermediateNodeVisitor visitor)
    where TNode : ExtensionIntermediateNode
{
    var typedVisitor = visitor as IExtensionIntermediateNodeVisitor<TNode>;
    if (typedVisitor == null)
    {
        visitor.VisitExtension(node);
    }
    else
    {
        typedVisitor.VisitExtension(node);
    }
}
```

We could move `ReportMissingCodeTargetExtension` somewhere else and create an `IIntermediateNodeExtension` interface.

rynowak:
We discussed this and think that this isn't finished yet.

Neither the current approach nor this one really support our future needs for multi-targeting.

PR #1529, opened 2017-07-07
Made `RazorCodeGenerationOptionsBuilder.DesignTime` getter only

ajaybhargavb:
- Made RazorCodeGenerationOptions consistent with RazorParserOptions

#1510 

@NTaylorMullen @rynowak

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
@@ -13,11 +13,11 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     internal class DefaultRazorIntermediateNodeLoweringPhase : RazorEnginePhaseBase, IRazorIntermediateNodeLoweringPhase
     {
-        private IRazorCodeGenerationOptionsFeature[] _optionsCallbacks;
+        private IRazorCodeGenerationOptionsFeature _optionsFeature;
 
         protected override void OnIntialized()
         {
-            _optionsCallbacks = Engine.Features.OfType<IRazorCodeGenerationOptionsFeature>().OrderBy(f => f.Order).ToArray();
+            _optionsFeature = GetRequiredFeature<IRazorCodeGenerationOptionsFeature>();
         }
 
         protected override void ExecuteCore(RazorCodeDocument codeDocument)
@@ -31,7 +31,7 @@ namespace Microsoft.AspNetCore.Razor.Language
             var document = new DocumentIntermediateNode();
             var builder = IntermediateNodeBuilder.Create(document);
 
-            document.Options = CreateCodeGenerationOptions();
+            document.Options = _optionsFeature.GetOptions();
 
             var namespaces = new Dictionary<string, SourceSpan?>(StringComparer.Ordinal);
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -13,11 +13,12 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     internal class DefaultRazorIntermediateNodeLoweringPhase : RazorEnginePhaseBase, IRazorIntermediateNodeLoweringPhase
     {
-        private IRazorCodeGenerationOptionsFeature[] _optionsCallbacks;
+        private IRazorCodeGenerationOptionsFeature _optionsFeature;
+
```
NTaylorMullen: extra line.

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorIntermediateNodeLoweringPhaseIntegrationTest.cs
```diff
@@ -431,18 +433,35 @@ public void Lower_WithImports_IgnoresBlockDirective()
...
         {
             tagHelpers = tagHelpers ?? new TagHelperDescriptor[0];
 
-            var engine = RazorEngine.Create(b =>
+            RazorEngine engine = null;
+            if (designTime)
             {
-                builder?.Invoke(b);
+                engine = RazorEngine.CreateDesignTime(b =>
+                {
```
NTaylorMullen: Lift this into a private function: `ConfigureEngine` and re-use.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/CodeGeneration/TestCodeRenderingContext.cs
```diff
@@ -15,7 +15,7 @@ public static class TestCodeRenderingContext
         {
             var codeWriter = new CodeWriter();
             var documentNode = new DocumentIntermediateNode();
-            var options = RazorCodeGenerationOptions.CreateDesignTimeDefault();
+            var options = RazorCodeGenerationOptions.CreateDesignTime(c => { });
```
NTaylorMullen: Ah, this is why you wanted a `CreateDesignTimeDefault`. It's a little ugly for tests but I think it's worth the lesser public API (people other than us will never need to do this).

As a side note, for lambdas where you don't use the variable it's common practice to use the `_` character

rynowak: > As a side note, for lambdas where you don't use the variable it's common practice to use the _ character

This is absolutely not the case in C# and has never been part of our teams' guidelines. `_` is now a reserved word in some contexts, we shouldn't sprinkle it all over

src/Microsoft.AspNetCore.Razor.Language/RazorCodeGenerationOptions.cs
```diff
@@ -1,23 +1,43 @@
...
+            }
+
+            var builder = new DefaultRazorCodeGenerationOptionsBuilder(designTime: false);
+            configure(builder);
+            var options = builder.Build();
+
+            return options;
         }
 
-        public static RazorCodeGenerationOptions CreateDesignTimeDefault()
```
rynowak: 👎

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/DefaultCodeTargetTest.cs
```diff
@@ -31,7 +31,7 @@ public void Constructor_CreatesDefensiveCopy()
         public void CreateWriter_DesignTime_CreatesDesignTimeNodeWriter()
         {
             // Arrange
-            var options = RazorCodeGenerationOptions.CreateDesignTimeDefault();
```
rynowak: Why take this away? If we're going to have `CreateDefault` we should have this

Issue #1107, opened 2017-03-22
Track recent breaking changes in `exceptions.net*.json` files


dougbu:
This repo contains untracked breaking changes. To find and address them
1. Revert 6fbbd5d
2. Run `.\build.cmd /t:Pack` in projects 6fbbd5d touched or `.\build.cmd /t:Package` in the repo root
3. Update `exceptions.net45.json` and `exceptions.netcore.json` to track the changes
4. If not already announced, suggest filing issues at https://github.com/aspnet/Announcements

dougbu:
In addition the following projects lack baselines. The details below are reformatted from a recent Universe build on the CI. (Might make sense to disable API Check for these projects until they are released to the public.)
```
[Razor\src\Microsoft.AspNetCore.Mvc.Razor.Extensions\Microsoft.AspNetCore.Mvc.Razor.Extensions.csproj]
  No baseline file for net46 found in Microsoft.AspNetCore.Mvc.Razor.Extensions.
  No baseline file for netstandard1.3 found in Microsoft.AspNetCore.Mvc.Razor.Extensions.
[Razor\src\Microsoft.AspNetCore.Razor.Evolution\Microsoft.AspNetCore.Razor.Evolution.csproj]
  No baseline file for netstandard1.3 found in Microsoft.AspNetCore.Razor.Evolution.
[Razor\src\Microsoft.CodeAnalysis.Razor.Workspaces\Microsoft.CodeAnalysis.Razor.Workspaces.csproj]
  No baseline file for netstandard1.3 found in Microsoft.CodeAnalysis.Razor.Workspaces.
[Razor\src\Microsoft.CodeAnalysis.Razor\Microsoft.CodeAnalysis.Razor.csproj]
  No baseline file for netstandard1.3 found in Microsoft.CodeAnalysis.Razor.
[Razor\src\Microsoft.CodeAnalysis.Remote.Razor\Microsoft.CodeAnalysis.Remote.Razor.csproj]
  No baseline file for net46 found in Microsoft.CodeAnalysis.Remote.Razor.
[Razor\src\Microsoft.VisualStudio.LanguageServices.Razor\Microsoft.VisualStudio.LanguageServices.Razor.csproj]
  No baseline file for net46 found in Microsoft.VisualStudio.LanguageServices.Razor.
[Razor\src\RazorPageGenerator\RazorPageGenerator.csproj]
  No baseline file for netcoreapp1.0 found in RazorPageGenerator.
```

[...]

PR #1531, opened 2017-07-07
Update repo to utilize ApiCheck.

NTaylorMullen:
- Re-enabled api check for `Microsoft.AspNetCore.Razor` and `Microsoft.AspNetCore.Razor.Runtime`. This resulted in me adding known breaking changes for the packages.
- Added empty baseline files for `Microsoft.AspNetCore.Razor.Language`, `Microsoft.AspNetCore.Mvc.Razor.Extensions` and `Microsoft.CodeAnalysis.Razor`.
- Disabled ApiCheck for `Microsoft.CodeAnalysis.Razor.Workspaces`, `Microsoft.CodeAnalysis.Remote.Razor`, `RazorPageGenerator` and `Microsoft.VisualStudio.LanguageServices.Razor` to prevent ApiCheck warnings about missing baselines (we don't ship these packages).

#1107 

Also as a side note, there's a baseline.net45.json in the `Microsoft.AspNetCore.Razor` project; it doesn't seem to be used at all (the project doesn't cross-compile). Thought about removing it but I'm not aware of any cross-cutting concerns with it.


rynowak:
Can we get rid of the old net45 baseline file? It's not doing anything if it's always ignored

NTaylorMullen:
> Can we get rid of the old net45 baseline file? It's not doing anything if it's always ignored

I'd like to but want to make sure there's not a cross-cutting/cross-repo piece I'm missing. Who would be the right person to run that by?

NTaylorMullen:
> Can we get rid of the old net45 baseline file? It's not doing anything if it's always ignored

Removed it. Looked around/asked around.

rynowak:
:shipit:

PR #1532, opened 2017-07-10
Update our VSIX version to 15.5

rynowak:


rynowak:
FYI only - will merge once CI passes

PR #1549, opened 2017-07-12
Re-introduce RazorEditorParser.

NTaylorMullen:
- Revived `RazorEditorParser`.
- Made `PartialParseResult` internal and renamed it to `PartialParseResultInternal`. This fell in line with other syntax tree types.
- Moved the `RazorEditorParser` implementation away from `TextChange` and `ITextBuffer`. Instead it now relies on `SourceChange` and the VS contract `ITextSnapshot`.
- Added `RazorEditorParserTest` to ensure the changes in implementation did not impact previous functionality.
- Removed some obvious tests that unnecessarily re-tested behavior that was already verified.
- Updated tests.
- Moved several Language.Test types to the common test project so they could be reused.

#1259


src/Microsoft.VisualStudio.LanguageServices.Razor/TextSnapshotSourceDocument.cs
```diff
@@ -0,0 +1,79 @@
...
+                return;
+            }
+
+            for (var i = 0; i < count; i++)
+            {
+                destination[destinationIndex + i] = this[sourceIndex + i];
+            }
+        }
+
+        public override byte[] GetChecksum() => throw new NotImplementedException();
```
rynowak: Will this just not happen because it's design time?

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Test/TestBoundAttributeDescriptorBuilderExtensions.cs
```diff
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNetCore.Razor.Language;
 
-namespace Microsoft.AspNetCore.Razor.Language
+namespace Microsoft.VisualStudio.LanguageServices.Razor.Test
```
rynowak: You can drop `.Test`

src/Microsoft.VisualStudio.LanguageServices.Razor/TextSnapshotSourceDocument.cs
NTaylorMullen: Correct.

PR #1556, opened 2017-07-16
Fix #1555

### LINKED ISSUE #1555: Consider reverting diagnostic from placing a file-scoped directive after a code block
We currently add a diagnostic for cases like this:
```
@{
}
@model
```

Unfortunately we also have item templates like:
```
@*
*@
@{
}
```

So if you're going to start by adding a directive to this file, you have to know to put it before the code block. I'm not sure that we've had enough bake time for this change, and I'm worried about breaking apps that are upgrading, as well as surprising authors of new code.

rynowak:
Revert validation that file-scoped directives appear before HTML or
code.


test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -35,7 +35,7 @@ protected IntegrationTestBase()
 #if GENERATE_BASELINES
         protected bool GenerateBaselines { get; set; } = true;
 #else
-        protected bool GenerateBaselines { get; set; } = false;
+        protected bool GenerateBaselines { get; set; } = true;
```
NTaylorMullen: undo

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1787,39 +1787,6 @@ private void ValidateDirectiveUsage(DirectiveDescriptor descriptor)
...
-                    }
-
-                    UsageError(Resources.FormatDirectiveMustExistBeforeMarkupOrCode(descriptor.Directive));
-                    return;
-                }
-            }
-
-            return;
-
             void UsageError(string message)
```
ajaybhargavb: Could just inline this now.

PR #1557, opened 2017-07-18
Generate full pdbs for desktop packages

### LINKED ISSUE #1548: Ensure we have full pdbs for binaries that we ship in VS
The SDK defaults to producing portable pdbs. VS currently requires regular pdbs for symbol archival. We'll need to ensure we produce the right kind of pdbs for these for the 15.5 Razor release.

pranavkm:
Fixes #1548

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -12,6 +12,9 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+
+    <!-- Generate full pdbs for desktop targeting projects on platforms that support generating full pdbs. -->
+    <DebugType Condition="'$(OS)'=='Windows_NT' AND '$(TargetFramework)'=='net46'">full</DebugType>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


build/common.props
```diff
@@ -12,6 +12,9 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+
+    <!-- Generate full pdbs for desktop targeting projects on platforms that support generating full pdbs. -->
+    <DebugType Condition="'$(OS)'=='Windows_NT' AND '$(TargetFramework)'=='net46'">full</DebugType>
```
rynowak: Is it possible to make it so that this will make noise if we target another `net4X`?

rynowak:
We need to port this to rel/15.3 also

build/common.props
pranavkm: Oddly using `"TargetFrameworkIdentifier"==".NETFramework"` didn't generate full pdbs. `TargetFramework` was the only thing that worked. We have other instances of `net46` in our proj files so it's unlikely we'll miss this when we change update.

pranavkm:
Ported to 15.3 here: https://github.com/aspnet/Razor/commit/e30b09f08e554e0b860d8e1fdf4dd9e08e6f5293

PR #1559, opened 2017-07-19
Update product code to work with editor.

NTaylorMullen:
Primary points in this PR are:
- Revert Roslyn dev version back to 2.3.0-bet4-*.
- Do work to enable the Razor editor for 1.0.0 targeting projects (remove the `RazorViewAttribute`).
- Change `TagHelperDescriptorProvider`s to not add `null` to the list of descriptors.


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ViewComponentTagHelperDescriptorProvider.cs
```diff
@@ -55,7 +55,12 @@ public void Execute(TagHelperDescriptorProviderContext context)
             var factory = new ViewComponentTagHelperDescriptorFactory(compilation);
             for (var i = 0; i < types.Count; i++)
             {
-                context.Results.Add(factory.CreateDescriptor(types[i]));
+                var descriptor = factory.CreateDescriptor(types[i]);
+
+                if (descriptor != null)
```
NTaylorMullen: Today this can never be `null`. Just being overly careful.

rynowak: OK. Seems good to do anyway

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs
```diff
@@ -29,10 +29,13 @@ public static void Register(IRazorEngineBuilder builder)
             builder.Features.Add(new ViewComponentTagHelperPass());
             builder.Features.Add(new RazorPageDocumentClassifierPass());
             builder.Features.Add(new MvcViewDocumentClassifierPass());
-            builder.Features.Add(new AssemblyAttributeInjectionPass());
 
             if (!builder.DesignTime)
             {
+                // This is a temporary workaround until we have support for multi-targeting. If this were to be added
+                // then in 1.0.0 applications you'd get an error about missing attributes at design time.
+                builder.Features.Add(new AssemblyAttributeInjectionPass());
```
rynowak: I actually think this is fine as a non-workaround. We don't need this in the design time code.

test/Microsoft.CodeAnalysis.Razor.Test/DefaultTagHelperDescriptorProviderTest.cs
```diff
@@ -0,0 +1,38 @@
...
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class DefaultTagHelperDescriptorProviderTest
+    {
+        private static readonly Assembly _assembly = typeof(DefaultTagHelperDescriptorProviderTest).GetTypeInfo().Assembly;
+
+        [Fact]
+        public void Execute_DoesNotAddEditorBrowsableNeverDescriptors()
```
rynowak: 👍

pranavkm:
We used to have `2.3.0-*` which got lifted to 2.3.0 for the ones that were on NuGet.org resulting in the downgrade. Maybe the `-beta4-*` makes the difference.

NTaylorMullen:
> So do we really have no warnings with the 2.3.0-beta4 packages? @pranavkm told me we did before

No warnings, sorry forgot to add that in the PR comments.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs
NTaylorMullen: Alright, will remove comment 😄

Issue #1497, opened 2017-07-01
Make Razor Codegen warning safe


rynowak:
We weren't checking for warnings in the Razor tests that do compilation, which allows some warnings to slip through and break the build. 

We need to do some work to make sure we're clean.

Currently known issues:
- [ ] test:_ViewImports_DesignTime
- [ ] test:_ViewImports_DesignTime
- [ ] find a way to run with actual compilation in more of our tests so we can find more

ajaybhargavb:
From @NTaylorMullen,

For additional testing of Microsoft.AspNetCore.Razor.Language.Test. I can see two paths for enabling compilation in our language tests:

Re-use mvc shim and compile the language.test cases.
Create a new, non-mvc shim to compile the language.tests against.
The second is technically more correct but is a much larger time investment; i'd stick with number 1 since it'd have a large benefit for little time investment. Lets talk about it more on Monday.

rynowak:
my preference would be to invest in making another shim, but doing it far in the future. I'm going to put this item back in the backlog for now.

PR #1585, opened 2017-07-29
Removed warning in ViewImports integration test

ajaybhargavb:
Part of #1497 

Get rid of warning for _ViewImports test.


ajaybhargavb:
Agreed. I'll move this comment to the issue so this doesn't get lost after the merge.

PR #1590, opened 2017-07-31
Roslyn-ize our tooling contracts.

NTaylorMullen:
### General

Contracts aren't in their final locations. This PR is more of an intermediary step in enabling Roslyn services ontop of our current contracts and then start discussing where contracts should live (see next steps below).

- Added `ILanguageServiceFactory` types for all serviceable contracts.
- Added temporary MEF compatible types to VS.LanguageServices.Razor. These will enable the editor to slowly consume new bits.
- Added a Workspaces.Test project since some implementations moved.
- Updated the `RazorSyntaxFactsService` to not depend on VisualStudio. Added an extension type to the VS.LanguageServices.Razor project to enable easy-access from VisualStudio.

Platform dependent services were **entirely** moved from VisualStudio.Razor => Razor.Workspaces. They're in `ServiceLayer.Editor` and are:
- `RazorSyntaxFactsService`, this included mutating the API to not depend on VisualStudio and moving some primitive types such as `AcceptedCharacters`.
- `TagHelperCompletionService`
- `TagHelperFactsService`

Project dependent services were partially moved. They sit in `ServiceLayer.Default` and are:
- `RazorTemplateEngineFactoryService`. The Default implementation and its corresponding factory still live in VisualStudio.Razor. This way Razor.Workspaces can get by without a reference to Mvc.Razor.Extensions.
- `ITagHelperResolver` used to exist in VisualStudio.Razor. Removed the type and replaced its usage with the already-existing `TagHelperResolver` type in Razor.Workspaces. Both contracts were nearly identical.

### Potential Improvements/next steps

The biggest take away from this work is that we can separate our contracts into two categories: 
1. Platform dependent tooling; i.e. `RazorSyntaxFactsService`, `TagHelperCompletionService`, `TagHelperFactsService`
2. Project dependent tooling; i.e. `TagHelperResolver`, `RazorTemplateEngineFactoryService`

We don't currently have any plans to enable extensibility for platform dependent tooling so these contracts can sit next to their default implementations. 

As for project dependent tooling, it's important that we separate core contracts and their implementations; reason being, the default implementations will be tied to MVC and we don't want our core tooling packages to require the Mvc extension.

Given these two categories I think it makes sense to have the following structure to support both categories but also make things like API multitargeting clean:
image

In the diagram above it shows a dependency from Remote -> Mvc.Workspaces. This would be temporary until we built a system that didn't require it. 

This would be follow up work but I wanted to get some groundwork in PR form before I commit to such drastic changes 😄 

@rynowak @ajaybhargavb I'm most curious about your reactions to these proposed "next steps".


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(TagHelperResolver), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultTagHelperResolverFactory : ILanguageServiceFactory
+    {
+        private readonly VisualStudioWorkspace _workspace;
+        private readonly SVsServiceProvider _services;
+
+        [ImportingConstructor]
+        public DefaultTagHelperResolverFactory(VisualStudioWorkspace workspace, SVsServiceProvider services)
```
NTaylorMullen: When this guy gets moved out of the VS package we'll make an abstraction for the `SVsServiceProvider` that's non-VS specific.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorSyntaxFactsServiceExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    public static class RazorSyntaxFactsServiceExtensions
+    {
+        public static int? GetDesiredIndentation(
```
NTaylorMullen: Given that we have the VS types to our disposal; might as well do more work on their behalf. This is consistent with the contract they work with today.

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorSyntaxFactsService.cs
```diff
@@ -1,18 +1,19 @@
...
 {
-    public abstract class RazorSyntaxFactsService
+    public abstract class RazorSyntaxFactsService : ILanguageService
     {
         public abstract IReadOnlyList<ClassifiedSpan> GetClassifiedSpans(RazorSyntaxTree syntaxTree);
 
         public abstract IReadOnlyList<TagHelperSpan> GetTagHelperSpans(RazorSyntaxTree syntaxTree);
 
-        public abstract int? GetDesiredIndentation(RazorSyntaxTree syntaxTree, ITextSnapshot syntaxTreeSnapshot, ITextSnapshotLine line, int indentSize, int tabSize);
+        public abstract int? GetDesiredIndentation(RazorSyntaxTree syntaxTree, int previousLineEnd, Func<int, string> lineProvider, int indentSize, int tabSize);
```
rynowak: I don't think this method fits on this service. We should think about splitting it off

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -17,8 +17,8 @@
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    [Export(typeof(ITagHelperResolver))]
-    internal class DefaultTagHelperResolver : ITagHelperResolver
+    [Export(typeof(TagHelperResolver))]
```
rynowak: Should this be removed?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
rynowak: This doesn't need to be moved out of the VS package though.. This **is** the VS-specific implementation

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+            _services = services;
+        }
+
+        public ILanguageService CreateLanguageService(HostLanguageServices languageServices)
+        {
+            return new DefaultTagHelperResolver()
+            {
+                Workspace = _workspace,
+                Services = _services,
+            };
```
rynowak: We should use constructor injection

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
rynowak: The *right* way to do this is to provide a non-OOP implementation in `ServiceLayer.Default` and then provide the VS implementation in `ServiceLayer.Host`.

src/Microsoft.VisualStudio.LanguageServices.Razor/VisualStudioRazorSyntaxFactsService.cs
```diff
@@ -1,15 +1,13 @@
...
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using Microsoft.CodeAnalysis;
+using System.Composition;
 using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    public interface ITagHelperResolver
+    [Export(typeof(RazorSyntaxFactsService))]
```
rynowak: It would be nice to have a big scary banner here saying that this is legacy and should be transitioned away from (on all of these)

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
rynowak: AH nvm, I get it.

tooling/Microsoft.VisualStudio.RazorExtension/DocumentInfo/RazorDocumentInfoWindowControl.xaml
```diff
@@ -29,6 +35,18 @@
...
+            <Label Grid.Column="0" Grid.Row="5" Content="Resolver"/>
+            <Label Grid.Column="1" Grid.Row="5" Content="{Binding TagHelperResolver}"/>
+            <Label Grid.Column="0" Grid.Row="6" Content="Syntax Facts"/>
+            <Label Grid.Column="1" Grid.Row="6" Content="{Binding RazorSyntaxFactsService}"/>
+            <Label Grid.Column="0" Grid.Row="7" Content="Template Eng"/>
+            <Label Grid.Column="1" Grid.Row="7" Content="{Binding RazorTemplateEngineFactoryService}"/>
+            <Label Grid.Column="0" Grid.Row="8" Content="Completion Service"/>
+            <Label Grid.Column="1" Grid.Row="8" Content="{Binding TagHelperCompletionService}"/>
+            <Label Grid.Column="0" Grid.Row="9" Content="TH Facts"/>
+            <Label Grid.Column="1" Grid.Row="9" Content="{Binding TagHelperFactsService}"/>
```
rynowak: lol y

src/Microsoft.CodeAnalysis.Razor.Workspaces/AcceptedCharacters.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.VisualStudio.LanguageServices.Razor
+namespace Microsoft.CodeAnalysis.Razor
```
rynowak: This is a binary breaking change FYI, this will require a recompile.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: The goal is to eventually remove the import/export pieces once we move the editor to only utilize Roslyn services. That being said, I agree that this is a little out of place since I wrote VisualStudioXYZ components to maintain the MEFability of other contracts.

I'll build a `VisualStudioTagHelperResolver` type for consistency/easy removal once the editor is using our Roslyn services.

src/Microsoft.CodeAnalysis.Razor.Workspaces/AcceptedCharacters.cs
rynowak: Due to ^^^^ this issue, I'm not sure that your strategy here is going to work VS just copy-pasting services in the `Microsoft.CodeAnalysis.Razor` namespace.

I want to make sure that when we make these changes that we are **done** and don't have to keep asking Barry's team to integrate more small changes.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
NTaylorMullen: > The right way to do this is to provide a non-OOP implementation in ServiceLayer.Default and then provide the VS implementation in ServiceLayer.Host.

Sure, could totally do that.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
NTaylorMullen: Hmm? We have to return this type constructed. Do you mean have the `DefaultTagHelperResolver` take in these bits as ctor arguments?

tooling/Microsoft.VisualStudio.RazorExtension/DocumentInfo/RazorDocumentInfoWindowControl.xaml
NTaylorMullen: To verify the correct services are flowing 😄. Mostly just used this to validate what I did worked. If we don't think it's valuable can nuke.

src/Microsoft.CodeAnalysis.Razor.Workspaces/AcceptedCharacters.cs
NTaylorMullen: Fair enough. If that's the case I'd like to go with my proposed solution and make one breaking change that needs to be reacted to.

rynowak: Your strategy assumes/requires that we don't make any changes to the services as they are currently specified. I'm not sure that's the case.

NTaylorMullen: > Your strategy assumes/requires that we don't make any changes to the services as they are currently specified. I'm not sure that's the case.

How does it make that assumption?

If there's one thing that's for-sure in my mind it's that the syntax tree like contracts don't belong in the VisualStudio.LanguageServices.Razor assembly. What those contracts are could definitely change; heck, we could end up having an entirely different cross-plat SyntaxTree introspection API. However, there's no current plan to overhaul what these APIs currently are (at least not concretely) so it'd be nice to at least put them in assemblies that make sense and iterate from there.

The `Microsoft.VisualStudio.LanguageServices.Razor` assembly has become a dumping ground for all things tooling. In order to get things like API multi-targeting, TagHelper resolution, VS for mac and editor loop ownership in the right form. Having the structure in the (currently) right form will allow us to question tooling based decisions to ensure we don't limit ourselves in the long run.

rynowak: > iterate from there

I don't want to do that. I want to move things to the appropriate places, and clean them up (if needed) in one step

rynowak: I agree with everything else you are saying

rynowak:
FYI here's a kinda neat thing that the project system does to avoid leaking `SVsServiceProvider` into all of their codez.

https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/VsService%601.cs

NTaylorMullen:
🆙 📅 

- Created copies of existing services and their dependencies to prevent binary incompatibilities. Example: There's two `AcceptedCharacters` enum's, one in `Microsoft.VisualStudio.LanguageServices.Razor/Legacy` under its binary compatible namespace and one in `Microsoft.CodeAnalysis.Razor.Workspaces` under the `Microsoft.CodeAnalysis.Razor` namespace.
- Removed MEF exports/imports from existing default implementations. They now expect to be retrieved via Roslyn services. Example: `Microsoft.VisualStudio.LanguageServices.Razor/Legacy/DefaultRazorSyntaxFactsService` still has an export for `RazorSyntaxFactsService` but `Microsoft.CodeAnalysis.Razor.Workspaces/DefaultRazorSyntaxFactsService` does not.
- Changed all `ServiceLayer.Editor` services to `ServiceLayer.Default`
- Could not change the `DefaultTagHelperResolverFactory` to make use of importing constructors for VS types. That didn't seem to be supported, tried multiple variants with no luck.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Shell;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [ExportLanguageServiceFactory(typeof(TagHelperResolver), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultTagHelperResolverFactory : ILanguageServiceFactory
+    {
+        [Import]
```
NTaylorMullen: Could not change the DefaultTagHelperResolverFactory to make use of importing constructors for VS types. That didn't seem to be supported, tried multiple variants with no luck.

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultRazorSyntaxFactsService.cs
```diff
@@ -0,0 +1,342 @@
...
+                {
+                    toProcess.InsertRange(i + 1, blockChildren);
+                    blockChildren.Clear();
+                }
+            }
+
+            return results;
+        }
+
+        public override int? GetDesiredIndentation(RazorSyntaxTree syntaxTree, int previousLineEndIndex, Func<int, string> getLineContent, int indentSize, int tabSize)
```
rynowak: Can you log a follow up to find a new home for this?

src/Microsoft.CodeAnalysis.Razor.Workspaces/DefaultTagHelperCompletionService.cs
```diff
@@ -0,0 +1,273 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Razor
```
rynowak: We might want to create a 'completion' namespace

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorTemplateEngineFactoryService.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Host;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public abstract class RazorTemplateEngineFactoryService : ILanguageService
```
rynowak: I don't really care where this goes for now, because eventually it will be an implementation detail/internal

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
rynowak: That's a little strange but #YOLO

Issue #1580, opened 2017-07-24
DefaultTagHelperContent has incorrect DebuggerDisplayAttribute


pranavkm:
![image](https://user-images.githubusercontent.com/174281/28536910-fad51cd4-705d-11e7-812b-ac92585f10c2.png)

The line in question: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Runtime/TagHelpers/DefaultTagHelperContent.cs#L17. One thing we could do is standardize on `nameof` in these strings. It looks uglier (https://github.com/aspnet/Mvc/blob/b796bc0f39d08bdbf0ce36279db3148ed9c9e888/src/Microsoft.AspNetCore.Mvc.RazorPages/PageActionDescriptor.cs#L10) since you cannot use string interpolation, but it guards against these values breaking during refactoring.

ajaybhargavb:
ce689b0449db04a9fed4250c9465d4be7beb7cc5

PR #1594, opened 2017-08-01
Fixed DebuggerDisplay in DefaultTagHelperContent

ajaybhargavb:
- Also refactored it in few other places

#1580 

@rynowak @NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptor.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptor.cs
@@ -8,7 +8,6 @@ using System.Linq;
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    [DebuggerDisplay("{DisplayName,nq}")]
     public abstract class AllowedChildTagDescriptor : IEquatable<AllowedChildTagDescriptor>
     {
         public string Name { get; protected set; }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptor.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    [DebuggerDisplay("{DisplayName,nq}")]
+    [DebuggerDisplay("{" + nameof(DisplayName) + ",nq}")]
```
rynowak: This was better before 👎

ajaybhargavb: It **looked** better before. But as @pranavkm [said](https://github.com/aspnet/Razor/issues/1580#issue-245158823),  this prevents bugs like these from happening again.

pranavkm: Strongly typed > Stringly typed

rynowak: After some discussion about this, we should make the following changes:
1. Override ToString where we have a display name property
2. Make a real `DebuggerToString()` for the places that don't use a display name property

ajaybhargavb:
🆙 📅

Issue #1535, opened 2017-07-11
Implement Multitargeting for MVC codegen


NTaylorMullen:
This is enabling us to have Mvc version x and x+1 in the box.

NTaylorMullen:
We will need to conditionally render `RazorViewAttribute` when generating Razor code for a 1.0.0 application.

NTaylorMullen:
Before starting this guy, going to wait for conversations surround proposed improvements in: https://github.com/aspnet/Razor/pull/1590#issue-246644259

ajaybhargavb:
4fcad1548ee9b222fffc72fae8bd9b641ea19824

ajaybhargavb:
dbcac41be7ca731d07ea296423758ccfbd3bf383

Issue #1183, opened 2017-04-06
Add multi-targeting support to RazorTemplateEngineFactoryService


rynowak:
When we create a RazorTemplateEngine it should a version of the Razor codegeneration support that matches the project's dependencies. We have the ability to generate ASP.NET Core 1.X code for 1.X proejcts and ASP.NET Core 2.X code for 2.X projects. 

This items tracks the following improvements:
- Create a template engine with the appropriate version targeted when a new document is opened
- Update the template engine with the appropriate version when project references change
- Treat newest version as the 'fallback', if the version is unknown we just assume latest

rynowak:
This isn't needed in Preview2

Eilon:
@rynowak - what milestone should this be in?

rynowak:
This is 2.1

ajaybhargavb:
Do we have a design for it already? If not, we should discuss tomorrow.

PR #1599, opened 2017-08-04
Get the correct version of Mvc for multitargeting

ajaybhargavb:
#1535 

This is temporary code to make multitargeting light up until we figure out the design for #1183.

@NTaylorMullen @rynowak 

Created https://github.com/aspnet/Razor/issues/1600 to track adding unit tests.

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
@@ -2,7 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
+using System.Linq;
 using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Host;
 using Mvc1_X = Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X;
 using MvcLatest = Microsoft.AspNetCore.Mvc.Razor.Extensions;
@@ -11,6 +14,7 @@ namespace Microsoft.CodeAnalysis.Razor
 {
     internal class DefaultTemplateEngineFactoryService : RazorTemplateEngineFactoryService
     {
+        private const string MvcAssemblyName = "Microsoft.AspNetCore.Mvc.Razor";
         private static readonly Version LatestSupportedMvc = new Version(2, 1, 0);
 
         private readonly HostLanguageServices _services;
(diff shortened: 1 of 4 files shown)


tooling/Microsoft.VisualStudio.RazorExtension/Properties/BindingRedirectAttributes.cs
```diff
@@ -12,6 +12,14 @@
     NewVersion = "2.1.0.0")]
 
 [assembly: ProvideBindingRedirection(
+    AssemblyName = "Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X",
```
ajaybhargavb: This is not necessary but it doesn't hurt and might be needed in the future.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
```diff
@@ -11,6 +13,7 @@ namespace Microsoft.CodeAnalysis.Razor
 {
     internal class DefaultTemplateEngineFactoryService : RazorTemplateEngineFactoryService
     {
+        private const string MvcAssemblyName = "Microsoft.AspNetCore.Mvc";
```
NTaylorMullen: This is a meta package. We should be looking for `Microsoft.AspNetCore.Mvc.Razor` since that's what actually enables Razor like features.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
```diff
@@ -70,8 +73,36 @@ public override RazorTemplateEngine Create(string projectPath, Action<IRazorEngi
...
+                var directory = Path.GetDirectoryName(p.FilePath);
+                return string.Equals(
+                    NormalizeDirectoryPath(directory),
+                    NormalizeDirectoryPath(projectPath),
+                    StringComparison.OrdinalIgnoreCase);
+            });
+
+            if (project != null)
+            {
+                var compilation = project.GetCompilationAsync().Result;
```
NTaylorMullen: @rynowak I can see this whole `GetMvcVersion` bit being as is and also being in the remote client. Ryan did you have a preference?

rynowak: We can't use `.Result` in tooling.

tooling/Microsoft.VisualStudio.RazorExtension/Properties/BindingRedirectAttributes.cs
rynowak: Sure, seems good. We want this in the VSIX

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
rynowak: Does the project have what we need? Or do we need to create a compilation?

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
```diff
@@ -70,8 +73,36 @@ public override RazorTemplateEngine Create(string projectPath, Action<IRazorEngi
 
         private Version GetMvcVersion(string projectPath)
         {
-            // TODO: Need to figure out the actual referenced Mvc version in the project. https://github.com/aspnet/Razor/issues/1183
+            var workspace = _services.WorkspaceServices.Workspace;
+
+            var project = workspace.CurrentSolution.Projects.FirstOrDefault(p =>
```
rynowak: We can also think about just making an API that takes the `Project` and transition to that.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
ajaybhargavb: Actually the project has `MetadataReferences`. I'll try using that instead.

ajaybhargavb: I still need to compilation to get the `IAssemblySymbol` from the `MetadataReference`.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTemplateEngineFactoryService.cs
ajaybhargavb: I assumed that we want to have such a thing eventually https://github.com/aspnet/Razor/issues/1183 but we want a temporary solution until that can be done.

ajaybhargavb:
🆙 📅

Issue #1565, opened 2017-07-19
Unhelpful additional error message for empty bound tag attribute


pakrym:
Repro:
1. File->New->Web App
2. Go to Index.cshtml
3. Add tag helper with `int` attribute
4. Add attribute but keep value empty

![image](https://user-images.githubusercontent.com/1697911/28382224-4b120f4e-6c72-11e7-925f-d8a5ba825c04.png)

rynowak:
@ajaybhargavb is there something we do here to prevent the invalid codegen in this case? Let's only fix this if it's straightforward

ajaybhargavb:
After investigation and a hacky fix in https://github.com/aspnet/Razor/pull/1604, we decided that there is no easy way to fix this without breaking intellisense.

PR #1604, opened 2017-08-08
Stop generating invalid code for empty non-string bound attributes

ajaybhargavb:
#1565 

This seems like a straightforward enough fix.
Note: This is NOT a design PR.

@NTaylorMullen @rynowak

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EmptyAttributeTagHelpers_DesignTime.mappings.txt
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EmptyAttributeTagHelpers_DesignTime.mappings.txt
@@ -3,18 +3,3 @@ Source Location: (14:0,14 [15] TestFiles/IntegrationTests/CodeGenerationIntegrat
 Generated Location: (683:13,38 [15] )
 |*, TestAssembly|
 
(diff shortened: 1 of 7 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EmptyAttributeTagHelpers_DesignTime.mappings.txt
```diff
@@ -3,18 +3,3 @@ Source Location: (14:0,14 [15] TestFiles/IntegrationTests/CodeGenerationIntegrat
 Generated Location: (683:13,38 [15] )
 |*, TestAssembly|
 
-Source Location: (66:3,26 [0] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/EmptyAttributeTagHelpers.cshtml)
-||
```
NTaylorMullen: Hmm, worried this will bork IntelliSense at design time. Could you verify?

ajaybhargavb: It shouldn't. The syntax tree isn't changed. But I'll double check.

ajaybhargavb:
Closing this PR because we can't do this in a straightforward way without breaking intellisense.

Issue #1607, opened 2017-08-09
Razor view @model bug in Preview 2


ajaybhargavb:
From @nickalbrecht,

I have a view that uses the following model definition
`@model KeyValuePair<string, string>?`

I use ? because KeyValuePair is a struct and I need it to be nullable. This worked in 1.1 but I just tried to upgrade my site to Preview 2 to see if there are any bugs, and I'm getting the following error on the `?` character

> Unexpected literal following the 'model' directive. Expected 'line break'.

Note, that as a work around, I can define my model as `@model Nullable<KeyValuePair<string, string>>` instead, it just means I have to change all of my views and this seems like an unintended change in the new version.

ajaybhargavb:
364fc1a55d6fa47dca795491189be4c6a354974c

PR #1608, opened 2017-08-09
Allow nullable types with generic arguments in directives

ajaybhargavb:
#1607 

@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -959,7 +959,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
         {
             if (Optional(CSharpSymbolType.Identifier) || Optional(CSharpSymbolType.Keyword))
             {
-                Optional(CSharpSymbolType.QuestionMark); // Nullable
                 if (Optional(CSharpSymbolType.DoubleColon))
                 {
                     if (!Optional(CSharpSymbolType.Identifier))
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -993,6 +993,7 @@ private void TypeArgumentList()
             Assert(CSharpSymbolType.LessThan);
             Balance(BalancingModes.None);
             Optional(CSharpSymbolType.GreaterThan);
+            Optional(CSharpSymbolType.QuestionMark);
```
rynowak: This doesn't seem like it should be here. It's not part of a type argument list.

rynowak: What about types like `int?` do they have the same bug?

ajaybhargavb: They don't. Types like `int?` are handled by https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs#L962

ajaybhargavb: > It's not part of a type argument list.

Sure. I can move this to the line after this method call.

pranavkm: Are generic arrays `Foo<int>[]` similarly covered?

ajaybhargavb: Nice catch. Need to support that too.

ajaybhargavb:
🆙 📅

Issue #1572, opened 2017-07-19
Unable to use nested tag helpers when prefix is enabled


pakrym:
Consider following tag helpers:

``` c#
    [HtmlTargetElement("grid")]
    [OutputElementHint("table")]
    [RestrictChildren("cell")]
    public class GridTagHelper : TagHelper
    {
    }

    [HtmlTargetElement("cell", ParentTag = "grid")]
    public class CellTagHelper : TagHelper
    {
    }
```

When there is no tag helper prefix everything works fine:

![image](https://user-images.githubusercontent.com/1697911/28391536-466042ec-6c93-11e7-9a93-1dd815a2946c.png)

But if I add `@tagHelperPrefix T:` there is no way to use nested `cell` as tag helper anymore because `T:cell` is not allowed by `RestrictChildren` and `cell` is not recognized as tag helper:

![image](https://user-images.githubusercontent.com/1697911/28391593-7cd1e466-6c93-11e7-8971-bf114f0b05e2.png)

Should `RestrictChildren` be considering `tagHelperPrefix` while filtering?

NTaylorMullen:
Great find, looks like we regressed this: https://github.com/aspnet/Razor/issues/532

ajaybhargavb:
45732a5dd38d987acbf294b2df7aea6c0f5162c4

PR #1610, opened 2017-08-10
Fix: Nested tag helpers do not work when tag helper prefix is set

ajaybhargavb:
#1572

@NTaylorMullen @rynowak 

When `ParentTag` is set, we need to compare with the current parent tag without prefix.


src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -511,7 +511,7 @@ private void ValidateParentAllowsPlainTag(Block tagBlock, ErrorSink errorSink)
             var currentTracker = _trackerStack.Count > 0 ? _trackerStack.Peek() : null;
 
             if (HasAllowedChildren() &&
-                !_currentTagHelperTracker.AllowedChildren.Contains(tagName, StringComparer.OrdinalIgnoreCase))
+                !_currentTagHelperTracker.PrefixedAllowedChildren.Contains(tagName, StringComparer.OrdinalIgnoreCase))
```
ajaybhargavb: This is not part of this fix. But when I looked at it I didn't understand why we are comparing with `AllowedChildren` instead of `PrefixedAllowedChildren`. Just wanted to bring this up.

rynowak: Can you come up with a unit test that covers this code?

src/Microsoft.AspNetCore.Razor.Language/TagHelperBinder.cs
```diff
@@ -74,11 +74,20 @@ public TagHelperBinder(string tagHelperPrefix, IEnumerable<TagHelperDescriptor>
             }
 
             var tagNameWithoutPrefix = _tagHelperPrefix != null ? tagName.Substring(_tagHelperPrefix.Length) : tagName;
+
+            var parentTagNameWithoutPrefix = parentTagName;
+            if (!string.IsNullOrEmpty(_tagHelperPrefix) &&
+                !string.IsNullOrEmpty(parentTagName) &&
+                parentTagName.Length > _tagHelperPrefix.Length)
+            {
+                parentTagNameWithoutPrefix = parentTagName.Substring(_tagHelperPrefix.Length);
```
NTaylorMullen: parent tag might not be a TagHelper. I'd imagine this would break in the following scenario:
```HTML
@tagHelperPrefix th

<parent>
    <th:requires-parent-as-parent>
</parent>
```

ajaybhargavb: Yup. Forgot to do a startswith check here.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
ajaybhargavb: Figured out why. I'll add a comment.

ajaybhargavb:
🆙 📅 

Ended up having to handle 2 different cases which weren't handled properly.

All the below cases are with a non-empty tag helper prefix.
**1. Both parent and child are tag helpers (Valid structure)**
Expected: Both are recognized as tag helpers without errors
Actual: The child tag is not recognized as a tag helper.
Fix: Compare with unprefixed parent tag. ([fix](https://github.com/aspnet/Razor/blob/ajbaaska/nested-prefix/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs#L193-L195))

**2. Both parent and child are tag helpers (Child has invalid structure)**
Expected: Only one error about the child being a malformed tag helper.
Actual: Extra error about the child tag not being allowed by parent (Even though it is)
Fix: Use `PrefixedAllowedChildren` for comparison if a `TagHelperBinding` exists. ([fix1](https://github.com/aspnet/Razor/blob/ajbaaska/nested-prefix/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs#L260-L265) and [fix2](https://github.com/aspnet/Razor/blob/ajbaaska/nested-prefix/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs#L521-L529))

**3. Parent is a tag helper and child is a regular tag. (Valid structure)**
This was correctly handled before. Just added a test for this case to make sure I don't break it.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -45,6 +45,7 @@ internal class TagHelperParseTreeRewriter
         private TagHelperBlockTracker _currentTagHelperTracker;
         private BlockBuilder _currentBlock;
         private string _currentParentTagName;
```
NTaylorMullen: Instead of this and the `_currentParentIsTagHelper` variable just make use of the tracker stack's current `TagBlockTracker`. We do something simpler with tag helper block trackers (_currentTagHelperTracker).

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -508,10 +513,21 @@ private void ValidateParentAllowsPlainTag(Block tagBlock, ErrorSink errorSink)
...
+            if (!HasAllowedChildren())
+            {
+                return;
+            }
 
-            if (HasAllowedChildren() &&
-                !_currentTagHelperTracker.AllowedChildren.Contains(tagName, StringComparer.OrdinalIgnoreCase))
+            var parentTagName = GetCurrentParentTagName();
+
+            var tagHelperBinding = _tagHelperBinder.GetBinding(
```
NTaylorMullen: Should still update the APIs in the `GetBinding` to reflect that the parent tag name is expected to be prefixless. It's internal so don't have to worry about API breaking changes there.

ajaybhargavb:
🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ElementCompletionContext.cs
```diff
@@ -15,6 +15,7 @@ public sealed class ElementCompletionContext
             string containingTagName,
             IEnumerable<KeyValuePair<string, string>> attributes,
             string containingParentTagName,
+            bool containingParentIsTagHelper,
```
rynowak: This is a breaking change, you can't do this.

src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperFactsService.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.CodeAnalysis.Razor
 {
     public abstract class TagHelperFactsService : ILanguageService
     {
-        public abstract TagHelperBinding GetTagHelperBinding(TagHelperDocumentContext documentContext, string tagName, IEnumerable<KeyValuePair<string, string>> attributes, string parentTag);
+        public abstract TagHelperBinding GetTagHelperBinding(TagHelperDocumentContext documentContext, string tagName, IEnumerable<KeyValuePair<string, string>> attributes, string parentTag, bool parentIsTagHelper);
```
rynowak: Breaking change, cannot do

src/Microsoft.CodeAnalysis.Razor.Workspaces/ElementCompletionContext.cs
NTaylorMullen: We didn't ship this and VS doesn't use this type yet, it uses the one in `LanguageServices.Razor`.

src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperFactsService.cs
NTaylorMullen: ditto

src/Microsoft.VisualStudio.LanguageServices.Razor/Legacy/TagHelperFactsService.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor
     // ----------------------------------------------------------------------------------------------------
     public abstract class TagHelperFactsService
     {
-        public abstract TagHelperBinding GetTagHelperBinding(TagHelperDocumentContext documentContext, string tagName, IEnumerable<KeyValuePair<string, string>> attributes, string parentTag);
+        public abstract TagHelperBinding GetTagHelperBinding(TagHelperDocumentContext documentContext, string tagName, IEnumerable<KeyValuePair<string, string>> attributes, string parentTag, bool parentIsTagHelper);
```
ajaybhargavb: Actually this is a breaking change. I'll undo.

ajaybhargavb:
🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ElementCompletionContext.cs
rynowak: OK fine.

Issue #1592, opened 2017-07-31
@model does not support the C# syntax for tuple


javiercn:
I'd like to define the model below

```Razor
@model (id, playingOn)
```

so that I can do `@Model.id` and `@Model.playingOn`

This is very useful  to easily pack more than one parameter when rendering a partial.

rynowak:
I would not expect that to work, I would imagine you would need to do `(int id, int playingOn)` - does that work?

ajaybhargavb:
This is currently not supported. We don't consider tuples when we parse type names in directives https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs#L960.

rynowak:
Cool beans, let's fix it.

javiercn:
Yeah, I forgot to update the bug. I messed up when opening it, but it doesn't work (tried again with proper syntax)

ajaybhargavb:
8d2a9e59291df7358b62fc9e51ea21d9cfd110f6

PR #1613, opened 2017-08-12
Support tuples for type directive tokens

ajaybhargavb:
#1592 

@NTaylorMullen @rynowak 

I've also verified this change in design time in VS.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -957,7 +957,34 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
         protected bool NamespaceOrTypeName()
         {
-            if (Optional(CSharpSymbolType.Identifier) || Optional(CSharpSymbolType.Keyword))
+            if (Optional(CSharpSymbolType.LeftParenthesis))
+            {
+                while (!Optional(CSharpSymbolType.RightParenthesis) && !EndOfFile)
+                {
+                    Optional(CSharpSymbolType.WhiteSpace);
+
+                    if (!NamespaceOrTypeName())
+                    {
+                        return false;
+                    }
+
+                    Optional(CSharpSymbolType.WhiteSpace);
+                    Optional(CSharpSymbolType.Identifier);
+                    Optional(CSharpSymbolType.WhiteSpace);
+                    Optional(CSharpSymbolType.Comma);
+                }
+
+                if (At(CSharpSymbolType.WhiteSpace) && NextIs(CSharpSymbolType.QuestionMark))
+                {
+                    // Only accept the whitespace if we are going to consume the next token.
+                    AcceptAndMoveNext();
+                }
+
+                Optional(CSharpSymbolType.QuestionMark); // Nullable
+
+                return true;
+            }
+            else if (Optional(CSharpSymbolType.Identifier) || Optional(CSharpSymbolType.Keyword))
             {
                 if (Optional(CSharpSymbolType.DoubleColon))
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -957,7 +957,27 @@ protected bool QualifiedIdentifier(out int identifierLength)
 
         protected bool NamespaceOrTypeName()
         {
-            if (Optional(CSharpSymbolType.Identifier) || Optional(CSharpSymbolType.Keyword))
+            if (Optional(CSharpSymbolType.LeftParenthesis))
+            {
+                while (!Optional(CSharpSymbolType.RightParenthesis) && !EndOfFile)
+                {
```
NTaylorMullen: `(   string val, int val2)` is a valid tuple. Therefore, need `Optional(CSharpSymbolType.WhiteSpace)` here (means you can 
 also remove the trailing one at the end of this while.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -811,6 +811,30 @@ public void DirectiveDescriptor_AllowsNullableTypes(string expectedType)
                     Factory.Span(SpanKindInternal.Code, expectedType, markup: false).AsDirectiveToken(descriptor.Tokens[0])));
         }
 
+        [Theory]
+        [InlineData("(int aa, string bb)?")]
```
NTaylorMullen: Add a spaced out tuple case: 
```
(    int     aa     ,    string     bb    )
```

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
ajaybhargavb: Added.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
ajaybhargavb: Added.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -957,7 +957,28 @@ protected bool QualifiedIdentifier(out int identifierLength)
...
+                    {
+                        return false;
+                    }
+
+                    Optional(CSharpSymbolType.WhiteSpace);
+                    Optional(CSharpSymbolType.Identifier);
+                    Optional(CSharpSymbolType.WhiteSpace);
+                    Optional(CSharpSymbolType.Comma);
+                }
+
```
NTaylorMullen: Oops, missed this. Can have optional whitespace before the questionmark too. That's a little more tricky though. You only want to consume the whitespace if the next symbol is a question mark. Should do the same thing with our other `nullable` support below.

ajaybhargavb: Huh?! Didn't realize that is valid.

ajaybhargavb: Looks like even `int   ?aa = null;` is valid.

rynowak: Yep, this is why languages don't usually treat whitespace as a token.

PR #1615, opened 2017-08-14
Fix ModelExpression's in section directive blocks.

NTaylorMullen:
- The `DirectiveRemovalPass` runs at `DefaultFeatureOrder + 50` in order to allow the design time directive tokens pass to consume directive intermediate nodes; however, this has a side effect of leaving around the original `@section` nodes a bit too long. The `ModelExpressionPass` would see two references to the sections body (`SectionIntermediateNode` and `DirectiveIntermediateNode`) and then do its work twice. To combat this behavior I bumped the ModelExpressionPass order to be 50 points higher than the directive removal pass.
- Added MVC tests for current and 1_X extensions.

#1614


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ModelExpressionPass.cs
```diff
@@ -13,6 +13,8 @@ public class ModelExpressionPass : IntermediateNodePassBase, IRazorOptimizationP
     {
         private const string ModelExpressionTypeName = "Microsoft.AspNetCore.Mvc.ViewFeatures.ModelExpression";
 
+        public override int Order => DefaultFeatureOrder + 100;
```
ajaybhargavb: Add a comment here to explain the `+ 100`

rynowak:
A high level suggestion I have to make this more flexible would be to add an annotation at the document level inside the model pass. That's something I considered when bringing up this feature anyway.

NTaylorMullen:
> A high level suggestion I have to make this more flexible would be to add an annotation at the document level inside the model pass. That's something I considered when bringing up this feature anyway.

Like mark nodes that have already been handled by the `ModelExpressionPass`?

rynowak:
IIRC you said there was some trickiness about the ordering of passes wrt determining the model type. That's what I was referring to

NTaylorMullen:
> IIRC you said there was some trickiness about the ordering of passes wrt determining the model type. That's what I was referring to

Ah, ya if we wanted to take that approach we could mark the document with an annotation stating the model type. Honestly wouldn't be that awful 😄

NTaylorMullen:
🆙 📅 to have the section directive pass move non-token section nodes from the directive intermediate node to the section intermediate node. This way there's no double reference of those nodes.

src/Microsoft.AspNetCore.Razor.Language/Extensions/SectionDirectivePass.cs
```diff
@@ -37,7 +37,9 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
 
                 for (; i < directive.Node.Children.Count; i++)
                 {
+                    // Move non-token children over to the section node so we don't have double references to children nodes.
                     section.Children.Add(directive.Node.Children[i]);
+                    directive.Node.Children.RemoveAt(i--);
```
rynowak: It would be better to change this to a while loop. This is just.... an abomination.

rynowak: #TaylorControlFlow

PR #1623, opened 2017-08-16
Pinning versions for 2.0.0

pranavkm:


build/dependencies.props
```diff
@@ -1,24 +1,25 @@
...
     <StreamJsonRpcVersion>1.1.92</StreamJsonRpcVersion>
-    <TestSdkVersion>15.3.0-*</TestSdkVersion>
+    <TestSdkVersion>15.3.0</TestSdkVersion>
 
     <!-- Must keep in sync with tooling\Microsoft.VisualStudio.RazorExtension\project.json -->
     <VsShellVersion>15.0.26606</VsShellVersion>
     
-    <XunitVersion>2.3.0-beta2-*</XunitVersion>
+    <XunitVersion>2.3.0-beta2-build3683</XunitVersion>
+    <XunitRunnerVisualStudioVersion>2.3.0-beta2-build1317</XunitRunnerVisualStudioVersion>
```
pranavkm: These are two different versions. Had to add these to distinguish between the two packages we reference. The remainder are pretty straighforward.

version.props
```diff
@@ -2,6 +2,5 @@
 <Project>
     <PropertyGroup>
         <VersionPrefix>2.0.0</VersionPrefix>
-        <VersionSuffix>rtm</VersionSuffix>
```
pranavkm: Makes `2.0.0` versioned package.

NuGet.config
```diff
@@ -2,7 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
-    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-release/api/v3/index.json" />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json" />
     <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
```
smitpatel: Do we still need tools in NuGet feed?

pranavkm: The SDK is in the tools feed. We treat it the same as master and do not have a archival policy configured on it.

Issue #1492, opened 2017-06-30
Investigate why VisualStudio.LanguageServices.Razor.Test can't take a dependency on Test.Common


ajaybhargavb:
This seems to work fine on windows but fails on linux/mac. We need to investigate why this happens.
```
/Razor/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Microsoft.VisualStudio.LanguageServices.Razor.Test.csproj : error NU1201: Project Microsoft.AspNetCore.Razor.Test.Common is not compatible with net461 (.NETFramework,Version=v4.6.1). Project Microsoft.AspNetCore.Razor.Test.Common supports: netcoreapp2.0 (.NETCoreApp,Version=v2.0) [/Users/asplab/TeamCity/work/169923259d6dcfec/Universe/.r/Razor/.build/KoreBuild.proj]
```

ajaybhargavb:
Looks like this issue has been resolved in the newer CLI. Actually, `VisualStudio.LanguageServices.Razor.Test` has already been depending on `Test.Common` for a while. I've just sent a PR to remove the duplicate files that I added as a workaround.

ajaybhargavb:
d726be6dd1f1a31cd38692554be0c5d230199cc1

PR #1629, opened 2017-08-17
Removed duplicate extension methods

ajaybhargavb:
#1492 

https://github.com/aspnet/Razor/issues/1492#issuecomment-323225672

@NTaylorMullen


rynowak:
:shipit:

PR #1640, opened 2017-08-23
Upgrade to xunit 2.3.0-beta4

natemcmaster:
Includes changes to test for issues discovered by new analyzers added in this upgrade.

Also, there were three tests with unused theory parameters. I changed one test to actually use them, and suppressed the warning in the other.


test/Microsoft.VisualStudio.LanguageServices.Razor.Test/RazorEditorParserTest.cs
```diff
@@ -85,7 +85,12 @@ public static TheoryData TagHelperPartialParseRejectData
 
         [Theory]
         [MemberData(nameof(TagHelperPartialParseRejectData))]
-        public void TagHelperTagBodiesRejectPartialChanges(object editObject, object expectedDocument)
+        public void TagHelperTagBodiesRejectPartialChanges(object editObject,
+            // This parameter is not used in this tests. This disables the build warning
+            // that says 'Theory methods should use all of their parameters'
+#pragma warning disable xUnit1026
+            object expectedDocument)
```
natemcmaster: Instead of suppressing, would it be better to split the MemberData so that it doesn't result in used theory params?

NTaylorMullen: Go for it! 😄

natemcmaster:
:up: :date:

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/RazorEditorParserTest.cs
natemcmaster: Ok. I've updated this PR. Deleted this param and the unused theory data.

NTaylorMullen:
Looks good broseph 👍

PR #1642, opened 2017-08-23
Use PackageReference instead of project.json

natemcmaster:
VS 15.3 supports PackageReference in classic csproj files now.

cc @rynowak


NTaylorMullen:
@rynowak you should take a look since you were the one who moved this to use project.json.

pranavkm:
LKGTM. Please get sign off from @rynowak though.

PR #1645, opened 2017-08-24
Fixes #1632 - track textviews in the document tracker

### LINKED ISSUE #1632: Change RazorDocumentTracker to track `ITextBuffer`s.
VisualStudio associates 1 Razor document->many text views (ITextBuffer->ITextView). Today we ignore the ITextBuffer when creating DocumentTrackers. Instead we track the `ITextViews` that are created which is misleading given that you can have many views of the same text buffer (which only has 1 parser today).

rynowak:
I've stripped out some of the dead code and complexity from the document
tracker in an attempt to simplify it. I will bring this back as part of
the multi-targeting work.


src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/VisualStudioRazorParser.cs
```diff
@@ -145,7 +146,7 @@ private void Onidle(object sender, ElapsedEventArgs e)
         {
             _foregroundThreadAffinitizedObject.AssertIsBackground();
 
-            var textViews = DefaultTextViewRazorDocumentTrackerService.GetTextViews(_textBuffer);
+            var textViews = Array.Empty<ITextView>();
```
rynowak: Temporary hack until the next work item is done.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/VisualStudioDocumentTracker.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor.Editor
+{
+    public abstract class VisualStudioDocumentTracker
+    {
+        public abstract event EventHandler ContextChanged;
+
+        public abstract bool IsSupportedProject { get; }
```
NTaylorMullen: Why does the tracker need to know this guy?

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Microsoft.VisualStudio.LanguageServices.Razor.Test.csproj
```diff
@@ -3,6 +3,10 @@
   <PropertyGroup>
     <TargetFramework>net461</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
+    
+    <!-- The binding redirects are needed due to a conflict between the test host and Roslyn regarding System.Collections.Immutable -->
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
```
NTaylorMullen: This needed now because of the dependency on Moq?

tooling/Microsoft.VisualStudio.RazorExtension/Microsoft.VisualStudio.RazorExtension.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <VsixVersion>15.5</VsixVersion>
     <VsixVersion Condition="'$(BuildNumber)'!=''">$(VsixVersion).$(BuildNumber)</VsixVersion>
+    <VsixVersion Condition="'$(BuildNumber)'==''">$(VsixVersion).999999</VsixVersion>
```
ajaybhargavb: Why?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/VisualStudioDocumentTracker.cs
rynowak: The primary purpose of the document tracker is to track what project we're associated with and *if* that project is a dotnet core project. If we end up getting our own content type then we don't need this.

tooling/Microsoft.VisualStudio.RazorExtension/Microsoft.VisualStudio.RazorExtension.csproj
rynowak: I was running into an issue building/running the VSIX. Since I have 15.5.<something> installed, that's a higher number than 15.5. Using 15.5.999999999999999 means that a local build is always 'higher' or at least the same.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -0,0 +1,108 @@
...
+            _workspace = workspace;
+
+            _textViews = new List<ITextView>();
+            
+            Update();
+        }
+
+        public override bool IsSupportedProject => _isSupportedProject;
+
+        public override ProjectId ProjectId => null;
```
rynowak: hack until the n+2 work item is done

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Microsoft.VisualStudio.LanguageServices.Razor.Test.csproj
NTaylorMullen: Sure, but why is this new?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/VisualStudioDocumentTracker.cs
NTaylorMullen: ok, figured that.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Microsoft.VisualStudio.LanguageServices.Razor.Test.csproj
rynowak: `AdHocWorkspace` uses MEF, which does an `Assembly.GetExportedTypes`, which changes the assembly load order, which exposes the conflict between versions of `System.Collections.Immutable`.

rynowak: The simple fact is that all of our test projects should just have this. There's no good reason not to.

pranavkm: The test SDK is supposed to generate this for you though: https://github.com/Microsoft/vstest/blob/6924402e13817bd2c749d6ff5027db62ca61225c/src/package/nuspec/Microsoft.NET.Test.Sdk.targets#L51.

rynowak: That's super cool and interesting

rynowak: We need it for core as well. This is a dependency of the test host, so we need to lift it.

rynowak:
d87e0f7

PR #1648, opened 2017-08-24
Add a default VS foregrounddispatcher

rynowak:


rynowak:
@NTaylorMullen signed off in person

Issue #1617, opened 2017-08-14
Trailing whitespace after type name in directives should not be metacode


ajaybhargavb:
For example, if you type `@model int` and add some whitespace at the end, the whitespace will be highlighted. 

![image](https://user-images.githubusercontent.com/1579269/29296128-26556d0e-810d-11e7-9686-910d32561487.png)

This is because the trailing whitespace is classified as `Metacode` by the parser. It shouldn't be.

ajaybhargavb:
b12683a35d7c73065103b09e33cdd781f1a18935

PR #1656, opened 2017-08-29
Mark trailing whitespace after type name in directives as markup

ajaybhargavb:
#1617 

Also verified this in VS.
image

@NTaylorMullen


ajaybhargavb:
> Add/use SpanKindInternal.None / SpanKind.None and see if the editor code properly ignores it.

Nice idea! It worked. I wonder if there are other places we can use this where we don't want any intellisense.

Issue #1221, opened 2017-04-13
Extensible directives do not provide IntelliSense when typing


NTaylorMullen:
This looks to be due to the lack of a line mapping prior to a directives token being fully written out.

rynowak:
Discussed with @NTaylorMullen and the fix for this is high risk, moving this out to preview 2

ajaybhargavb:
4b68a48f1d05c48f208040fc9e005ff113831953

PR #1659, opened 2017-08-29
Fix: Broken intellisense for extensible directives

ajaybhargavb:
#1221 

Also verified this in VS.
image

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
@@ -57,6 +57,13 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
                 {
                     case DirectiveTokenKind.Type:
 
+                        if (string.IsNullOrEmpty(node.Content))
+                        {
+                            // This is most likely a marker symbol.
+                            WriteMarkerSymbol(context, node);
+                            break;
+                        }
+
                         // {node.Content} __typeHelper = default({node.Content});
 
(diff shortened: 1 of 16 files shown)


src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
```diff
@@ -133,5 +157,12 @@ private void WriteDesignTimeDirectiveToken(CodeRenderingContext context, Directi
             }
             context.CodeWriter.WriteLine("))();");
         }
+
+        private void WriteMarkerSymbol(CodeRenderingContext context)
+        {
+            // We want to map marker symbols to a location in the generated document
+            // that will provide CSharp intellisense.
+            context.CodeWriter.Write(" ");
```
NTaylorMullen: This can do the add source mapping part too.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1665,6 +1665,15 @@ private void HandleDirective(DirectiveDescriptor descriptor)
                     {
                         Span.ChunkGenerator = SpanChunkGenerator.Null;
                         Output(SpanKindInternal.Code, AcceptedCharactersInternal.WhiteSpace);
+
+                        if (EndOfFile || At(CSharpSymbolType.NewLine))
+                        {
+                            // Add a marker symbol to provide CSharp intellisense.
```
NTaylorMullen: More specific in the comment so when we come back and revisit the hack we'll know 😄

PR #1666, opened 2017-08-31
Suppress API Check for the MVC 1.X shim

rynowak:
This is a tooling-only package. We have no compatibility requirement for
it.


rynowak:
FYI only

PR #1671, opened 2017-09-06
Add a subsystem for detecting versions

rynowak:
This adds the beginning of an API for detecting versions from the
project. We will flesh out this API more when we do tooling
extensiblity.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
@@ -1,0 +1,101 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    // This is hardcoded for now. A more complete design would fan out to a list of providers.
+    internal class DefaultProjectExtensibilityConfigurationFactory : ProjectExtensibilityConfigurationFactory
+    {
+        private const string MvcAssemblyName = "Microsoft.AspNetCore.Mvc.Razor";
+        private const string RazorV1AssemblyName = "Microsoft.AspNetCore.Razor";
+        private const string RazorV2AssemblyName = "Microsoft.AspNetCore.Razor.Language";
+
+        // Using MaxValue here so that we ignore patch and build numbers. We only want to compare major/minor.
+        private static readonly Version MaxSupportedRazorVersion = new Version(2, 0, Int32.MaxValue, Int32.MaxValue);
+        private static readonly Version MaxSupportedMvcVersion = new Version(2, 0, Int32.MaxValue, Int32.MaxValue);
+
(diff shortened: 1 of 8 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
```diff
@@ -0,0 +1,104 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    // This is hardcoded for now. A more complete design would fan out to a list of providers.
+    internal class DefaultProjectExtensibilityConfigurationFactory : ProjectExtensibilityConfigurationFactory
```
rynowak: Pay close attention to the **implementation** here. This will likely be split out into a set of providers in the future, but for now this is hardcoded/good enough.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectExtensibilityConfiguration.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    internal abstract class ProjectExtensibilityConfiguration
```
rynowak: For now this is pretty bare.

You can get the list of extensibility assemblies and whether or not it was a fallback or a close match.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectExtensibilityConfigurationFactory.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Host;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    internal abstract class ProjectExtensibilityConfigurationFactory : ILanguageService
```
rynowak: This will be called on a background thread by the project system. It's async because we want it to be called from any thread, but this might touch VS services with threading requirements in the future.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        // internal/separate for testing.
+        internal ProjectExtensibilityConfiguration GetConfiguration(IEnumerable<AssemblyIdentity> references)
+        {
+            // Avoiding ToDictionary here because we don't want a crash if there is a duplicate name.
+            var assemblies = new Dictionary<string, AssemblyIdentity>();
+            foreach (var assembly in references)
+            {
+                if (assembly.Name != null)
```
NTaylorMullen: Can't be null

NTaylorMullen:
Restarted the failed travis build because it looked like it failed due to unrelated reasons

Issue #1672, opened 2017-09-06
Make Razor directive completion not load Razor types in non-Razor scenarios.


NTaylorMullen:

NTaylorMullen:
67f255adca773e12391bd404a98d4a4a01d903de

PR #1673, opened 2017-09-06
Change completion provider to be lazy if not Razor.

NTaylorMullen:
image

#1672


src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorCodeDocumentProvider.cs
```diff
@@ -2,10 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis;
 
-namespace Microsoft.CodeAnalysis.Razor
+namespace Microsoft.VisualStudio.LanguageServices.Razor.Editor
```
NTaylorMullen: Moved this type to VisualStudio.LanguageServices.Razor so we don't load the workspaces dll (needed for the lazy import statement).

PR #1675, opened 2017-09-06
Update our diagnostics window

rynowak:
Updates our diagnostics window to use the 'in the box' version of the
assembly/version discovery logic.


rynowak:
00dc95098f862079b1dc03965446dbb53e75e629

PR #1676, opened 2017-09-07
Add [ForegroundTheory]

rynowak:


test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Xunit/ForegroundTheoryAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor.Test.Xunit
```
ajaybhargavb: Should this be in `Xunit` namespace?

rynowak: Yes

PR #1681, opened 2017-09-12
Change notifications for the project manager

rynowak:
There's still nothing processing the notifications in the background.
This is all the plumbing for dirty checking and publishing updates.


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManagerBase.cs
```diff
@@ -11,6 +11,8 @@ internal abstract class ProjectSnapshotManagerBase : ProjectSnapshotManager
 
         public abstract void ProjectChanged(Project underlyingProject);
 
+        public abstract void ProjectChanged(ProjectSnapshotUpdateContext update);
```
ajaybhargavb: Where is this called? Or is that coming in the next PR?

rynowak: Next PR

Issue #1678, opened 2017-09-07
Implementing quirking of parser and language features for Razor


rynowak:
We're planning some other work items for Razor 2.1.0 that will require us to add features or relax restrictions specific to Razor 2.1.0

The time has come for us to implement version support in Razor. Make it so.

PR #1683, opened 2017-09-15
Added version support to Razor parser

ajaybhargavb:
#1678 

@rynowak 

- Added `Version` property to `RazorParserOptions`
- Added an internal `RazorParserFeatureContext` that contains a flag for all features we want to enable/disable in the parser
- Reason for having a centralized location is so we don't have to look at the version every time we want to decide whether to light-up a feature
- ~~I'll send a separate PR that will use `AllowMinimizedBooleanTagHelperAttributes` to enable/disable that feature.~~ https://github.com/aspnet/Razor/pull/1695


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorParserOptions.cs
```diff
@@ -18,12 +18,18 @@ public DefaultRazorParserOptions(DirectiveDescriptor[] directives, bool designTi
             Directives = directives;
             DesignTime = designTime;
             ParseLeadingDirectives = parseLeadingDirectives;
+            Version = version ?? RazorParserFeatureContext.LatestRazorVersion;
```
ajaybhargavb: Should this throw ArgNull instead?

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureContext.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal abstract class RazorParserFeatureContext
```
ajaybhargavb: Couldn't think of a better that isn't already used

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorParserOptions.cs
```diff
@@ -18,12 +18,18 @@ public DefaultRazorParserOptions(DirectiveDescriptor[] directives, bool designTi
...
 
         public override bool DesignTime { get; }
 
         public override IReadOnlyCollection<DirectiveDescriptor> Directives { get; }
 
         public override bool ParseLeadingDirectives { get; }
+
+        public override Version Version { get; }
+
+        internal override RazorParserFeatureContext FeatureContext { get; }
```
rynowak: I would drop the name `Context` from this. It's not really acting as a context parameter. 

So maybe: `internal override RazorParserFeatureFlags FeatureFlags { get; set; }`

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureContext.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        internal static readonly Version LatestRazorVersion = new Version(2, 1, 0);
```
rynowak: Think about using an enum here instead. Version allows you to express a huge range of stuff including versions that done exist.

Think continuous vs discrete.

We never need/want to handle a version we didn't ship.

src/Microsoft.AspNetCore.Razor.Language/RazorParserOptions.cs
```diff
@@ -54,5 +54,9 @@ public static RazorParserOptions CreateDesignTime(Action<RazorParserOptionsBuild
         /// In a future release this may be updated to include all leading directive content.
         /// </remarks>
         public abstract bool ParseLeadingDirectives { get; }
+
+        public abstract Version Version { get; }
```
rynowak: this is a breaking change

src/Microsoft.AspNetCore.Razor.Language/RazorParserOptions.cs
```diff
@@ -54,5 +54,9 @@ public static RazorParserOptions CreateDesignTime(Action<RazorParserOptionsBuild
...
+
+        internal abstract RazorParserFeatureContext FeatureContext { get; }
```
rynowak: this is a breaking change

src/Microsoft.AspNetCore.Razor.Language/RazorParserOptionsBuilder.cs
```diff
@@ -13,6 +14,8 @@ public abstract class RazorParserOptionsBuilder
 
         public abstract bool ParseLeadingDirectives { get; set; }
 
+        public abstract Version Version { get; set; }
```
rynowak: breaking change,

src/Microsoft.AspNetCore.Razor.Language/RazorParserOptions.cs
rynowak: These need to be virtual

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/RazorParserVersion.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public enum RazorParserVersion
+    {
+        Version1_0,
+        Version1_1,
+        Version2_0,
+        Version2_1,
```
ajaybhargavb: Will there be a need to add more here other than for every major release? Eg. We'll need different parser behavior for patch releases?

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureFlags.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal abstract class RazorParserFeatureFlags
+    {
+        internal static readonly RazorParserVersion LatestRazorParserVersion = RazorParserVersion.Version2_1;
+
+        public static RazorParserFeatureFlags Create(RazorParserVersion version)
+        {
+            if (version == LatestRazorParserVersion)
```
rynowak: This should be a >= `Version2_1` - we do don't want to keep changing this code if we add more versions

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureFlags.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal abstract class RazorParserFeatureFlags
```
rynowak: This doesn't really need to be abstract if its internal

src/Microsoft.AspNetCore.Razor.Language/RazorParserVersion.cs
rynowak: No, not for patches

rynowak: We might want to do these the way Roslyn does - https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/LanguageVersion.cs

src/Microsoft.AspNetCore.Razor.Language/RazorParserVersion.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public enum RazorParserVersion
```
rynowak: RazorLanguageVersion

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureFlags.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        internal static readonly RazorParserVersion LatestRazorParserVersion = RazorParserVersion.Version2_1;
+
+        public static RazorParserFeatureFlags Create(RazorParserVersion version)
+        {
+            if (version == LatestRazorParserVersion)
+            {
+                return new DefaultRazorParserFeatureFlags(allowMinimizedBooleanTagHelperAttributes: true);
```
rynowak: Imagine you have 4-5 different features to control. Think about just having them as variables and then call the constructor at the end.

src/Microsoft.AspNetCore.Razor.Language/RazorParserFeatureFlags.cs
ajaybhargavb: I didn't want this to be constructible without using the Create method.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorParserOptionsBuilder.cs
```diff
@@ -19,9 +20,11 @@ public DefaultRazorParserOptionsBuilder(bool designTime)
 
         public override bool ParseLeadingDirectives { get; set; }
 
+        public override RazorParserVersion Version { get; set; }
```
NTaylorMullen: We should treat `Version` like we do `DesignTime`, aka required at construction of the builder.

ajaybhargavb:
🆙 📅

Issue #431, opened 2015-06-23
Support bound and minimized boolean (or bool?) TagHelper attributes.


NTaylorMullen:
Today `TagHelper`s do not support bound minimized attributes because typically it doesn't make sense to have a bound attribute that will have 0 value. This isn't the case for `boolean` attributes though; we could special case `boolean` and even `bool?` values to allow situations such as:

``` C#
public class InputTagHelper : TagHelper
{
    public bool Checked { get; set; }

    public bool? Extra { get; set; }

    public override void Process(...)
    {
        // <input checked />                 Checked = true, Extra = null
        // <input checked="1 + 1 == 2" />    Checked = true, Extra = null
        // <input />                         Checked = false, Extra= null
        // <input extra />                   Checked = false, Extra = true
        // <input extra="1 + 3 == 2" />      Checked = false, Extra = false
        // <input extra checked />           Checked = true, Extra = true
    }
}
```

NTaylorMullen:
Updated issue to showcase `bool?`

dougbu:
Note this works today: The `TagHelper` could _not_ bind the bool attribute. Instead check whether it appears in `TagHelperOutput.Attributes`. Check `TagHelperAttribute.Minimized` to differentiate the minimized case from `checked="checked"` (for some reason).

dougbu:
Related to #1589 as well. That probably is _not_ more advanced.

ajaybhargavb:
Assigning this to myself as I am already working on this.

NTaylorMullen:
https://github.com/aspnet/Razor/commit/bd8e9ecc318946e70e33e406a43e31b3ea307d9c

PR #1695, opened 2017-09-21
Added support for minimized boolean bound tag helper attributes

ajaybhargavb:
#431 

Added support for both `bool` and `Dictionary<*, bool>` types.
Have not added support for `bool?`.


src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -407,7 +407,18 @@ public void WriteTagHelperProperty(CodeRenderingContext context, DefaultTagHelpe
                             context.CodeWriter.WriteStartAssignment(GetPropertyAccessor(node));
                         }
 
-                        RenderTagHelperAttributeInline(context, node, node.Source);
+                        if (node.Children.Count == 0 &&
+                            node.AttributeStructure == AttributeStructure.Minimized &&
+                            (node.BoundAttribute.IsBooleanProperty ||
+                            (node.IsIndexerNameMatch && node.BoundAttribute.IsIndexerBooleanProperty)))
```
ajaybhargavb: Considering moving this and the line above to an extension method. Thoughts?

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
```diff
@@ -61,10 +63,14 @@ internal static class TagHelperBlockRewriter
...
-                    // is null or whitespace.
-                    if ((result.IsBoundAttribute && result.AttributeValueNode == null) ||
+                    // Check if it's a non-boolean bound attribute that is minimized or if it's a bound
+                    // non-string attribute that is null or whitespace.
+                    var isMinimizedBoundAttribute = result.IsBoundAttribute && result.AttributeValueNode == null;
+                    var isValidMinimizedAttribute = allowMinimizedBooleanTagHelperAttributes && result.IsBoundBooleanAttribute;
+                    if ((isMinimizedBoundAttribute &&
+                        !isValidMinimizedAttribute) ||
                         (result.IsBoundNonStringAttribute &&
+                        result.AttributeValueNode != null &&
```
ajaybhargavb: This is kind of ugly. Can we split the below error message into two? 
`Attribute '{0}' on tag helper element '{1}' requires a value. Tag helper bound attributes of type '{2}' cannot be empty or contain only whitespace.`
We can have one for null and one for whitespace. That will make the code here simpler. Thoughts?

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -684,18 +686,28 @@ private void AddTagHelperAttributes(IList<TagHelperAttributeNode> attributes, Ta
...
+                                continue;
+                            }
+
                             var setTagHelperProperty = new TagHelperPropertyIntermediateNode()
                             {
                                 AttributeName = attribute.Name,
                                 BoundAttribute = associatedAttributeDescriptor,
                                 TagHelper = associatedDescriptor,
                                 AttributeStructure = attribute.AttributeStructure,
                                 Source = BuildSourceSpanFromNode(attributeValueNode),
```
ajaybhargavb: SourceSpan will be null for minimized bound attributes.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -53,7 +53,10 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
             }
 
             var errorSink = new ErrorSink();
-            var rewriter = new TagHelperParseTreeRewriter(tagHelperPrefix, descriptors);
+            var rewriter = new TagHelperParseTreeRewriter(tagHelperPrefix, descriptors)
```
NTaylorMullen: Should just have this guy take in the feature flags.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
NTaylorMullen: You already do this logic in the intermediate node lowering phase; you can store that result on the `TagHelperPropertyIntermediateNode`. Kind of like how that node has the `IsIndexerNameMatch` today.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
```diff
@@ -16,13 +16,14 @@ internal static class TagHelperBlockRewriter
         public static TagHelperBlockBuilder Rewrite(
             string tagName,
             bool validStructure,
+            bool allowMinimizedBooleanTagHelperAttributes,
```
NTaylorMullen: Have this take in the feature flags. Future TagHelper changes will be much easier to complete.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
NTaylorMullen: Eh, can just leave the error message as is. You should factor your if statement into two if's for more clarity:
1. if feature is disabled - act like the old way
2. else act like new way

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
ajaybhargavb: That's the thing. It is not just done in IRL phase. It is also computed in the block rewriter.

NTaylorMullen: The block rewriter part is fine. It's done slightly differently. Lets do it in two places instead of three 😄. That's also consistent with how we pass information like this elsewhere

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
NTaylorMullen: Actually, can't you move this error entirely to the intermediate phase? It's crazy that we did these types of checks here to begin with.

ajaybhargavb: Yup. I like that. If I can move this error to lowering phase, this will get simplified.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -684,6 +686,14 @@ private void AddTagHelperAttributes(IList<TagHelperAttributeNode> attributes, Ta
                                 return TagHelperMatchingConventions.CanSatisfyBoundAttribute(attribute.Name, a);
                             });
 
+                            var isValidMinimizedAttribute = associatedAttributeDescriptor.ExpectsBooleanValue(attribute.Name);
```
NTaylorMullen: expectsBooleanValue

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorExtensions.cs
```diff
@@ -27,5 +27,27 @@ public static bool IsDefaultKind(this BoundAttributeDescriptor attribute)
 
             return string.Equals(attribute.Kind, TagHelperConventions.DefaultKind, StringComparison.Ordinal);
         }
+
+        public static bool ExpectsStringValue(this BoundAttributeDescriptor attribute, string name)
```
NTaylorMullen: We able to mark these internal? Don't want to make these guys public unless we absolutely have to.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -407,7 +407,17 @@ public void WriteTagHelperProperty(CodeRenderingContext context, DefaultTagHelpe
                             context.CodeWriter.WriteStartAssignment(GetPropertyAccessor(node));
                         }
 
-                        RenderTagHelperAttributeInline(context, node, node.Source);
+                        if (node.Children.Count == 0 &&
+                            node.AttributeStructure == AttributeStructure.Minimized &&
+                            (node.BoundAttribute.ExpectsBooleanValue(node.AttributeName)))
```
NTaylorMullen: no need for extra parenthesis

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -429,7 +439,17 @@ public void WriteTagHelperProperty(CodeRenderingContext context, DefaultTagHelpe
                                 .Write(".");
                         }
 
-                        RenderTagHelperAttributeInline(context, node, node.Source);
+                        if (node.Children.Count == 0 &&
+                            node.AttributeStructure == AttributeStructure.Minimized &&
+                            (node.BoundAttribute.ExpectsBooleanValue(node.AttributeName)))
```
NTaylorMullen: ditto

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/MinimizedTagHelpers.cshtml
```diff
@@ -15,4 +15,5 @@
                 input-unbound-required="hello2" 
            catchall-unbound-required 
            input-bound-required-string="world" />
+    <div boundbool booldict-prefix-key></div>
```
NTaylorMullen: Do we have tests validating TagHelpers with an unmatched bound boolean attribute? Technically there should be no codegen for that but want to verify we're not missing any explosion cases

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
NTaylorMullen: Whoy not utilize the ctor?

ajaybhargavb: No reason now. Did that when I wasn't passing in featureFlags directly

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/MinimizedTagHelpers.cshtml
ajaybhargavb: Just added a line below this.

ajaybhargavb:
🆙 📅

Issue #1696, opened 2017-09-21
Create a VS for Mac Plugin for Razor binaries.


NTaylorMullen:

NTaylorMullen:
fca4f238951974dd5ce0c29e77a8a69f1d46cb2c

PR #1701, opened 2017-09-27
Add a VS for Mac compatible Razor addin

NTaylorMullen:
- Add Microsoft.MonoDevelop.RazorAddin.
- As part of this work I also added `Microsoft.MonoDevelop.LanguageServices.Razor` to be the monodevelop specific Razor code.
- Added MSBuild infrastructure to automate creation of MonoDevelop addins (MPacks). This work enables us to not have a dependency on a specific version of monodevelop and does not require us to have tool-prerequisites on the box. Every build outputs the mpacks into the artifacts/build directory.
- Built in build-level metadata pieces to workaround how addins are typically developed. They are usually authored C# first and then config files are generated after the fact; with this changeset we auto-generate the addin.info and its corresponding assembly attributes. Both of these take information directly from the build system.

#1696 

/cc @ToddGrun @KirillOsenkov

### DIFF
--- a/build/MPack.targets
+++ b/build/MPack.targets
@@ -1,0 +1,64 @@
+<Project>
+  <PropertyGroup>
+    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);GenerateMPack</PackageDependsOn>
+    <AddinName>Microsoft.VisualStudio.Mac.RazorAddin</AddinName>
+    <AddinDirectory>$(RepositoryRoot)tooling\$(AddinName)\</AddinDirectory>
+  </PropertyGroup>
+
+  <Import Project="$(AddinDirectory)AddinMetadata.props" />
+
+  <Target
+    Name="GenerateMPack"
+    Condition="'$(OS)'=='Windows_NT'">
+    <!--
+      In our case the mpack archive requires the following:
+      1. An addin.info
+      2. An addin binary (Microsoft.VisualStudio.Mac.RazorAddin.dll)
+        a. _Manifest.addin.xml embedded
+        b. Addin assembly attributes for metadata 
+      3. All language service binaries
+    -->
+
+    <PropertyGroup>
+      <MSBuildArtifactsDir>$(ArtifactsDir)msbuild\</MSBuildArtifactsDir>
+      <MPackSourcesDir>$(MSBuildArtifactsDir)sources\</MPackSourcesDir>
+      <AddinOutputPath>$(AddinDirectory)bin\$(Configuration)\net461\</AddinOutputPath>
+      <LanguageServiceName>Microsoft.VisualStudio.Mac.LanguageServices.Razor</LanguageServiceName>
+      <LanguageServiceOutputPath>$(RepositoryRoot)src\$(LanguageServiceName)\bin\$(Configuration)\net46\</LanguageServiceOutputPath>
+      <MPackName>$(AddinName)_$(AddinVersion)</MPackName>
+      <MPackFileName>$(MPackName).mpack</MPackFileName>
+      <MPackOutputPath>$(BuildDir)$(MPackFileName)</MPackOutputPath>
+      <MPackZipFile>$(BuildDir)$(MPackName).zip</MPackZipFile>
+      <MPackManifest>$(AddinDirectory)Properties\_Manifest.addin.xml</MPackManifest>
+      <AddinInfoFilePath>$(MPackSourcesDir)addin.info</AddinInfoFilePath>
+    </PropertyGroup>
+    
+    <MakeDir Directories="$(MPackSourcesDir)" Condition="!Exists('$(MPackSourcesDir)')" />
+
+    <!-- We need to resolve the language service assemblies to generate an addin.info for the mpack -->
+    <XmlPeek XmlInputPath="$(MPackManifest)" Query="/ExtensionModel/Runtime/Import/@assembly">
+      <Output TaskParameter="Result" ItemName="LanguageServiceAssemblies" />
+    </XmlPeek>
+
+    <ItemGroup>
+      <AddinInfoLines Include="&lt;Addin id=&quot;$(AddinId)&quot; namespace=&quot;$(AddinNamespace)&quot; version=&quot;$(AddinVersion)&quot; name=&quot;$(AddinDetailedName)&quot; author=&quot;$(Authors)&quot; description=&quot;$(Description)&quot; category=&quot;$(AddinCategory)&quot;&gt;" />
+      <AddinInfoLines Include="  &lt;Runtime&gt;" />
+      <AddinInfoLines Include="    &lt;Import assembly=&quot;%(LanguageServiceAssemblies.Identity)&quot; /&gt;" />
+      <AddinInfoLines Include="    &lt;Import assembly=&quot;$(AddinName).dll&quot; /&gt;" />
+      <AddinInfoLines Include="  &lt;/Runtime&gt;" />
+      <AddinInfoLines Include="  &lt;Dependencies&gt;" />
+      <AddinInfoLines Include="  &lt;/Dependencies&gt;" />
+      <AddinInfoLines Include="&lt;/Addin&gt;" />
+    </ItemGroup>
+
+    <!-- Generate the addin.info and gather sources for mpack zipping-->
+    <WriteLinesToFile File="$(AddinInfoFilePath)" Lines="@(AddinInfoLines)" Overwrite="true" />
+    <Copy SourceFiles="$(LanguageServiceOutputPath)\%(LanguageServiceAssemblies.Identity)" DestinationFolder="$(MPackSourcesDir)" />
+    <Copy SourceFiles="$(AddinOutputPath)$(AddinName).dll" DestinationFolder="$(MPackSourcesDir)" />
+
+    <!-- We cannot use the ZipArchive task due to how it functions in CoreCLR. The archive it generates is unreadable by Visual Studio for Mac. -->
+    <Exec Command="powershell.exe -NonInteractive -command &quot;&amp; { Compress-Archive -Path $(MPackSourcesDir)* -DestinationPath $(MPackZipFile) -Force; } &quot;" />
+    <Move SourceFiles="$(MPackZipFile)" DestinationFiles="$(MPackOutputPath)" />
+  </Target>
+
+</Project>
--- a/tooling/Microsoft.VisualStudio.Mac.RazorAddin/AddinMetadata.props
+++ b/tooling/Microsoft.VisualStudio.Mac.RazorAddin/AddinMetadata.props
@@ -1,0 +1,12 @@
+﻿<Project>
+  <PropertyGroup>
+    <VSForMacVersion>7.0</VSForMacVersion>
+    <AddinVersion Condition="'$(BuildNumber)'!=''">$(VSForMacVersion).$(BuildNumber)</AddinVersion>
+    <AddinVersion Condition="'$(BuildNumber)'==''">$(VSForMacVersion).999999</AddinVersion>
+    <AddinId>RazorAddin</AddinId>
+    <AddinNamespace>Microsoft.VisualStudio.Mac</AddinNamespace>
+    <AddinDetailedName>Razor Language Services</AddinDetailedName>
+    <AddinCategory>Web Development</AddinCategory>
+    <Description>Language services for ASP.NET Core Razor</Description>
+  </PropertyGroup>
+</Project>
--- a/tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
+++ b/tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
@@ -1,0 +1,72 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="AddinMetadata.props" />
+  
+  <PropertyGroup>
+    <TargetFramework>net461</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="Properties\_Manifest.addin.xml" LogicalName="_Manifest.addin.xml" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- 
+      By default the MonoDevelop.Addins package has restore tasks that depend on monodevelop existing on the machine. 
+      We can avoid this requirement by not letting it auto-restore monodevelop dependencies because we know which ones
+      we depend on and can bring them in manually.
+    -->
+    <PackageReference Include="MonoDevelop.Addins" Version="$(MonoDevelopAddinsVersion)" NoWarn="KRB4002" ExcludeAssets="build" />
+    <Reference Include="Mono.Addins">
+      <HintPath>$(NuGetPackageRoot)monodevelop.addins\$(MonoDevelopAddinsVersion)\build\Mono.Addins.dll</HintPath>
+    </Reference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- 
+      The extension project can not have a direct reference to its language service pieces. They are included via the manifest above.
+      This piece ensures that transitive builds/restores occurr for this project.
+    -->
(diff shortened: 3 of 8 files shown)


src/Microsoft.MonoDevelop.LanguageServices.Razor/Microsoft.MonoDevelop.LanguageServices.Razor.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
NTaylorMullen: Nothing in this project today. It's just a placeholder for VS for mac specific code for when it arises.

KirillOsenkov:
cc @mhutch

build/MPack.targets
```diff
@@ -0,0 +1,64 @@
...
+
+    <ItemGroup>
+      <AddinInfoLines Include="&lt;Addin id=&quot;$(AddinId)&quot; namespace=&quot;$(AddinNamespace)&quot; version=&quot;$(AddinVersion)&quot; name=&quot;$(AddinDetailedName)&quot; author=&quot;$(Authors)&quot; description=&quot;$(Description)&quot; category=&quot;$(AddinCategory)&quot;&gt;" />
+      <AddinInfoLines Include="  &lt;Runtime&gt;" />
+      <AddinInfoLines Include="    &lt;Import assembly=&quot;%(LanguageServiceAssemblies.Identity)&quot; /&gt;" />
+      <AddinInfoLines Include="    &lt;Import assembly=&quot;$(AddinName).dll&quot; /&gt;" />
+      <AddinInfoLines Include="  &lt;/Runtime&gt;" />
+      <AddinInfoLines Include="  &lt;Dependencies&gt;" />
+      <AddinInfoLines Include="  &lt;/Dependencies&gt;" />
+      <AddinInfoLines Include="&lt;/Addin&gt;" />
```
rynowak: using XML to generate XML is LOL

build/MPack.targets
```diff
@@ -0,0 +1,64 @@
...
+    </ItemGroup>
+
+    <!-- Generate the addin.info and gather sources for mpack zipping-->
+    <WriteLinesToFile File="$(AddinInfoFilePath)" Lines="@(AddinInfoLines)" Overwrite="true" />
+    <Copy SourceFiles="$(LanguageServiceOutputPath)\%(LanguageServiceAssemblies.Identity)" DestinationFolder="$(MPackSourcesDir)" />
+    <Copy SourceFiles="$(AddinOutputPath)$(AddinName).dll" DestinationFolder="$(MPackSourcesDir)" />
+
+    <!-- We cannot use the ZipArchive task due to how it functions in CoreCLR. The archive it generates is unreadable by MonoDevelop. -->
+    <Exec Command="powershell.exe -NonInteractive -command &quot;&amp; { Compress-Archive -Path $(MPackSourcesDir)* -DestinationPath $(MPackZipFile) -Force; } &quot;" />
```
pranavkm: You could add a custom build task to the repo and do all of this work in there. Might be cleaner than rendering xml and shelling out to powershell. 

https://github.com/aspnet/Coherence/blob/dev/build/tasks/ has an example of what it takes to write a task.

src/Microsoft.MonoDevelop.LanguageServices.Razor/Microsoft.MonoDevelop.LanguageServices.Razor.csproj
```diff
@@ -0,0 +1,13 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>net46</TargetFrameworks>
+    <Description>Razor is a markup syntax for adding server-side logic to web pages. This package contains the Razor design-time infrastructure for Mono Develop.</Description>
+    <EnableApiCheck>false</EnableApiCheck>
```
pranavkm: empty baseline instead?

src/Microsoft.MonoDevelop.LanguageServices.Razor/Microsoft.MonoDevelop.LanguageServices.Razor.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net46</TargetFrameworks>
+    <Description>Razor is a markup syntax for adding server-side logic to web pages. This package contains the Razor design-time infrastructure for Mono Develop.</Description>
```
pranavkm: Make sure we have the right branding etc for Mono Develop.

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
```diff
@@ -0,0 +1,72 @@
...
+    </Reference>
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- 
+      The extension project can not have a direct reference to its language service pieces. They are included via the manifest above.
+      This piece ensures that transitive builds/restores occurr for this project.
+    -->
+    <ProjectReference Include="..\..\src\Microsoft.MonoDevelop.LanguageServices.Razor\Microsoft.MonoDevelop.LanguageServices.Razor.csproj">
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
```
pranavkm: Might also need `PrivateAssets="true"` to avoid a package reference.

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
```diff
@@ -0,0 +1,72 @@
...
+  </ItemGroup>
+
+  <!-- We need to generate assembly attributes with build time information to let MonoDevelop know RazorAddins metadata at runtime. -->
+  <Target Name="PreBuild" BeforeTargets="PreBuildEvent">
+    <PropertyGroup>
+      <GeneratedAddinAssemblyInfo>$(IntermediateOutputPath)$(MSBuildProjectFile).AddinInfo.cs</GeneratedAddinAssemblyInfo>
+    </PropertyGroup>
+    
+    <ItemGroup>
+      <AssemblyAttributes Include="Mono.Addins.AddinAttribute">
```
pranavkm: Could we not do this in C# (like add an `AssemblyInfo.cs`)?

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
```diff
@@ -0,0 +1,72 @@
...
+        <_Parameter1>$(AddinId)</_Parameter1>
+        <Namespace>$(AddinNamespace)</Namespace>
+        <Version>$(AddinVersion)</Version>
+      </AssemblyAttributes>
+      <AssemblyAttributes Include="Mono.Addins.AddinNameAttribute">
```
pranavkm: Maybe call this `RazorAssemblyAttribute`? It's kinda close to `AssemblyAttribute` which is the name the SDK uses for it's itemgroup so might be confusing.

tooling/Microsoft.MonoDevelop.RazorAddin/Properties/_Manifest.addin.xml
```diff
@@ -0,0 +1,15 @@
...
+    <Import assembly="Microsoft.AspNetCore.Razor.Language.dll" />
+    <Import assembly="Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.dll" />
+    <Import assembly="Microsoft.AspNetCore.Mvc.Razor.Extensions.dll" />
+    <Import assembly="Microsoft.CodeAnalysis.Razor.dll" />
+    <Import assembly="Microsoft.CodeAnalysis.CSharp.dll" />
+    <Import assembly="Microsoft.MonoDevelop.LanguageServices.Razor.dll" />
+  </Runtime>
+  <Dependencies>
+  </Dependencies>
+</ExtensionModel>
```
pranavkm: New line

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <EmbeddedResource Include="Properties\_Manifest.addin.xml" LogicalName="_Manifest.addin.xml" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- 
+      By default the MonoDevelop.Addins package has restore tasks that depend on monodevelop existing on the machine. 
+      We can avoid this requirement by not letting it auto-restore monodevelop dependencies because we know which ones
+      we depend on and can bring them in manually.
+    -->
+    <PackageReference Include="MonoDevelop.Addins" Version="$(MonoDevelopAddinsVersion)" NoWarn="KRB4002" ExcludeAssets="build" />
```
pranavkm: Should this go in the lineup?

mhutch:
What exactly were the issues with MonoDevelop.Addins build tools?

It seems like you're (partly) replicating their functionality. I think it would make more sense to change them to work for your scenario.

FWIW, you can build against a specific copy of VS for Mac by setting [various properties](https://github.com/mhutch/MonoDevelop.AddinMaker/wiki/Build-System). This enables you, for example, to build a VS for Mac extension on Windows.

rynowak:
@mhutch our requirements are:
- no installed dependencies or SDKs (nuget only)
- works on windows
- project can be opened/build in VS

Ideally it's possible to build on non-windows but it's not a requirement.

If you have a solution that checks all of these boxes today we'll gladly use it. 

---

I believe @NTaylorMullen started with using a set of attributes in C# code to generate the XML manifest, but ran into a blocker when we needed to incorporate a build number coming from msbuild.  I suggested generating that attribute with C# and including it in the build but he wanted to go a different way and wrote the solution we have here.

Maybe you two should talk offline.

mhutch:
@rynowak all of those are true for the MonoDevelop.Addins targets. Just copy VS for Mac over from a Mac and set the `MDBinDir` MSBuild property to point to its `bin` directory.

The inability to open in VS comes from the project flavor (which is used to load the correct extension in VS for Mac) not the targets.

rynowak:
>  Just copy VS for Mac over from a Mac and set the MDBinDir MSBuild property to point to its bin directory.

> no installed dependencies or SDKs (nuget only)

mhutch:
@rynowak doesn't need to be installed, you just need to copy and unpack a zip file.

How exactly do you plan to be able to reference VS for Mac's libraries from this extension?

rynowak:
> How exactly do you plan to be able to reference VS for Mac's libraries from this extension?

Also nuget

The hope here is that we can build the mpack in every possible environment without any additional machine dependencies. Adding any dependency that is required to build and can't be delivered via nuget grows our 'minimum' configuration - this impacts a lot more people beyond the 3-4 that regularly work on to this repo.

mhutch:
@rynowak you plan to ship the VS for Mac assemblies on public NuGet?

rynowak:
no

build/MPack.targets
NTaylorMullen: Arent MSBuild tasks executed in a CoreCLR context?

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
NTaylorMullen: You could but the assembly info would not be consistent with addin info unless you duplicated all that information.

build/MPack.targets
pranavkm: Yup. I thought you had an issue with the specific structure of the produced zip rather than the format (the format is pretty standard, no?)

NTaylorMullen: The CoreCLR implementation of ZipArchive was re-written and has had a significant number of bugs already logged/fixed (heck our own ZIpArchive task works around one of the bugs). I spent some time investigating what could be wrong but didn't get far.

rynowak: I think we should pursue whatever issues are blocking us in parallel to getting this work done. 

We want this to be able to build everywhere, the hack with powershell is a speed bump we have to cross.

tooling/Microsoft.MonoDevelop.RazorAddin/Microsoft.MonoDevelop.RazorAddin.csproj
NTaylorMullen: It could but I'd still need to have the version so I can HintPath below. Going to hold off on having this in the LineUp until another project needs it.

src/Microsoft.MonoDevelop.LanguageServices.Razor/Microsoft.MonoDevelop.LanguageServices.Razor.csproj
NTaylorMullen: Gonna leave for now. This is how the rest of the VS specific assemblies are interacting with API check today. We'll update them all together when its time 😄

rynowak: Please log an issue for this. It's well past time.

NTaylorMullen:
🆙 📅

src/Microsoft.MonoDevelop.LanguageServices.Razor/Microsoft.MonoDevelop.LanguageServices.Razor.csproj
NTaylorMullen: We're still making API breaking changes to these assemblies. Are you just suggesting empty baselines for all of the VS specific assemblies? Or to have real baselines?

tooling/Microsoft.VisualStudio.Mac.RazorAddin/AddinMetadata.props
```diff
@@ -0,0 +1,12 @@
+﻿<Project>
+  <PropertyGroup>
+    <VSForMacVersion>7.0</VSForMacVersion>
+    <AddinVersion Condition="'$(BuildNumber)'!=''">$(VSForMacVersion).$(BuildNumber)</AddinVersion>
+    <AddinVersion Condition="'$(BuildNumber)'==''">$(VSForMacVersion).999999</AddinVersion>
+    <AddinId>RazorAddin</AddinId>
+    <AddinNamespace>Microsoft.VisualStudio.Mac</AddinNamespace>
```
rynowak: Is this right?

tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
```diff
@@ -0,0 +1,72 @@
...
+    <ProjectReference Include="..\..\src\Microsoft.VisualStudio.Mac.LanguageServices.Razor\Microsoft.VisualStudio.Mac.LanguageServices.Razor.csproj">
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <PrivateAssets>true</PrivateAssets>
+      <OutputItemType>Content</OutputItemType>
+      <Targets>Build</Targets>
+    </ProjectReference>
+  </ItemGroup>
+
+  <!-- We need to generate assembly attributes with build time information to let MonoDevelop know RazorAddins metadata at runtime. -->
+  <Target Name="PreBuild" BeforeTargets="PreBuildEvent">
```
rynowak: Can we give this a better target name?

tooling/Microsoft.VisualStudio.Mac.RazorAddin/AddinMetadata.props
NTaylorMullen: Yes, this results in the final mpack being:

`Microsoft.VisualStudio.Mac.RazorAddin.mpack`

tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
NTaylorMullen: Sure

PR #1702, opened 2017-09-27
Preserve the build number when building the VSIX

natemcmaster:
Changes 
 - Pass $(BuildNumber) to msbuild.exe /t:Restore. I was hitting weird issues building this locally because nupkg version is determined by /t:Restore
 - Restore the VSIX project when executing `build /t:Restore`

### DIFF
--- a/build/VSIX.targets
+++ b/build/VSIX.targets
@@ -1,13 +1,14 @@
 <Project>
   <PropertyGroup>
-    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);GenerateVSIX</PackageDependsOn>
+    <RestoreDependsOn Condition="'$(OS)'=='Windows_NT'">$(RestoreDependsOn);RestoreVSIX</RestoreDependsOn>
+    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);PackageVSIX</PackageDependsOn>
     <VSIXName>Microsoft.VisualStudio.RazorExtension</VSIXName>
     <VSIXProject>$(RepositoryRoot)tooling\$(VSIXName)\$(VSIXName).csproj</VSIXProject>
   </PropertyGroup>
 
   <Target
     Name="GenerateVSIX"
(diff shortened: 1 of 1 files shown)


build/VSIX.targets
```diff
@@ -1,13 +1,14 @@
...
   <PropertyGroup>
-    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);GenerateVSIX</PackageDependsOn>
+    <RestoreDependsOn Condition="'$(OS)'=='Windows_NT'">$(RestoreDependsOn);RestoreVSIX</RestoreDependsOn>
+    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);PackageVSIX</PackageDependsOn>
     <VSIXName>Microsoft.VisualStudio.RazorExtension</VSIXName>
     <VSIXProject>$(RepositoryRoot)tooling\$(VSIXName)\$(VSIXName).csproj</VSIXProject>
   </PropertyGroup>
 
   <Target
     Name="GenerateVSIX"
```
rynowak: Where is this target called now? Is this obsolete?

rynowak: Or is this just left here as a "meta-target"

build/VSIX.targets
```diff
@@ -1,13 +1,14 @@
 <Project>
```
rynowak: @natemcmaster - is it worth trying to move this set of targets to aspnet/BuildTools given that we're going to add another VSIX in the future?

build/VSIX.targets
natemcmaster: I left it as a 'meta-target', as CLI sugar if you only want to build the vsix project. I can remove it you'd rather clean it out.

build/VSIX.targets
rynowak: Given that we just spun our last build for 15.5 this is a safe time to churn stuff. We won't need to build from `dev` for a little while.

natemcmaster: Yes. It'll require a little more design, but I'm thinking it is worth providing a way for a repo to say "I need to build X and Y with MSBuild.exe, not dotnet-msbuild".

Issue #1538, opened 2017-07-11
Implement brace smart indent completion.


NTaylorMullen:
Brace completion should not be part of the core Razor syntax tree contract. For example [this](https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/EditorFeatures/Core/Implementation/AutomaticCompletion/BraceCompletionSessionProvider.cs) is how Roslyn does it for CSharp.

NTaylorMullen:
After a significant amount of investigation, decoupling brace completion from the editing loop is not currently possible without more drastic editor changes. For the time being we should implement brace completion as part of our current VisualStudioRazorParser.

NTaylorMullen:
This depends on https://github.com/aspnet/Razor/issues/1633

rynowak:
This is moving out to a future 15.X release. Sticking it in 2.1 until we have that scoped out.

NTaylorMullen:
31c16af40bbe10c9fb924be559bd42955aa4e246

PR #1713, opened 2017-10-06
Add smart indentation for brace completion.

NTaylorMullen:
This is a middle-ground approach to implementing brace completion smart indent without refactoring the entire `VisualStudioRazorParser`. Decided to use the "smart indent" verbiage because that's what the visual studio editor uses.

Here's what this PR accomplishes. Pipe | indicates cursor

## Razor code block smart indentation
A user types `@{|`, Roslyn autocompletes it to `@{ |}`. User then hits enter and **without** this PR it becomes:
```
@{
    |}
```
This PR fixes that and makes it:
```
@{
    |
}
```

## Razor directive block smart indentation
A user types `@functions {|`, Roslyn autocompletes it to `@functions {|}`. User then hits enter and **without** this PR it becomes:
```
@functions {
    |}
```
This PR fixes that and makes it:
```
@functions {
    |
}
```

One significant note is that the cursor position in the above examples don't have spaces before it. That's the editor auto-indent that will auto-add those spaces if a user then attempts to add content.

---------------------------------------------------------------------------------

- Added a standalone brace smart indenter that listens to `ITextBuffer` changed events to determine when a brace completion event needs to be handled.
- Added extension methods to handle interaction with the `ITextBuffer` and document trackers and text views.
- Added a `BraceSmartIndenterTest` to verify all parts of the smart indenter.
- Moved private test infrastructure classes into their own files and expanded on their functionality to enable the brace completion smart indent scenarios.

#1538


src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,218 @@
...
+    /// <example>
+    /// Attempting to insert a newline (pipe indicates the cursor):
+    /// @{ |}
+    /// Should result in the text buffer looking like the following:
+    /// @{
+    ///     |
+    /// }
+    /// This is also true for directive block scenarios.
+    /// </example>
+    internal class BraceSmartIndenter : IDisposable
```
NTaylorMullen: `IDisposable` for now since the `VisualStudioRazorParser` is. This will probably change once we own the lifetime of the parser (my next item).

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,218 @@
...
+                var edit = textView.TextBuffer.CreateEdit();
+                var indent = CalculateIndent(textView.TextBuffer, context.ChangePosition);
+                edit.Insert(caret.Position, indent);
+                edit.Apply();
+
+                // Place the caret inbetween the braces (before our indent).
+                var currentSnapshotPoint = new SnapshotPoint(textView.TextBuffer.CurrentSnapshot, caret.Position);
+                textView.Caret.MoveTo(currentSnapshotPoint);
+
+                // For Razor metacode cases the editors smart indent wont kick in automatically. This forces the smart indent.
```
rynowak: `editor's`

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,218 @@
...
+                // For example attempting to do enter a newline between the functions directive: @functions {} will not auto-indent
+                // inbetween the braces unless we forcefully move to end of line.
```
rynowak: `in between`

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,218 @@
...
+            if (contentChange.Changes.Count == 0)
+            {
+                return;
+            }
+
+            var firstChange = contentChange.Changes[0];
+            var lastChange = contentChange.Changes[contentChange.Changes.Count - 1];
+            var oldLength = lastChange.OldEnd - firstChange.OldPosition;
+            var newLength = lastChange.NewEnd - firstChange.NewPosition;
+            var newText = contentChange.After.GetText(firstChange.NewPosition, newLength);
```
rynowak: Can we put this in a private method called `GetNormalizedChange` - if this is duplicated with the parser maybe an extension method?

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+
+        // Internal for testing
+        internal static bool TryCreateIndentationContext(int changePosition, int changeLength, string finalText, ITextBuffer textBuffer, out BraceIndentationContext context)
+        {
+            var focusedTextView = textBuffer.GetFocusedTextView();
+            if (focusedTextView != null && ParserHelpers.IsNewLine(finalText))
+            {
+                var currentSnapshot = textBuffer.CurrentSnapshot;
```
rynowak: I don't feel like I super understand what this code is doing

src/Microsoft.VisualStudio.Editor.Razor/TextBufferExtensions.cs
```diff
@@ -17,5 +20,34 @@ public static bool IsRazorBuffer(this ITextBuffer textBuffer)
 
             return textBuffer.ContentType.IsOfType(RazorLanguage.ContentType);
         }
+
+        public static VisualStudioDocumentTracker GetDocumentTracker(this ITextBuffer textBuffer)
```
rynowak: Why do we need this? Can you just use the service?

rynowak: I really really don't like the idea of hanging lots of interdependent functionality on extensions, I'd prefer that it use services.

For instance you could add `GetFocusedTextView` to the document tracker service

test/Microsoft.VisualStudio.Editor.Razor.Test.Common/StringTextSnapshot.cs
```diff
@@ -4,15 +4,32 @@
...
         {
             Content = content;
+            _lines = new List<ITextSnapshotLine>();
+
+            var start = 0;
+            var delimiterIndex = 0;
+            while (delimiterIndex != -1)
+            {
+                delimiterIndex = Content.IndexOf(Environment.NewLine, start);
+                var nextLineStartIndex = delimiterIndex != -1 ? delimiterIndex + Environment.NewLine.Length : Content.Length;
```
rynowak: This isn't right, we need to be able to work with all line break types in all environements

test/Microsoft.VisualStudio.Editor.Razor.Test/Infrastructure/TestEdit.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.Editor.Razor.Test.Infrastructure
```
rynowak: Don't use gross namespaces, just put this in `Microsoft.VisualStudio.Test`

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
```diff
@@ -0,0 +1,382 @@
...
+            var caret = CreateCaretFrom(3 + change.Length, afterChangeSnapshot);
+            TestTextBuffer textBuffer = null;
+            var focusedTextView = CreateFocusedTextView(() => textBuffer, caret);
+            textBuffer = CreateTrackedTextBuffer(initialSnapshot, focusedTextView);
+            var editorOperations = new Mock<IEditorOperations>();
+            editorOperations.Setup(operations => operations.MoveToEndOfLine(false));
+            var editorOperationsFactory = new Mock<IEditorOperationsFactoryService>();
+            editorOperationsFactory.Setup(factory => factory.GetEditorOperations(It.IsAny<ITextView>()))
+                .Returns(editorOperations.Object);
+            var braceSmartIndenter = new BraceSmartIndenter(Dispatcher, textBuffer, editorOperationsFactory.Object);
```
rynowak: This is really opaque and hard to understand, can we move some of this boilerplate to properties

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
```diff
@@ -0,0 +1,382 @@
...
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public class BraceSmartIndenterTest : ForegroundDispatcherTestBase
+    {
+        [Fact]
+        public void TextBuffer_OnPostChanged_MovesToEndOfLine()
+        {
+            // Arrange
+            var change = Environment.NewLine;
```
rynowak: This isn't right, we need to work with all newline types in all places, we shouldn't test based on `Environment.NewLine`

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
```diff
@@ -0,0 +1,382 @@
...
+
+            // Act
+            textBuffer.ApplyEdit(edit);
+
+            // Assert
+            editorOperations.VerifyAll();
+        }
+
+        [Fact]
+        public void TextBuffer_OnPostChanged_PlacesCursorBackToOriginalLocation()
```
rynowak: `TextBuffer_OnPostChanged_MovesToEndOfLine` and `TextBuffer_OnPostChanged_PlacesCursorBackToOriginalLocation` - what's different about these cases?

NTaylorMullen: They're exactly the same just verifying two different aspects of what should happen. In editor land those aspects have impacts on what the user sees. Would rather have 1 test for each so if we break one of those expectations down the line we have 1 specific test failing rather than 1 multi-purpose test.

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
NTaylorMullen: Ya, getting all the test infrastructure to work with all variants of newline is painful which is why I stuck to `Environment.NewLine`. The product code on the other hand does understand all aspects. I'll put in some time to make the test infrastructure more resilient.

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
NTaylorMullen: Why not have it opaque? It's a standalone component.

src/Microsoft.VisualStudio.Editor.Razor/TextBufferExtensions.cs
NTaylorMullen: Sure we can move these

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
NTaylorMullen: I'll pull pieces out into methods to be more descriptive.

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
rynowak: > They're exactly the same just verifying two different aspects of what should happen. 

But why though

> Would rather have 1 test for each so if we break one of those expectations down the line we have 1 specific test failing rather than 1 multi-purpose test.

This just proves that the factoring is wrong. 

These tests (as written) are integration tests. Unit tests have the goal that you're stating, but these are not that.

NTaylorMullen:
🆙 📅 

Addressed all feedback. Key changes are:
- Added a BraceSmartIndenter integration test class (removed the move to end of line and caret tests)
- Refactored the smart indenter class to be more unit testable. This also involved adding a common base class to share with the integration test variant.
- Updated VisualStudioRazorParser to consume document trackers only on foreground threads. This was required when we added GetTracker to the document tracker services for ITextBuffer
- Moved text buffer change logic into an extension method.

src/Microsoft.VisualStudio.Editor.Razor/TextContentChangedEventArgsExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.Text
+{
+    internal static class TextContentChangedEventArgsExtensions
+    {
+        public static bool TextChangeOccurred(this TextContentChangedEventArgs args, out (ITextChange firstChange, ITextChange lastChange, string newText, string oldText) changeInformation)
```
NTaylorMullen: Add tests.

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,266 @@
...
+                // }
+
+                // For Razor metacode cases the editor's smart indent wont kick in automatically. 
+                TriggerSmartIndent(textView);
+
+                // @{
+                //     |
+                // }
+            }
+        }
```
rynowak: This is solid 👍

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,266 @@
...
+                    {
+                        context = new BraceIndentationContext(focusedTextView, changePosition);
+                        return true;
+                    }
+                }
+            }
+
+            context = null;
+            return false;
+        }
```
rynowak: This makes way more sense now 👍

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
```diff
@@ -79,6 +108,8 @@ public VisualStudioRazorParser(ForegroundDispatcher dispatcher, ITextBuffer buff
 
         public string FilePath { get; }
 
+        private readonly ErrorReporter _errorReporter;
```
rynowak: put this with the fields yo

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTrackerFactory.cs
```diff
@@ -114,6 +114,25 @@ public override VisualStudioDocumentTracker GetTracker(ITextView textView)
...
+                throw new ArgumentNullException(nameof(textBuffer));
+            }
+
+            _foregroundDispatcher.AssertForegroundThread();
+
+            // A little bit of hardening here, to make sure our assumptions are correct.
+            DefaultVisualStudioDocumentTracker tracker;
+            if (!textBuffer.Properties.TryGetProperty(typeof(VisualStudioDocumentTracker), out tracker))
+            {
+                Debug.Fail("The document tracker should be initialized");
```
rynowak: This isn't right, what if the buffer isn't a razor buffer? This should just return null gracefully.

The ITextView version has the asserts because at that point we've already determined that it's a razor buffer

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterIntegrationTest.cs
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.VisualStudio.Test;
+using Microsoft.VisualStudio.Text;
+using Xunit;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public partial class BraceSmartIndenterIntegrationTest : BraceSmartIndenterTestBase
```
rynowak: Which is it? are they using inheritance or partials?

test/Microsoft.VisualStudio.Editor.Razor.Test/Infrastructure/TestTextBuffer.cs
```diff
@@ -0,0 +1,149 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Utilities;
+
+namespace Microsoft.VisualStudio.Test
+{
+    public class TestTextBuffer : ITextBuffer
```
rynowak: Did you ask about those mocks?

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,266 @@
...
+
+        internal static bool BeforeClosingBrace(int linePosition, ITextSnapshotLine lineSnapshot)
+        {
+            var lineText = lineSnapshot.GetText();
+            while (linePosition < lineSnapshot.Length)
+            {
+                if (!char.IsWhiteSpace(lineText[linePosition]))
+                    break;
```
rynowak: braces

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,266 @@
...
+            return beforeClosingBrace;
+        }
+
+        internal static bool AfterOpeningBrace(int linePosition, ITextSnapshotLine lineSnapshot)
+        {
+            var lineText = lineSnapshot.GetText();
+            while (linePosition >= 0)
+            {
+                if (!char.IsWhiteSpace(lineText[linePosition]))
+                    break;
```
rynowak: braces

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -0,0 +1,266 @@
...
+                linePosition--;
+            }
```
rynowak: This loop might be simpler written as a `for`

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
NTaylorMullen: #thanksvs

test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterIntegrationTest.cs
NTaylorMullen: lmao, well lets make this a `public partial abstract internal protected facepalm class`

test/Microsoft.VisualStudio.Editor.Razor.Test/Infrastructure/TestTextBuffer.cs
NTaylorMullen: Talked to the WTE folks and they rely on a privately constructed MEF service catalog (probing paths etc.) to create buffer factories; insufficient for our usecase. Will need to reach out to others.

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
NTaylorMullen: #taylorcontrolflow

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTrackerFactory.cs
NTaylorMullen: Added a RazorBuffer check above and added tests.

NTaylorMullen:
🆙 📅

Issue #1547, opened 2017-07-12
Span nodes are not linked outside of tests.


NTaylorMullen:
This means that each `Span`'s Next member is always null.

ajaybhargavb:
@NTaylorMullen, could you give me a little more context on this? I looked at `SyntaxTreeNode` and there is no `Next` property. Am I looking in the wrong place?

NTaylorMullen:
Sorry, this issue is misleading. I was referring to `Span`'s `Next`/`Previous` properties.

ajaybhargavb:
f4e9ddad22f78a704317dcfcfa655bedb6a35096

PR #1717, opened 2017-10-10
Link syntax tree Span nodes

ajaybhargavb:
#1547 

@NTaylorMullen 

Turned out to be an easy fix.


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/RazorParserTest.cs
```diff
@@ -60,5 +61,29 @@ public void ParseMethodUsesProvidedParserListenerIfSpecified()
                                .Accepts(AcceptedCharactersInternal.NonWhiteSpace)),
                     factory.Markup(" baz")));
         }
+
+        [Fact]
+        public void Parse_SyntaxTreeSpansAreLinked()
+        {
+            var factory = new SpanFactory();
```
NTaylorMullen: throw in arrange.

rynowak:
Are there tests that need to be enabled because of this?

Issue #1630, opened 2017-08-18
Manage VisualStudioRazorParser lifetime


NTaylorMullen:
Today we rely on the WTE's RazorCodeGenerator to create and dispose our parser; instead, we should detect when a Razor document opens and create/dispose ourselves accordingly.

In order to accomplish this task we'll need to expose some extensibility points to enable VisualStudio to configure the Razor parser in the correct way.

rynowak:
This will be done in a future 15.X release. Moving to 2.1 so we don't lose track of it.

NTaylorMullen:
212d97e51170d1df49a6a7569e6a063e66a89c99

PR #1722, opened 2017-10-13
Manage VisualStudioRazorParser lifetime.

NTaylorMullen:
- Exposed `VisualStudioRazorParser`, `DocumentStructureChangedEventArgs` and `ICanHasContextChangedListener` as ways to consume the new parser for a Razor document.
- Split the `VisualStudioRazorParser` into an abstract base and an implementation to avoid internal constructors.
- Changed the parser and corresponding smart indenter to take in document trackers, template engine factories and parser context change listeners. Of these additions the parser context change listeners will be deprecated once we own the TagHelper discovery mechanisms.
- Changed how the parser manages its internal parsing life cycle. It now creates template engines when the document tracker tells it to. So when project changes happen or new documents are opened the parser will re-instantiate its internal parsers to ensure that they're parsing against the correct configurations.
- Split the `VisualStudioDocumentTrackerFactory` into a provider and a factory. The provider exposes the old `GetTracker` methods where the factory is the text view connection listener implementation.
- Changed the code document provider to now use the parser provider in order to locate code documents associated with buffers. Prior to this that logic was hard coded.
- Removed old template engine reconstruction logic in the document tracker now that the parser owns that piece.

#1630


src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+using System;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class DefaultVisualStudioDocumentTrackerProvider : VisualStudioDocumentTrackerProvider
```
NTaylorMullen: Had to split this from the factory in order to prevent a circular dependency with the parser factory.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+            // other functionality for this class will maintain state correctly for the other buffers.
+            var textBuffer = textView.BufferGraph.GetRazorBuffers().FirstOrDefault();
+            if (textBuffer == null)
+            {
+                // No Razor buffer, nothing to track.
+                return null;
+            }
+
+            // A little bit of hardening here, to make sure our assumptions are correct.
+            if (!textBuffer.Properties.TryGetProperty(typeof(VisualStudioDocumentTracker), out VisualStudioDocumentTracker tracker))
```
davidfowl: nit: needs more var

NTaylorMullen: lol could do `TryGetProperty<VisualStudioDocumentTracker>` to use `var` but doesn't buy anything imo. Properties bag is `object -> object`

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
rynowak: This is a gross hack - is this necessary because the document tracker also needs to be able to create parsers?

It might be less of a gross hack to just build **all** of the initialization into the tracker factory.

Either way you decide, document this decision in comments so it's not lost.

rynowak: Also as long as it's internal, I don't super care

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+{
+    internal class DefaultVisualStudioRazorParser : VisualStudioRazorParser, IDisposable
+    {
+        // Internal for testing.
+        internal TimeSpan IdleDelay = TimeSpan.FromSeconds(3);
+        internal Timer _idleTimer;
+
+        private readonly object IdleLock = new object();
+        private readonly ICompletionBroker _completionBroker;
+        private readonly IEnumerable<ICanHasContextChangedListener> _parserContextChangedListeners;
```
rynowak: nah

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+        private readonly VisualStudioDocumentTracker _documentTracker;
+        private readonly ForegroundDispatcher _dispatcher;
+        private readonly RazorTemplateEngineFactoryService _templateEngineFactory;
+        private readonly string _projectDirectory;
```
rynowak: Is this part of the tracker? should it be?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+
+            if (templateEngineFactory == null)
+            {
+                throw new ArgumentNullException(nameof(templateEngineFactory));
+            }
+
+            if (string.IsNullOrEmpty(filePath))
+            {
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(filePath));
+            }
```
rynowak: Should this be part of the tracker?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+
+            if (string.IsNullOrEmpty(projectDirectory))
+            {
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(projectDirectory));
+            }
```
rynowak: Should this be part of the tracker?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
rynowak: ## Another Idea

Should we just have a **BufferIntializer** service? And then have all of the other things call into it? It might be more straightforward to have all of the wire-ups in one place.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        public override RazorTemplateEngine TemplateEngine => _templateEngine;
+
+        public override string FilePath { get; }
+
+        public override RazorCodeDocument CodeDocument => _codeDocument;
+
+        public override ITextSnapshot Snapshot => _snapshot;
+
+        public override ITextBuffer TextBuffer { get; }
```
rynowak: nit: I don't super-love mixing field-backed properties and auto-props. It makes it a bit harder to chase down what's going the on.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+
+        public override async void Reparse()
+        {
+            // Can be called from any thread
```
rynowak: Who is calling this from a background thread?

rynowak: This should return `Task`

rynowak: BTW the problem with this approach is that callers can't look at the results due to threading.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,372 @@
...
+
+            _parser.Start();
+
+            // This is temporary until we own the TagHelper resolution system. At that point the parser will push out updates
+            // via DocumentStructureChangedEvents when contexts change. For now, listeners need to know more information about
+            // the parser.
+            foreach (var contextChangeListener in _parserContextChangedListeners)
+            {
+                contextChangeListener.OnContextChanged(this);
+            }
```
rynowak: I'm willing to accept this as a temporary thing.

src/Microsoft.VisualStudio.Editor.Razor/ICanHasContextChangedListener.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    /// <summary>
+    /// This class will ceases to be useful once the Razor runtime owns TagHelper discovery
+    /// </summary>
+    public interface ICanHasContextChangedListener
```
rynowak: OK, I'm willing to accept this as a temporary thing

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioDocumentTrackerProvider.cs
```diff
@@ -1,12 +1,13 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.CodeAnalysis.Host;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
 
 namespace Microsoft.VisualStudio.Editor.Razor
 {
-    public abstract class VisualStudioDocumentTrackerFactory
+    public abstract class VisualStudioDocumentTrackerProvider : ILanguageService
```
rynowak: Can we just delete the work provider from the world? I'm not happy to see it popping up all over. It doesn't mean anything

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
```diff
@@ -2,266 +2,25 @@
...
-                    return;
-                }
-
-                CodeDocument = args.CodeDocument;
-                Snapshot = args.Snapshot;
-                _partialParser = new RazorSyntaxTreePartialParser(CodeDocument.GetSyntaxTree());
-                DocumentStructureChanged(this, args);
-            }
-        }
+        public abstract void Reparse();
```
rynowak: This should be `Task` and called `ReparseAsync`

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParserFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class VisualStudioRazorParserFactory
```
rynowak: The public one should be factory

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParserProvider.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class VisualStudioRazorParserProvider : ILanguageService
```
rynowak: Can we use `Accessor` instead of `Provider`

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTrackerFactory.cs
```diff
@@ -19,30 +17,37 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor.Editor
     [ContentType(RazorLanguage.ContentType)]
     [TextViewRole(PredefinedTextViewRoles.Document)]
     [Export(typeof(IWpfTextViewConnectionListener))]
-    [Export(typeof(VisualStudioDocumentTrackerFactory))]
-    internal class DefaultVisualStudioDocumentTrackerFactory : VisualStudioDocumentTrackerFactory, IWpfTextViewConnectionListener
+    internal class DefaultVisualStudioDocumentTrackerFactory : IWpfTextViewConnectionListener
```
rynowak: This should have a different name YO

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
davidfowl: Ah,

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: It kind of is via its `Project` Roslyn property but it explodes if accessed prior to the Roslyn project system being created. We could make an exception for this directory and make it always available. on the tracker.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: It could be I'd need to look into how that information gets refreshed when the document moves and when that information is available (today we get it from the Roslyn text document associated with the buffer).

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: > Who is calling this from a background thread?

When Idle determines it needs to be called and when the WTE code determines TagHelpers have changed. Eventually this method wont need to be exposed at all once we own the entire TagHelper decision stack.



> BTW the problem with this approach is that callers can't look at the results due to threading.

Hmm? not sure I follow, could you expand.

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioDocumentTrackerProvider.cs
NTaylorMullen: lol, i'll rename it `VisualStudioHelperDocumentTrackerHelperHelper`. If you don't like provider please provide a good alternative.

NTaylorMullen: nm, you did with Accessor. Will rename

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTrackerProvider.cs
NTaylorMullen: > Should we just have a BufferIntializer service? And then have all of the other things call into it? It might be more straightforward to have all of the wire-ups in one place

You thinking the service would be the bit that inserts all the tracker + the parser? Or were you thinking it'd be a one stop shop for `GetXYZFromBuffer`?

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParserFactory.cs
NTaylorMullen: Hu? There doesn't need to be a public one. Could you expand?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTrackerFactory.cs
NTaylorMullen: How do you feel about having a different service that manages creation of document trackers?

```C#
DefaultVisualStudioDocumentTrackerFactory
    VisualStudioDocumentTracker Create(ITextBuffer textBuffer)
```

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: Investigated this and can definitely track that data on the document tracker. Also found that we don't track document movements correctly in our document tracker: https://github.com/aspnet/Razor/issues/1725

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -129,58 +138,17 @@ public void Subscribe()
         public void Unsubscribe()
         {
             _projectManager.Changed -= ProjectManager_Changed;
+
+            // Detached from project.
+            _isSupportedProject = false;
+            _project = null;
+            OnContextChanged(project: null);
```
NTaylorMullen: Notify listeners on unsubscribe so the parser can shut down its internal background parser.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorTextViewConnectionListener.cs
```diff
@@ -0,0 +1,156 @@
...
+                }
+
+                if (!_textBufferFactoryService.TryGetDocumentTracker(textBuffer, out var documentTracker) ||
+                    !(documentTracker is DefaultVisualStudioDocumentTracker tracker))
+                {
+                    Debug.Fail("Tracker should always be available given our expectations of the VS workflow.");
+                    return;
+                }
+
+                // REVIEWERS: Thinking of removing tracker.TextViewsInternal and replacing it with two methods:
```
NTaylorMullen: 👀

src/Microsoft.VisualStudio.Editor.Razor/TextBufferFactoryService.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class TextBufferFactoryService
```
NTaylorMullen: Chose this name to correspond with some of the other VS/Roslyn types that create/retrieve TextDocument/ITextDocuments. This is the text buffer initializer idea.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultTextBufferFactoryService.cs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        private static readonly object RazorTextBufferInitializationKey = new object();
+
+        private readonly VisualStudioDocumentTrackerFactory _documentTrackerFactory;
+        private readonly VisualStudioRazorParserFactory _parserFactory;
+        private readonly BraceSmartIndenterFactory _braceSmartIndenterFactory;
+
+        [ImportingConstructor]
+        public DefaultTextBufferFactoryService(
+            VisualStudioDocumentTrackerFactory documentTrackerFactory,
```
NTaylorMullen: There are now factories for each of the initialization pieces (tracker, parser, smartindenter). I felt that it was much cleaner to isolate the dependencies of factories and ultimately keep this class straightforward.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultTextBufferFactoryService.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        public override bool TryGetDocumentTracker(ITextBuffer textBuffer, out VisualStudioDocumentTracker documentTracker)
+        {
+            if (textBuffer == null)
+            {
+                throw new ArgumentNullException(nameof(textBuffer));
+            }
+
+            if (!textBuffer.IsRazorBuffer())
```
NTaylorMullen: Could normalize all of these methods into a single generic method but wanted to keep them separate for clarity.

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenterFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class BraceSmartIndenterFactory
```
rynowak: Why is this needed?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,387 @@
...
+            // the parser. In the case that the tracker does not belong to a supported project the editor will tear down its
+            // attachment to the parser when it recognizes the document closing.
+            foreach (var contextChangeListener in _parserContextChangedListeners)
+            {
+                contextChangeListener.OnContextChanged(this);
+            }
+
+            if (_parser.IsIdle)
+            {
+                ReparseAsync().GetAwaiter().GetResult();
```
rynowak: You should not block here. This will wait for the reparse to complete right?

src/Microsoft.VisualStudio.Editor.Razor/TextBufferFactoryService.cs
rynowak: OK. I'm not crazy about the name but it's internal so #YOLO

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioDocumentTrackerFactory.cs
```diff
@@ -2,14 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Text.Editor;
 
 namespace Microsoft.VisualStudio.Editor.Razor
 {
-    public abstract class VisualStudioDocumentTrackerFactory
+    internal abstract class VisualStudioDocumentTrackerFactory
```
rynowak: Is this needed?

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
```diff
@@ -2,266 +2,26 @@
...
-                    return;
-                }
-
-                CodeDocument = args.CodeDocument;
-                Snapshot = args.Snapshot;
-                _partialParser = new RazorSyntaxTreePartialParser(CodeDocument.GetSyntaxTree());
-                DocumentStructureChanged(this, args);
-            }
-        }
+        public abstract Task ReparseAsync();
```
rynowak: Does this task complete when the parse is completed? Or when then the parse is enqued?

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParserFactory.cs
rynowak: I'm not sure this comment makes snes

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultBraceSmartIndenterFactory.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (workspace == null)
+            {
+                throw new ArgumentNullException(nameof(workspace));
+            }
+
+            _editorOperationsFactory = editorOperationsFactory;
+            _dispatcher = workspace.Services.GetRequiredService<ForegroundDispatcher>();
+        }
+
+        public override BraceSmartIndenter Create(VisualStudioDocumentTracker documentTracker)
```
rynowak: I'm really not a fan of this kind of signature, because if we add more concepts, it will change more. 

I think it's fine because it's internal, but I want to at least point out the scalability issues with this design.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultTextBufferFactoryService.cs
rynowak: ok

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultTextBufferFactoryService.cs
rynowak: The way you have it is best

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -107,19 +110,25 @@ public void Subscribe()
...
+                (hierarchy = _projectService.GetHierarchy(_textBuffer)) != null &&
+                _textDocumentFactory.TryGetTextDocument(_textBuffer, out var textDocument))
             {
                 projectPath = _projectService.GetProjectPath(hierarchy);
                 isSupportedProject = _projectService.IsSupportedProject(hierarchy);
             }
+            else
+            {
+                return;
+            }
```
rynowak: Why is the else necessary? What case are you handling now that wasn't handled before?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -107,19 +110,25 @@ public void Subscribe()
 
                 // We expect the document to have a hierarchy even if it's not a real 'project'.
                 // However the hierarchy can be null when the document is in the process of closing.
-                (hierarchy = _projectService.GetHierarchy(_textBuffer)) != null)
+                (hierarchy = _projectService.GetHierarchy(_textBuffer)) != null &&
+                _textDocumentFactory.TryGetTextDocument(_textBuffer, out var textDocument))
```
rynowak: This seems suspect.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -79,6 +78,10 @@ internal class DefaultVisualStudioDocumentTracker : VisualStudioDocumentTracker
 
         public override Workspace Workspace => _workspace;
 
+        public override string FilePath => _filePath;
```
rynowak: Should we just pass in the filename on construction?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
rynowak: ??? We need to discuss this

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioRazorParserFactory.cs
```diff
@@ -0,0 +1,72 @@
...
+using Microsoft.VisualStudio.Language.Intellisense;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor.Editor
+{
+    [System.Composition.Shared]
+    [Export(typeof(VisualStudioRazorParserFactory))]
+    internal class DefaultVisualStudioRazorParserFactory : VisualStudioRazorParserFactory
+    {
+        private readonly ForegroundDispatcher _dispatcher;
+        private readonly CodeAnalysis.Razor.RazorTemplateEngineFactoryService _templateEngineFactoryService;
```
rynowak: Add an alias-ed `using`

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorTextViewConnectionListener.cs
rynowak: Add those methods on `DefaultVisualStudioDocumentTracker` - not on the abstract class and we're fine 👍

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: Nope, just blocks till the parse request is enqueued.

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
NTaylorMullen: https://github.com/aspnet/Razor/pull/1722#discussion_r145479745

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
NTaylorMullen: If we eventually want to control what happens when a file moves we'll need to have direct control of the file path.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
NTaylorMullen: totally didn't mean to commit this piece; should have been undone.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
NTaylorMullen: Talked offline, this is fine.

NTaylorMullen:
🆙 📅, no longer in design

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -0,0 +1,423 @@
...
+        public override event EventHandler<DocumentStructureChangedEventArgs> DocumentStructureChanged;
+
+        // Internal for testing.
+        internal TimeSpan IdleDelay = TimeSpan.FromSeconds(3);
+        internal Timer _idleTimer;
+        internal BackgroundParser _parser;
+
+        private readonly object IdleLock = new object();
+        private readonly ICompletionBroker _completionBroker;
+        private readonly IEnumerable<ICanHasContextChangedListener> _parserContextChangedListeners;
```
rynowak: We will need a more serious name than this, even if it's temporary. #NoFunRule

Consider this a blocker. `IContextChangedListener` would be fine

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioDocumentTrackerFactory.cs
rynowak: Leave this public.

src/Microsoft.VisualStudio.Editor.Razor/TextBufferFactoryService.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        public abstract bool TryGetParser(ITextBuffer textBuffer, out VisualStudioRazorParser parser);
+
+        public abstract bool TryGetDocumentTracker(ITextBuffer textBuffer, out VisualStudioDocumentTracker documentTracker);
+
+        public abstract bool TryGetSmartIndenter(ITextBuffer textBuffer, out BraceSmartIndenter braceSmartIndenter);
+    }
```
rynowak: Let's call this `RazorEditorFactoryService` and make it public with just the `TryGetDocumentTracker` public

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
```diff
@@ -14,63 +16,32 @@
...
-                    string.Empty, 
-                    new DefaultErrorReporter(),
-                    new TestCompletionBroker(),
-                    new Mock<VisualStudioDocumentTrackerFactory>().Object,
-                    new Mock<IEditorOperationsFactoryService>().Object));
-        }
-
-        // [Fact] Silent skip to avoid warnings. Skipping until we can control the parser more directly.
+        [ForegroundFact]
         private void BufferChangeStartsFullReparseIfChangeOverlapsMultipleSpans()
```
rynowak: Is this passing? We don't want 'silent skips' We want visible skips with bugs associated.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
```diff
@@ -234,8 +205,8 @@ void ApplyAndVerifyPartialChange(TestEdit testEdit, string expectedCode)
             }
         }
 
-        // [Fact] Silent skip to avoid warnings. Skipping until we can control the parser more directly.
-        private void ImplicitExpressionProvisionallyAcceptsDotlessCommitInsertions()
+        [ForegroundFact]
+        private async Task ImplicitExpressionProvisionallyAcceptsDotlessCommitInsertions()
```
rynowak: These are all private YO.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
```diff
@@ -463,106 +434,114 @@ private void ImplicitExpressionAcceptsIdentifierTypedAfterDotIfLastChangeWasProv
...
+                documentTracker,
+                templateEngineFactory,
                 new DefaultErrorReporter(),
-                new TestCompletionBroker(), 
-                new Mock<VisualStudioDocumentTrackerFactory>().Object,                
-                new Mock<IEditorOperationsFactoryService>().Object);
+                new TestCompletionBroker(),
+                Enumerable.Empty<ICanHasContextChangedListener>())
+            {
+                IdleDelay = TimeSpan.FromMilliseconds(1),
```
rynowak: What's the impact of this for testing? Just a small value so it's fast but the timer is still involved?

rynowak: (add a comment)

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
```diff
@@ -680,10 +677,10 @@ public void ApplyEditAndWaitForParse(TestEdit edit)
                 WaitForParse();
             }
 
-            public void ApplyEditAndWaitForReparse(TestEdit edit)
+            public async Task ApplyEditAndWaitForReparseAsync(TestEdit edit)
```
rynowak: Delicious

test/Microsoft.VisualStudio.Editor.Razor.Test/Infrastructure/TestTextBuffer.cs
```diff
@@ -49,12 +56,27 @@ public void ApplyEdits(params TestEdit[] edits)
...
+        {
+            add
+            {
+                _attachedChangedEvents.Add(value);
+            }
+            remove
+            {
+                _attachedChangedEvents.Remove(value);
+            }
+        }
```
rynowak: lol why

rynowak: "The Taylor" 2.0

NTaylorMullen: Unit testability. We have several classes that attach/detach to buffers based on criteria happening. It'd be awful to do an `Assert.Throws(InvokeChangedEvent)` to determine of a class has attached to the buffer.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
NTaylorMullen: Yup, somehow it's running private tests lol; totally missed it. Will make public.

PR #1727, opened 2017-10-18
RazorPageGenerator: extracted creating RazorEngine into a method [refactoring]

evil-shrike:
RazorPageGenerator: extracted creating RazorEngine into a method to simplify reusing logic in custom generators
RazorPageGenerator: added optional cli argument with base directory  (previously current dir always was used)

### DIFF
--- a/src/RazorPageGenerator/Program.cs
+++ b/src/RazorPageGenerator/Program.cs
@@ -14,17 +14,25 @@ namespace RazorPageGenerator
     {
         public static int Main(string[] args)
         {
-            if (args == null || args.Length != 1)
+            if (args == null || args.Length < 1)
             {
                 Console.WriteLine("Invalid argument(s).");
-                Console.WriteLine("Usage:   dotnet razorpagegenerator <root-namespace-of-views>");
-                Console.WriteLine("Example: dotnet razorpagegenerator Microsoft.AspNetCore.Diagnostics.RazorViews");
+                Console.WriteLine(@"Usage:   
+    dotnet razorpagegenerator <root-namespace-of-views> [path]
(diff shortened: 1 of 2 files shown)


rynowak:
@pranavkm can you take a look?

@evil-shrike - thanks for the PR. This is a tool that we use to produce some generated code that we ship as part of the diagnostics page (among others). There's no real support commitment from us, but we're happy to accept a changes like this if it makes your life better 😄

src/RazorPageGenerator/Program.cs
```diff
@@ -52,8 +60,15 @@ public static IList<RazorPageGeneratorResult> MainCore(string rootNamespace, str
...
                 builder.Features.Add(new SuppressChecksumOptionsFeature());
+                if (configure != null)
+                {
+                    configure(builder);
+                }
             });
-
+            return razorEngine;
+        }
+        public static IList<RazorPageGeneratorResult> MainCore(RazorEngine razorEngine, string targetProjectDirectory)
```
pranavkm: New line before this.

src/RazorPageGenerator/Program.cs
```diff
@@ -37,7 +45,7 @@ public static int Main(string[] args)
             return 0;
         }
 
-        public static IList<RazorPageGeneratorResult> MainCore(string rootNamespace, string targetProjectDirectory)
+        public static RazorEngine CreateRazor(string rootNamespace, Action<IRazorEngineBuilder> configure = null)
```
pranavkm: `CreateRazorEngine`

pranavkm:
Thanks for the PR!

PR #1728, opened 2017-10-18
[2.0.x] Ensure the VSIX project is skipped when executing restore and build

natemcmaster:
The syntax of the Update in repo.targets was not actually matching Razor.sln. Something either changed in MSBuild or this never worked. Either way, this ensures we don't actually execute build and restore on the VSIX csproj when building patches.

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -1,7 +1,7 @@
 <Project>
   <Import Project="VSIX.targets" />
   <ItemGroup>
-    <Solutions Update="..\Razor.sln">
+    <Solutions Update="$(RepositoryRoot)Razor.sln">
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -1,7 +1,8 @@
 <Project>
   <Import Project="VSIX.targets" />
   <ItemGroup>
-    <Solutions Update="..\Razor.sln">
+    <!-- It's important to use RepositoryRoot in the file path, not relative paths. Relative paths are relative to the build context of KoreBuild, not this file. -->
```
NTaylorMullen: No need for the comment, we shouldn't have been using a relative path here to begin with.

natemcmaster: ok.

Issue #1718, opened 2017-10-10
Harvest editor settings from the project and trigger project changes accordingly.


NTaylorMullen:
Today the editor settings are hard coded today. We need to enable the ability to read from the editor settings directly or indirectly and then trigger project change/updates to force reparses for Razor documents.

NTaylorMullen:
fb8aff12f130e39664c097fe94e358a15ae7013e

PR #1732, opened 2017-10-21
Add EditorSettings management to workspaces.

NTaylorMullen:
- Built a design where there's a singleton `EditorSettingsManager` that handles the "current" settings state in the world. When it detects that settings have changed via an update method being called it dispatches a `Changed` event.
- Exposed editor settings on the document tracker. When the editor settings change the document tracker dispatches to any listeners that its context has changed.
- Added tests to validate all the various settings management.

#1718

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
@@ -10,6 +10,8 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Microsoft.CodeAnalysis.Razor;
+using Microsoft.CodeAnalysis.Razor.Editor;
+using Microsoft.CodeAnalysis.Razor.ProjectSystem;
 using Microsoft.VisualStudio.Language.Intellisense;
 using Microsoft.VisualStudio.Text;
 using ITextBuffer = Microsoft.VisualStudio.Text.ITextBuffer;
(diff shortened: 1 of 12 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultEditorSettingsManager.cs
```diff
@@ -0,0 +1,54 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
```
rynowak: This should be a different namespace, it's not really the project system

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/EditorSettings.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    public abstract class EditorSettings : IEquatable<EditorSettings>
```
rynowak: sealed makes more sense here. We don't anticipate having drastically different editor behavior across platforms

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/EditorSettingsManager.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using Microsoft.CodeAnalysis.Host;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    public abstract class EditorSettingsManager : ILanguageService
+    {
+        public abstract event EventHandler<EditorSettingsChangedEventArgs> Changed;
+
+        public abstract EditorSettings Settings { get; }
```
rynowak: It might be nicer to call this `Current` or `CurrentSettings`

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -376,21 +377,25 @@ private void OnResultsReady(object sender, DocumentStructureChangedEventArgs arg
...
+            }
+
             public int Order { get; set; }
 
             public void Configure(RazorCodeGenerationOptionsBuilder options)
             {
-                options.IndentSize = 4;
-                options.IndentWithTabs = false;
+                options.IndentSize = _settings.IndentSize;
+                options.IndentWithTabs = _settings.IndentingWithTabs;
```
rynowak: Nice.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/EditorSettingsManager.cs
NTaylorMullen: I like `Current`

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioDocumentTracker.cs
```diff
@@ -16,6 +16,8 @@ public abstract class VisualStudioDocumentTracker
 
         internal abstract ProjectExtensibilityConfiguration Configuration { get; }
 
+        public abstract EditorSettings Settings { get; }
```
rynowak: This property should be `EditorSettings`

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
rynowak: Make these property names the same.

Issue #1733, opened 2017-10-23
Protect against NREs in DirectiveCompletionProvider


NTaylorMullen:
Today we attempt to locate the owner of a change and if its `null` we NRE. That NRE crashes VS.

Should fix this for the 15.5 release and then port back to latest.

NTaylorMullen:
Discovered the root cause of the NRE in VS. Repro:

1. Be on a slower machine so Razor reparses take a significant amount of time.
2. At the end of a .cshtml file in a .net core project type `@{`.
3. Quickly hit enter then invoke IntelliSense with ctrl + space.

Basically what happens is the Razor directive completion provider attempts to resolve a completion at a location that hasn't been parsed yet. Therefore the owner for the given completion point ends up being `null`.

NTaylorMullen:
6ac0137a9ff9f42ad4ee89df2e7891401a575f23
d0276973897e70c7d1793d0c5611c972018ccdc5

PR #1734, opened 2017-10-24
Make AtDirectiveCompletionPoint resilient to null owners.

NTaylorMullen:
- Added a test to verify the new owner == null case.

#1733

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorDirectiveCompletionProvider.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorDirectiveCompletionProvider.cs
@@ -147,12 +147,19 @@ namespace Microsoft.VisualStudio.LanguageServices.Razor.Editor
             return completionItems;
         }
 
-        private bool AtDirectiveCompletionPoint(RazorSyntaxTree syntaxTree, CompletionContext context)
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/RazorDirectiveCompletionProvider.cs
```diff
@@ -147,12 +147,19 @@ internal virtual IEnumerable<CompletionItem> GetCompletionItems(RazorSyntaxTree
             return completionItems;
         }
 
-        private bool AtDirectiveCompletionPoint(RazorSyntaxTree syntaxTree, CompletionContext context)
+        // Internal for testing
+        internal bool AtDirectiveCompletionPoint(RazorSyntaxTree syntaxTree, CompletionContext context)
```
rynowak: Can you log a bug also to fix the underlying problem here? I think this is the right fix for VS15.5, but we need to ultimately make this work.

NTaylorMullen: Interesting, are you thinking of a fix where the directive completion provider awaits the completion of a parse in order to determine completion?

PR #1736, opened 2017-10-25
React to API feedback on VisualStudioRazorParser.

NTaylorMullen:
- Changed `ReparseAsync` to be `QueueReparse`. It's now async void to not give the misconception that it blocks until a reparse has been completed.
- Removed `IContextChangedListener`. People can get the same effect of the interface by retrieving the document tracker interface via the `RazorEditorFactoryService` and then when its context changes getting the parser.
- Exposed `TryGetParser` to aid in replacing `IContextChangedListener`.
- Updated tests to not rely on `IContextChangedListener`.


src/Microsoft.VisualStudio.Editor.Razor/IContextChangedListener.cs
```diff
@@ -1,13 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-namespace Microsoft.VisualStudio.Editor.Razor
-{
-    /// <summary>
-    /// This class will cease to be useful once the Razor tooling owns TagHelper discovery
-    /// </summary>
-    public interface IContextChangedListener
```
rynowak: So did it turn out that we don't need this at? Cool beans.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -110,7 +102,7 @@ internal DefaultVisualStudioRazorParser(RazorCodeDocument codeDocument)
         // Used in unit tests to ensure we can block background idle work.
         internal ManualResetEventSlim BlockBackgroundIdleWork { get; set; }
 
-        public override async Task ReparseAsync()
+        public override async void QueueReparse()
         {
             // Can be called from any thread
```
rynowak: You can safely remove the 'await' from this method. There's no need to await that task

PR #1745, opened 2017-10-27
Pin tool and package versions to make builds more repeatable

natemcmaster:
Part of https://github.com/aspnet/Universe/issues/575 and https://github.com/aspnet/KoreBuild/issues/239

This removes floating versions of dependencies. This includes the korebuild version and versions used by PackageReference's.

Due to Razor's unique requirement for multiple versions of Roslyn packages, I've prefixed any of the 'dev' versions with VSIX_. These versions are also set here so our external dependency analysis doesn't fail in Universe. https://github.com/aspnet/Universe/blob/dev/build/dependencies.props#L50-L97

One more change: I moved a couple properties for the VS for Mac addin up to the version.props file. As much as possible, we're trying to make version.props the only file you need to manipulate to change the versions of build artifacts.


natemcmaster:
AppVeyor failed on this test. Seems unrelated to my change. Flaky tests? @rynowak 
```
Failed   ProcessUpdateAsync_DoesntBlockForegroundThread
Error Message:
 Assert.Same() Failure
Expected: Mock<Microsoft.CodeAnalysis.Razor.ProjectSystem.ProjectExtensibilityConfiguration:00000001>.Object
Actual:   (null)
Stack Trace:
   at Microsoft.CodeAnalysis.Razor.ProjectSystem.DefaultProjectSnapshotWorkerTest.<ProcessUpdateAsync_DoesntBlockForegroundThread>d__10.MoveNext() in C:\projects\razor\test\Microsoft.VisualStudio.LanguageServices.Razor.Test\ProjectSystem\DefaultProjectSnapshotWorkerTest.cs:line 0
```

natemcmaster:
:bell:

korebuild-lock.txt
```diff
@@ -0,0 +1,2 @@
+version:2.1.0-preview1-15549
```
NTaylorMullen: What's this file do? Are we going to have to re-commit it every time a new KoreBuild pieces gets built?

src/Microsoft.CodeAnalysis.Razor/Microsoft.CodeAnalysis.Razor.csproj
```diff
@@ -10,8 +10,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Common" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="$(MicrosoftCodeAnalysisCommonPackageVersion)" />
```
NTaylorMullen: These package is also used at runtime. Why are we not utilizing the lineup versions to ensure MVC + this are consistent?

build/dependencies.props
```diff
@@ -0,0 +1,50 @@
...
+    <VSIX_MicrosoftCodeAnalysisCSharpPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisCSharpPackageVersion>
+    <VSIX_MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
+    <VSIX_MicrosoftCodeAnalysisEditorFeaturesTextPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisEditorFeaturesTextPackageVersion>
+    <VSIX_MicrosoftCodeAnalysisRemoteRazorServiceHubPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisRemoteRazorServiceHubPackageVersion>
+    <VSIX_MicrosoftCodeAnalysisWorkspacesCommonPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisWorkspacesCommonPackageVersion>
+    <VSIX_MicrosoftVisualStudioLanguageServicesPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftVisualStudioLanguageServicesPackageVersion>
+    <VSIX_MicrosoftVisualStudioLanguageServicesRazorRemoteClientPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftVisualStudioLanguageServicesRazorRemoteClientPackageVersion>
+    <XunitAnalyzersPackageVersion>0.7.0</XunitAnalyzersPackageVersion>
+    <XunitPackageVersion>2.3.0</XunitPackageVersion>
```
NTaylorMullen: Why have things like xunit? Does our repo differ from others in the org?

korebuild-lock.txt
natemcmaster: This file pins the version of KoreBuild so a clean repo doesn't just download 'latest' everytime you run build.cmd. It will get updated about once a week along with the rest of the versions in build\dependencies.props. Of course, you can also update it sooner manually if you need a newer version of KoreBuild.

src/Microsoft.CodeAnalysis.Razor/Microsoft.CodeAnalysis.Razor.csproj
natemcmaster: These versions should still be consistent. The automation that sets versions in build/dependencies.props uses Internal.AspNetCore.Universe.Lineup to ensure we are consistent across repos.

build/dependencies.props
natemcmaster: Two reasons: (1) update automation - it makes our update logic much simpler. It only changes the build/dependencies.props file. and (2) build-from-source. We need the ability to override the xunit version when building repos in https://github.com/dotnet/source-build

src/Microsoft.CodeAnalysis.Remote.Razor/Microsoft.CodeAnalysis.Remote.Razor.csproj
```diff
@@ -17,11 +17,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(RoslynDevVersion)" NoWarn="KRB4002" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="$(RoslynDevVersion)" NoWarn="KRB4002" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Remote.Razor.ServiceHub" Version="$(RoslynDevVersion)" NoWarn="KRB4002" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="$(RoslynDevVersion)" NoWarn="KRB4002" />
-    <PackageReference Include="StreamJsonRpc" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(VSIX_MicrosoftCodeAnalysisCSharpPackageVersion)" />
```
ryanbrandenburg: Is removing the NoWarn intentional because we fixed the issue?

src/Microsoft.VisualStudio.Editor.Razor/Microsoft.VisualStudio.Editor.Razor.csproj
```diff
@@ -7,9 +7,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="$(RoslynDevVersion)" NoWarn="KRB4002" />
```
ryanbrandenburg: Same `NoWarn` question on all of these.

natemcmaster: Right, NoWarn="KRB4002" isn't needed anymore because we aren't using package lineup. KRB4002 was our own custom warning for packages that had a version on them even though lineups were used. That doesn't apply now.

natemcmaster:
FYI - I plan to merge this later this afternoon, so please take a look now if you want to add additional feedback.

There were some concerns discussed in-person about how this affects cross-repo changes, especially when there are breaking changes.  (See https://github.com/aspnet/Coherence-Signed/issues/680). I don't believe these concerns should prevent us from striving for reproducible builds. It will require devs to take explicit actions to upgrade downstream repos when breaking API changes are made, but this is actually something that I believe we want (as opposed to repos that suddenly can't build because dependencies changed.)

cc @mikeharder @muratg

NTaylorMullen:
Talked offline. Can go in.

Issue #1748, opened 2017-11-01
Make VisualStudioRazorParser dispatch document structure changed events on foreground thread.


NTaylorMullen:
There's a lot of ceremony that needs to go on when notifying listeners of a document structure changed event, lets own that.

rynowak:
A good general rule for IDE code is that .NET `event`s are always raised on the UI thread. Do we have any other cases like this?

NTaylorMullen:
> Do we have any other cases like this?

In practice no. However, there's a few classes were we don't assert the current thread prior to dispatching an event and we should:
- DefaultProjectSnapshotManager
- DefaultEditorSettingsManager
- DefaultVisualStudioDocumentTracker

NTaylorMullen:
a2972ebf1ce0b7bde9da4444523503330f10898d

PR #1752, opened 2017-11-02
Dispatch DocumentStructureChanged event on foreground thread.

NTaylorMullen:
- Had to add extra logic to track document structure changes so listeners could know if an event was on its way or not.
- Added and fixed some tests.

#1748

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
@@ -11,7 +11,6 @@ using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Microsoft.CodeAnalysis.Razor;
 using Microsoft.CodeAnalysis.Razor.Editor;
-using Microsoft.CodeAnalysis.Razor.ProjectSystem;
 using Microsoft.VisualStudio.Language.Intellisense;
 using Microsoft.VisualStudio.Text;
 using ITextBuffer = Microsoft.VisualStudio.Text.ITextBuffer;
@@ -27,6 +26,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
         internal TimeSpan IdleDelay = TimeSpan.FromSeconds(3);
         internal Timer _idleTimer;
         internal BackgroundParser _parser;
+        internal ChangeReference _latestChangeReference;
 
         private readonly object IdleLock = new object();
         private readonly ICompletionBroker _completionBroker;
@@ -95,6 +95,8 @@ namespace Microsoft.VisualStudio.Editor.Razor
         public override ITextSnapshot Snapshot => _snapshot;
 
         public override ITextBuffer TextBuffer => _documentTracker.TextBuffer;
+
+        public override bool HasPendingChanges => _latestChangeReference != null;
 
         // Used in unit tests to ensure we can be notified when idle starts.
         internal ManualResetEventSlim NotifyForegroundIdleStart { get; set; }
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -96,6 +96,8 @@ internal DefaultVisualStudioRazorParser(RazorCodeDocument codeDocument)
 
         public override ITextBuffer TextBuffer => _documentTracker.TextBuffer;
 
+        public override bool IsIdle => _pendingStructureChanges == 0 && _parser.IsIdle;
```
NTaylorMullen: WTE code is highly coupled to the idea that they can query their RazorCodeGenerator for its "readiness" and react in different ways based on the outcome. Prior to this change we'd notify WTE on the background thread and they'd determined readiness by when a document structure change was processed on the foreground thread. Now that we dispatch on the foreground thread, so they have less ceremony, we have to provide them a "readiness" replacement.

I would have rathered an "ensure ready" async equivalent to avoid all the IsReady ceremony but from my investigation it looks like that'd require a significant amount of WTE changes. I'm open to other ideas here if you have any.

rynowak: Can this be called `HasPendingChanges` or something similar?

Seeing code like:
```
public override bool IsIdle => _pendingStructureChanges ...
```

Sadly off my *badly-designed-contept* early warning system

rynowak: I want to understand exactly what it they do that's stateful here. I think the code changes are fine, I just want to learn more

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -322,32 +324,43 @@ private void OnStartingBackgroundIdleWork()
...
             _dispatcher.AssertBackgroundThread();
 
-            if (DocumentStructureChanged != null)
-            {
-                if (args.Snapshot != TextBuffer.CurrentSnapshot)
-                {
-                    // A different text change is being parsed.
-                    return;
-                }
+            Interlocked.Increment(ref _pendingStructureChanges);
```
rynowak: Does this happen before or after a change to `_parser.IsIdle`?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
rynowak: Can the state of this property change on a background thread? If that's possible then this isn't super useful

NTaylorMullen: > I want to understand exactly what it they do that's stateful here.

Not sure I understand the ask. One example of them being stateful based off of the result of this property is they determine ranges in the document that they call "artifacts". Those artifacts are stateful and get updated whenever document structure changes flow through the system.

> Can this be called HasPendingChanges or something similar?

Yup

NTaylorMullen: It can because it's based off of the background parser that's how it's always been. This is something we can definitely change by enforcing changes to be enqueued on the foreground thread and then tracking the "latest" change until its completion.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: after

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
rynowak: This just seems like it's a big buggy by definition if that's the case.

Running a work item on the foreground thread is a bit like taking a lock. You expect that the things you are interacting with won't change while you're *holding the lock*

NTaylorMullen:
🆙 📅

Issue #1750, opened 2017-11-01
Make VisualStudioRzorParser more resilient to Background->Foreground thread dispatching after text buffer disposal


NTaylorMullen:
Today we re-dispatch to the foreground thread in several cases; this isn't tolerant of a text buffer being disposed and all its underlying resources being disposed.

NTaylorMullen:
399dcca4f2a6b4edfc882a6837c81a3d2a41dcd6

PR #1756, opened 2017-11-07
Make Foreground dispatchable methods noop if parser is disposed.

NTaylorMullen:
- Added tests to validate each of the new noopable methods.

#1750


src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -283,13 +292,13 @@ private void OnIdle(object state)
...
-        private void ReparseOnForeground(object state)
+        // Internal for testing
+        internal void ReparseOnForeground(object state)
         {
             _dispatcher.AssertForegroundThread();
 
-            if (_parser == null)
+            if (_disposed)
             {
-                Debug.Fail("Reparse being attempted after the parser has been disposed.");
```
NTaylorMullen: Bad assumption before.

Issue #1735, opened 2017-10-24
Transition RazorTextViewConnectionListener to use ITextViewConnectionListener


NTaylorMullen:
It currently depends on the WpfTextViewConnectionListener. This is windows specific; therefore, we should go with an interface that works in VS for mac.

NTaylorMullen:
Blocked by WTE's addin being functional in VS for mac.

rynowak:
How's this blocked? Doesn't it work in VS for windows?

NTaylorMullen:
> How's this blocked? Doesn't it work in VS for windows?

Yes it works in windows; however, the original purpose of this issue was to have a text view connection listener that could work in both Windows and VS for mac. Given that the VS for mac side of the equation is blocked and will definitely influence the design of the shared connection listener i'd rather not do throw away work.

NTaylorMullen:
182714c324331ddafd0d79d6f5bb6704010fe2ce

PR #1758, opened 2017-11-08
Move away from WPF text view connection listener.

NTaylorMullen:
- Had to add a new NuGet feed to access latest VS bits.
- Updated tests to react to new ITextViewConnectionListener.

#1735

### DIFF
--- a/Razor.sln
+++ b/Razor.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26917.3000
+VisualStudioVersion = 15.0.26927.3
 MinimumVisualStudioVersion = 15.0.26730.03
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{3C0D6505-79B3-49D0-B4C3-176F0F1836ED}"
 	ProjectSection(SolutionItems) = preProject
@@ -16,9 +16,10 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{F8C12DD6-659D-405A-AA27-FB22AD92A010}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		build\dependencies.props = build\dependencies.props
 		Directory.Build.props = Directory.Build.props
 		Directory.Build.targets = Directory.Build.targets
-		NuGet.config = NuGet.config
+		build\sources.props = build\sources.props
 	EndProjectSection
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RazorPageGenerator", "src\RazorPageGenerator\RazorPageGenerator.csproj", "{7BE58880-36AD-4CD5-9E16-2A5AFEA790EF}"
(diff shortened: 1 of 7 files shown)


Razor.sln
```diff
@@ -16,6 +16,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{F8C12DD6-659D-405A-AA27-FB22AD92A010}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		build\dependencies.props = build\dependencies.props
```
ajaybhargavb: How come we didn't need this before?

NTaylorMullen: Still don't need it. It's just nice to have. This makes the dependencies.props file show up in the `Solution Items`

ajaybhargavb: Oh duh. Just realized it

build/dependencies.props
```diff
@@ -17,7 +17,7 @@
     <MicrosoftNETTestSdkPackageVersion>15.3.0</MicrosoftNETTestSdkPackageVersion>
     <MicrosoftVisualStudioComponentModelHostPackageVersion>15.0.26606</MicrosoftVisualStudioComponentModelHostPackageVersion>
     <MicrosoftVisualStudioEditorPackageVersion>15.0.26606</MicrosoftVisualStudioEditorPackageVersion>
-    <MicrosoftVisualStudioLanguageIntellisensePackageVersion>15.0.26606</MicrosoftVisualStudioLanguageIntellisensePackageVersion>
+    <MicrosoftVisualStudioLanguageIntellisensePackageVersion>15.6.161-preview</MicrosoftVisualStudioLanguageIntellisensePackageVersion>
```
rynowak: If we're updating one of these VS packages we should update all of them.

@natemcmaster - what's the value of listing all of these external dependencies with their own package versions? Can we just have a single variable for all of the things that version together (Roslyn, VS editor)

rynowak: @NTaylorMullen is this a concern for VS4Mac

src/Microsoft.VisualStudio.Editor.Razor/InternalVisualStudioDocumentTracker.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.VisualStudio.Text.Editor;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class InternalVisualStudioDocumentTracker : VisualStudioDocumentTracker
```
rynowak: Do we need this? Can we just use the **implementation**. These methods aren't an abstraction, they are part of an implementation.

NTaylorMullen:
Hold off on reviewing this. Started plugging it into VS for Mac and ran into some issues that will change the way this PR goes in.

rynowak:
>Hold off on reviewing this.

Too late bruh

rynowak:
Tell me all of your problems

build/dependencies.props
natemcmaster: Condensing variables into logical groups of packages that version together might make the dependencies.props file more user-friendly, but it makes it less bot-friendly. And at the moment, this would break our auto-update automation and the Universe build. We designed this file for simplicity: variables should be `packageId.ToPascalCase() + "PackageVersion"` and there should be one version variable per unique package. See https://github.com/dotnet/source-build/blob/dev/release/2.0/Documentation/auto-dependency-flow/api.md. 

@NTaylorMullen make sure to update this: https://github.com/aspnet/Universe/blob/dev/build/dependencies.props#L171 when you merge. Otherwise the auto-update bot will rollback to a previous version. (cc @ryanbrandenburg we should probably make sure the Update-Repos step ensures repos compile before pushing any changes.)

ryanbrandenburg: I filed aspnet/Universe#636 to track making our update script check that everything compiles.

NTaylorMullen:
This is good for review now 😄

Issue #1744, opened 2017-10-26
Add Tag Helper Project System


rynowak:
This item tracks us taking control of the Tag Helper discovery end-to-end in tooling

ajaybhargavb:
9a2e3ceab64f04f933b3a4ab2886d7293f508eb3

ajaybhargavb:
90120f6a3b713b08aac9c9abb442e17c80e36dfc
0a76ad70172f29721173b07ef2707bc558a4c367

ajaybhargavb:
8abbaa46ccc28abf30d39f841a73a69df9b9a3bc

PR #1759, opened 2017-11-09
Added TagHelper discovery to Razor project system

ajaybhargavb:
#1744 

@rynowak @NTaylorMullen 

Pending:
- ~Need to add the `OnBuildComplete` trigger for the project system.~
- React to these changes on WTE
- ~Tests~

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
@@ -2,6 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
 
 namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 {
@@ -41,6 +44,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
             ComputedVersion = other.ComputedVersion;
             Configuration = other.Configuration;
+            TagHelpers = other.TagHelpers;
         }
 
         private DefaultProjectSnapshot(ProjectSnapshotUpdateContext update, DefaultProjectSnapshot other)
@@ -59,11 +63,14 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
             ComputedVersion = update.UnderlyingProject.Version;
             Configuration = update.Configuration;
+            TagHelpers = update.TagHelpers ?? Array.Empty<TagHelperDescriptor>();
         }
 
         public override ProjectExtensibilityConfiguration Configuration { get; }
 
         public override Project UnderlyingProject { get; }
+
+        public override IReadOnlyList<TagHelperDescriptor> TagHelpers { get; } = Array.Empty<TagHelperDescriptor>();
 
         // This is the version that the computed state is based on.
         public VersionStamp? ComputedVersion { get; set; }
@@ -92,7 +99,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
             return new DefaultProjectSnapshot(update, this);
         }
 
-        public bool HasChangesComparedTo(ProjectSnapshot original)
+        public bool HasConfigurationChanged(ProjectSnapshot original)
         {
             if (original == null)
             {
@@ -100,6 +107,16 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
             }
 
             return !object.Equals(Configuration, original.Configuration);
+        }
+
+        public bool HaveTagHelpersChanged(ProjectSnapshot original)
+        {
+            if (original == null)
+            {
+                throw new ArgumentNullException(nameof(original));
+            }
+
+            return !Enumerable.SequenceEqual(TagHelpers, original.TagHelpers);
         }
     }
 }
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
@@ -149,9 +149,14 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
                 // Now we need to know if the changes that we applied are significant. If that's the case then 
                 // we need to notify listeners.
-                if (snapshot.HasChangesComparedTo(original))
+                if (snapshot.HasConfigurationChanged(original))
                 {
                     NotifyListeners(new ProjectChangeEventArgs(snapshot, ProjectChangeKind.Changed));
+                }
+
+                if (snapshot.HaveTagHelpersChanged(original))
+                {
+                    NotifyListeners(new ProjectChangeEventArgs(snapshot, ProjectChangeKind.TagHelpersChanged));
                 }
             }
         }
@@ -169,6 +174,25 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
                 // We need to notify listeners about every project removal.
                 NotifyListeners(new ProjectChangeEventArgs(snapshot, ProjectChangeKind.Removed));
+            }
+        }
+
+        public override void ProjectBuildComplete(Project underlyingProject)
+        {
+            if (underlyingProject == null)
+            {
+                throw new ArgumentNullException(nameof(underlyingProject));
+            }
+
+            if (_projects.TryGetValue(underlyingProject.Id, out var original))
+            {
+                // Doing an update to the project should keep computed values, but mark the project as dirty if the
+                // underlying project is newer.
+                var snapshot = original.WithProjectChange(underlyingProject);
+                _projects[underlyingProject.Id] = snapshot;
+
+                // Notify the background worker so it can trigger tag helper discovery.
+                NotifyBackgroundWorker(underlyingProject);
             }
         }
 
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
@@ -387,7 +387,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
         private void ConfigureTemplateEngine(IRazorEngineBuilder builder)
         {
             builder.Features.Add(new VisualStudioParserOptionsFeature(_documentTracker.EditorSettings));
-            builder.Features.Add(new VisualStudioTagHelperFeature(TextBuffer));
+            builder.Features.Add(new VisualStudioTagHelperFeature(_documentTracker.TagHelpers));
         }
 
         private class VisualStudioParserOptionsFeature : RazorEngineFeatureBase, IConfigureRazorCodeGenerationOptionsFeature
@@ -408,29 +408,20 @@ namespace Microsoft.VisualStudio.Editor.Razor
             }
         }
 
-        /// <summary>
-        /// This class will cease to be useful once we control TagHelper discovery. For now, it delegates discovery
-        /// to ITagHelperFeature's that exist on the text buffer.
-        /// </summary>
         private class VisualStudioTagHelperFeature : ITagHelperFeature
         {
-            private readonly ITextBuffer _textBuffer;
+            private readonly IReadOnlyList<TagHelperDescriptor> _tagHelpers;
 
-            public VisualStudioTagHelperFeature(ITextBuffer textBuffer)
+            public VisualStudioTagHelperFeature(IReadOnlyList<TagHelperDescriptor> tagHelpers)
             {
-                _textBuffer = textBuffer;
+                _tagHelpers = tagHelpers;
             }
 
             public RazorEngine Engine { get; set; }
 
             public IReadOnlyList<TagHelperDescriptor> GetDescriptors()
             {
-                if (_textBuffer.Properties.TryGetProperty(typeof(ITagHelperFeature), out ITagHelperFeature feature))
-                {
-                    return feature.GetDescriptors();
-                }
-
-                return Array.Empty<TagHelperDescriptor>();
+                return _tagHelpers;
             }
         }
 
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel.Composition;
+using Microsoft.CodeAnalysis.Razor.ProjectSystem;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Editor.Razor;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Export(typeof(ProjectSnapshotChangeTrigger))]
+    internal class VsSolutionUpdatesProjectSnapshotChangeTrigger : ProjectSnapshotChangeTrigger, IVsUpdateSolutionEvents2
+    {
+        private readonly IServiceProvider _services;
+        private readonly TextBufferProjectService _projectService;
+
+        private ProjectSnapshotManagerBase _projectManager;
+
+        [ImportingConstructor]
+        public VsSolutionUpdatesProjectSnapshotChangeTrigger(
+            [Import(typeof(SVsServiceProvider))] IServiceProvider services,
+            TextBufferProjectService projectService)
+        {
+            _services = services;
+            _projectService = projectService;
+        }
+
+        public override void Initialize(ProjectSnapshotManagerBase projectManager)
+        {
+            _projectManager = projectManager;
+
+            // Attach the event sink to solution update events.
+            var solutionBuildManager = _services.GetService(typeof(SVsSolutionBuildManager)) as IVsSolutionBuildManager;
+            if (solutionBuildManager != null)
+            {
+                // We expect this to be called only once. So we don't need to Unadvise.
+                var hr = solutionBuildManager.AdviseUpdateSolutionEvents(this, out var cookie);
+                Marshal.ThrowExceptionForHR(hr);
+            }
+        }
+
+        public int UpdateSolution_Begin(ref int pfCancelUpdate)
+        {
+            return VSConstants.S_OK;
+        }
+
+        public int UpdateSolution_Done(int fSucceeded, int fModified, int fCancelCommand)
+        {
+            return VSConstants.S_OK;
+        }
+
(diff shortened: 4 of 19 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -101,5 +108,20 @@ public bool HasChangesComparedTo(ProjectSnapshot original)
...
+        {
+            if (original == null)
+            {
+                throw new ArgumentNullException(nameof(original));
+            }
+
+            if (TagHelpers == null || original.TagHelpers == null)
+            {
+                return TagHelpers != original.TagHelpers;
+            }
```
rynowak: Use object.ReferenceEquals

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -388,7 +387,7 @@ internal void OnDocumentStructureChanged(object state)
         private void ConfigureTemplateEngine(IRazorEngineBuilder builder)
         {
             builder.Features.Add(new VisualStudioParserOptionsFeature(_documentTracker.EditorSettings));
-            builder.Features.Add(new VisualStudioTagHelperFeature(TextBuffer));
+            builder.Features.Add(new VisualStudioTagHelperFeature(_documentTracker.TagHelpers));
```
rynowak: ❤️

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -27,7 +28,7 @@ internal class DefaultVisualStudioDocumentTracker : VisualStudioDocumentTracker
         private ProjectSnapshot _project;
         private string _projectPath;
 
-        public override event EventHandler ContextChanged;
+        public override event EventHandler<ContextChangeEventArgs> ContextChanged;
```
rynowak: This is a breaking change. Does WTE use this?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -409,29 +408,20 @@ public void Configure(RazorCodeGenerationOptionsBuilder options)
...
 
             public IReadOnlyList<TagHelperDescriptor> GetDescriptors()
             {
-                if (_textBuffer.Properties.TryGetProperty(typeof(ITagHelperFeature), out ITagHelperFeature feature))
-                {
-                    return feature.GetDescriptors();
-                }
-
-                return Array.Empty<TagHelperDescriptor>();
+                return _tagHelpers ?? Array.Empty<TagHelperDescriptor>();
```
rynowak: Do we need a lock here?

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
NTaylorMullen: Yes. When we insert these changes we'll need to update WTE anyways.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: This seems wrong. Every time the TagHelpers are updated we don't plan on mutating that list do we? 

With this approach the feature will be stuck to 1 set of un-updateable TagHelpers.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: This is interesting. We technically need to lock the TagHelpers for 1 full parse; meaning, if we retrieve the TagHelpers in two different phases we want those TagHelpers to be identical.

Today this isn't an issue because we only ever fetch TagHelpers once; however, i'd like to solve the problem so we don't shoot ourselves in the foot later on.

src/Microsoft.VisualStudio.Editor.Razor/ContextChangeEventArgs.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public class ContextChangeEventArgs : EventArgs
```
NTaylorMullen: sealed

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -3,7 +3,6 @@
```
NTaylorMullen: Do we want the entire parser to be torn down/re-created every time tag helpers change?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
ajaybhargavb: I see your point. Right now this works because we recreate the parser for any type of `ContextChange`. But you are right, we should be able to update TagHelpers without having to create a new parser.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
ajaybhargavb: Same answer as https://github.com/aspnet/Razor/pull/1759/files#r150061201.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
ajaybhargavb: @NTaylorMullen, not sure I fully understand. What do you mean by a phase? 

> we only ever fetch TagHelpers once

How? I thought we fetched it every time a project reference change or build happens.

NTaylorMullen: Phase in the sense of:
```C#
builder.Phases.Add(new DefaultRazorParsingPhase());
builder.Phases.Add(new DefaultRazorSyntaxTreePhase());
builder.Phases.Add(new DefaultRazorTagHelperBinderPhase());
builder.Phases.Add(new DefaultRazorIntermediateNodeLoweringPhase());
builder.Phases.Add(new DefaultRazorDocumentClassifierPhase());
builder.Phases.Add(new DefaultRazorDirectiveClassifierPhase());
builder.Phases.Add(new DefaultRazorOptimizationPhase());
builder.Phases.Add(new DefaultRazorCSharpLoweringPhase());
```

Today we only call `GetTagHelpers` in the `DefaultRazorTagHelperBinderPhase`.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
ajaybhargavb: Just discussed with @NTaylorMullen. We just want to keep the existing behavior here (recreate the parser whenever there are new TH descriptors). We can revisit this if we see a perf impact in the future.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
ajaybhargavb: Spoke to @NTaylorMullen. Since we are keeping the existing behavior, there is no need for a lock here as this will be configured only once per parser.

ajaybhargavb:
🆙 📅

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            _services = services;
+            _projectService = projectService;
+        }
+
+        public override void Initialize(ProjectSnapshotManagerBase projectManager)
+        {
+            var eventSink = new SolutionUpdateEventSink(projectManager, _projectService);
+
+            // Attach the event sink to solution update events.
```
NTaylorMullen: Do we have to worry about unattaching?

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+                var projectName = _projectService.GetProjectName(pHierProj);
+                var projectPath = _projectService.GetProjectPath(pHierProj);
+
+                // Get the corresponding roslyn project by matching the project name and the project path.
+                foreach (var project in _projectManager.Workspace.CurrentSolution.Projects)
+                {
+                    if (string.Equals(projectName, project.Name, StringComparison.Ordinal) &&
+                        string.Equals(projectPath, project.FilePath, StringComparison.Ordinal))
+                    {
+                        _projectManager.ProjectBuildComplete(project);
```
NTaylorMullen: Does this fire if build failed/was canceled?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -101,5 +108,20 @@ public bool HasChangesComparedTo(ProjectSnapshot original)
 
             return !object.Equals(Configuration, original.Configuration);
         }
+
+        public bool HasTagHelpersChanged(ProjectSnapshot original)
```
NTaylorMullen: Add a test

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: Lets assume this is non-null.

src/Microsoft.VisualStudio.Editor.Razor/ContextChangeEventArgs.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public sealed class ContextChangeEventArgs : EventArgs
+    {
+        public ContextChangeEventArgs(ContextChangeKind kind)
+        {
+            Kind = kind;
+        }
+
+        public ContextChangeKind Kind { get; set; }
```
NTaylorMullen: `{ get; }`

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -172,6 +177,20 @@ public override void ProjectRemoved(Project underlyingProject)
...
+        public override void ProjectBuildComplete(Project underlyingProject)
+        {
+            if (underlyingProject == null)
+            {
+                throw new ArgumentNullException(nameof(underlyingProject));
+            }
+
+            if (_projects.ContainsKey(underlyingProject.Id))
+            {
+                // Notify the background worker so it can trigger tag helper dicovery.
```
NTaylorMullen: `discovery`

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
NTaylorMullen: Should be `HaveTagHelpersChanged`, `TagHelpers` is plural

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -101,5 +108,20 @@ public bool HasChangesComparedTo(ProjectSnapshot original)
...
+        {
+            if (original == null)
+            {
+                throw new ArgumentNullException(nameof(original));
+            }
+
+            if (TagHelpers == null || original.TagHelpers == null)
```
NTaylorMullen: `TagHelpers` wont be able to be `null`. `original.TagHelpers` on the other hand might be in a far-off land 😉

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -149,10 +149,15 @@ public override void ProjectUpdated(ProjectSnapshotUpdateContext update)
```
NTaylorMullen: Above here, we're checking if the snapshot is dirty and if so to re-notify background workers. Should we be notifying anyone? Seems like we'd be notifying not up-to-date information.

ajaybhargavb: I think you might be mixing this up with `NotifyListeners`. We need to `NotifyBackgroundWorker` if the snapshot is dirty so we can compute up to date values.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
ajaybhargavb: Don't think so. We only even attach it once. Same as Workspace_Changed events.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+            _projectService = projectService;
+        }
+
+        public override void Initialize(ProjectSnapshotManagerBase projectManager)
+        {
+            var eventSink = new SolutionUpdateEventSink(projectManager, _projectService);
+
+            // Attach the event sink to solution update events.
+            var solutionBuildManager = _services.GetService(typeof(SVsSolutionBuildManager)) as IVsSolutionBuildManager;
+            var result = solutionBuildManager.AdviseUpdateSolutionEvents(eventSink, out var cookie);
```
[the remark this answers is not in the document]
ajaybhargavb: It is used for unattaching.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            public int UpdateProjectCfg_Done(IVsHierarchy pHierProj, IVsCfg pCfgProj, IVsCfg pCfgSln, uint dwAction, int fSuccess, int fCancel)
+            {
+                var projectName = _projectService.GetProjectName(pHierProj);
+                var projectPath = _projectService.GetProjectPath(pHierProj);
+
+                // Get the corresponding roslyn project by matching the project name and the project path.
+                foreach (var project in _projectManager.Workspace.CurrentSolution.Projects)
+                {
+                    if (string.Equals(projectName, project.Name, StringComparison.Ordinal) &&
```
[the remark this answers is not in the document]
ajaybhargavb: I did this extra check because I looked at `Project` in roslyn and there is not guarantee that this will be non-null. But for name they throw an ArgNull.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
ajaybhargavb: Yes it does. Since WTE used the same event, there is no change in behavior.

ajaybhargavb:
🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -172,6 +177,25 @@ public override void ProjectRemoved(Project underlyingProject)
...
+            }
+
+            if (_projects.TryGetValue(underlyingProject.Id, out var original))
+            {
+                // Doing an update to the project should keep computed values, but mark the project as dirty if the
+                // underlying project is newer.
+                var snapshot = original.WithProjectChange(underlyingProject);
+                _projects[underlyingProject.Id] = snapshot;
+
+                // Notify the background worker so it can trigger tag helper discovery.
```
NTaylorMullen: Leave a comment here as to why we don't check is dirty to remind our future selves.

NTaylorMullen:
Looks good, get it in once @rynowak signs off.

src/Microsoft.VisualStudio.LanguageServices.Razor/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+            _services = services;
+            _projectService = projectService;
+        }
+
+        public override void Initialize(ProjectSnapshotManagerBase projectManager)
+        {
+            var eventSink = new SolutionUpdateEventSink(projectManager, _projectService);
+
+            // Attach the event sink to solution update events.
+            var solutionBuildManager = _services.GetService(typeof(SVsSolutionBuildManager)) as IVsSolutionBuildManager;
```
rynowak: null check this.

src/Microsoft.VisualStudio.LanguageServices.Razor/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+            var result = solutionBuildManager.AdviseUpdateSolutionEvents(eventSink, out var cookie);
+            Debug.Assert(result == VSConstants.S_OK);
```
rynowak: This should be stronger than a `Debug.Assert`. 

https://msdn.microsoft.com/en-us/library/0z72ax64(v=vs.110).aspx

src/Microsoft.VisualStudio.LanguageServices.Razor/VsSolutionUpdatesProjectSnapshotChangeTrigger.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        // Internal for testing.
+        internal class SolutionUpdateEventSink : IVsUpdateSolutionEvents2
```
rynowak: Consider just implementing these events on the class itself. Roslyn has a nice pattern for this. When they have to implement an API like this, they put that in a partial class in another file and then call protected/internal methods with the relevant data.

ajaybhargavb:
🆙 📅

Issue #1762, opened 2017-11-10
Cleanup legacy services in favor for CodeAnalysis.Razor alternatives


NTaylorMullen:
Today we have copies of several of our public API services in `Microsoft.VisualStudio.LanguageServices.Razor`. We should remove those and export the newer variants of those from `Microsoft.CodeAnalysis.Razor.Workspaces`.

- [x] EditorSettingsManager
- [x] RazorIndentationFactsService
- [x] RazorSyntaxFactsService
- [x] TagHelperCompletionService
- [x] TagHelperFactsService
- [x] Internalize RazorTemplateEngineFactoryService, TagHelperResolver, RazorDiagnosticJsonConverter
- [x] Delete RazorEditorParser

NTaylorMullen:
44a47182b229ef998c9e520eb5d3a4f5d8c7fae1
de23788019fc83400be7f60335eed6032c60162c
44e58740290f8622f8f83ecf7c3ff5c7728b64b6
44b4bf6bc607a3748feb6b1b8a91c42b5e539579
520876e55d3bb86d90b3bf22e4c71db07283dd58
458111ad03c8b46a8a1ebab472f3b23f845415c3
6cf78ceb1bbdfd0a8d573925f09fd540f5f146c1

PR #1763, opened 2017-11-10
Use CodeAnalysis.Razor services instead of legacy ones.

NTaylorMullen:
- Deleted Legacy/* services that were not TagHelper related (that will be deleted once the TagHelper project system is in place) and moved to using CodeAnalysis.Razor services.
- Deleted corresponding legacy tests.
- Exported public API services in `Microsoft.VisualStudio.Editor.Razor` so external components can utilize MEF to bring them in. We use Roslyn workspace services internally to pass our own services around.
- Made changes on the WTE side to consume these bits and all their tests pass.

#1762


NTaylorMullen:
The decision made here is that for each of the exported types I'm going to lift it into `Microsoft.VisualStudio.Editor.Razor`. If that type is used internally we'll split its functionality into one Roslyn service driven and the other MEF exported.

Closing this PR in favor of the several smaller ones incoming.

PR #1768, opened 2017-11-13
Add EditorSettingsUpdateSource as an exported MEF component.

NTaylorMullen:
- Internalized `EditorSettingsManager` since it's now consumed by the update source.
- Added a `VisualStudioWorkspaceAccessor` to enable the `Microsoft.VisualStudio.Editor.Razor` project to have access to the current workspace. Only added the Windows specific implementation of this type since the VS for mac implementation will require infrastructural work to make happen.

#1762


src/Microsoft.VisualStudio.Editor.Razor/VisualStudioWorkspaceAccessor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class VisualStudioWorkspaceAccessor
```
rynowak: Yeah, this is exactly what I had in mind to paper over the differences between VS and VS4Mac

src/Microsoft.VisualStudio.Editor.Razor/EditorSettingsUpdateSource.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Razor.Editor;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public abstract class EditorSettingsUpdateSource
+    {
+        public abstract void Update(EditorSettings updateSettings);
+    }
```
rynowak: I don't think this is better than just exposing the same API as EditorSettingsManager. Having a write-only interface is wierd

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var razorLanguageServices = workspaceAccessor.Workspace.Services.GetLanguageServices(RazorLanguage.Name);
+            _editorSettingsManager = razorLanguageServices.GetRequiredService<EditorSettingsManagerInternal>();
+        }
+
+        public override event EventHandler<EditorSettingsChangedEventArgs> Changed
+        {
+            add => _editorSettingsManager.Changed += value;
+            remove => _editorSettingsManager.Changed -= value;
+        }
```
rynowak: Do we need this explicit event definition? Just do the default

PR #1771, opened 2017-11-13
Cleanup legacy and unintentionally public bits.

NTaylorMullen:
- Made `RazorTemplateEngineFactoryService`, `TagHelperResolver`, `RazorDiagnosticJsonConverter` and `RazorLanguageServiceException` internal.
- Deleted all items in the Legacy folder in VisualStudio.LanguageServices.Razor.

#1762


src/Microsoft.VisualStudio.LanguageServices.Razor/RazorLanguageServiceException.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    public sealed class RazorLanguageServiceException : Exception
+    internal sealed class RazorLanguageServiceException : Exception
```
NTaylorMullen: A little odd to have this internal but I felt that it was better than removing it entirely. The extra pieces of information it exposes are valuable given a circumstance when we need to investigate an issue. Feel free to object and change my mind.

ajaybhargavb: IMO custom exceptions make sense if some external code might want to catch it. But if it is not exposed, isn't it the same as throwing an `InvalidOperationException` with our custom message and the info we need?

NTaylorMullen: Only real difference between the two is you get the specific exception type as an added detail and allows you to not specify all the extra details everywhere. I'll go ahead and change.

PR #1772, opened 2017-11-14
Update samples and tests to target netcoreapp2.1

pranavkm:


test/Microsoft.AspNetCore.Razor.Test.Common/Microsoft.AspNetCore.Razor.Test.Common.csproj
```diff
@@ -4,7 +4,7 @@
     <!-- To generate baselines, run tests with /p:GenerateBaselines=true -->
     <DefineConstants Condition="'$(GenerateBaselines)'=='true'">$(DefineConstants);GENERATE_BASELINES</DefineConstants>
     <DefineConstants>$(DefineConstants);__RemoveThisBitTo__GENERATE_BASELINES</DefineConstants>
-    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.0;net46</TargetFrameworks>
```
pranavkm: I don't think this project has tests. We could possibly leave it cross-compiling for netcoreapp2.0 + net46 since 2.1 would be compatible with 2.0

PR #1780, opened 2017-11-20
FYI: Change code document provider to work in old Razor. 

NTaylorMullen:
- Prior to this the old code document provider would utilize the editor factory service which would incorrectly create a parser for older Razor documents. This was wrong on many levels.
- Updated tests to react to new expectations.

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultTextBufferCodeDocumentProvider.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultTextBufferCodeDocumentProvider.cs
@@ -12,19 +12,6 @@ namespace Microsoft.VisualStudio.Editor.Razor
     [Export(typeof(TextBufferCodeDocumentProvider))]
     internal class DefaultTextBufferCodeDocumentProvider : TextBufferCodeDocumentProvider
     {
-        private readonly RazorEditorFactoryService _editorFactoryService;
-
-        [ImportingConstructor]
-        public DefaultTextBufferCodeDocumentProvider(RazorEditorFactoryService editorFactoryService)
-        {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultTextBufferCodeDocumentProvider.cs
```diff
@@ -12,27 +12,14 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
-        }
-
         public override bool TryGetFromBuffer(ITextBuffer textBuffer, out RazorCodeDocument codeDocument)
         {
             if (textBuffer == null)
             {
                 throw new ArgumentNullException(nameof(textBuffer));
             }
 
-            if (_editorFactoryService.TryGetParser(textBuffer, out var parser) && parser.CodeDocument != null)
```
NTaylorMullen: Was a blatant mistake before to use the editor factory service here. Basically there was a chance that this call path (through the completion provider) in a valid Razor Core scenario would create all the text buffer initialization dependencies; in that case it would have tried to do that on a background thread exploding spectacularly 😄. Ultimately this code path really doesn't have any reason to want to construct the parser anyhow; it really just wants to consume it if it`s available.

rynowak: This was the fix I was going to suggest. Looks good.

Did you check if other call sites are broken by this change in semantics?

NTaylorMullen: Yup. This class today is only used (indirectly) by the directive completion provider.

PR #1781, opened 2017-11-20
[Design] Added ViewImports tracking to TagHelper project system

ajaybhargavb:
#1744 

@rynowak @NTaylorMullen 

Unit tests are on the way


src/Microsoft.VisualStudio.Editor.Razor/ImportDocumentInfo.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public abstract class ImportDocumentInfo
+    {
+        public abstract string FilePath { get; }
+
+        public abstract RazorSourceDocument Import { get; }
```
ajaybhargavb: I added this property because for cases like in-memory default imports it didn't make sense to have an entry here with a `null` file path.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        public override void AddDocument(VisualStudioDocumentTracker tracker)
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            if (!_documents.Contains(tracker))
+            {
+                _documents.Add(tracker);
+
+                if (!tracker.TextBuffer.Properties.TryGetProperty<VisualStudioRazorParser>(typeof(VisualStudioRazorParser), out var parser))
```
ajaybhargavb: Couldn't use `RazorEditorFactoryService` to get the parser here because #circularimport

src/Microsoft.VisualStudio.LanguageServices.Razor/ViewImportChangeTracker.cs
```diff
@@ -0,0 +1,289 @@
...
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Editor;
+using Microsoft.VisualStudio.Editor.Razor;
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.TextManager.Interop;
+
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    internal class ViewImportChangeTracker : IVsFileChangeEvents, IVsRunningDocTableEvents, IDisposable
```
ajaybhargavb: This was originally a private class in `DefaultVisualStudioOpenDocumentManager` but I moved this out and made it internal for the sake of testing and clarity.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+            _editorAdaptersFactoryService = editorAdaptersFactoryService;
+
+            _documents = new List<VisualStudioDocumentTracker>();
+            _viewImportChangeTrackerCache = new Dictionary<string, ViewImportChangeTracker>(StringComparer.OrdinalIgnoreCase);
+
+            _foregroundDispatcher = workspace.Services.GetRequiredService<ForegroundDispatcher>();
+            _errorReporter = workspace.Services.GetRequiredService<ErrorReporter>();
+        }
+
+        public override IReadOnlyList<VisualStudioDocumentTracker> OpenDocuments => _documents;
```
ajaybhargavb: This is currently not being used for anything but I thought it is good info to expose.

src/Microsoft.VisualStudio.Editor.Razor/BackgroundParser.cs
```diff
@@ -343,12 +347,24 @@ private void WorkerLoop()
...
                 var sourceDocument = new TextSnapshotSourceDocument(snapshot, _filePath);
-                var imports = _templateEngine.GetImports(_filePath);
+
+                IEnumerable<RazorSourceDocument> imports = null;
+
+                try
+                {
+                    imports = _templateEngine.GetImports(_filePath);
+                }
+                catch (IOException)
```
rynowak: What if instead of this silliness we used different options when creating the stream?

http://index/?leftProject=Microsoft.CodeAnalysis.Workspaces&leftSymbol=bjlw2xzo5g0k&rightProject=Microsoft.CodeAnalysis.Workspaces.Desktop&file=Workspace%5CFileTextLoader.cs&line=171

VS

https://msdn.microsoft.com/en-us/library/system.io.file.openread(v=vs.110).aspx#Anchor_2

src/Microsoft.VisualStudio.Editor.Razor/ImportDocumentInfo.cs
rynowak: Sorry, I don't super get this comment.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
rynowak: :+1:

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                throw new ArgumentNullException(nameof(editorAdaptersFactoryService));
+            }
+
+            _fileChangeService = serviceProvider.GetService(typeof(SVsFileChangeEx)) as IVsFileChangeEx;
+            _runningDocumentTable = serviceProvider.GetService(typeof(SVsRunningDocumentTable)) as IVsRunningDocumentTable;
+            _editorAdaptersFactoryService = editorAdaptersFactoryService;
+
+            _documents = new List<VisualStudioDocumentTracker>();
+            _viewImportChangeTrackerCache = new Dictionary<string, ViewImportChangeTracker>(StringComparer.OrdinalIgnoreCase);
```
rynowak: We should not hardcode this everywhere, we should start writing all of our code assuming xplat. It's just a good habit.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+                {
+                    var importFilePath = import.FilePath;
+                    if (importFilePath == null)
+                    {
+                        // This is probably an in-memory view import. We can't track it.
+                        continue;
+                    }
+
+                    ViewImportChangeTracker viewImportChangeTracker;
+                    if (!_viewImportChangeTrackerCache.ContainsKey(importFilePath))
```
rynowak: use TryGetValue

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        public override void RemoveDocument(VisualStudioDocumentTracker tracker)
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            if (_documents.Contains(tracker))
+            {
+                _documents.Remove(tracker);
+            }
```
rynowak: You don't need this, you can just call remove directly

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+                    if (!_viewImportChangeTrackerCache.ContainsKey(importFilePath))
+                    {
+                        // First time seeing this view import. Create a change tracker for it.
+                        viewImportChangeTracker = new ViewImportChangeTracker(
+                            _fileChangeService,
+                            _runningDocumentTable,
+                            _editorAdaptersFactoryService,
+                            _foregroundDispatcher,
+                            _errorReporter,
+                            importFilePath);
```
rynowak: Consider making these 'dumb' - like just have all the data. Then you don't need to pass around all of these services.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultVisualStudioOpenDocumentManager.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        private void EnsureUndisposedTrackers()
+        {
+            foreach (var key in _viewImportChangeTrackerCache.Keys.ToList())
+            {
+                var value = _viewImportChangeTrackerCache[key];
+                if (value.IsDisposed)
+                {
+                    // This could happen if the file was deleted or renamed.
+                    _viewImportChangeTrackerCache.Remove(key);
```
rynowak: If that happens don't we get notified on another code path? If this is possible then I think we're missing something

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -193,6 +206,20 @@ private void Subscribe()
             _editorSettingsManager.Changed += EditorSettingsManager_Changed;
 
             OnContextChanged(_project, ContextChangeKind.ProjectChanged);
+
+            if (_textBuffer.Properties.TryGetProperty<VisualStudioRazorParser>(typeof(VisualStudioRazorParser), out var parser) &&
```
rynowak: This doesn't feel right.

src/Microsoft.VisualStudio.LanguageServices.Razor/Editor/DefaultVisualStudioDocumentTrackerFactory.cs
```diff
@@ -69,7 +77,7 @@ public override VisualStudioDocumentTracker Create(ITextBuffer textBuffer)
             }
 
             var filePath = textDocument.FilePath;
-            var tracker = new DefaultVisualStudioDocumentTracker(filePath, _projectManager, _projectService, _editorSettingsManager, _workspace, textBuffer);
+            var tracker = new DefaultVisualStudioDocumentTracker(filePath, _projectManager, _projectService, _editorSettingsManager, _workspace, textBuffer, _documentManager);
```
rynowak: The document tracker is becoming a god-object. https://en.wikipedia.org/wiki/God_object 

This isn't super working.

Issue #1778, opened 2017-11-16
Add an API to determine if a TagHelper exists at a location.


NTaylorMullen:
We'll have to do some extra work to resolve if the parents location associates with a TagHelper in the fact service but that will prevent callers from having to do that.

NTaylorMullen:
ba496bf1aff436fbb17e5b1badd35ba5d1131c59

PR #1786, opened 2017-11-29
Add a way for tooling to determine if a span corresponds to a TagHelper.

NTaylorMullen:
- This enables the editor to ask us if an HTML parent happens to be a TagHelper.

#1778

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public static class RazorSyntaxFactsServiceExtensions
+    {
+        public static bool IsTagHelperSpan(this RazorSyntaxFactsService syntaxFactsService, RazorSyntaxTree syntaxTree, SourceSpan span)
+        {
+            if (syntaxFactsService == null)
+            {
+                throw new ArgumentNullException(nameof(syntaxFactsService));
+            }
+
+            if (syntaxTree == null)
+            {
+                // Extra hardening for the case that tooling hasn't retrieved a SyntaxTree yet.
+                return false;
+            }
+
+            var tagHelperSpans = syntaxFactsService.GetTagHelperSpans(syntaxTree);
+            for (var i = 0; i < tagHelperSpans.Count; i++)
+            {
+                var tagHelperSpan = tagHelperSpans[i].Span;
+                if (tagHelperSpan.AbsoluteIndex == span.AbsoluteIndex &&  tagHelperSpan.Length == span.Length)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+    }
+}
--- a/test/Microsoft.VisualStudio.Editor.Razor.Test/RazorSyntaxFactsServiceExtensionsTest.cs
+++ b/test/Microsoft.VisualStudio.Editor.Razor.Test/RazorSyntaxFactsServiceExtensionsTest.cs
@@ -1,0 +1,82 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Xunit;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public class RazorSyntaxFactsServiceExtensionsTest
+    {
+        [Fact]
+        public void IsTagHelperSpan_ReturnsTrue()
+        {
+            // Arrange
+            var syntaxTree = GetSyntaxTree(
+@"<div>
+    <taghelper />
+</div>");
+            var location = new SourceSpan(9 + Environment.NewLine.Length, 13);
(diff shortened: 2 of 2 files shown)


src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public static class RazorSyntaxFactsServiceExtensions
+    {
+        public static bool IsTagHelperSpan(this RazorSyntaxFactsService syntaxFactsService, SourceSpan span, RazorSyntaxTree syntaxTree)
```
ajaybhargavb: Why an extension? Why not make it part of `RazorSyntaxFactsService`?

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            if (syntaxFactsService == null)
+            {
+                throw new System.ArgumentNullException(nameof(syntaxFactsService));
```
ajaybhargavb: using System;

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            }
+
+            if (syntaxTree == null)
+            {
+                return false;
```
ajaybhargavb: Why not throw? That is what we do in `GetTagHelperSpans` and `GetClassifiedSpans`

ajaybhargavb: Also null-check `span`.

NTaylorMullen: > Why not throw? That is what we do in GetTagHelperSpans and GetClassifiedSpans

> Why not throw? That is what we do in GetTagHelperSpans and GetClassifiedSpans

It is kind of icky to not throw for a null syntax tree but it's possible that the tooling side of the equation doesn't yet have a SyntaxTree. This is just being extra protective.



> Also null-check span.

It's a struct, can't be `null`.

ajaybhargavb: > This is just being extra protective.

Okay. I suggest leaving a comment then.

ajaybhargavb: > It's a struct, can't be null.

💯 PranavPoints will be awarded if you still manage to null check this.  😆

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
NTaylorMullen: I thought about putting it on the service directly but didn't for the reasons:
1. This method utilizes the services `GetTagHelperSpans` public implementation. Basically feels weird for a helper method to sit on a public contract.
2. It's totally different than the rest of the `RazorSyntaxFactsService` APIs.

ajaybhargavb: > This method utilizes the services GetTagHelperSpans public implementation. Basically feels weird for a helper method to sit on a public contract.

Correct me if I am wrong, but I don't think we should be concerned about this. They are both just implementation details. If someone re-implements this, they have to re-implement all the methods.

> It's totally different than the rest of the RazorSyntaxFactsService APIs.

It looks pretty similar to me but I don't have a strong opinion on this. If you think this should be an extension just for this reason, I think that's fine.

NTaylorMullen: > Correct me if I am wrong, but I don't think we should be concerned about this. They are both just implementation details. If someone re-implements this, they have to re-implement all the methods.

As an extension method that depends on the `GetTagHelperSpans` API they don't need to/can't re-implement `IsTagHelperSpan` because it acts off of the output of the `GetTagHelperSpans` method. Basically it works as long as `GetTagHelperSpans` works 😉

ajaybhargavb: Alright. That makes sense. 👍

test/Microsoft.VisualStudio.Editor.Razor.Test/RazorSyntaxFactsServiceExtensionsTest.cs
```diff
@@ -0,0 +1,67 @@
...
+            var service = new DefaultRazorSyntaxFactsService();
+
+            // Act
+            var result = service.IsTagHelperSpan(location, syntaxTree);
+
+            // Assert
+            Assert.True(result);
+        }
+        [Fact]
+        public void IsTagHelperSpan_ReturnsFalse()
```
ajaybhargavb: Add a test for null syntax tree case.

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxFactsServiceExtensions.cs
ajaybhargavb: nit: Put `syntaxTree` before `span` in the parameters.

Issue #1789, opened 2017-12-01
Implement Mac version of Microsoft.VisualStudio.LanguageServices.Razor.


NTaylorMullen:
- [x] Add `MonoDevelop.Core` and `MonoDevelop.Ide` dependencies.
- [x] Expose Razor binaries to MEF composition in Mac world.
- [x] Implement `ErrorReporter`
- [x] Implement `ForegroundDispatcher`
- [x] Implement `VisualStudioWorkspaceAccessor`
- [x] Implement `DefaultBraceSmartIndenterFactory`
- [x] Remove `Microsoft.VisualStudio.Language.IntelliSense` dependency from `Microsoft.VisualStudio.Editor.Razor`
- [x] Implement `DefaultVisualStudioRazorParserFactory`
- [x] Implement `DefaultVisualStudioDocumentTrackerFactory`
- [x] Implement `DefaultFileChangeTrackerFactory`
- [x] Implement `DefaultTextBufferProjectService`
- [x] Implement `RazorEditorFactoryService`
- [x] Implement `TagHelperResolver`

NTaylorMullen:
Going to use this issue as a meta issue to track all VS for mac implementation tasks.

NTaylorMullen:
cd1d57f97b9051e058fdd800bfdcaa3cf9f22e53

PR #1791, opened 2017-12-04
Add Mac implementation of VisualStudioWorkspaceAccessor.

NTaylorMullen:
#1789 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultVisualStudioWorkspaceAccessor.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor
+{
+    [System.Composition.Shared]
+    [Export(typeof(VisualStudioWorkspaceAccessor))]
+    internal class DefaultVisualStudioWorkspaceAccessor : VisualStudioWorkspaceAccessor
+    {
+        public DefaultVisualStudioWorkspaceAccessor()
+        {
+            Workspace = TypeSystemService.AllWorkspaces.First();
```
NTaylorMullen: Is my expectation that there will always be at least 1 workspace and that the first is the one we care about (we want registered Roslyn services)?

KirillOsenkov: Try http://source.monodevelop.com/#MonoDevelop.Ide/MonoDevelop.Ide.TypeSystem/TypeSystemService_WorkspaceHandling.cs,f1acb7fbdbce0331

NTaylorMullen:
🆙 📅

PR #1792, opened 2017-12-04
Add Mac implementation of VisualStudioForegroundDispatcher.

NTaylorMullen:
#1789 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/VisualStudioForegroundDispatcher.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [Shared]
+    [ExportWorkspaceService(typeof(ForegroundDispatcher), ServiceLayer.Host)]
+    internal class VisualStudioForegroundDispatcher : ForegroundDispatcher
+    {
+        public override TaskScheduler BackgroundScheduler { get; } = TaskScheduler.Default;
+
+        public override TaskScheduler ForegroundScheduler { get; } = TaskScheduler.FromCurrentSynchronizationContext();
+
+        public override bool IsForegroundThread => MonoDevelop.Core.Runtime.IsMainThread;
```
NTaylorMullen: Can't utilize a using here. It expects `Microsoft.Runtime` if I try to.

KirillOsenkov: We can't guarantee that the MEF composition will happen on the UI thread.

Use the MainTaskScheduler instead:
http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/Runtime.cs,d50b51ce3b3aed98

KirillOsenkov: Sorry, this one: http://source.monodevelop.com/#MonoDevelop.Core/MonoDevelop.Core/Runtime.cs,66bf11b517f553a5

NTaylorMullen: The MainTaskScheduler doesn't seem to align with the "ForegroundThread" in practice. For example, in the `ITextViewConnectionListener.SubjectBuffersConnected` event handler (expected to be invoked on the foreground thread) this is what I'm seeing:

image

KirillOsenkov: and the full call stack at that point indicates that we're indeed the UI thread? I think there can be more than one TaskSchedulers for the UI Thread. Weird.

You can commit whatever works for you, for now. I should investigate this - it might be an actual bug in MonoDevelop.

KirillOsenkov: Filed https://.../DevDiv/_workitems/edit/533116 to track this.

Issue #1795, opened 2017-12-04
Use windows VS error reporter instead of the default error reporter.


NTaylorMullen:
Right now the Windows RazorExtension is using the `DefaultErrorReporter` (does nothing on error). Instead we should be registering the VS windows specific one that logs information to the activity log.

NTaylorMullen:
Biggest part of this item by the way will be verifying that errors work as we expect them to in VisualStudio since we haven't had the luxury of trying out the API we thought we were 😉

ajaybhargavb:
6be80e7bede5f9a8e29690f29b96e577ece398b3

PR #1796, opened 2017-12-05
Add Mac implementation of ErrorReporter.

NTaylorMullen:
- Moved the existing error reporter factory temporarily into the windows specific language service assembly so it did not conflict with the Mac specific implementation. The Windows VS API will be exposed later via https://github.com/aspnet/Razor/issues/1795.

#1789 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultErrorReporterFactory.cs
```diff
@@ -4,8 +4,9 @@
 using System.Composition;
 using Microsoft.CodeAnalysis.Host;
 using Microsoft.CodeAnalysis.Host.Mef;
+using Microsoft.CodeAnalysis.Razor;
 
-namespace Microsoft.CodeAnalysis.Razor
+namespace Microsoft.VisualStudio.LanguageServices.Razor
```
NTaylorMullen: @ajaybhargavb this is temporary until you take care of #1795

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/VisualStudioErrorReporter.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.CodeAnalysis.Razor;
+using MonoDevelop.Core;
+
+namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor
+{
+    internal class VisualStudioErrorReporter : ErrorReporter
+    {
+        public override void ReportError(Exception exception)
+        {
+            if (exception == null)
```
mkArtakMSFT: Should this really be ignored ? Is this the functionality of the base class?
My point is that a method shouldn't even be called, if it doesn't make sense. So, in that case, throwing would make more sense.
Again, I understand that it may be a breaking change if implemented that way.

NTaylorMullen: In our cases yes. This is a non-critical public API that's available to any extension in VisualStudio (that depends on Razor) and throwing from here results in the IDE crashing. In the tooling side of Razor we default to being more protective in non-critical APIs so we don't tarnish the user experience.

In any case. I will definitely add a `Debug.Assert` here so in a debugging VS scenario we get proper notifications that something unexpected is happening. 👍

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/VisualStudioErrorReporter.cs
```diff
@@ -0,0 +1,40 @@
...
+        public override void ReportError(Exception exception, Project project)
+        {
+            if (exception == null)
+            {
+                Debug.Fail("Null exceptions should not be reported.");
+                return;
+            }
+
+            LoggingService.LogError(
+                Resources.FormatRazorLanguageServiceProjectError(project?.Name),
```
ajaybhargavb: Should this fallback to the general error if `project` is null?

NTaylorMullen: Nope, this will at the very least enable us to know that a project had an error from the log.

PR #1797, opened 2017-12-05
Use windows VS error reporter instead of the no-op default

ajaybhargavb:
#1795 

Related to https://github.com/aspnet/Razor/pull/1796

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/VisualStudioErrorReporterFactory.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/VisualStudioErrorReporterFactory.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Composition;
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.LanguageServices.Razor/VisualStudioErrorReporterFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.VisualStudio.LanguageServices.Razor
+{
+    [Shared]
+    [ExportWorkspaceServiceFactory(typeof(ErrorReporter), ServiceLayer.Host)]
+    internal class VisualStudioErrorReporterFactory : IWorkspaceServiceFactory
+    {
+        private readonly IServiceProvider _serviceProvider;
+
+        [ImportingConstructor]
+        public VisualStudioErrorReporterFactory(SVsServiceProvider serviceProvider)
```
NTaylorMullen: Do the `[Import(typeof(SvsServiceProvider))] IServiceProvider serviceProvider` to be explicit about our expectations. This is also consistent with what we do elsewhere.

PR #1798, opened 2017-12-06
Make BraceSmartIndenterFactory VisualStudio agnostic.

NTaylorMullen:
- Moved the `BraceSmartIndenterFactory` into the `Microsoft.VisualStudio.Editor.Razor` assembly so it can be used on Windows and Mac.
- Changed how the factory is exported since its only used internally. It now uses Roslyn services in the context of a Razor language service to be consistent with how the rest of Razor works.

#1789 

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultBraceSmartIndenterFactory.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultBraceSmartIndenterFactory.cs
@@ -2,38 +2,32 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.ComponentModel.Composition;
-using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Razor;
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultBraceSmartIndenterFactory.cs
```diff
@@ -2,38 +2,32 @@
...
     internal class DefaultBraceSmartIndenterFactory : BraceSmartIndenterFactory
     {
         private readonly IEditorOperationsFactoryService _editorOperationsFactory;
         private readonly ForegroundDispatcher _dispatcher;
 
-        [ImportingConstructor]
         public DefaultBraceSmartIndenterFactory(
             IEditorOperationsFactoryService editorOperationsFactory,
-            [Import(typeof(VisualStudioWorkspace))] Workspace workspace)
+            ForegroundDispatcher dispatcher)
```
ajaybhargavb: https://github.com/aspnet/Razor/pull/1800#discussion_r155406244 😝

Issue #1799, opened 2017-12-07
Move Internal services to use roslyn services instead of Mef


ajaybhargavb:
We want to use roslyn services for internal services.

Checklist:
- [x] RazorDocumentManager
- [x] RazorCodeDocumentProvider
- [x] RazorTextBufferProvider
- [x] TextBufferProjectService
- [x] ~~ProjectSnapshotChangeTrigger~~ ([Not supported](https://github.com/aspnet/Razor/pull/1816#discussion_r156259692))

ajaybhargavb:
f708c463d9a7603786a8e5d670dd9901e48c0a74

ajaybhargavb:
fb68a31ad5ec921935f6eca12eaa151f03f1d064

ajaybhargavb:
6b8223b5442a13c11bbc83aaabbb9bf6c53bf5c8

ajaybhargavb:
f208b27bd73c5e3944d4395b5bfd1e656bc65c0a

PR #1800, opened 2017-12-07
Export RazorDocumentManager using roslyn services

ajaybhargavb:
Fixes https://github.com/aspnet/Razor/issues/1799

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManager.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManager.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.ComponentModel.Composition;
 using System.Diagnostics;
 using Microsoft.CodeAnalysis.Razor;
 using Microsoft.VisualStudio.Text;
@@ -11,20 +10,22 @@ using Microsoft.VisualStudio.Text.Editor;
 
 namespace Microsoft.VisualStudio.Editor.Razor
 {
-    [System.Composition.Shared]
-    [Export(typeof(RazorDocumentManager))]
     internal class DefaultRazorDocumentManager : RazorDocumentManager
     {
         private readonly ForegroundDispatcher _foregroundDispatcher;
         private readonly RazorEditorFactoryService _editorFactoryService;
         private readonly TextBufferProjectService _projectService;
 
-        [ImportingConstructor]
         public DefaultRazorDocumentManager(
+            ForegroundDispatcher dispatcher,
             RazorEditorFactoryService editorFactoryService,
-            TextBufferProjectService projectService,
-            VisualStudioWorkspaceAccessor workspaceAccessor)
+            TextBufferProjectService projectService)
         {
+            if (dispatcher == null)
+            {
+                throw new ArgumentNullException(nameof(dispatcher));
+            }
+
             if (editorFactoryService == null)
             {
                 throw new ArgumentNullException(nameof(editorFactoryService));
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManager.cs
```diff
@@ -3,27 +3,23 @@
...
         private readonly ForegroundDispatcher _foregroundDispatcher;
         private readonly RazorEditorFactoryService _editorFactoryService;
         private readonly TextBufferProjectService _projectService;
 
-        [ImportingConstructor]
         public DefaultRazorDocumentManager(
             RazorEditorFactoryService editorFactoryService,
             TextBufferProjectService projectService,
-            VisualStudioWorkspaceAccessor workspaceAccessor)
+            ForegroundDispatcher dispatcher)
```
NTaylorMullen: put dispatcher first for consistency across our types. You can utilize the refactor method dialog to do this so you don't have to change all references manually.

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManagerFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+    [ExportLanguageServiceFactory(typeof(RazorDocumentManager), RazorLanguage.Name, ServiceLayer.Default)]
+    internal class DefaultRazorDocumentManagerFactory : ILanguageServiceFactory
+    {
+        private readonly RazorEditorFactoryService _editorFactoryService;
+        private readonly TextBufferProjectService _projectService;
+
+        [ImportingConstructor]
+        public DefaultRazorDocumentManagerFactory(
+            RazorEditorFactoryService editorFactoryService,
+            TextBufferProjectService projectService)
```
NTaylorMullen: This type should also come from Roslyn services. Can be a follow up.

src/Microsoft.VisualStudio.Editor.Razor/RazorDocumentManager.cs
```diff
@@ -2,12 +2,13 @@
...
 
 using System.Collections.Generic;
+using Microsoft.CodeAnalysis.Host;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
 
 namespace Microsoft.VisualStudio.Editor.Razor
 {
-    internal abstract class RazorDocumentManager
+    internal abstract class RazorDocumentManager : ILanguageService
```
NTaylorMullen: Shower Thought: A general rule of thumb with the approach we're taking in Razor is, if we have an `internal abstract class` then it should always be an `ILanguageService` 😄

src/Microsoft.VisualStudio.Editor.Razor/RazorTextViewConnectionListener.cs
```diff
@@ -20,22 +20,17 @@ internal class RazorTextViewConnectionListener : ITextViewConnectionListener
         private readonly RazorDocumentManager _documentManager;
 
         [ImportingConstructor]
-        public RazorTextViewConnectionListener(
-            VisualStudioWorkspaceAccessor workspaceAccessor,
-            RazorDocumentManager documentManager)
+        public RazorTextViewConnectionListener(VisualStudioWorkspaceAccessor workspaceAccessor)
```
NTaylorMullen: Beautiful 👍

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManagerFactory.cs
ajaybhargavb: Added it to the checklist https://github.com/aspnet/Razor/issues/1799#issue-279962189

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorDocumentManager.cs
ajaybhargavb: Done

PR #1801, opened 2017-12-07
Make VisualStudioRazorParserFactory VisualStudio agnostic.

NTaylorMullen:
- Removed the `Microsoft.VisualStudio.Language.IntelliSense` dependency from `Microsoft.VisualStudio.Editor.Razor` because it isn't supported in VS for mac.
- Replaced `ICompletionBroker` usage in the VS agnostic dll with a new abstract type `VisualStudioCompletionBroker`. This also enables us to implement completion
- Added Mac implementation of new `VisualStudioCompletionBroker`.

#1789 

FYI @mkArtakMSFT


build/dependencies.props
```diff
@@ -27,6 +27,7 @@
     <MicrosoftVisualStudioShellInterop80PackageVersion>8.0.50727</MicrosoftVisualStudioShellInterop80PackageVersion>
     <MicrosoftVisualStudioShellInterop90PackageVersion>9.0.30729</MicrosoftVisualStudioShellInterop90PackageVersion>
     <MicrosoftVisualStudioShellInteropPackageVersion>7.10.6071</MicrosoftVisualStudioShellInteropPackageVersion>
+    <MicrosoftVisualStudioTextUIPackageVersion>15.6.161-preview</MicrosoftVisualStudioTextUIPackageVersion>
```
NTaylorMullen: This already exists in Universe with this version so did not need to do a PR there.

src/Microsoft.VisualStudio.Editor.Razor/Microsoft.VisualStudio.Editor.Razor.csproj
```diff
@@ -9,7 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Features" Version="$(VSIX_MicrosoftCodeAnalysisCSharpFeaturesPackageVersion)" />
     <PackageReference Include="Microsoft.CodeAnalysis.EditorFeatures.Text" Version="$(VSIX_MicrosoftCodeAnalysisEditorFeaturesTextPackageVersion)" />
-    <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" Version="$(MicrosoftVisualStudioLanguageIntellisensePackageVersion)" />
```
NTaylorMullen: This isn't supported in VS for mac, had to downgrade to Text.UI

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.8.0" Version="$(MicrosoftVisualStudioShellInterop80PackageVersion)" />
     <PackageReference Include="Microsoft.VisualStudio.Shell.Interop.9.0" Version="$(MicrosoftVisualStudioShellInterop90PackageVersion)" />
     <PackageReference Include="Microsoft.VisualStudio.Shell.Interop" Version="$(MicrosoftVisualStudioShellInteropPackageVersion)" />
+    <PackageReference Include="Microsoft.VisualStudio.Language.Intellisense" Version="$(MicrosoftVisualStudioLanguageIntellisensePackageVersion)" />
```
NTaylorMullen: Now that VS.Editor.Razor doesn't have this dependency this needs to (it used to get this transitively).

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/Editor/DefaultVisualStudioCompletionBroker.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using Microsoft.VisualStudio.Editor.Razor;
+using Microsoft.VisualStudio.Text.Editor;
+using MonoDevelop.Ide.CodeCompletion;
+
+namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor.Editor
+{
+    internal class DefaultVisualStudioCompletionBroker : VisualStudioCompletionBroker
+    {
+        public override bool IsCompletionActive(ITextView textView)
```
NTaylorMullen: @ToddGrun @KirillOsenkov would appreciate if you guys could review since you have more monodevelop knowledge. In the WebTooling code you guys don't utilize the `textView.HasAggregateFocus` so wanted your feedback.

build/dependencies.props
rynowak: Are there other VS dependencies that should be versioned along with this? We really want to try and keep a consistent set of deps if possible.

NTaylorMullen: Yes, we only list top level dependencies in the Razor repo but Universe has the entire graph (all versioned consistently).

Issue #1804, opened 2017-12-07
Refactor ImportDocumentManager to not be so dependent on Windows


NTaylorMullen:
The import document manager has a lot of logic in it that's highly dependent on Windows but also has a lot of vital logic in it that's non-windows specific. We should build out a file change abstraction that can be re-implemented on Mac (instead of the entire import document manager).

rynowak:
@NTaylorMullen done?

PR #1805, opened 2017-12-08
Refactor `ImportDocumentManager` to not be windows specific.

NTaylorMullen:
- Added a `FileChangeTracker`, `FileChangeTrackerFactory` and corresponding implementations. These types now enable us to implement Windows and Mac file change tracking instead of replacing the entire import manager.
- Changed the import manager to be a Roslyn based service.
- Moved import manager implementation to the editor.razor assembly now since it no longer depends on windows.
- Updated import manager unit tests.
- Added import manager integration test
- Added file change tracking tests.

#1804 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Editor.Razor/DefaultImportDocumentManager.cs
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class DefaultImportDocumentManager : ImportDocumentManager
+    {
+        private readonly FileChangeTrackerFactory _fileChangeTrackerFactory;
+        private readonly ForegroundDispatcher _foregroundDispatcher;
+        private readonly ErrorReporter _errorReporter;
+        private readonly RazorTemplateEngineFactoryService _templateEngineFactoryService;
+        private readonly Dictionary<string, ImportTracker> _importTrackerCache;
```
mkArtakMSFT: Minor: I personally prefer taking dependencies on interfaces. So this would be "IDictionary<string, ImportTracker>" in this case

NTaylorMullen: The only problem with using the interface variants for private fields is that it limits you to what APIs you can use when making modifications in the future. By not using the interface here it's a bet that we'll more often than not be using the field in some manner; basically i'll have more APIs to make use of and wont have to go searching.

If I were to use the interface I'd be betting on us needing to change the type in the future which is typically a less common scenario.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultFileChangeTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+            _foregroundDispatcher = foregroundDispatcher;
+            _errorReporter = errorReporter;
+            _fileChangeService = fileChangeService;
+            _fileChangeCookie = VSConstants.VSCOOKIE_NIL;
+        }
+
+        public override string FilePath { get; }
+
+        public override void StartListening()
+        {
```
NTaylorMullen: Should assert foreground thread here or lock.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultFileChangeTracker.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                _errorReporter.ReportError(exception);
+            }
+        }
+
+        public int FilesChanged(uint fileCount, string[] filePaths, uint[] fileChangeFlags)
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            Debug.Assert(filePaths.Length == 1, "We only subscribed to 1 file change therefore we should only have 1 filepath.");
```
ajaybhargavb: Technically, it is not a guarantee that the length here will be 1. Looking at [this](http://index/?query=SVsFileChangeEx&rightProject=CloudTools(Repo)&file=Main.CAT%5Csrc%5CCommon%5CExternals%5CVSSDK.VS140%5CVisualStudioIntegration%5CTools%5CWizards%5CTemplates%5C1033%5CCS%5CEditorPane.cs&line=1844), I feel we should leave this as [before](https://github.com/aspnet/Razor/pull/1805/files#diff-d269dc5729aa0fa13d3ebb38462253efL155) (loop through all the filePaths). Although, I don't know if we want to go the extent of having a timer delay.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultFileChangeTracker.cs
ajaybhargavb: 👍.  Here and in `StopListening`

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultFileChangeTracker.cs
NTaylorMullen: Hmm, going to remove the length check then and just fire events for everything

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.Editor.Razor/DefaultImportDocumentManager.cs
```diff
@@ -0,0 +1,158 @@
...
+                var args = new ImportChangedEventArgs(importPath, kind, importTracker.AssociatedDocuments);
+                handler(this, args);
+            }
+        }
+
+        private void FileChangeTracker_Changed(object sender, FileChangeEventArgs e)
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            FireImportChanged(e.FilePath, e.Kind);
```
ajaybhargavb: Minor but you could just inline the contents of this method. It doesn't seem to be used for testing as well.

NTaylorMullen: I like to have 1 degree of separation between firing an event and listening to an event. Makes things like constructing args for the new event cleaner (don't have to pick  different name) also we've followed the pattern so far to have an `OnEventName` which is responsible for triggering a certain event. That being said, going to rename `FireImportChanged` to `OnChanged`

NTaylorMullen: Result:

```C#
private void OnChanged(ImportTracker importTracker, FileChangeKind changeKind)
{
    _foregroundDispatcher.AssertForegroundThread();

    if (Changed == null)
    {
        return;
    }

    var args = new ImportChangedEventArgs(importTracker.FilePath, changeKind, importTracker.AssociatedDocuments);
    Changed.Invoke(this, args);
}

private void FileChangeTracker_Changed(object sender, FileChangeEventArgs args)
{
    _foregroundDispatcher.AssertForegroundThread();

    if (_importTrackerCache.TryGetValue(args.FilePath, out var importTracker))
    {
        OnChanged(importTracker, args.Kind);
    }
}
```

PR #1814, opened 2017-12-12
Moved RazorTextBufferProvider from Mef to ILanguageService

ajaybhargavb:
#1799 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Editor.Razor/DefaultTextBufferProviderFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            if (bufferGraphService == null)
+            {
+                throw new ArgumentNullException(nameof(bufferGraphService));
+            }
+
+            _bufferGraphService = bufferGraphService;
+        }
+
+        public ILanguageService CreateLanguageService(HostLanguageServices languageServices)
```
NTaylorMullen: might as well null check language services here.

PR #1815, opened 2017-12-12
Add Mac implementation of FileChangeTracker.

NTaylorMullen:
- Added a LanguageServices Mac test project.
- Added a few tests for the new `DefaultFileChangeTracker`. There's currently an issue with the MonoDevelop.Core binaries we're compiling against which doesn't allow me to unit test other pieces of the class (they depend on instantiating mono develop objects).
- Added IVT from product projects to new mac testing project.

#1789 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultFileChangeTracker.cs
```diff
@@ -0,0 +1,130 @@
...
+            FileService.FileRemoved -= FileService_FileRemoved;
+        }
+
+        private void FileService_FileChanged(object sender, FileEventArgs args) => HandleFileChangeEvent(FileChangeKind.Changed, args);
+
+        private void FileService_FileCreated(object sender, FileEventArgs args) => HandleFileChangeEvent(FileChangeKind.Added, args);
+
+        private void FileService_FileRemoved(object sender, FileEventArgs args) => HandleFileChangeEvent(FileChangeKind.Removed, args);
+
+        private void HandleFileChangeEvent(FileChangeKind changeKind, FileEventArgs args)
```
NTaylorMullen: There's currently an issue with the MonoDevelop.Core binaries we're compiling against which doesn't allow me to unit test other pieces of the class (they explode on instantiation because of signing goodness). Working on getting this resolved offline.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultFileChangeTracker.cs
```diff
@@ -0,0 +1,130 @@
...
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(filePath));
+            }
+
+            if (foregroundDispatcher == null)
+            {
+                throw new ArgumentNullException(nameof(foregroundDispatcher));
+            }
+
+            FilePath = filePath;
+            _normalizedFullPath = NormalizePath(FilePath);
```
rynowak: This should `_normalizedFilePath` - let's not mix *full path* and *file path*

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultFileChangeTracker.cs
rynowak: I think that's fine for now. It's a static and it will be very hard to test. This class is also pretty trivial. The only thing here that I really care about tests for is this method.

NTaylorMullen:
This good?

test/Microsoft.VisualStudio.Mac.LanguageServices.Razor.Test/DefaultFileChangeTrackerTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [ForegroundFact]
+        public void StartListening_AlreadyListening_DoesNothing()
+        {
+            // Arrange
+            var tracker = new TestFileChangeTracker("C:/_ViewImports.cshtml", Dispatcher);
+            tracker.StartListening();
+
+            // Act
+            tracker.StartListening();
```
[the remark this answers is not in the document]
ajaybhargavb: Ah apparently I can't see. 🙈

PR #1820, opened 2017-12-12
Make DefaultTagHelperResolver use the correct GetTagHelpersAsync.

NTaylorMullen:
- Prior to this we weren't overriding the `GetTagHelpersAsync` method resulting in our TagHelper discovery being executed in-process.
- Removed legacy `ITagHelperResolver` types.

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -28,7 +25,7 @@ public DefaultTagHelperResolver(ErrorReporter errorReporter, Workspace workspace
             _workspace = workspace;
         }
 
-        public async Task<TagHelperResolutionResult> GetTagHelpersAsync(Project project)
+        public override async Task<TagHelperResolutionResult> GetTagHelpersAsync(Project project, CancellationToken cancellationToken)
```
ajaybhargavb: Shouldn't this CancellationToken be passed in to all the async calls below?

ajaybhargavb: I see the same issue here https://github.com/aspnet/Razor/blob/dev/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotWorker.cs#L51

NTaylorMullen: Going to skip out on the second part because that would require us to pass the cancellation token as part of `ProjectSnapshotUpdateContext`.

NTaylorMullen:
🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotWorker.cs
```diff
@@ -63,7 +63,7 @@ public override Task ProcessUpdateAsync(ProjectSnapshotUpdateContext update, Can
             var configuration = await _configurationFactory.GetConfigurationAsync(update.UnderlyingProject);
             update.Configuration = configuration;
 
-            var result = await _tagHelperResolver.GetTagHelpersAsync(update.UnderlyingProject);
+            var result = await _tagHelperResolver.GetTagHelpersAsync(update.UnderlyingProject, CancellationToken.None);
```
rynowak: I think this is right. If we end up needing cancellation here we'll have to do work to support it, but now all of the signatures are correct 👍

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -43,7 +40,7 @@ public DefaultTagHelperResolver(ErrorReporter errorReporter, Workspace workspace
                 // when it's disconnected (user stops the process).
                 //
                 // This will change in the future to an easier to consume API but for VS RTM this is what we have.
```
rynowak: These comments are a bit out of date. Can we can update this to make it seem less temporary? I think we'll have this complexity here forever.

NTaylorMullen: Will do

PR #1823, opened 2017-12-13
Add VS agnostic non-OOP DefaultTagHelperResolver.

NTaylorMullen:
- Added tests for the VisualStudio.Editor tag helper resolver.

#1789 

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -2,27 +2,37 @@
...
+            }
+
+            if (workspace == null)
+            {
+                throw new ArgumentNullException(nameof(workspace));
+            }
+
             _errorReporter = errorReporter;
             _workspace = workspace;
+            _inProcessResolver = new InProcessTagHelperResolver(_errorReporter);
```
NTaylorMullen: A little icky doing this but didn't want to use inheritance either. Feel free to suggest something else.

src/Microsoft.VisualStudio.Editor.Razor/DefaultTagHelperResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class DefaultTagHelperResolver : TagHelperResolver
+    {
+        private readonly ErrorReporter _errorReporter;
+
+        // Hack for testability. The view component visitor will normally just no op if we're not referencing
+        // an appropriate version of MVC.
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
rynowak: We don't need this attribute here. This is only visible in our own code anyway. This just makes working on the tests more annoying.

src/Microsoft.VisualStudio.Editor.Razor/DefaultTagHelperResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public override TagHelperResolutionResult GetTagHelpers(Compilation compilation)
+        {
+            var descriptors = new List<TagHelperDescriptor>();
+
+            var providers = new ITagHelperDescriptorProvider[]
+            {
+                new DefaultTagHelperDescriptorProvider() { DesignTime = true, },
+                new ViewComponentTagHelperDescriptorProvider() { ForceEnabled = ForceEnableViewComponentDiscovery },
```
rynowak: Does this have a `DesignTime` flag also?

src/Microsoft.VisualStudio.Editor.Razor/DefaultTagHelperResolver.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            try
+            {
+                var compilation = await project.GetCompilationAsync(cancellationToken).ConfigureAwait(false);
+                var result = GetTagHelpers(compilation);
+                return result;
+            }
+            catch (Exception exception)
+            {
+                _errorReporter.ReportError(exception, project);
```
rynowak: I don't think we need to report this exception here, it's going to bubble. Whomever is the caller should be handling this and logging it.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
rynowak: I think this is fine.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -75,32 +74,21 @@ public DefaultTagHelperResolver(ErrorReporter errorReporter, Workspace workspace
...
-                new DefaultTagHelperDescriptorProvider() { DesignTime = true, },
-                new ViewComponentTagHelperDescriptorProvider(),
-            };
-
-            var results = new List<TagHelperDescriptor>();
-            var context = TagHelperDescriptorProviderContext.Create(results);
-            context.SetCompilation(compilation);
-
-            for (var i = 0; i < providers.Length; i++)
+            if (result == null)
```
rynowak: I think we can safely remove the call to errorReporter now that are the caller for this code. The logging here was important back when WTE was calling us.

rynowak: I think we can remove the error logging here. This was important when WTE was calling this API. Now we're the only caller.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
```diff
@@ -75,32 +74,21 @@ public DefaultTagHelperResolver(ErrorReporter errorReporter, Workspace workspace
...
+                // Was unable to get tag helpers OOP, fallback to default behavior.
+                result = await _inProcessResolver.GetTagHelpersAsync(project, cancellationToken);
             }
 
-            var diagnostics = new List<RazorDiagnostic>();
-            var resolutionResult = new TagHelperResolutionResult(results, diagnostics);
+            return result;
+        }
+
+        public override TagHelperResolutionResult GetTagHelpers(Compilation compilation)
```
rynowak: We should look at this together and figure out if we still need both of these methods.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
```diff
@@ -1,14 +1,15 @@
...
 using Microsoft.CodeAnalysis.Host;
 using Microsoft.CodeAnalysis.Host.Mef;
 using Microsoft.CodeAnalysis.Razor;
 
 namespace Microsoft.VisualStudio.LanguageServices.Razor
 {
-    [ExportLanguageServiceFactory(typeof(TagHelperResolver), RazorLanguage.Name, ServiceLayer.Default)]
+    [Shared]
+    [ExportLanguageServiceFactory(typeof(TagHelperResolver), RazorLanguage.Name, ServiceLayer.Host)]
     internal class DefaultTagHelperResolverFactory : ILanguageServiceFactory
```
rynowak: This isn't the default one anymore, this is VS (Windows) specific.

test/Microsoft.VisualStudio.Editor.Razor.Test/xunit.runner.json
```diff
@@ -0,0 +1,3 @@
+{
+  "shadowCopy": false
```
rynowak: Use the same naming format here as the other projects too 👍

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: We don't

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultTagHelperResolverFactory.cs
NTaylorMullen: Isn't that portrayed by the assembly?

src/Microsoft.VisualStudio.Editor.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: No, there's no design time specific code in view component TagHelper discovery

src/Microsoft.VisualStudio.Editor.Razor/DefaultTagHelperResolver.cs
NTaylorMullen: I'm fine with not having any throw cases in the editor.razor variant. However, in the windows resolver going to remove the error reporter log but keep the try/catch => throw unexpected error since we're calling into Roslyn bits that deal with calling oop.

NTaylorMullen:
🆙 📅

src/Microsoft.CodeAnalysis.Remote.Razor/DefaultTagHelperResolver.cs
```diff
@@ -19,7 +19,7 @@ public DefaultTagHelperResolver(bool designTime)
 
         public bool DesignTime { get; }
 
-        public override TagHelperResolutionResult GetTagHelpers(Compilation compilation)
+        private TagHelperResolutionResult GetTagHelpers(Compilation compilation)
```
NTaylorMullen: We can't move the default TagHelperResolver implementation into CodeAnalysis.Workspaces because it doesn't depend on the Mvc extensions.

Issue #1827, opened 2017-12-14
Get rid of LegacyRazorDiagnostic


rynowak:
This is coming up again now that I'm dealing with `RazorDiagnostic` in the build-time-Razor work, and it also deserves to just be fixed.

We should port all of our usage of old Razor errors to the new `RazorDiagnostic` class and get rid of the legacy stuff. 

Checklist:

Razor.Language:
- [x] DefaultDirectiveSyntaxTreePass
- [x] CSharpTokenizer
- [x] CSharpCodeParser
- [x] HtmlMarkupParser
- [x] TokenizerBackedParser
- [x] TagHelperParseTreeRewriter
- [x] TagHelperBlockRewriter - Taylor
- [x] DefaultTagHelperTargetExtension - Taylor

rynowak:
Note that this is on the path for the build-time-Razor work. If it's feasible to get rid of this without too much pain we should do it now.

ajaybhargavb:
6bbd063e263ac2bf79fcfa1240cceffeb0c4ca71

PR #1831, opened 2017-12-15
Use DefaultRazorDiagnostic in DefaultDirectiveSyntaxTreePass

ajaybhargavb:
#1827 

This is just one of many usages that must be moved. Just sent this one first to make sure every one is on the same page as to what needs to be done. More PRs will follow this one.

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Razor.Language
 
         // General Errors ID Offset = 0
 
-        public static readonly RazorDiagnosticDescriptor Directive_BlockDirectiveCannotBeImported =
+        internal static readonly RazorDiagnosticDescriptor Directive_BlockDirectiveCannotBeImported =
         new RazorDiagnosticDescriptor(
             $"{DiagnosticPrefix}0000",
             () => Resources.BlockDirectiveCannotBeImported,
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
```diff
@@ -55,6 +55,16 @@ public static RazorDiagnostic CreateParsing_DuplicateDirective(string directive,
             return RazorDiagnostic.Create(Parsing_DuplicateDirective, location, directive);
         }
 
+        internal static readonly RazorDiagnosticDescriptor Parsing_SectionsCannotBeNested =
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}2002",
+                () => LegacyResources.ParseError_Sections_Cannot_Be_Nested,
```
ajaybhargavb: Should this move from `LegacyResources` to `Resources`?

NTaylorMullen: Yes

ajaybhargavb: I'll keep then in LegacyResources for now. I'll move them all together in a separate PR once we get rid of the legacy diagnostic.

rynowak: yes, these should eventually move. The goal is to nuke `LegacyResources`

PR #1832, opened 2017-12-15
Make ErrorSink and Tokenizer depend on RazorDiagnostic instead of RazorError

ajaybhargavb:
#1827 

Progress...

@mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpTokenizer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpTokenizer.cs
@@ -343,7 +343,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             return base.GetSymbolContent(type);
         }
 
-        protected override CSharpSymbol CreateSymbol(string content, CSharpSymbolType type, IReadOnlyList<RazorError> errors)
+        protected override CSharpSymbol CreateSymbol(string content, CSharpSymbolType type, IReadOnlyList<RazorDiagnostic> errors)
         {
             return new CSharpSymbol(content, type, errors);
         }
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/ErrorSink.cs
```diff
@@ -6,40 +6,41 @@
...
 
         /// <summary>
-        /// Creates and tracks a new <see cref="RazorError"/>.
+        /// Creates and tracks a new <see cref="RazorDiagnostic"/>.
         /// </summary>
         /// <param name="location"><see cref="SourceLocation"/> of the error.</param>
         /// <param name="message">A message describing the error.</param>
         /// <param name="length">The length of the error.</param>
+        /// <remarks>This is temporary. It will be removed once we get rid of <see cref="LegacyRazorDiagnostic"/>.</remarks>
         public void OnError(SourceLocation location, string message, int length)
```
ajaybhargavb: Need this for now so I can send PRs in chunks

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpTokenizer.cs
```diff
@@ -547,10 +547,11 @@ private StateResult VerbatimStringLiteral()
             else if (EndOfFile)
             {
                 CurrentErrors.Add(
-                    new RazorError(
-                        LegacyResources.ParseError_Unterminated_String_Literal,
-                        CurrentStart,
-                        length: 1 /* end of file */));
+                    RazorDiagnostic.Create(
```
NTaylorMullen: Shouldn't these be created from RazorDiagnosticFactory? Same question elsewhere

ajaybhargavb: Yes they should be. I will send a separate PR for that because it is viral and needs a lot more to change for the tests to build

NTaylorMullen: kk

PR #1833, opened 2017-12-15
Moved the test infrastructure to depend on RazorDiagnostic instead of RazorError

ajaybhargavb:
#1827 

This will let me move errors into `RazorDiagnosticFactory` in chunks.

FYI @mkArtakMSFT

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperBlockRewriterTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperBlockRewriterTest.cs
@@ -156,7 +156,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             };
 
             // Act & Assert
-            EvaluateData(descriptors, documentContent, (MarkupBlock)expectedOutput, expectedErrors: new RazorError[0]);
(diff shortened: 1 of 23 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperBlockRewriterTest.cs
```diff
@@ -3612,12 +3612,15 @@ public static TheoryData MinimizedAttributeData_CSharpBlock
...
+                    var errors = data[2] as RazorDiagnostic[];
 
                     for (var i = 0; i < errors.Length; i++)
                     {
                         var error = errors[i];
-                        error.Location = SourceLocationTracker.Advance(error.Location, "@{");
+                        var currentErrorLocation = new SourceLocation(error.Span.AbsoluteIndex, error.Span.LineIndex, error.Span.CharacterIndex);
+                        var newErrorLocation = SourceLocationTracker.Advance(currentErrorLocation, "@{");
+                        var newError = new RazorError(error.GetMessage(), newErrorLocation, error.Span.Length);
+                        errors[i] = RazorDiagnostic.Create(newError);
```
ajaybhargavb: Some super-hackery here because `RazorError` is mutable and `RazorDiagnostic` is not. I'll clean this up when I get to this.

Issue #1793, opened 2017-12-04
Don't partial parse on already returned SyntaxTrees.


NTaylorMullen:
Today we partial parse on `SyntaxTree`'s that we've sent over to consumers via `DocumentStructureChanged` in our `VisualStudioRazorParser`. We should clone the returned syntax trees and partial parse on the cloned version of them so older document structure changed results can be passed to the facts services properly.

NTaylorMullen:
463e11b73941a9b6a26ad771c50f92ab770de974

PR #1834, opened 2017-12-15
Make SyntaxTree mutations not leak.

NTaylorMullen:
- Prior to this when the partial parser would successfully parse a change we'd mutate the returned syntax tree so any data inquired about the tree would then be wrong. We now isolate mutations to copied versions of the syntax tree.
- Added copy tests to ensure that we were appropriately copying all the various syntax node types.

#1793 

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Span = Microsoft.AspNetCore.Razor.Language.Legacy.Span;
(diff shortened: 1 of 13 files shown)


src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
```diff
@@ -9,13 +11,17 @@ namespace Microsoft.VisualStudio.Editor.Razor
 {
     internal class RazorSyntaxTreePartialParser
     {
-        private readonly RazorSyntaxTree _syntaxTree;
+        // Internal for testing
+        internal Block _syntaxTreeRoot;
```
rynowak: If you're using this in tests, just make it a property

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
```diff
@@ -65,5 +71,61 @@ private PartialParseResultInternal GetPartialParseResult(SourceChange change)
...
+                    {
+                        var clonedChild = DeepCopy(block.Children[i]);
+                        blockBuilder.Children.Add(clonedChild);
+                    }
+
+                    return blockBuilder.Build();
+                case Span span:
+                    var spanBuilder = new SpanBuilder(span);
+                    return spanBuilder.Build();
+                default:
```
rynowak: This functionality should be virtual methods on these classes, not a switch statement.

src/Microsoft.VisualStudio.Editor.Razor/Resources.resx
```diff
@@ -120,4 +120,7 @@
   <data name="ArgumentCannotBeNullOrEmpty" xml:space="preserve">
     <value>Value cannot be null or an empty string.</value>
   </data>
+  <data name="UnknownSyntaxNodeType" xml:space="preserve">
+    <value>Unknown syntax node type '{0}'.</value>
+  </data>
```
rynowak: If you make these virtual methods then you literally can't hit this case.

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
```diff
@@ -9,13 +11,17 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
     internal class RazorSyntaxTreePartialParser
     {
-        private readonly RazorSyntaxTree _syntaxTree;
+        // Internal for testing
+        internal Block _syntaxTreeRoot;
+
         private Span _lastChangeOwner;
         private bool _lastResultProvisional;
 
         public RazorSyntaxTreePartialParser(RazorSyntaxTree syntaxTree)
```
mkArtakMSFT: Please validate that the syntaxTree is not null first

src/Microsoft.VisualStudio.Editor.Razor/RazorSyntaxTreePartialParser.cs
NTaylorMullen: Gonna make it abstract on SyntaxTreeNode

NTaylorMullen:
🆙 📅

PR #1839, opened 2017-12-15
Adds support for Razor compilation at build-time

rynowak:
This PR adds two new tools as well as a tasks project. None of these
projects produce a package and they ship as part of
Microsoft.AspNetCore.Razor.Design. For now this is a 'fat' package that
contains all of the dependencies, but we plan to strip them out in the
future.

The support for compilation at build-time will start as **off** by
default. The immediate goal here is to get this to flow through the
build so that we can test it as part of the inner loop effort. We will
enable this feature by default once we've done more thorough testing.

Since this is mostly a code dump, I plan to address blocking and minor
feedback only. If there are design issues that are non-critical, I will
open follow up items.

The next step will be to start adding more detailed tests.


src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+      <Link>tasks\net461\Microsoft.AspNetCore.Razor.Tasks.dll</Link>
+      <Hidden>true</Hidden>
+    </Content>
+    <Content Include="..\Microsoft.AspNetCore.Razor.Tasks\bin\$(Configuration)\netstandard2.0\Microsoft.AspNetCore.Razor.Tasks.dll">
+      <PackagePath>tasks\netstandard2.0\</PackagePath>
+      <Link>tasks\netstandard2.0\Microsoft.AspNetCore.Razor.Tasks.dll</Link>
+      <Hidden>true</Hidden>
+    </Content>
+
+    <Content Include="..\Microsoft.AspNetCore.Razor.GenerateTool\bin\$(Configuration)\netcoreapp2.0\publish\**\*.*">
```
rynowak: This uses the published output because we need all of the dependencies. It's not ideal, but I don't see it as a blocker right now

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -0,0 +1,240 @@
...
+      OutputPath="$(RazorGenerateOutputPath)" />
+
+    <ItemGroup>
+      <FileWrites Include="@(_RazorGenerated)" />
+    </ItemGroup>
+  </Target>
+
+  <!--
+    What follows here was copied and modified from the XamlPreCompile target in 
+    Microsoft.CSharp.CurrentVersion.targets
```
rynowak: I suspect that this will ultimately migrate to the CSharp targets. Xaml has its own call to CSC in a target that they own.

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        private class StaticTagHelperFeature : ITagHelperFeature
+        {
+            public RazorEngine Engine { get; set; }
+
+            public IReadOnlyList<TagHelperDescriptor> TagHelpers { get; set; }
+
+            public IReadOnlyList<TagHelperDescriptor> GetDescriptors() => TagHelpers;
+        }
```
rynowak: This file basically contains what precompilation does today, but without emitting the assembly. Most of the logic for managing the file paths will need to move into MSBuild so we know which files to update/generate in a single place but this isn't a blocker.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -47,9 +47,10 @@ internal static ProjectDirectory Project
 #endif
         }
 
-        internal Task<MSBuildResult> DotnetMSBuild(string target)
+        internal Task<MSBuildResult> DotnetMSBuild(string target, string args = null, bool debug = false)
         {
-            return MSBuildProcessManager.RunProcessAsync($"/t:{target}", Project.DirectoryPath);
+            var timeout = debug ? (TimeSpan?)Timeout.InfiniteTimeSpan : null;
+            return MSBuildProcessManager.RunProcessAsync($"/t:{target} {args}", Project.DirectoryPath, timeout);
```
rynowak: We need this or else MSBuild gets killed. I wanted to separate it from `Debugger.IsAttached` so I can use it at the command line.

test/Microsoft.AspNetCore.Razor.Test.MvcShim/Microsoft.AspNetCore.Mvc.Razor/RazorPage.cs
```diff
@@ -16,5 +19,30 @@ public override void BeginContext(int position, int length, bool isLiteral)
...
+
+        public Task<HtmlString> RenderSectionAsync(string name)
+        {
+            return null;
+        }
+
+        public Task<HtmlString> RenderSectionAsync(string name, bool required)
+        {
+            return null;
+        }
```
rynowak: These aren't targeted by Razor code, but are used in the template's layout. I need them here so I can compile against the MvcShim

test/testapps/SimpleMvc/Program.cs
```diff
@@ -5,6 +5,9 @@ public class Program
     {
         public static void Main(string[] args)
         {
+            // Just make sure we have a reference to the MvcShim
+            var t = typeof(Microsoft.AspNetCore.Mvc.IActionResult);
+            System.Console.WriteLine(t.FullName);
```
rynowak: This is a nice canary for tests. I want the build to fail creating the SimpleMvc assembly if something is wrong.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+      <PackagePath>tools\%(RecursiveDir)\%(Filename)%(Extension)</PackagePath>
+      <Link>tools\%(RecursiveDir)\%(Filename)%(Extension)</Link>
+      <Hidden>true</Hidden>
+    </Content>
+
+    <Content Include="..\Microsoft.AspNetCore.Razor.TagHelperTool\bin\$(Configuration)\netcoreapp2.0\publish\**\*.*">
+      <PackagePath>tools\%(RecursiveDir)\%(Filename)%(Extension)</PackagePath>
+      <Link>tools\%(RecursiveDir)\%(Filename)%(Extension)</Link>
+      <Hidden>true</Hidden>
+    </Content>
```
rynowak: I'm getting a warning about duplicate warning because of the overlap of these two includes, which I need to address.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+      <PackagePath>tasks\netstandard2.0\</PackagePath>
+      <Link>tasks\netstandard2.0\Microsoft.AspNetCore.Razor.Tasks.dll</Link>
+      <Hidden>true</Hidden>
+    </Content>
+
+    <Content Include="..\Microsoft.AspNetCore.Razor.GenerateTool\bin\$(Configuration)\netcoreapp2.0\publish\**\*.*">
+      <PackagePath>tools\%(RecursiveDir)\%(Filename)%(Extension)</PackagePath>
+      <Link>tools\%(RecursiveDir)\%(Filename)%(Extension)</Link>
+      <Hidden>true</Hidden>
+    </Content>
```
rynowak: This is also having an issue on clean builds, I'll dig in here.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -37,10 +37,9 @@ public static ProjectDirectory Create(string projectName)
 
                 CopyDirectory(new DirectoryInfo(projectRoot), new DirectoryInfo(destinationPath));
 
-                foreach (var project in Directory.EnumerateFiles(destinationPath, "*.csproj"))
-                {
-                    RewriteCsproj(projectRoot, project);
-                }
+                CreateDirectoryProps(projectRoot, destinationPath);
+                RewriteCsproj(projectRoot, Path.Combine(destinationPath, projectName + ".csproj"));
```
rynowak: I should be able to remove this now.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
natemcmaster: Most likely because the files don't exist when the glob is evaluated. You have to delay the glob evaluation after /t:BuildProjectReferences are built and before /t:GenerateNuspec

natemcmaster: Or, use nuspec.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Razor.Tasks\Microsoft.AspNetCore.Razor.Tasks.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Razor.GenerateTool\Microsoft.AspNetCore.Razor.GenerateTool.csproj">
+      <Targets>Publish</Targets>
+    </ProjectReference>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Razor.TagHelperTool\Microsoft.AspNetCore.Razor.TagHelperTool.csproj">
+      <Targets>Publish</Targets>
+    </ProjectReference>
+
+    <!-- We don't need to really reference anything, we just want these other projects to build first -->
```
natemcmaster: If you want to avoid TFM messiness, this can also be done by setting an explicit build order in the sln file, or with a target that invokes `<MSBuild Targets="Publish">` on dependency projects

src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+        [Required]
+        public string TagHelperManifest { get; set; }
+        
+        protected override bool SkipTaskExecution()
+        {
+            if (Assemblies.Length == 0)
+            {
+                // If for some reason there are no assemblies, we have nothing to do. Let's just
+                // skip running the tool.
+                File.WriteAllText(TagHelperManifest, "{ }");
```
NTaylorMullen: Is it even possible in practice for there to be 0 assemblies? I'd lean towards blowing up if we have 0

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/CleanProjectIntegrationTest.cs
```diff
@@ -12,9 +12,11 @@ public class CleanProjectIntegrationTest : MSBuildIntegrationTestBase
         [InitializeTestProject("SimpleMvc")]
         public async Task CleanProject_RunBuild()
         {
-            var result = await DotnetMSBuild("Restore;Build"); // Equivalent to dotnet build
+            var result = await DotnetMSBuild("Restore;Build", "/p:RazorCompileOnBuild=true");
 
             Assert.BuildPassed(result);
+            Assert.FileExists(Project, @"bin\Debug\netcoreapp2.0\SimpleMvc.dll");
```
NTaylorMullen: On the CI do we always build tests in debug? Also, is FileExists case insensitive?

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -0,0 +1,238 @@
...
+    <!-- Prefer32Bit was introduced in .NET 4.5. Set it to false if we are targeting 4.0 -->
+    <PropertyGroup Condition="('$(TargetFrameworkVersion)' == 'v4.0')">
+      <Prefer32Bit>false</Prefer32Bit>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(UseSharedCompilation)' == ''">
+      <UseSharedCompilation>true</UseSharedCompilation>
+    </PropertyGroup>
+
+    <Csc
```
natemcmaster: We should probably add `Deterministic` to this. Are there any other parameters that were left off intentionally, or was this just a copy-paste? I'm mildly concerned about the future-proofness of using this task directly from an NuGet package and keepign it aligned with parameters available from https://github.com/dotnet/roslyn/blob/version-2.3.2/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets.

Also, we should probably also consider how to enable developers who really care about determinism, even in symbol generation. To enable that, we'd need to provide a value for the `SourceLink` and `PathMap`  parameters that makes symbol contents insensitive to the folder path containing the source code. Example: https://github.com/dotnet/roslyn/blob/5c3dff5ba9c7905b8b194378acca7f0b288e366e/build/Targets/Settings.props#L124-L136

src/Microsoft.AspNetCore.Razor.GenerateTool/Application.cs
```diff
@@ -0,0 +1,85 @@
...
+            foreach (var arg in args)
+            {
+                if (!arg.StartsWith("@", StringComparison.Ordinal))
+                {
+                    expandedArgs.Add(arg);
+                }
+                else
+                {
+                    var fileName = arg.Substring(1);
+                    expandedArgs.AddRange(File.ReadLines(fileName));
```
natemcmaster: It's pretty common in rsp files to skip lines that begin with '#'....not sure if this is something that matters.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -0,0 +1,238 @@
...
+
+  <!--
+    What follows here was copied and modified from the XamlPreCompile target in 
+    Microsoft.CSharp.CurrentVersion.targets
+
+    The XamlPreCompile target must remain identical to
+    the CoreCompile target in Microsoft.CSharp.Core.targets.
+    Any updates to one must be made to the other.
+  -->
+  <Target
```
natemcmaster: How well does this play with DesignTime builds? Do we need a `DesignTimeBuild != 'true'` condition?

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
rynowak: Good to know, I assumed that calling publish explicitly would be broken because I would have to forward lots of properties to it.

This is working for me just fine right now so I'd like to leave it unless we have a strong reason to change.

build/dependencies.props
```diff
@@ -7,8 +7,19 @@
...
     <MicrosoftAspNetCoreHtmlAbstractionsPackageVersion>2.1.0-preview1-27807</MicrosoftAspNetCoreHtmlAbstractionsPackageVersion>
     <MicrosoftAspNetCoreTestingPackageVersion>2.1.0-preview1-27807</MicrosoftAspNetCoreTestingPackageVersion>
+
+    <!--
+      Attempt to keep these mostly aligned with https://github.com/dotnet/cli/blob/master/build/DependencyVersions.props
+      and with the version of MSBuild used by KoreBuild.
+      MSBuild will prefer the assemblies that ship in the .NET Core SDK and MSbuild. The dependency versions here don't matter
+      as long as the version we compile for is binary compatible with what the .NET Core SDK uses.
+    -->
+    <MicrosoftBuildFrameworkPackageVersion>15.5.0-preview-000172-1103937</MicrosoftBuildFrameworkPackageVersion>
```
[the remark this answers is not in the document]
rynowak: I copied this from build tools. @natemcmaster want to explain what I should do?

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+  -->
+  <ItemGroup>
+    <Content Include="build\**\*.props" PackagePath="build\" />
+    <Content Include="build\**\*.targets" PackagePath="build\" />
+    <Content Include="buildMultiTargeting\*.props" PackagePath="buildMultiTargeting\" />
+    <Content Include="buildMultiTargeting\*.targets" PackagePath="buildMultiTargeting\" />
+
+    <Content Include="..\Microsoft.AspNetCore.Razor.Tasks\bin\$(Configuration)\net46\Microsoft.AspNetCore.Razor.Tasks.dll">
+      <PackagePath>tasks\net46\</PackagePath>
+      <Link>tasks\net461\Microsoft.AspNetCore.Razor.Tasks.dll</Link>
```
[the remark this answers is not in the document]
rynowak: This puts it in the output directory so I can use it from tests

src/Microsoft.AspNetCore.Razor.GenerateTool/Application.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        private string GetInformationalVersion()
+        {
+            var assembly = typeof(Application).GetTypeInfo().Assembly;
+            var attribute = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+            return attribute.InformationalVersion;
+        }
+
+        private static string[] ExpandResponseFiles(string[] args)
```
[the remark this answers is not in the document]
rynowak: Yes

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        private List<SourceItem> GetRazorFiles(string projectDirectory, string[] sources)
+        {
+            var trimLength = projectDirectory.EndsWith("/") ? projectDirectory.Length - 1 : projectDirectory.Length;
+
+            var items = new List<SourceItem>(sources.Length);
+            for (var i = 0; i < sources.Length; i++)
+            {
+                var fullPath = Path.Combine(projectDirectory, sources[i]);
+                if (fullPath.StartsWith(projectDirectory, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
rynowak: Not going to fix this now. All of this code needs to move into the MSBuild

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+                }
+            }
+
+            return items;
+        }
+
+        private OutputItem[] GenerateCode(RazorTemplateEngine templateEngine, IReadOnlyList<SourceItem> sources)
+        {
+            var outputs = new OutputItem[sources.Count];
+            Parallel.For(0, outputs.Length, new ParallelOptions() { MaxDegreeOfParallelism = 4 }, i =>
```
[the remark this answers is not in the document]
rynowak: copied from precompilation, sure will remove

src/Microsoft.AspNetCore.Razor.TagHelperTool/Application.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Reflection;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.AspNetCore.Razor.TagHelperTool
+{
+    internal class Application : CommandLineApplication
+    {
+        public Application()
+        {
+            Name = "razor-taghelpertool";
```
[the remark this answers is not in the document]
rynowak: This is copied from the prototype, we can polish this later once we get beyond hello world.

src/Microsoft.AspNetCore.Razor.TagHelperTool/Microsoft.AspNetCore.Razor.TagHelperTool.csproj
```diff
@@ -0,0 +1,31 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.CommandLineUtils.Sources" Version="$(MicrosoftExtensionsCommandLineUtilsSourcesPackageVersion)" />
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonPackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Mvc.Razor.Extensions\Microsoft.AspNetCore.Mvc.Razor.Extensions.csproj" />
+    <ProjectReference Include="..\Microsoft.AspNetCore.Razor.Language\Microsoft.AspNetCore.Razor.Language.csproj" />
```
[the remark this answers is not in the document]
rynowak: Sure

src/Microsoft.AspNetCore.Razor.TagHelperTool/RunCommand.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            if (!ValidateArguments(application))
+            {
+                application.ShowHelp();
+                return 1;
+            }
+
+            return ExecuteCore(
+                outputFilePath: application.TagHelperManifest.Value(),
+                assemblies: application.Assemblies.Values.ToArray());
```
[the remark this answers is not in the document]
rynowak: Correct

src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
rynowak: It's more msbuild friendly to noop

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/CleanProjectIntegrationTest.cs
rynowak: We're launching the build, so it's whatever we want.

rynowak:
> Only reviewed like half of this. It's friday

👍 😆 :canada: 🎄 👍

src/Microsoft.AspNetCore.Razor.GenerateTool/Application.cs
rynowak: I copied this code from our other tools, I'm not super concerned about it since we generate the `.rsp`

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: This is a copy paste. We need to talk to the relevant Roslyn folks about all of this.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: Design time calls `CoreCompile` not build, so it doesn't come up.

rynowak: We will need to deal with this at some point because we will eventually have our own design time build.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
     <Description>Razor is a markup syntax for adding server-side logic to web pages. This package contains MSBuild support for Razor.</Description>
-    <TargetFrameworks>net46;netstandard2.0</TargetFrameworks>
+
+    <!-- 
+      Using netcoreapp2.0 here is a workaround for https://github.com/Microsoft/msbuild/issues/2661
+
+      We need to use a P2P reference to build some other projects that need to target netcoreapp2.0
+      and even though we don't actually want to reference the output, MSBuild will not allow it.
+    -->
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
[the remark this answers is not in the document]
rynowak: We don't have anything in lib. @pranavkm assured me that this would do the right thing.

I can't manually author the `.nuspec` because it depends on the publish output of the tools, which can include corefx assemblies, that would be really fragile.

I'm going to create an item for this and move on. My only concern right now is that this doesn't break our build system.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
rynowak: I don't want to write the `.nuspec` because it will contain files we don't own.

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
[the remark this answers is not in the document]
pranavkm: I think the limit was partly based on the wisdom of some folks and partly from experimenting. With precompilation, for projects with a large number of views, having it be unbound was capping the CPU and freezing up the publish dialog.

src/Microsoft.AspNetCore.Razor.Tasks/Microsoft.AspNetCore.Razor.Tasks.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- This needs to support desktop and CoreCLR to run in msbuild -->
+    <TargetFrameworks>net46;netstandard2.0</TargetFrameworks>
```
pranavkm: `net461` should suffice

rynowak: Are we doing that? I'm being consistent with other stuff. Everything else in Razour is net46

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -0,0 +1,238 @@
...
+      <Output
+        TaskParameter="TagHelperManifest"
+        ItemName="FileWrites"/>
+    </RazorTagHelper>
+
+  </Target>
+
+  <Target Name="_RazorResolveSourceFiles">
+    <ItemGroup>
+      <RazorCompile Include="@(Content)" Condition="'%(Content.Extension)'=='.cshtml'" />
```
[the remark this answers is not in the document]
rynowak: I'll log an item for it. I don't want to use those names, but we could be support compatibility with the old things (if we have to).

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -2,8 +2,68 @@
...
+    We're using CopyToOutputDirectory and Link here to mimic shape of what goes into the package in the output
+    directory. This is used by tests.
+  -->
+  <ItemGroup>
+    <Content Include="build\**\*.props" PackagePath="build\" />
+    <Content Include="build\**\*.targets" PackagePath="build\" />
+    <Content Include="buildMultiTargeting\*.props" PackagePath="buildMultiTargeting\" />
+    <Content Include="buildMultiTargeting\*.targets" PackagePath="buildMultiTargeting\" />
+
+    <Content Include="..\Microsoft.AspNetCore.Razor.Tasks\bin\$(Configuration)\net46\Microsoft.AspNetCore.Razor.Tasks.dll">
```
rynowak: Using the net46 reference is currently broken right now because the P2P reference doesn't build twice. I'm going to remove net46 for now since it's not on the critical path and log an issue. I'll put my head together with @natemcmaster and @pranavkm and see if we can figure this out next week.

rynowak: Logged #1843 to track this.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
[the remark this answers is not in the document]
rynowak: Logged #1843 to track this.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
rynowak: Logged #1843 to track this.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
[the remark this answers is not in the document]
rynowak: https://github.com/aspnet/Razor/issues/1844

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: https://github.com/aspnet/Razor/issues/1845

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: https://github.com/aspnet/Razor/issues/1845

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: https://github.com/aspnet/Razor/issues/1846

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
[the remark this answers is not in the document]
rynowak: https://github.com/aspnet/Razor/issues/1847

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                var source = sources[i];
+                OutputItem compilationInfo;
+                using (var fileStream = source.CreateReadStream())
```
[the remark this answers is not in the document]
rynowak: :+1:

src/Microsoft.AspNetCore.Razor.GenerateTool/RunCommand.cs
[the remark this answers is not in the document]
rynowak: I'm going to change this to use `Math.Min(1, MaxCPUs / 2)`. Hardcoding any number other than 1 will always cause issues. Some machines still have one core (antares).

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -0,0 +1,69 @@
...
+    public abstract class DotNetToolTask : ToolTask
+    {
+        public bool Debug { get; set; }
+
+        public bool DebugTool { get; set; }
+
+        [Required]
+        public string ToolAssembly { get; set; }
+
+        protected override string ToolName => "dotnet";
```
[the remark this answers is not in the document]
rynowak: Doesn't support net46.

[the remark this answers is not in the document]
rynowak: https://github.com/aspnet/Razor/issues/1848

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -79,6 +78,37 @@ void RewriteCsproj(string originalProjectRoot, string filePath)
...
+#elif RELEASE
+                var configuration = "Release";
+#else
+#error Unknown Configuration
+#endif
+                var text = $@"
+<Project>
+  <PropertyGroup>
+    <OriginalProjectRoot>{originalProjectRoot}</OriginalProjectRoot>
+    <_RazorMSBuildRoot>$(OriginalProjectRoot)\..\..\..\src\Microsoft.AspNetCore.Razor.Design\bin\{configuration}\netcoreapp2.0\</_RazorMSBuildRoot>
```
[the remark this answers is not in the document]
rynowak: What is that and how does it work?

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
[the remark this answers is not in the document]
rynowak: @natemcmaster - I checked and you're right! OK, we'll have to deal with this somehow.

build/dependencies.props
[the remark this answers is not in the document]
rynowak: Ok, I will use the one in universe

---
In reply to: [157332865](https://github.com/aspnet/Razor/pull/1839#discussion_r157332865) [](ancestors = 157332865)

rynowak:
OK, I met with @pranavkm and @natemcmaster to do some brainstorming about the `.csproj` issues that I'm having. I'm going to first break out the tasks and tools and get those in since there is no more feedback on those.

rynowak:
c0eecc87e77fd2cfd83342bf0eee4f62e66b20a3 and a3fef5eeaa0463b739ea5d0b1c61d894ba5ef548

rynowak:
This was merged FYI. See ^^^^^^^^^^^ for commits

PR #1842, opened 2017-12-16
Moved CSharpTokenizer errors out of legacy diagnostic

ajaybhargavb:
#1827 

More progress.. Will keep 'em coming

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpTokenizer.cs
```diff
@@ -547,11 +547,8 @@ private StateResult VerbatimStringLiteral()
             else if (EndOfFile)
             {
                 CurrentErrors.Add(
-                    RazorDiagnostic.Create(
-                        new RazorError(
-                            LegacyResources.ParseError_Unterminated_String_Literal,
-                            CurrentStart,
-                            length: 1 /* end of file */)));
+                    RazorDiagnosticFactory.CreateParsing_UnterminatedStringLiteral(
+                        new SourceSpan(CurrentStart, contentLength: 1 /* end of file */)));
```
rynowak: Suggestion - turn the `1` into a named constant or static factory on `SourceSpan` like `ToEndOfFile`. Then the comments aren't needed here, the code will be self documenting. This is common enough that it seems worth it.

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorTagHelperBinderPhaseTest.cs
```diff
@@ -117,11 +111,8 @@ public void Execute_CanHandleSingleLengthTagHelperPrefix()
             };
             var expectedDiagnostics = new[]
             {
-                RazorDiagnostic.Create(
-                    new RazorError(
-                        LegacyResources.ParseError_Unterminated_String_Literal,
-                        new SourceLocation(17 + Environment.NewLine.Length, 1, 17),
-                        length: 1)),
+                RazorDiagnosticFactory.CreateParsing_UnterminatedStringLiteral(
```
rynowak: **non-blocking:**

@NTaylorMullen and @ajaybhargavb do you guys like this pattern of `Assert.Equals` + expected?

Another approach would be to write an assert method for it that takes the descriptor and span.

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives_DesignTime.diagnostics.txt
```diff
@@ -2,18 +2,18 @@ TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cs
...
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(4,2): Error RZ9999: Directive 'addTagHelper' must have a value.
 TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(4,15): Error RZ9999: Invalid tag helper directive look up text ''. The correct look up text format is: "name, assemblyName".
-TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(5,15): Error RZ9999: Unterminated string literal.  Strings that start with a quotation mark (") must be terminated before the end of the line.  However, strings that start with @ and a quotation mark (@") can span multiple lines.
+TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml(5,15): Error RZ1000: Unterminated string literal.  Strings that start with a quotation mark (") must be terminated before the end of the line.  However, strings that start with @ and a quotation mark (@") can span multiple lines.
```
rynowak: :+1:

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorTagHelperBinderPhaseTest.cs
NTaylorMullen: I'm personally a fan of verifying the entire error for 1 reason, if the diagnostic takes in parameters you end up needing to substitute those into the descriptor anyways to have accurate verification.

rynowak: In that case you just verify the parameters as well.

ajaybhargavb: This works for tests where we just verify the errors directly but in a lot of other cases where we build the expected root of a big syntax tree with a bunch of expected errors and compare it with the actual root like [here](https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpBlockTest.cs#L502), it might complicate our test infrastructure more than we want.
Plus, we don't have access to the individual parameters once the `RazorDiagnostic` has been created. So I'm not sure how they can be compared.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpTokenizer.cs
ajaybhargavb: As far as I can see we only do this in 2 places in our source. I don't think it's worth special casing this. I'll leave this as is for now. I'll do it if I find more places where this is needed.

PR #1853, opened 2017-12-18
Refocus Razor runtime assemblies

rynowak:
Microsoft.AspNetCore.Razor becomes the home for TagHelper-related types
that you use in your code.

Microsoft.AspNetCore.Razor.Runtime becomes the home for types you need
to build a view engine. User code should not need this package anymore.

None of these are breaking changes due to typeforwards.


shared/Microsoft.AspNetCore.Razor.TagHelpers.Testing.Sources/CaseSensitiveBoundAttributeComparer.cs
```diff
@@ -35,11 +34,7 @@ public bool Equals(TagHelperAttribute attributeX, TagHelperAttribute attributeY)
 
         public int GetHashCode(TagHelperAttribute attribute)
         {
-            var hashCodeCombiner = HashCodeCombiner.Start();
-            hashCodeCombiner.Add(attribute.GetHashCode());
-            hashCodeCombiner.Add(attribute.Name, StringComparer.Ordinal);
-
-            return hashCodeCombiner.CombinedHash;
```
rynowak: Adding a source project to tests makes things get nasty. This isn't important so I just simplified it.

Issue #1851, opened 2017-12-18
Create a Mac implementation of our project build trigger.


NTaylorMullen:
Today the project build trigger is responsible for re-finding TagHelpers once a project has been built. This does not exist in the Mac world today.

NTaylorMullen:
4b93741610ad4df73a1fcde4106931f261d35b6c

PR #1854, opened 2017-12-19
Add Mac project build change trigger.

NTaylorMullen:
- Added a Mac specific implementation of the project build change trigger. It applies to all types of project builds so we need to do a little extra filtering to ensure that we're not operating on a non-ASP.NET Core project.
- Added tests to validate that the project build event fires correctly.

#1851

FYI @mkArtakMSFT


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/Editor/DefaultTextBufferProjectService.cs
```diff
@@ -81,7 +81,7 @@ public override string GetProjectPath(object project)
         }
 
         // VisualStudio for Mac only supports ASP.NET Core Razor.
-        public override bool IsSupportedProject(object project) => true;
+        public override bool IsSupportedProject(object project) => project is DotNetProject;
```
NTaylorMullen: Made this a little more restrictive since technically there's a lot of project types in VS for Mac and we only understand one of them 😄.

rynowak: OK, sounds good

rynowak:
Looks nice and straightforward 👍

Issue #1852, opened 2017-12-18
Move to new RazorCSharpCore content type.


NTaylorMullen:
There's a new RazorCSharpCore content type that inherits from the RazorCSharp content type. We should move all of our content type based logic/code to use it.

NTaylorMullen:
So the way I'd like to complete this issue is by changing the [RazorLanguage.ContentType](https://github.com/aspnet/Razor/blob/06c2cf31d48b59b8ab85328d088c0206a8026f87/src/Microsoft.CodeAnalysis.Razor/RazorLanguage.cs#L10) value to `"RazorCoreCSharp"`; however, that's a breaking change since `RazorLanguage.ContentType` is a `public const string` 😢.

@mkArtakMSFT & @rynowak how do you feel about making this breaking change? I can't imagine great use-cases for why others would use the `RazorLanguage.ContentType` anyways.

If we choose not to do the breaking change (just have a non-public representation of the content type) then that property will be semi-invalid for the next foundational tooling release until we release a runtime version of Razor that we feel comfortable breaking.

rynowak:
We should add a new 'core' content type and use that IMO

NTaylorMullen:
8d6b86c76a17067b8fce656cfbab03750b779b15

PR #1856, opened 2017-12-19
Add `RazorLanguage.CoreContentType`.

NTaylorMullen:
- Going to wait to merge this until the web tooling side has the core content type changes in VS for mac. That way we can continue to test against VS for mac without changing content type. /cc @ToddGrun 

#1852 

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
@@ -205,7 +205,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
                 }
 
                 var mappedPoints = projectionSnapshot.MapToSourceSnapshots(context.CompletionListSpan.Start);
-                var htmlSnapshotPoints = mappedPoints.Where(p => p.Snapshot.TextBuffer.ContentType.IsOfType(RazorLanguage.ContentType));
+                var htmlSnapshotPoints = mappedPoints.Where(p => p.Snapshot.TextBuffer.IsRazorBuffer());
 
                 if (!htmlSnapshotPoints.Any())
                 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
```diff
@@ -205,7 +205,7 @@ protected virtual bool TryGetRazorSnapshotPoint(CompletionContext context, out S
                 }
 
                 var mappedPoints = projectionSnapshot.MapToSourceSnapshots(context.CompletionListSpan.Start);
-                var htmlSnapshotPoints = mappedPoints.Where(p => p.Snapshot.TextBuffer.ContentType.IsOfType(RazorLanguage.ContentType));
```
rynowak: nit: this doesn't really need to change. We'll never really operate on non-core, but this is still true for both content types.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorDocumentManagerTest.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
 {
     public class DefaultRazorDocumentManagerTest : ForegroundDispatcherTestBase
     {
-        private IContentType RazorContentType { get; } = Mock.Of<IContentType>(c => c.IsOfType(RazorLanguage.ContentType) == true);
+        private IContentType RazorContentType { get; } = Mock.Of<IContentType>(c => c.IsOfType(RazorLanguage.CoreContentType) == true);
```
rynowak: it would be better to update these names to `RazorCoreContentType`

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
NTaylorMullen: Should just be using the `IsRazorBuffer()` here.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorDocumentManagerTest.cs
NTaylorMullen: Sure

PR #1857, opened 2017-12-19
Make RazorCodeGen track file renames \ deletes

### LINKED ISSUE #1810: Target MvcRazorPrecompile is skipped on file rename/deletions for incremental builds.
_From @dasMulli on August 24, 2017 19:40_

The `MvcRazorPrecompile` target uses all razor files as inputs for incremental compilation, but msbuild really only tracks the highest modification timestamp of the passed inputs, so any change to the closure of the files - especially renames and deletions - do not cause the target to be rebuilt.

See https://github.com/Microsoft/msbuild/issues/2470 for more details.

A possible workaround would be to write the list of all files or a hash of all the items' identities to a cache file (using msbuild 15's `<WriteLinesToFile WriteOnlyWhenDifferent="True" …/>`) that is then used as an input to the target.

cc @pranavkm @Eilon 
I think this some users complained a few months ago on a GitHub issue (and I couldn't figure it out then), now I can't find the issue/comment 😢 

_Copied from original issue: aspnet/MvcPrecompilation#194_

pranavkm:
Fixes #1810

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
@@ -91,10 +91,34 @@
     </ItemGroup>
   </Target>
 
+   <Target Name="_RazorCreateSourceHashFiles" DependsOnTargets="_RazorResolveSourceFiles">
+    <PropertyGroup>
+      <_RazorSourcesHashFile>$(IntermediateOutputPath)$(MSBuildProjectName).RazorSourceHash.cache</_RazorSourcesHashFile>
+    </PropertyGroup>
+
+    <Hash ItemsToHash="@(RazorCompile)">
+      <Output TaskParameter="HashResult" PropertyName="_RazorSourcesHash" />
+    </Hash>
+
+    <MakeDir
+      Directories="$(IntermediateOutputPath)"
+      Condition="!Exists('$(IntermediateOutputPath)')" />
+
+    <WriteLinesToFile
+      Lines="$(_RazorSourcesHash)"
+      File="$(_RazorSourcesHashFile)"
+      Overwrite="True"
+      WriteOnlyWhenDifferent="True" />
+
+    <ItemGroup>
+      <FileWrites Include="$(_RazorSourcesHashFile)" />
+    </ItemGroup>
+  </Target>
+
   <Target
     Name="RazorCoreGenerate"
-    DependsOnTargets="_RazorResolveSourceFiles;_RazorResolveTagHelpers"
-    Inputs="$(MSBuildAllProjects);@(RazorCompile);$(_RazorTagHelperOutputCache)"
+    DependsOnTargets="_RazorResolveSourceFiles;_RazorCreateSourceHashFiles;_RazorResolveTagHelpers"
+    Inputs="$(MSBuildAllProjects);$(_RazorSourcesHashFile);@(RazorCompile);$(_RazorTagHelperOutputCache)"
     Outputs="@(_RazorGenerated)">
 
     <RemoveDir
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
@@ -2,10 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Moq;
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
@@ -67,6 +69,38 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
             }
 
             File.WriteAllText(filePath, content, Encoding.UTF8);
+        }
+
+        /// <summary>
+        /// Locks all files, discovered at the time of method invocation, under the
+        /// specified <paramref name="directory" /> from reads or writes.
+        /// </summary>
+        public IDisposable LockDirectory(string directory)
+        {
+            directory = Path.Combine(Project.DirectoryPath, directory);
+            var disposables = new List<IDisposable>();
+            foreach (var file in Directory.EnumerateFiles(directory, "*", SearchOption.AllDirectories))
+            {
+                disposables.Add(LockFile(file));
+            }
+
+            var disposable = new Mock<IDisposable>();
+            disposable.Setup(d => d.Dispose())
+                .Callback(() => disposables.ForEach(d => d.Dispose()));
+
+            return disposable.Object;
+        }
+
+        public IDisposable LockFile(string path)
+        {
+            path = Path.Combine(Project.DirectoryPath, path);
+            return File.Open(path, FileMode.Open, FileAccess.Read, FileShare.None);
+        }
+
+        public FileThumbPrint GetThumbPrint(string path)
+        {
+            path = Path.Combine(Project.DirectoryPath, path);
+            return FileThumbPrint.Create(path);
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
@@ -2,18 +2,27 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using Microsoft.AspNetCore.Testing;
+using Moq;
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
     internal class ProjectDirectory : IDisposable
     {
+#if PRESERVE_WORKING_DIRECTORY
+        public bool PreserveWorkingDirectory { get; set; } = true;
+#else
+        public bool PreserveWorkingDirectory { get; set; }
+#endif
+
         public static ProjectDirectory Create(string projectName)
         {
-            var destinationPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            var destinationPath = Path.Combine(Path.GetTempPath(), "Razor", Path.GetRandomFileName());
             Directory.CreateDirectory(destinationPath);
 
             try
@@ -80,6 +89,7 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 #else
 #error Unknown Configuration
 #endif
+
                 var text = $@"
 <Project>
   <PropertyGroup>
@@ -113,7 +123,14 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 
         public void Dispose()
         {
-            CleanupDirectory(DirectoryPath);
+            if (PreserveWorkingDirectory)
+            {
+                Console.WriteLine($"Skipping deletion of working directory {DirectoryPath}");
+            }
+            else
+            {
+                CleanupDirectory(DirectoryPath);
+            }
         }
 
         private static void CleanupDirectory(string filePath)
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -8,11 +11,13 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
     public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
     {
+        private const string RazorGenerateTarget = "RazorGenerate";
+
         [Fact]
         [InitializeTestProject("SimpleMvc")]
         public async Task RazorGenerate_Success_GeneratesFilesOnDisk()
         {
-            var result = await DotnetMSBuild("RazorGenerate");
+            var result = await DotnetMSBuild(RazorGenerateTarget);
 
             Assert.BuildPassed(result);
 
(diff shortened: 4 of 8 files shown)


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -52,7 +52,7 @@
 
   <Target
     Name="_RazorResolveTagHelpers"
-    DependsOnTargets="CoreCompile;RazorResolveAssemblyReferences"
+    DependsOnTargets="Compile;RazorResolveAssemblyReferences"
```
pranavkm: So `CoreCompile` doesn't run `ResolveReferences` which causes the compilation to fail. Is running `/t:RazorCodeGen` by itself a scenario we would support?

rynowak: Why?

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -91,10 +91,34 @@
     </ItemGroup>
   </Target>
 
+   <Target Name="_RazorCreateSourceHashFiles" DependsOnTargets="_RazorResolveSourceFiles">
+    <PropertyGroup>
+      <RazorSourcesHashFile>$(IntermediateOutputPath)$(MSBuildProjectName).RazorSourceHash.cache</RazorSourcesHashFile>
```
rynowak: underscore this guy yeah?

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -91,10 +91,34 @@
...
+    </PropertyGroup>
+
+    <Hash ItemsToHash="@(RazorCompile)">
+      <Output TaskParameter="HashResult" PropertyName="RazorSourcesHash" />
```
rynowak: This is an implementation detail right?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -3,24 +3,14 @@
...
-#endif
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
     public abstract class MSBuildIntegrationTestBase
     {
-#if NET461
-#elif NETCOREAPP2_0 || NETCOREAPP2_1
         private static readonly AsyncLocal<ProjectDirectory> _project = new AsyncLocal<ProjectDirectory>();
```
pranavkm: You lied to me.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -91,10 +91,34 @@
     </ItemGroup>
   </Target>
 
+   <Target Name="_RazorCreateSourceHashFiles" DependsOnTargets="_RazorResolveSourceFiles">
+    <PropertyGroup>
+      <RazorSourcesHashFile>$(IntermediateOutputPath)$(MSBuildProjectName).RazorSourceHash.cache</RazorSourcesHashFile>
+    </PropertyGroup>
+
+    <Hash ItemsToHash="@(RazorCompile)">
```
rynowak: Is the idea that we hash the content or that we hash the file names? I expected that we were supposed to hash the file names, and used a separate up-to-date check for the contents?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -29,28 +19,19 @@ protected MSBuildIntegrationTestBase()
...
 
-        internal Task<MSBuildResult> DotnetMSBuild(string target, string args = null, bool debug = false)
+        internal Task<MSBuildResult> DotnetMSBuild(string target, string args = null, bool restore = true, bool debug = false)
         {
             var timeout = debug ? (TimeSpan?)Timeout.InfiniteTimeSpan : null;
-            return MSBuildProcessManager.RunProcessAsync(Project, $"/t:{target} {args}", timeout);
+            var msbuildArgs = $"/t:{target} {args}";
+            if (restore)
+            {
+                msbuildArgs = "/restore " + msbuildArgs;
```
pranavkm: I think they recommend doing this now. I can't find the email thread, but there were some scenarios where they advise against calling the target and instead using this switch. We could make the parameter `false` by default.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -2,18 +2,26 @@
...
 using System.Linq;
 using System.Threading;
 using Microsoft.AspNetCore.Testing;
+using Moq;
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
     internal class ProjectDirectory : IDisposable
     {
+#if PRESERVE_WORKING_DIR
```
rynowak: what is this for?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -2,18 +2,26 @@
...
+        public bool PreserveWorkingDir { get; set; } = true;
+#else
+        public bool PreserveWorkingDir { get; set; }
```
rynowak: no abbreviations

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -2,18 +2,26 @@
...
+#endif
+
         public static ProjectDirectory Create(string projectName)
         {
-            var destinationPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            var destinationPath = Path.Combine(Path.GetTempPath(), "RazorDesign", Path.GetRandomFileName());
```
rynowak: why not just `Razor`

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -108,9 +116,45 @@ private ProjectDirectory(string directoryPath)
 
         public string DirectoryPath { get; }
 
+        public string GetFullPath(params string[] pathSegments)
```
rynowak: please no

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -108,9 +116,45 @@ private ProjectDirectory(string directoryPath)
...
+            {
+                disposables.Add(LockForReadWrite(file));
+            }
+
+            var disposable = new Mock<IDisposable>();
+            disposable.Setup(d => d.Dispose())
+                .Callback(() => disposables.ForEach(d => d.Dispose()));
+
+            return disposable.Object;
+        }
```
rynowak: No, we shouldn't have/need this. Each test should operate on its own copy and be independeny

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
rynowak: No, the crosstargetting got removed.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: I don't understand this comment. We need `ResolveReferences` to run to get taghelpers.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
rynowak: no, I think this is good cleanup

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
pranavkm: Here's what the target call stack looks like:
```
RazorCodeGen 
  _RazorResolveTagHelpers
    CoreCompile
      Csc -> Errors since references haven't been resolved
      
RazorCodeGen 
  _RazorResolveTagHelpers
    Compile
        ResolveReferences
        CoreCompile
          Csc -> Works
```

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
pranavkm: I would think so too. We might need to review the embedded files thing too - https://github.com/aspnet/FileSystem/blob/dev/src/FS.Embedded/build/netstandard2.0/Microsoft.Extensions.FileProviders.Embedded.targets#L27. The pattern \ casing has been copied from the sdk - https://github.com/dotnet/sdk/blob/2995c8495082d1dd4bcad35c37007375edf45c6f/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.GenerateAssemblyInfo.targets#L99

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
pranavkm: Ahh, I guess this is broken for file edits (I'll add a test). We'll need this in addition to `@(RazorCompile)`. This file would track renames \ deletes while `@(RazorCompile)` would track edits and additions.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
pranavkm: Yeah, I'll use underscore.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
pranavkm: Sure

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
pranavkm: Right. This is meant to ensure incremental builds (see https://github.com/aspnet/Razor/pull/1857/files#diff-7ceaebc6353fabd862893aa68747108bR25)

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: Cool! Yeah this was just a bug.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
pranavkm: It's so 🐩 though

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
rynowak: Yeah I gets it now. This should be cleaned up, but the idea is good.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -78,12 +88,22 @@ void CreateDirectoryProps(string originalProjectRoot, string projectRoot)
...
+                    return $@"    <Reference Include=""{name}"" HintPath=""{path}"" />";
+                }));
                 var text = $@"
 <Project>
   <PropertyGroup>
     <OriginalProjectRoot>{originalProjectRoot}</OriginalProjectRoot>
     <_RazorMSBuildRoot>$(OriginalProjectRoot)\..\..\..\src\Microsoft.AspNetCore.Razor.Design\bin\{configuration}\netstandard2.0\</_RazorMSBuildRoot>
   </PropertyGroup>
+  <ItemGroup>
+{referencesItems}
```
pranavkm: @rynowak  - wip

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
rynowak: BTW we might end up using `RazorCompile` for the items that get passed to CSC (the generated `.cs` files). That's more consistent with the terms in the SDK

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -2,55 +2,115 @@
...
+        public IDisposable LockFile(string path)
         {
-            var timeout = debug ? (TimeSpan?)Timeout.InfiniteTimeSpan : null;
-            return MSBuildProcessManager.RunProcessAsync(Project, $"/t:{target} {args}", timeout);
+            return File.Open(path, FileMode.Open, FileAccess.Read, FileShare.None);
+        }
+
+        public FileThumbPrint GetFileThumbPrint(string path) => FileThumbPrint.Create(path);
+
+        public class FileThumbPrint : IEquatable<FileThumbPrint>
```
rynowak: Might as well make this a non-nested class if it's going to be public

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -0,0 +1,153 @@
...
+            Assert.FileExists(result, RazorIntermediateOutputPath, "Views", "Shared", "_ValidationScriptsPartial.cs");
+            Assert.FileExists(result, RazorIntermediateOutputPath, "Views", "Shared", "Error.cs");
+            Assert.FileCountEquals(result, 8, RazorIntermediateOutputPath, "*.cs");
+        }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task RazorGenerate_BuildsIncrementally()
+        {
+            // Act - 1
```
rynowak: FYI - I'm skipping the comments like this on things that aren't unit tests. If you think they are useful then go ahead.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
pranavkm: Btw, I get a msbuild unification warning about System.Diagnostics.DiagnosticSource. I didn't get a chance to dig in to it last night, but we do some special casing around the library in all other places where we use the binary that I might be missing

pranavkm:
🆙 📅

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -24,9 +26,9 @@ protected MSBuildIntegrationTestBase()
 #error Configuration not supported
 #endif
 
-        protected string IntermediateOutputPath => Path.Combine("obj", Configuration, TargetFramework);
+        protected string IntermediateOutputPath => Path.Combine(Project.DirectoryPath, "obj", Configuration, TargetFramework);
```
rynowak: These aren't really intended to be full paths, or at least it isn't important that they are.

pranavkm: It just ended up being repetitive. I can make the calling code specify it if you super hate this.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -46,5 +48,29 @@ internal Task<MSBuildResult> DotnetMSBuild(string target, string args = null, bo
...
         }
+
+        /// <summary>
+        /// Locks all files, discovered at the time of method invocation, under the
+        /// specified <paramref name="directory" /> from reads or writes.
+        /// </summary>
+        public IDisposable LockDirectory(string directory)
+        {
+            var disposables = new List<IDisposable>();
+            foreach (var file in Directory.EnumerateFiles(directory, "*", SearchOption.AllDirectories))
```
rynowak: It would be nice to do a `Path.Combine(ProjectDirectory.DirectoryPath, directory)` here - this is the convention that I've been doing in other cases. This means that it will work in both cases (if `directory` is rooted or relative to the project).

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -10,11 +11,13 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
...
     public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
     {
+        private const string RazorGenerateTarget = "RazorGenerate";
+
         [Fact]
         [InitializeTestProject("SimpleMvc")]
         public async Task RazorGenerate_Success_GeneratesFilesOnDisk()
         {
-            var result = await DotnetMSBuild("RazorGenerate");
+            var result = await DotnetMSBuild(RazorGenerateTarget);
```
rynowak: gross

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
rynowak: No, you should handle it in all of the other cases. I want the APIs in the tests to be able to use relative paths - that's the part that gets repetative. Look at how Assert.File... works

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -52,5 +57,124 @@ public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
...
+            var generatedFile = Path.Combine(RazorIntermediateOutputPath, "Views", "Home", "Contact.cs");
+
+            // Assert - 1
+            Assert.BuildPassed(result);
+            var fileThumbPrint = GetThumbPrint(generatedFile);
+
+            // Act - 2
+            // Update the source content and build. We should expect the outputs to be regenerated.
+            ReplaceContent("Uodated content", file);
+            await Task.Delay(TimeSpan.FromSeconds(4));
```
pranavkm: There's some flakiness on Travis here. We're just relying on good ol' MSBuild's timestamp calculation to work and it's not consistent. I don't have a OSX machine handy but I wasn't able to reproduce this on my Ubuntu VM. Playing around with delays to make it more stable

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -52,5 +57,126 @@ public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
...
+
+            // Assert - 1
+            Assert.BuildPassed(result);
+            var fileThumbPrint = GetThumbPrint(generatedFile);
+
+            // Act - 2
+            // Update the source content and build. We should expect the outputs to be regenerated.
+            // Timestamps on xplat are precise only to a second. Add a delay so we can ensure that MSBuild recognizes the
+            // file change. See https://github.com/dotnet/corefx/issues/26024
+            await Task.Delay(TimeSpan.FromSeconds(1));
```
rynowak: I don't feel good about this. We should use something like this as a workaround instead https://msdn.microsoft.com/en-us/library/system.io.file.setlastwritetime(v=vs.110).aspx

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -52,5 +57,126 @@ public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
...
+            ReplaceContent("Uodated content", file);
```
rynowak: typo

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
pranavkm: They do have a 1 second granularity, so this works pretty reliably. Setting a write time in the future, which is sort of what we'd do have to effectively do to get the roll up is a poorer approximation of what a user would do.

rynowak: Delays in tests are fragile, and don't scale well.

PR #1860, opened 2017-12-19
Move DefaultTagHelperTargetExtension errors to use RazorDiagnosticFactory.

NTaylorMullen:
- Added tests validating these errors because we didn't have any before.

#1827 

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
@@ -550,7 +550,8 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
             }
         }
 
-        private void RenderTagHelperAttributeInline(
+        // Internal for testing
+        internal void RenderTagHelperAttributeInline(
             CodeRenderingContext context,
             DefaultTagHelperPropertyIntermediateNode property,
             IntermediateNode node,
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
```diff
@@ -316,6 +316,32 @@ public static RazorDiagnostic CreateTagHelper_InvalidTargetedAttributeName(strin
             return diagnostic;
         }
 
+        internal static readonly RazorDiagnosticDescriptor TagHelper_CodeBlocksNotSupportedInAttributes =
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}3011",
+                () => LegacyResources.TagHelpers_CodeBlocks_NotSupported_InAttributes,
```
NTaylorMullen: @ajaybhargavb  left these in `LegacyResources` since you said that we'd do one pass to remove all of the `LegacyResources`.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -574,20 +575,16 @@ public void WriteTagHelperRuntime(CodeRenderingContext context, DefaultTagHelper
             }
             else if (node is CSharpCodeIntermediateNode)
             {
-                var error = new RazorError(
-                    LegacyResources.TagHelpers_CodeBlocks_NotSupported_InAttributes,
-                    SourceLocation.FromSpan(span),
-                    span == null ? -1 : span.Value.Length);
-                context.Diagnostics.Add(RazorDiagnostic.Create(error));
+                var location = span.HasValue ? span.Value : SourceSpan.Undefined;
```
rynowak: This could be simplified to `span ?? SourceSpan.Undefined` and then just inlined.

src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
```diff
@@ -316,6 +316,32 @@ public static RazorDiagnostic CreateTagHelper_InvalidTargetedAttributeName(strin
             return diagnostic;
         }
 
+        internal static readonly RazorDiagnosticDescriptor TagHelper_CodeBlocksNotSupportedInAttributes =
```
NTaylorMullen: Hmm, thinking. Should I move these into the semantic diagnostic factory section? Probably should.

ajaybhargavb: Probably should. Also, merge conflicts 😢

NTaylorMullen: Race you 😉

NTaylorMullen: Merging this, when you merge feel free to push the diagnostic #'s of these diagnostics to be after yours ( so you don't need to increment all of yours ).

PR #1861, opened 2017-12-19
Moved CSharpCodeParser out of LegacyRazorDiagnostic

ajaybhargavb:
#1827 

FYI @mkArtakMSFT

PR #1863, opened 2017-12-19
Include LineIndex and CharacterIndex when comparing SourceLocation

ajaybhargavb:
https://github.com/aspnet/Razor/pull/1861#discussion_r157897818

FYI @mkArtakMSFT

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/SourceLocationTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/SourceLocationTest.cs
@@ -61,7 +61,7 @@ namespace Microsoft.AspNetCore.Razor.Language
         }
 
         [Fact]
-        public void Equals_ReturnsTrue_FilePathsNullAndAbsoluteIndicesMatch()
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/SourceLocationTest.cs
```diff
@@ -61,7 +61,7 @@ public void GetHashCode_ReturnsSameValue_WhenEqual(string path)
         }
 
         [Fact]
-        public void Equals_ReturnsTrue_FilePathsNullAndAbsoluteIndicesMatch()
+        public void Equals_ReturnsFalse_FilePathsNullAndAbsoluteIndicesMatch()
```
rynowak: This test should be replaced with a good test.

ajaybhargavb: There is already a test for the True case

ajaybhargavb:
Looks like this is failing in Travis. Going to revert and investigate why.

PR #1865, opened 2017-12-20
Moved TokenizerBackerParser out of LegacyRazorDiagnostic

ajaybhargavb:
#1827 

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -435,44 +433,6 @@ protected internal bool Optional(TSymbolType type)
             return false;
         }
 
-        protected internal bool Required(TSymbolType expected, bool errorIfNotFound, Func<string, string> errorBase)
```
ajaybhargavb: Unused. Also removed [this](https://github.com/aspnet/Razor/commit/808f539297646ba69dda4356f3d1757c8ac502a9#diff-39391599f57e8191da58dc3bfc4cfbb8L2102) unused method in an earlier commit as well.

PR #1869, opened 2017-12-21
Ensure test apps use the same SDK as Razor

### LINKED ISSUE #1858: Ensure test applications use the same sdk as the one used by Razor \ KoreBuild
This should help us identify regressions in the SDK and ensure we're compatible with the shipping SDK

pranavkm:
Fixes #1858

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
@@ -50,6 +50,7 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 
                 CreateDirectoryProps(projectRoot, binariesRoot, destinationPath);
                 CreateDirectoryTargets(destinationPath);
+                CopyGlobalJson(solutionRoot, destinationPath);
 
                 return new ProjectDirectory(destinationPath);
             }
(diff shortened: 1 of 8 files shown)


rynowak:
👍 :shipit:

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -112,6 +113,14 @@ void CreateDirectoryTargets(string projectRoot)
 ";
                 File.WriteAllText(Path.Combine(projectRoot, "Directory.Build.targets"), text);
             }
+
+            void CopyGlobalJson(string solutionRoot, string projectRoot)
+            {
+                var srcGlobalJson = Path.Combine(solutionRoot, "global.json");
+                var destinationGlobalJson = Path.Combine(projectRoot, "global.json");
+
+                File.Copy(srcGlobalJson, destinationGlobalJson);
```
NTaylorMullen: The global.json is only created when running `./build.cmd` so this will blow up if you're doing `dotnet test` prior to doing a `./build`. I'd do a file exists check

pranavkm: I can add a check. But the scenario is kinda broken regardless. Without the right sdk downloaded \ being pinned, there are no guarantees builds would succeed.

PR #1870, opened 2017-12-21
Change TagHelperBlockRewriter's error usage to use RazorDiagnosticFactory.

NTaylorMullen:
- Changed `DefaultRazorDiagnostic` to expose a little more information for tests. This info is only available when casted to the `DefaultRazorDiagnostic` type and only available internally.
- Changed parameter order of a `DefaultTagHelperTargetExtension` diagnostic.
- Updated baselines of files in `Razor.Language.Test`.
- Added a new convenience constructor on `SourceSpan`. The ctor is equivalent to calling the class with a SourceLocation.

#1827

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorDiagnostic.cs
```diff
@@ -8,26 +8,29 @@ namespace Microsoft.AspNetCore.Razor.Language
...
-        public override string Id => _descriptor.Id;
+        public override string Id => Descriptor.Id;
 
-        public override RazorDiagnosticSeverity Severity => _descriptor.Severity;
+        public override RazorDiagnosticSeverity Severity => Descriptor.Severity;
 
         public override SourceSpan Span { get; }
 
+        // Internal for testing
+        internal RazorDiagnosticDescriptor Descriptor { get; }
```
NTaylorMullen: Exposed a little more information here for tests. Basically in the tests we need to copy diagnostics and update the diagnostics location. Hate that the tests were written that way but it's more code than I'm willing to invest the time in moving.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -581,7 +581,7 @@ public void WriteTagHelperRuntime(CodeRenderingContext context, DefaultTagHelper
             else if (node is TemplateIntermediateNode)
             {
                 var expectedTypeName = property.IsIndexerNameMatch ? property.BoundAttribute.IndexerTypeName : property.BoundAttribute.TypeName;
-                var diagnostic = RazorDiagnosticFactory.CreateTagHelper_InlineMarkupBlocksNotSupportedInAttributes(expectedTypeName, span ?? SourceSpan.Undefined);
+                var diagnostic = RazorDiagnosticFactory.CreateTagHelper_InlineMarkupBlocksNotSupportedInAttributes(span ?? SourceSpan.Undefined, expectedTypeName);
```
NTaylorMullen: Noticed that the parameter order here was different than other diagnostics so I just updated it here in this PR.

src/Microsoft.AspNetCore.Razor.Language/SourceSpan.cs
```diff
@@ -30,6 +30,11 @@ public SourceSpan(string filePath, int absoluteIndex, int lineIndex, int charact
             FilePath = filePath;
         }
 
+        public SourceSpan(int absoluteIndex, int lineIndex, int characterIndex, int length)
```
NTaylorMullen: This is equivalent to using the ctor that uses `SourceLocation` but is more convenient when fixing tests.

src/Microsoft.AspNetCore.Razor.Language/RazorDiagnosticFactory.cs
```diff
@@ -422,6 +422,56 @@ public static RazorDiagnostic CreateTagHelper_InlineMarkupBlocksNotSupportedInAt
...
+            new RazorDiagnosticDescriptor(
+                $"{DiagnosticPrefix}2011",
+                () => LegacyResources.TagHelpers_CannotHaveCSharpInTagDeclaration,
+                RazorDiagnosticSeverity.Error);
+        public static RazorDiagnostic CreateTagHelper_CannotHaveCSharpInTagDeclaration(SourceSpan location, string tagName)
+        {
+            return RazorDiagnostic.Create(TagHelper_CannotHaveCSharpInTagDeclaration, location, tagName);
+        }
+
+        internal static readonly RazorDiagnosticDescriptor TagHelper_AttributesMustHaveAName =
```
ajaybhargavb: Should these also be called `Parsing_Blah` instead of `TagHelper_Blah`? 
Also, these are currently grouped under semantic errors. But we also have a region for TagHelper errors. I'm not sure where we should be putting TagHelper parsing errors.

NTaylorMullen: Ya you're right, these should all be Syntactic errors. That' what I get for continuing where I last added errors 😉. As for naming, I can update them to Parsing_TagHelper*

ajaybhargavb: `Parsing_EmptyTagHelperBoundAttribute` might be a semantic error though. What do you think?

NTaylorMullen: I could see that.

PR #1873, opened 2017-12-22
Use the Muxer path if available

### LINKED ISSUE #1848: Figure out if we can use DotnetMuxer in DotnetToolTask
This does a path search today, we could do better by using out DotnetMuxer utility library.

pranavkm:
Fixes #1848

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
@@ -6,9 +6,10 @@ using System.Diagnostics;
 using System.Threading;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
+using Microsoft.Extensions.CommandLineUtils;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -25,8 +26,18 @@ public abstract class DotNetToolTask : ToolTask
 
         protected override MessageImportance StandardErrorLoggingImportance => MessageImportance.High;
 
-        // use PATH to find dotnet
-        protected override string GenerateFullPathToTool() => ToolExe;
+        protected override string GenerateFullPathToTool()
+        {
+            var mainModule = Process.GetCurrentProcess().MainModule?.FileName;
+            if (!string.IsNullOrEmpty(mainModule)
```
rynowak: When would this be null? when would this be empty?

pranavkm:
🆙 📅 Didn't realize we weren't already referencing CommandLineUtils in the tasks project.

Issue #1698, opened 2017-09-22
Add unit tests for DefaultRazorIndentationFactsService


ajaybhargavb:
https://github.com/aspnet/Razor/pull/1697 adds some unit tests but we need a lot more coverage.

NTaylorMullen:
d4994eb4a451203f5cf6c5f7dc1543886f940b96

Issue #1876, opened 2017-12-22
Add API to RazorSourceLineCollection to enable better understanding of lines in a RazorSourceDocument


NTaylorMullen:
Today the `RazorSourceLineCollection` has too minimal of an API to get enough information about a source document. For instance, you can't go from line Index => line content; this is primarily driven by not being able to get the start position of a given line index.

Tried using this API to improve on our indentation facts service but it didn't have enough information.

NTaylorMullen:
Once this issue is resolved we should re-visit the `DefaultIndentationFactsService` APIs. Today the `GetDesiredIndentation` method relies on an `ITextSnapshot` + `ITextSnapshotLine` due to insufficient information from the `RazorSourceLineCollection`. If we had the ability to go from line index => line content we could change `GetDesiredIndentation` to look like:

```C#
bool TryGetDesiredIndentation(RazorSyntaxTree syntaxTree, int lineIndex, int indentSize, int tabSize, out int desiredIndentation)
```

rynowak:
I'd really prefer to not change the APIs that WTE uses - that's doesn't really add any value. 

This would add value for us if the **implementation** could use `RazorSourceDocument` because then it would be more portable, and not require to rely on mocks of VS services as much.

Issue #1461, opened 2017-06-22
Hitting <Enter> after "@page" in a Razor page indents unnecessarily


halter73:
![image](https://user-images.githubusercontent.com/54385/27452348-1c515fce-5747-11e7-974b-eebd95717ce7.png)

The screenshot shows the editor immediately after `<Enter>`. `<Backspace>` once puts the cursor back at the beginning of the line.

ajaybhargavb:
f41dc2aef94e873df96b3d72acdf1c8d50bf5119

halter73:
Finally! But aren't you supposed to be on a 🏖 right now @ajaybhargavb?

ajaybhargavb:
Haha! No I am back starting today.

PR #1878, opened 2017-12-22
Added DefaultRazorIndentationFactsService unit tests.

NTaylorMullen:
- Already had a good variety of integration tests so refactored the service to properly unit test each piece.
- Found several pieces of unneeded code (wasn't being used) so removed it.
- Removed the `LocateOwner` logic that was embedded in the service. We already have an equivalent locate owner on our `SyntaxTreeNode` items.

#1698

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.ComponentModel.Composition;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
@@ -15,6 +14,21 @@ namespace Microsoft.VisualStudio.Editor.Razor
     [Export(typeof(RazorIndentationFactsService))]
     internal class DefaultRazorIndentationFactsService : RazorIndentationFactsService
     {
+        // This method dives down a syntax tree looking for open curly braces, every time
+        // it finds one it increments its indent until it finds the provided "line". 
+        // 
+        // Examples:
+        // @{
+        //    <strong>Hello World</strong>
+        // }
+        // Asking for desired indentation of the @{ or } lines should result in a desired indentation of 4.
+        //
+        // <div>
+        //     @{
+        //         <strong>Hello World</strong>
+        //     }
+        // </div>
+        // Asking for desired indentation of the @{ or } lines should result in a desired indentation of 8.
         public override int? GetDesiredIndentation(
             RazorSyntaxTree syntaxTree,
             ITextSnapshot syntaxTreeSnapshot,
@@ -47,12 +61,9 @@ namespace Microsoft.VisualStudio.Editor.Razor
                 throw new ArgumentOutOfRangeException(nameof(tabSize));
             }
 
-            var previousLine = line.Snapshot.GetLineFromLineNumber(line.LineNumber - 1);
-            var trackingPoint = previousLine.Snapshot.CreateTrackingPoint(previousLine.End, PointTrackingMode.Negative);
-            var previousLineEndIndex = trackingPoint.GetPosition(syntaxTreeSnapshot);
-
+            var previousLineEndIndex = GetPreviousLineEndIndex(syntaxTreeSnapshot, line);
             var simulatedChange = new SourceChange(previousLineEndIndex, 0, string.Empty);
-            var owningSpan = LocateOwner(syntaxTree.Root, simulatedChange);
+            var owningSpan = syntaxTree.Root.LocateOwner(simulatedChange);
             if (owningSpan.Kind == SpanKindInternal.Code)
             {
                 // Example,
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
```diff
@@ -63,40 +59,17 @@ internal class DefaultRazorIndentationFactsService : RazorIndentationFactsServic
...
+                    if (IsCSharpOpenCurlyBrace(currentChild))
                     {
-                        var currentSpan = currentChild as Span;
-                        if (currentSpan.Symbols.Count == 1 &&
-                            currentSpan.Symbols[0] is CSharpSymbol symbol &&
-                            symbol.Type == CSharpSymbolType.LeftBrace)
-                        {
-                            var extraIndent = 0;
-
-                            // Dev11 337312: Only indent one level deeper if the item after the open curly brace is a markup block
```
NTaylorMullen: Wasn't being used and all their tests are passing so 👍

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
```diff
@@ -105,113 +78,20 @@ internal class DefaultRazorIndentationFactsService : RazorIndentationFactsServic
                     }
                 }
 
-                owningChild = owningParent;
-            }
-
-            return desiredIndentation;
-        }
-
-        private Span LocateOwner(Block root, SourceChange change)
```
NTaylorMullen: We do this better on our SYntaxTreeNodes this is basically just inlined duplicate code 🥂

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
```diff
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
NTaylorMullen: Would have loved to change this API to:
`bool TryGetDesiredIndentation(RazorSyntaxTree syntaxTree, int lineIndex, int indentSize, int tabSize)` but [this](https://github.com/aspnet/Razor/issues/1876) is preventing me from doing that.

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
```diff
@@ -63,40 +59,17 @@ internal class DefaultRazorIndentationFactsService : RazorIndentationFactsServic
...
-                            if (i < children.Count - 1)
-                            {
-                                var nextChild = children[i + 1];
-                                if (nextChild.IsBlock && ((nextChild as Block).Type == BlockKindInternal.Markup))
-                                {
-                                    extraIndent = indentSize;
-                                }
-                            }
-
-                            // We can't rely on the syntax trees representation of the source document because partial parses may have mutated
```
ajaybhargavb: Why nuke this comment?

NTaylorMullen: SyntaxTree no longer gets mutated publicly 😄

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
rynowak: Can you add a note to that bug explaining what to clean up once it's fixed?

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
rynowak: What about our tests? was this truly dead code?

We shouldn't count on their tests to have detailed coverage of these cases.

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
```diff
@@ -63,40 +59,17 @@ internal class DefaultRazorIndentationFactsService : RazorIndentationFactsServic
...
-                            // the underlying SyntaxTree text buffer. Because of this, if we want to provide accurate indentations we need to
-                            // operate on the current line representation as indicated by the provider.
-                            var lineText = line.Snapshot.GetLineFromLineNumber(currentSpan.Start.LineIndex).GetText();
-                            desiredIndentation = GetIndentLevelOfLine(lineText, tabSize) + indentSize;
-                        }
+                        var lineText = line.Snapshot.GetLineFromLineNumber(currentChild.Start.LineIndex).GetText();
```
rynowak: It would be great to have some comments documenting the cases we handle and what the expectations are (with examples)

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
NTaylorMullen: Truly dead code, `extraIndent` is never used

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
NTaylorMullen: Will do

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorIndentationFactsService.cs
rynowak: Oh wow, ok - great cleanup then 💯 🍺 🎄 :canada:

NTaylorMullen:
> If you haven't done it already, make sure to manually verify issues like #1461 doesn't happen again.

Verified

PR #1882, opened 2017-12-23
Make AspNetCoreSettings actually work

pranavkm:


Directory.Build.props
```diff
@@ -1,4 +1,8 @@
 ﻿<Project>
+  <Import
```
pranavkm: 😨

NTaylorMullen: I figured this would be something that would be included in KoreBuild no?

pranavkm: Well, a part of it. KoreBuild uses the files to pull in settings from the file. This is to get repo projects to also pull in settings

PR #1885, opened 2017-12-27
Removed LegacyRazorDiagnostic and RazorError

ajaybhargavb:
#1827 

This is the final piece
- Removed `LegacyRazorDiagnostic` and tests
- Removed `RazorError` and tests
- Updated few outdated references

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/Legacy/ErrorSink.cs
```diff
@@ -28,7 +28,7 @@ public ErrorSink()
         /// <summary>
         /// Tracks the given <paramref name="error"/>.
         /// </summary>
-        /// <param name="error">The <see cref="RazorError"/> to track.</param>
+        /// <param name="error">The <see cref="RazorDiagnostic"/> to track.</param>
         public void OnError(RazorDiagnostic error) =>_errors.Add(error);
```
rynowak: I'm fine with keeping the method name and parameter name as-is. Right now we don't issue warnings.

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorDiagnosticJsonConverter.cs
```diff
@@ -47,12 +46,6 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
                 return RazorDiagnostic.Create(descriptor, sourceSpan);
             }
-            else if (string.Equals(typeName, typeof(LegacyRazorDiagnostic).FullName, StringComparison.Ordinal))
```
rynowak: This code should be simplified further, we no longer need to deal with type names and we can just directly deserialize the `DefaultRazorDiagnostic`. I think it's OK if you do that in a follow up, but this was the issue that raised the priority of this.

rynowak: Actually a follow up is better, I think we'll want to add a microbenchmark for this and do some profiling.

rynowak:
:shipit: for this PR but let's follow up on https://github.com/aspnet/Razor/pull/1885#discussion_r158877010

src/Microsoft.VisualStudio.LanguageServices.Razor/RazorDiagnosticJsonConverter.cs
ajaybhargavb: Filed https://github.com/aspnet/Razor/issues/1886 to track this.

Issue #1836, opened 2017-12-15
Add RazorSyntaxFactsService tests


NTaylorMullen:

rynowak:
@ajaybhargavb and @NTaylorMullen - please divide these up between y'all

ajaybhargavb:
52a06e93bcbf1a6c7449d97646523da0f9208d3e

PR #1891, opened 2017-12-29
Added test for DefaultRazorSyntaxFactsService

ajaybhargavb:
#1836 

Added only a couple of basic tests. Didn't seem worth adding more as these tests cover pretty much everything this class does. 

FYI @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.csproj
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.csproj
@@ -14,10 +14,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsHashCodeCombinerSourcesPackageVersion)" />
-  </ItemGroup>
-
-  <ItemGroup>
     <ProjectReference Include="../Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X.csproj
```diff
@@ -13,10 +13,6 @@
     </Compile>
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsHashCodeCombinerSourcesPackageVersion)" />
```
ajaybhargavb: This is unnecessary as it is available from `Razor.Language`.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorSyntaxFactsServiceTest.cs
```diff
@@ -0,0 +1,83 @@
...
+            var service = new DefaultRazorSyntaxFactsService();
+
+            // Act
+            var spans = service.GetTagHelperSpans(syntaxTree);
+
+            // Assert
+            var actualSpan = Assert.Single(spans);
+            Assert.Equal(expectedSourceSpan, actualSpan.Span);
+            Assert.Equal(tagHelperContext.TagHelpers, actualSpan.TagHelpers);
+            Assert.Equal(tagHelperContext.Prefix, actualSpan.Binding.TagHelperPrefix);
```
ajaybhargavb: Intentionally not verifying the entire `TagHelperBinding` here because it is constructible from the information being verified here. This seemed cleaner.

rynowak:
> Added only a couple of basic tests. Didn't seem worth adding more as these tests cover pretty much everything this class does.

That's the idea 👍 We wanted some kind of basic coverage for these components.

ajaybhargavb:
🆙 📅

PR #1894, opened 2017-12-29
Add metadata to Razor

rynowak:
This PR introduces standard metadata to Razor. This change begins to
formalize the contract between generated code produced by Razor and
runtimes that want to load and interact with this code.

This is a step on making MVC a 'plugin' to Razor rather than the only
possible implementation. Since we're doing MSBuild work - this is the
right time to designate the current interaction between Razor and MVC as
'legacy' and move forward.

Additionally, we need the source thumbprinting to make re-compilation of
Razor a thing.


src/Microsoft.AspNetCore.Razor.Runtime/Hosting/IRazorSourceThumprintMetadata.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Hosting
+{
+    public interface IRazorSourceThumprintMetadata
```
ajaybhargavb: Why interface over abstract class?

rynowak:
> formalize the contract between generated code produced by Razor and runtimes that want to load and interact with this code.

I should mention that this is pretty similar to what we're doing today with the 'manifest'. Razor will generate a bunch of assembly attributes that identify which types are 'Razor' types. The `RazorCompiledItemAttribute` will include enough info to tell us what 'kind' of thing we're deal with and its 'identifier' - which is an analogue for the view engine path. 

This way we can easily scan and find, say, all of the pages, **OR** we can look up a specific one by path. From there we can fan out and look at metadata on the types themselves if we want to. The idea here is that at no point do we have to do an `Assembly.GetExportedTypes` 👍 

This system is also flexible enough for other non-MVC systems to use as long as the output is assemblies and types.

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorSourceThumbprintAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+    /// <para>
+    /// These attributes are added by the Razor infrastructure when generating code to assist runtime
+    /// implementations to determine the integrity of compiled items.
+    /// </para>
+    /// <para>
+    /// Runtime implementations should access the thumbprint metadata for an item using
+    /// <see cref="RazorCompiledItemExtensions.GetThumprintMetadata(RazorCompiledItem)"/>.
+    /// </para>
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
```
rynowak: I was torn on whether or not to include other type-like things here other than class, so YAGNI kicked in. If someone wants to make an argument, I can be convinced.

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorSourceThumbprintAttribute.cs
```diff
@@ -0,0 +1,74 @@
...
+        
+        /// <summary>
+        /// Gets the identifier of the source file associated with this thumbprint. May be <c>null</c>.
+        /// </summary>
+        public string Identifier { get; }
+
+        /// <summary>
+        /// Gets the kind of the source file associated with this thumbprint.
+        /// </summary>
+        public string Kind { get; }
```
ajaybhargavb: why not an enum for Kind? Same question for HashAlgorithm

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItemLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+            return new DefaultRazorCompiledItem(attribute.Type, attribute.Kind, attribute.Identifier);
+        }
+
+        /// <summary>
+        /// Retrieves the list of <see cref="RazorCompiledItemAttribute"/> attributes defined for the provided
+        /// <see cref="Assembly"/>.
+        /// </summary>
+        /// <param name="assembly">The <see cref="Assembly"/> to search.</param>
+        /// <returns>A list of <see cref="RazorCompiledItemAttribute"/> attributes.</returns>
+        protected IEnumerable<RazorCompiledItemAttribute> LoadAttributes(Assembly assembly)
```
ajaybhargavb: protected virtual?

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItem.cs
```diff
@@ -0,0 +1,40 @@
...
+        public abstract string Kind { get; }
+
+        /// <summary>
+        /// Gets a collection of arbitrary metadata associated with the item.
+        /// </summary>
+        /// <remarks>
+        /// For items loaded with the default implementation of <see cref="RazorCompiledItemLoader"/>, the 
+        /// metadata collection will return all attributes defined on the <see cref="Type"/>.
+        /// </remarks>
+        public IReadOnlyList<object> Metadata { get; }
```
ajaybhargavb: abstract?

rynowak: Yeah, good catch, this deffo should be abstract

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItemLoader.cs
rynowak: I didn't make this virtual because I don't think there's a different policy that works here. What's an example of a case where you'd override this?

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorSourceThumbprintAttribute.cs
rynowak: Because we don't know all of the kinds that exist. MVC defines it's own kinds.

For hashalgorithm it might be possible to use an enum, but it's more coupling. If MVC wants to use a hash algorithm that Razor doesn't know then that requires a code change in Razor. That's bad coupling.

Think of it this way, it's your data, we're just holding on to it for you 😆

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorSourceThumbprintAttribute.cs
rynowak: Other things would be enum, interface, struct, etc.

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItemLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// <para>
+    /// Inheriting from <see cref="RazorCompiledItemLoader"/> is useful when an implementation needs to consider
+    /// additional configuration or data outside of the <see cref="Assembly"/> being loaded.
+    /// </para>
+    /// <para>
+    /// Subclasses of <see cref="RazorCompiledItemLoader"/> can return subclasses of <see cref="RazorCompiledItem"/>
+    /// with additional data members by overriding <see cref="CreateItem(RazorCompiledItemAttribute)"/>.
+    /// </para>
+    /// </remarks>
+    public class RazorCompiledItemLoader
```
rynowak: Also want to mention here, this is not an abstraction. This is infrastructure.

rynowak:
Updated and removed [Design]. This is implemented now end-to-end. Do yourself a favor and look at it commit-by-commit. There's lots of baseline updates, every runtime baseline gets updated as part of this change.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorCodeGenerationOptionsFeature.cs
```diff
@@ -22,15 +22,15 @@ protected override void OnInitialized()
 
         public RazorCodeGenerationOptions GetOptions()
         {
-            var builder = new DefaultRazorCodeGenerationOptionsBuilder(_designTime);
+            return _designTime ? RazorCodeGenerationOptions.CreateDesignTime(ConfigureOptions) : RazorCodeGenerationOptions.Create(ConfigureOptions);
+        }
```
rynowak: This was calling the wrong API to create the code generation options

src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
```diff
@@ -3,13 +3,61 @@
...
 namespace Microsoft.AspNetCore.Razor.Language
 {
     public static class RazorCodeDocumentExtensions
     {
-        private static object TagHelperPrefixKey = new object();
+        // Internal for testing
+        internal static readonly string IdentifierKey = "identifier";
+        internal static readonly string ImportIdentifiersKey = "imports-identifiers";
+
+        public static string GetIdentifier(this RazorCodeDocument document)
```
rynowak: There's a big decision to make here about whether or not this should be part of the source document. After finishing this PR, I feel like it should be part of the source document. I'm planning to give that to @ajaybhargavb to play with. Thoughts?

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -35,7 +35,7 @@ protected IntegrationTestBase()
 #if GENERATE_BASELINES
         protected bool GenerateBaselines { get; set; } = true;
 #else
-        protected bool GenerateBaselines { get; set; } = false;
+        protected bool GenerateBaselines { get; set; } = true;
```
rynowak: OOPS lol

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -227,45 +232,5 @@ protected void AssertSourceMappingsMatchBaseline(RazorCodeDocument document)
 
             Assert.Equal(baseline, actual);
         }
-
-        protected class ApiSetsIRTestAdapter : IntermediateNodePassBase, IRazorOptimizationPass
```
rynowak: Doing this in an optimization pass is the wrong thing. I reworked this feature to use the default document classifier, which is how we'd tell users to do it.

rynowak: Also I had to do something here, because the `MetadataAttributePass` was seeing the uninitialized values.

src/Microsoft.AspNetCore.Razor.Language/RazorCodeGenerationOptionsBuilder.cs
```diff
@@ -11,8 +11,35 @@ public abstract class RazorCodeGenerationOptionsBuilder
...
+        /// <para>
+        /// The <c>Microsoft.AspNetCore.Razor.Runtime</c> package includes a default set of attributes intended
+        /// for runtimes to discover metadata about the compiled code.
+        /// </para>
+        /// <para>
+        /// The default metadata attributes should be suppressed if code generation targets a runtime without
+        /// a reference to <c>Microsoft.AspNetCore.Razor.Runtime</c>, or for testing purposes.
+        /// </para>
+        /// </remarks>
+        public virtual bool SuppressMetadataAttributes { get; set; }
```
rynowak: Added docs here because the metadata attributes also include checksums. I wanted to make it clear which one does which thing

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributePass.cs
```diff
@@ -0,0 +1,112 @@
...
+                Identifier = identifier,
+            });
+
+            // Now we need to add a [RazorSourceChecksum] for the source and for each import
+            // these are class attributes, so we need to find the insertion point to put them
+            // right before the class.
+            var insert = (int?)null;
+            for (var j = 0; j < @namespace.Children.Count; j++)
+            {
+                if (object.ReferenceEquals(@namespace.Children[j], @class))
```
NTaylorMullen: Should we just add an intermediate node collection called `Attributes` to our `ClassDeclarationIntermediateNode`? Could even add it to the `MemberDeclrationIntermediateNode` if we want to enable it everywhere 😉

src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
```diff
@@ -3,13 +3,61 @@
...
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            document.Items[IdentifierKey] = identifier;
+        }
+
+        public static IReadOnlyList<string> GetImportIdentifiers(this RazorCodeDocument document)
```
NTaylorMullen: This goes away once @ajaybhargavb moves the `GetIdentifier` to source document yes?

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItemLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <returns>A list of <see cref="RazorCompiledItem"/> objects.</returns>
+        public virtual IReadOnlyList<RazorCompiledItem> LoadItems(Assembly assembly)
+        {
+            if (assembly == null)
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            var items = new List<RazorCompiledItem>();
+            foreach (var attribute in LoadAttributes(assembly))
```
NTaylorMullen: nit: split the `LoadAttributes` into its own local variable for easier debugging.

src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
rynowak: The identifier is whatever you want. It's a way for the project and runtime to communicate. By default this will be a relative path.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -67,28 +68,52 @@ public override void VisitDocument(DocumentIntermediateNode node)
...
+                    // We only support algorithms that the debugger understands, which is currently SHA1 and SHA256.
 
-                    var sourceDocument = Context.SourceDocument;
-
-                    var checksum = sourceDocument.GetChecksum();
-                    var fileHashBuilder = new StringBuilder(checksum.Length * 2);
-                    foreach (var value in checksum)
+                    string algorithmId;
+                    var algorithm = Context.SourceDocument.GetChecksumAlgorithm();
+                    if (algorithm == HashAlgorithmName.SHA256.Name)
```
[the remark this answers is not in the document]
rynowak: Sure beans.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -67,28 +68,52 @@ public override void VisitDocument(DocumentIntermediateNode node)
...
+                            HashAlgorithmName.SHA1.Name,
+                            HashAlgorithmName.SHA256.Name
+                        });
+
+                        var message = Resources.FormatUnsupportedChecksumAlgorithm(
+                            algorithm,
+                            supportedAlgorithms,
+                            nameof(RazorCodeGenerationOptions) + "." + nameof(RazorCodeGenerationOptions.SuppressChecksum),
+                            bool.TrueString);
+                        throw new InvalidOperationException(message);
```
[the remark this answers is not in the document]
rynowak: This is a programming error though. You've written a source document implementation that can't generate checksums we support.

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributeTargetExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (node == null)
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            // [assembly: global::...RazorCompiledItem(typeof({node.TypeName}), @"{node.Kind}", @"{node.Identifier}")]
+            context.CodeWriter.Write("[assembly: ");
+            context.CodeWriter.Write(CompiledItemAttributeName);
+            context.CodeWriter.Write("(typeof(");
+            context.CodeWriter.Write(node.TypeName);
```
pranavkm: Does this need `global::`?

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItem.cs
```diff
@@ -0,0 +1,40 @@
...
+        public abstract string Kind { get; }
+
+        /// <summary>
+        /// Gets a collection of arbitrary metadata associated with the item.
+        /// </summary>
+        /// <remarks>
+        /// For items loaded with the default implementation of <see cref="RazorCompiledItemLoader"/>, the 
+        /// metadata collection will return all attributes defined on the <see cref="Type"/>.
+        /// </remarks>
+        public abstract IReadOnlyList<object> Metadata { get; }
```
pranavkm: Neat

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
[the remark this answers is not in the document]
rynowak: Discussed in person, and I still think I'm right. Going to leave this as an exception.

src/Microsoft.AspNetCore.Razor.Language/Extensions/IMetadataAttributeTargetExtension.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language.CodeGeneration;
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    internal interface IMetadataAttributeTargetExtension : ICodeTargetExtension
```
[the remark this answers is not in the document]
rynowak: internal because there's no reason for it to be public. Just following the pattern we use elsewhere

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributePass.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            // We need to be able to compute the data we need for the [RazorCompiledItem] attribute - that includes
+            // a full type name, and a document kind, and optionally an identifier.
+            // 
+            // If we can't use [RazorCompiledItem] then we don't care about the rest of the attributes.
+            var @namespace = documentNode.FindPrimaryNamespace();
+            if (@namespace == null || string.IsNullOrEmpty(@namespace.Content))
+            {
+                // No namespace node or it's incomplete. Skip.
```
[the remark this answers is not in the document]
rynowak: Yeah, I'm still not convinced that's right. We need to do **something**, but I don't love that either.

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributePass.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                TypeName = @namespace.Content + "." + @class.ClassName,
+                Kind = documentNode.DocumentKind,
+                Identifier = identifier,
+            });
+
+            // Now we need to add a [RazorSourceChecksum] for the source and for each import
+            // these are class attributes, so we need to find the insertion point to put them
+            // right before the class.
+            var insert = (int?)null;
```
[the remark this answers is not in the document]
rynowak: that's probably true.

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributePass.cs
rynowak: I started there but it's really a lot of work. Modeling attributes using a node has some challenges. You could do by adding the collection and not adding an Attribute node, but meh.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives_Runtime.codegen.cs
```diff
@@ -1,6 +1,7 @@
 #pragma checksum "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml" "{ff1816ec-aa5e-4d10-87f7-6f4963833460}" "fec5cf763044f842fa2114e997bb07e0bf280cd6"
 // <auto-generated/>
 #pragma warning disable 1591
+[assembly: global::Microsoft.AspNetCore.Razor.Hosting.RazorCompiledItemAttribute(typeof(AspNetCore.TestFiles_IntegrationTests_CodeGenerationIntegrationTest_IncompleteDirectives_cshtml), @"mvc.1.0.razor-page", @"TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives")]
 [assembly:global::Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.RazorPageAttribute(null, typeof(AspNetCore.TestFiles_IntegrationTests_CodeGenerationIntegrationTest_IncompleteDirectives_cshtml), null)]
```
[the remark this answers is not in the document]
rynowak: Yes. This is the new contract between Razor and MVC

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributePass.cs
```diff
@@ -0,0 +1,112 @@
...
+using System;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    internal class MetadataAttributePass : IntermediateNodePassBase, 
+        
+        // Optimization pass is the best choice for this class. It's not an optimization, but it also doesn't add semantically
+        // meaningful information.
+        IRazorOptimizationPass
```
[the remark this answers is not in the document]
rynowak: Sure

src/Microsoft.AspNetCore.Razor.Language/Extensions/MetadataAttributeTargetExtension.cs
rynowak: Not here. We apply global to `node.TypeName` instead.

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorCompiledItem.cs
rynowak: YEAHHHHHHHHHH BOI

rynowak:
Going to merge this and then take a look at that travis failure, which is not related to this PR.

PR #1902, opened 2018-01-03
Add support for running tests using desktop msbuild

pranavkm:


pranavkm:
WIP to see if this works e2e

pranavkm:
Cool, this works xplat. cc @rynowak \ @natemcmaster

natemcmaster:
The means you can build the .sln file on a clean repo anymore, right?

build/repo.targets
```diff
@@ -7,4 +7,27 @@
...
+  <Target Name="GenerateMSBuildLocationFile"
+      DependsOnTargets="GetToolsets"
+      Inputs="$(MSBuildLocationFileTemplate);$(VisualStudioMSBuildx86Path)"
+      Outputs="$(MSBuildLocationFileOutput)">
+
+    <PropertyGroup>
+      <TemplateProperties>MSBuildLocation=$(VisualStudioMSBuildx86Path)</TemplateProperties>
+    </PropertyGroup>
+
+    <GenerateFileFromTemplate
```
natemcmaster: FYI this target is also available in a project context. Sdk_GenerateFileFromTemplate. You might consider making this part of the normal project build so VS scenarios work. You could detect the path to MSBuild.exe when building in VS by looking at build-in MSBuild properties such as MSBuildToolsPath.

pranavkm:
Yeah. We already require a global.json to be generated for some of the tests, this is pilling on to it.

natemcmaster:
Ok. Change looks fine to me. :shipit: if @rynowak  is ok with this too

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs
```diff
@@ -15,9 +16,19 @@ public class RazorGenerateIntegrationTest : MSBuildIntegrationTestBase
...
+        [OSSkipCondition(OperatingSystems.Linux)]
+        [OSSkipCondition(OperatingSystems.MacOSX)]
+        [InitializeTestProject("SimpleMvc")]
+        public Task RazorGenerate_UsingDesktopMSBuild_Success_GeneratesFilesOnDisk()
+            => RazorGenerate_Success_GeneratesFilesOnDisk(MSBuildProcessKind.Desktop);
+
+        private async Task RazorGenerate_Success_GeneratesFilesOnDisk(MSBuildProcessKind msBuildProcessKind)
         {
-            var result = await DotnetMSBuild(RazorGenerateTarget);
+            var result = await DotnetMSBuild(RazorGenerateTarget, msBuildProcessKind: msBuildProcessKind);
```
rynowak: Can we limit this to just `Build_SimpleMvc_CanBuildSuccessfully`? I don't want to do cross product testing, and we have no reason to believe that our code will behave differently across different MSBuild kinds.

pranavkm: Sure

PR #1903, opened 2018-01-04
Use common BDN runner and config

ajaybhargavb:
aspnet/Coherence-Signed#684

FYI @mkArtakMSFT


rynowak:
Looks about right to me. /cc @pakrym

ajaybhargavb:
🆙 📅

Issue #1886, opened 2017-12-27
Simplify RazorDiagnostic deserialization


ajaybhargavb:
From https://github.com/aspnet/Razor/pull/1885#discussion_r158877067

Now that `LegacyRazorDiagnostic` is gone, we don't need to look at the type name when deserializing RazorDiagnostic. We can just directly deserialize to `DefaultRazorDiagnostic`.

Additionally, we need to do some profiling on this. This is a hot spot for Razor at build time

Here's a basic plan:
- [x] Use the `dotnet new mvc` template, add the serializers to project, and dump the set of taghelpers as JSON.
- [x] Add a new perf microbenchmark that serializes/deserializes this data
- [x] ~Do some profiling and make improvements to the serialization code~ (Not doing this because serialization doesn't seem to be high impact on perf as shown by the benchmark test)

rynowak:
@ajaybhargavb added some details, thanks!

ajaybhargavb:
98e74b9a69b7b277fc9a78c068de967185104d69

ajaybhargavb:
> This is a hot spot for Razor at build time

After looking at the [benchmark test](https://github.com/aspnet/Razor/pull/1904#issue-286087035), it looks like this is not the case.

rynowak:
@ajaybhargavb - done?

ajaybhargavb:
b68d9cf01fcc2ab39bedfb1d167e455de3f2efa9

PR #1904, opened 2018-01-04
Added TagHelperSerializationBenchmark

ajaybhargavb:
#1886 

- Added a `taghelper.json` input file which contains all the tag helpers from the default mvc template
- Added a benchmark test that serializes/deserializes above data

This takes an average of **5 ms**.
```
Runtime = .NET Core 2.1.0-preview1-26016-05 (Framework 4.6.26016.03), 64bit RyuJIT; GC = Concurrent Server
Mean = 5.4759 ms, StdErr = 0.0113 ms (0.21%); N = 30, StdDev = 0.0620 ms
```
  
FYI @mkArtakMSFT


test/Microsoft.AspNetCore.Razor.Performance/TagHelperSerializationBenchmark.cs
```diff
@@ -0,0 +1,53 @@
...
+            var serializer = new JsonSerializer();
+            serializer.Converters.Add(new RazorDiagnosticJsonConverter());
+            serializer.Converters.Add(new TagHelperDescriptorJsonConverter());
+
+            // Deserialize from json file.
+            IReadOnlyList<TagHelperDescriptor> tagHelpers;
+            using (var stream = File.OpenRead(_tagHelperFilePath))
+            {
+                var reader = new JsonTextReader(new StreamReader(stream));
+                tagHelpers = serializer.Deserialize<IReadOnlyList<TagHelperDescriptor>>(reader);
```
rynowak: I think it would be good read the taghelpers into memory in the constructor, and avoid any actual I/O as part of the test. 

That would look like file -> filestream -> copy to memory stream -> get the byte array

Then each test creates a new stream around that byte array

ajaybhargavb:
🆙 📅

benchmarks/Microsoft.AspNetCore.Razor.Performance/TagHelperSerializationBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+            IReadOnlyList<TagHelperDescriptor> tagHelpers;
+            using (var stream = new MemoryStream(_tagHelperBuffer))
+            {
+                var reader = new JsonTextReader(new StreamReader(stream));
+                tagHelpers = serializer.Deserialize<IReadOnlyList<TagHelperDescriptor>>(reader);
+            }
+
+            // Serialize back to json.
+            using (var stream = new MemoryStream())
+            using (var writer = new StreamWriter(stream, Encoding.UTF8, bufferSize: 4096, leaveOpen: true))
```
NTaylorMullen: don't need to leave open

benchmarks/Microsoft.AspNetCore.Razor.Performance/TagHelperSerializationBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            var serializer = new JsonSerializer();
+            serializer.Converters.Add(new RazorDiagnosticJsonConverter());
+            serializer.Converters.Add(new TagHelperDescriptorJsonConverter());
+
+            // Deserialize from json file.
+            IReadOnlyList<TagHelperDescriptor> tagHelpers;
+            using (var stream = new MemoryStream(_tagHelperBuffer))
+            {
+                var reader = new JsonTextReader(new StreamReader(stream));
```
NTaylorMullen: Should you be disposing this?

benchmarks/Microsoft.AspNetCore.Razor.Performance/TagHelperSerializationBenchmark.cs
```diff
@@ -0,0 +1,54 @@
...
+using BenchmarkDotNet.Attributes;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.VisualStudio.LanguageServices.Razor;
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.Razor.Performance
+{
+    public class TagHelperSerializationBenchmark
+    {
+        private byte[] _tagHelperBuffer;
```
NTaylorMullen: readonly

benchmarks/Microsoft.AspNetCore.Razor.Performance/TagHelperSerializationBenchmark.cs
ajaybhargavb: This doesn't matter as it will be disposed anyway but I agree there is no reason for this here.

PR #1908, opened 2018-01-06
Add relative path to source document

rynowak:
Adding this via a properties object that encompasses all of the optional
properties. This way if we need to add more items that are optional we
can continue to do so without overload explosion.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
@@ -18,14 +18,32 @@ namespace Microsoft.AspNetCore.Razor.Language
         internal static readonly RazorSourceDocument[] EmptyArray = new RazorSourceDocument[0];
 
         /// <summary>
-        /// Encoding of the file that the text was read from.
+        /// Gets the encoding of the text in the original source document.
         /// </summary>
+        /// <remarks>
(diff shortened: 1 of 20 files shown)


rynowak:
Will send a follow-up to actually **use** the new property. This just introduces it and sets it.

src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
```diff
@@ -72,6 +90,15 @@ public virtual string GetChecksumAlgorithm()
             return HashAlgorithmName.SHA1.Name;
         }
 
+        /// <summary>
+        /// Gets the file path in a format that should be used for display.
+        /// </summary>
+        /// <returns>The <see cref="RelativePath"/> if set, or the <see cref="FilePath"/>.</returns>
+        public virtual string GetFilePathForDisplay()
+        {
+            return RelativePath ?? FilePath;
```
NTaylorMullen: Hmm, is it worth handling the situation when both relative path and file path are null? I forget what we do in documents that have both of those values null.

rynowak: If that happens you just want get a file path in the display. Seems fine to me

rynowak:
@NTaylorMullen - btw we messed up kinda bad with the `RazorProject` API, an allowing `FilePath` to be defined in a wierd way. Currently working on fixing it - fallout from this PR

PR #1910, opened 2018-01-06
Add RelativePhysicalPath

rynowak:
The FilePath property on RPI isn't what we want for use in the compiler,
it's more of a view engine concept. Adding a property that does what we
want.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/FileSystemRazorProject.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/FileSystemRazorProject.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Razor.Language
             {
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(root));
             }
-
+            
             Root = root.Replace('\\', '/').TrimEnd('/');
         }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Razor.Language/FileSystemRazorProject.cs
```diff
@@ -36,16 +39,24 @@ public override IEnumerable<RazorProjectItem> EnumerateItems(string basePath)
...
+            var absoluteBasePath = NormalizeAndEnsureValidPath("/");
             var absolutePath = NormalizeAndEnsureValidPath(path);
 
-            return new FileSystemRazorProjectItem("/", path, new FileInfo(absolutePath));
+            var file = new FileInfo(absolutePath);
+
+            var relativePhysicalPath = file.FullName.Substring(absoluteBasePath.Length + 1); // Include leading separator
+            var filePath = "/" + relativePhysicalPath.Replace(Path.DirectorySeparatorChar, '/');
+
+            return new FileSystemRazorProjectItem("/", filePath, relativePhysicalPath, new FileInfo(absolutePath));
```
rynowak: Rather than reuse `path` here, I'm computing `filePath` from the rest of the data, this way we're sure everything is correct.

rynowak:
After merging this, I'll need to update MVC and then I can start using it. If I switch to this new property it will break until MVC adds it.

rynowak:
@NTaylorMullen, @ajaybhargavb - I'm to merge this to make progress but keep the PR open in case y'all have important feedback

src/Microsoft.AspNetCore.Razor.Language/RazorProjectItem.cs
```diff
@@ -18,15 +19,23 @@ public abstract class RazorProjectItem
...
-        /// The absolute path to the file, including the file name.
+        /// The absolute physical (file system) path to the file, including the file name.
         /// </summary>
         public abstract string PhysicalPath { get; }
 
+        /// <summary>
+        /// The relative physical (file system) path to the file, including the file name. Relative to the
+        /// physical path of the <see cref="BasePath"/>.
+        /// </summary>
+        public virtual string RelativePhysicalPath => null;
```
ajaybhargavb: I'm curious why we are putting file system concepts in `RazorProjectItem` instead of just on `FileSystemRazorProjectItem`. Shouldn't `RazorProjectItem` be generic enough to support items from places other than file system? Or am I misunderstanding the purpose of this abstraction?

rynowak: It already has physical path.

Also, I'm not sure anyone understands the purpose of this abstraction.... If we did it again, I would do it differently. This API has a bunch of MVC-isms built in.... but at the same time, file systems are pretty fundamantal. SO I decided to just embrace it.

PR #1911, opened 2018-01-06
Add an Identifier by default

rynowak:
Using the MVC view engine convention for identifiers seems to make the
most sense, and we already use that convention for RazorProject so I
guess we're stuck with it.


src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
```diff
@@ -3,62 +3,12 @@
...
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
     public static class RazorCodeDocumentExtensions
     {
-        // Internal for testing
-        internal static readonly string IdentifierKey = "identifier";
-        internal static readonly string ImportIdentifiersKey = "imports-identifiers";
-
-        public static string GetIdentifier(this RazorCodeDocument document)
```
rynowak: Now that we have relative paths on the source document, it just makes a lot more sense to compute the identifier on demand. I wasn't a big fan of putting it on the code document, and it's possible/easy to have a mismatch between the imports documents and identifiers.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultMetadataIdentifierFeature.cs
```diff
@@ -0,0 +1,40 @@
...
+                throw new ArgumentNullException(nameof(sourceDocument));
+            }
+
+            if (string.IsNullOrEmpty(sourceDocument.RelativePath))
+            {
+                return null;
+            }
+
+            var identifier = sourceDocument.RelativePath;
+            if (!identifier.StartsWith("/") && !identifier.StartsWith("\\"))
```
ajaybhargavb: Could get rid of the second part of this `if` if you `Replace("\\", "/")` first.

src/Microsoft.AspNetCore.Razor.Language/Extensions/IMetadataIdentifierFeature.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language.Extensions
+{
+    internal interface IMetadataIdentifierFeature : IRazorEngineFeature
+    {
+        string GetIdentifier(RazorCodeDocument codeDocument, RazorSourceDocument sourceDocument);
```
ajaybhargavb: Is `RazorCodeDocument` even necessary in this api?

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultMetadataIdentifierFeature.cs
rynowak: Yeah that might be clearer 👍

src/Microsoft.AspNetCore.Razor.Language/Extensions/IMetadataIdentifierFeature.cs
rynowak: It's not necessary in the default implementation

ajaybhargavb: Oops I meant `RazorSourceDocument`. Why not use `codeDocument.Source`?

rynowak: This applies to imports too

ajaybhargavb: Ah. Got it 👍

rynowak:
Looking into the appveyor failure. That's not related to this change at all

PR #1915, opened 2018-01-08
Update our version of the RazorCoreCompile

rynowak:
Updating RazorCoreCompile to be based on a much newer
version of the Core.CSharp.targets.

This adds support for some things we're missing like determinism and
analzyers.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
@@ -1,60 +1,85 @@
 <Project>
-  <!--
-    What follows here was copied and modified from the XamlPreCompile target in 
-    Microsoft.CSharp.CurrentVersion.targets
-
-    The XamlPreCompile target must remain identical to
-    the CoreCompile target in Microsoft.CSharp.Core.targets.
-    Any updates to one must be made to the other.
-  -->
-
-
   <ItemGroup>
     <!-- Used to creating the final compiled Razor dll -->
-    <_RazorIntermediateAssembly Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
+    <RazorIntermediateAssembly Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
 
     <!-- Similar to https://github.com/Microsoft/msbuild/blob/908cc9ccd4961441628f68e37a148183a87bb067/src/Tasks/Microsoft.Common.CurrentVersion.targets#L146-L153 -->
-    <_RazorIntermediatePdb 
+    <_RazorDebugSymbolsIntermediatePath 
       Condition="'$(DebugSymbols)'=='true' and '$(DebugType)'!='' and '$(DebugType)'!='none' and '$(DebugType)'!='embedded'" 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
```diff
@@ -1,76 +1,114 @@
...
-
-
   <ItemGroup>
     <!-- Used to creating the final compiled Razor dll -->
-    <_RazorIntermediateAssembly Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
+    <RazorIntermediateAssembly Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
 
     <!-- Similar to https://github.com/Microsoft/msbuild/blob/908cc9ccd4961441628f68e37a148183a87bb067/src/Tasks/Microsoft.Common.CurrentVersion.targets#L146-L153 -->
-    <_RazorIntermediatePdb 
+    <_RazorDebugSymbolsIntermediatePath
```
rynowak: Matching the naming used in the SDK

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
```diff
@@ -1,76 +1,114 @@
...
+    Remove Returns="@(CscCommandLineArgs)"
+    Remove @(_CoreCompileResourceInputs)
+    Remove $(ApplicationIcon) $(Win32Resource) $(Win32Manifest)
+    Remove @(EmbeddedDocumentation) and @(EmbeddedFiles)
+    Remove @(CustomAdditionalCompileInputs) and @(CustomAdditionalCompileOutputs)
+    Remove @(DocFileItem)
+    Remove PdbFile="$(PdbFile)"
+    Remove OutputRefAssembly="@(IntermediateRefAssembly)"
+
+    Remove EmbedAllSources="$(EmbedAllSources)" - not supported by our supported version of MSBuild
```
rynowak: This is a concern with us shipping outside of the SDK... We have to think kinda carefully about what versions of the SDK we can support. I don't want to block on it right now - but we need to know the answer before we ship 2.1.

My expectation is that we'll end up with these files split between the SDK and a nupkg.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
```diff
@@ -113,55 +162,14 @@
...
 
     <ItemGroup>
       <FileWrites Include="@(_RazorIntermediateAssembly)" Condition="Exists('@(_RazorIntermediateAssembly)')" />
-      <FileWrites Include="@(_RazorIntermediatePdb)" Condition="Exists('@(_RazorIntermediatePdb)')" />
+      <FileWrites Include="@(_RazorDebugSymbolsIntermediatePath)" Condition="Exists('@(_RazorDebugSymbolsIntermediatePath)')" />
     </ItemGroup>
-
-</Target>
-
-<Target Name="_CopyRazorAssemblies"
```
rynowak: Moved to the main targets file.

Issue #1918, opened 2018-01-09
Add MPack version constant for WebTooling to consume


NTaylorMullen:

NTaylorMullen:
e2703fe5f6f1cdcb158fbe867e2724655ebf7308

PR #1919, opened 2018-01-09
Added version constant so WebTooling can consume.

NTaylorMullen:
- Updated the VS for mac versions to 7.5.

#1918

FYI @mkArtakMSFT @ToddGrun

### DIFF
--- a/tooling/Microsoft.VisualStudio.Mac.RazorAddin/Properties/_Manifest.addin.xml
+++ b/tooling/Microsoft.VisualStudio.Mac.RazorAddin/Properties/_Manifest.addin.xml
@@ -10,8 +10,8 @@
     <Import assembly="Microsoft.VisualStudio.Mac.LanguageServices.Razor.dll" />
   </Runtime>
   <Dependencies>
-    <Addin id="::MonoDevelop.Core" version="7.4" />
-    <Addin id="::MonoDevelop.Ide" version="7.4" />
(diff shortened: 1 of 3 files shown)


tooling/Microsoft.VisualStudio.Mac.RazorAddin/Properties/_Manifest.addin.xml
```diff
@@ -10,8 +10,8 @@
     <Import assembly="Microsoft.VisualStudio.Mac.LanguageServices.Razor.dll" />
   </Runtime>
   <Dependencies>
-    <Addin id="::MonoDevelop.Core" version="7.4" />
-    <Addin id="::MonoDevelop.Ide" version="7.4" />
+    <Addin id="::MonoDevelop.Core" version="7.5" />
```
NTaylorMullen: These versions grow with the version of VS for mac

ajaybhargavb: Kinda sucks that we have to do this in 3 different places.

NTaylorMullen: Totally agree. The one in C# is totally a convenience piece. Eventually we could generate this manifest.addin.xml based off of msbuild if we find ourselves touching all this a lot.

PR #1920, opened 2018-01-09
Adding more tests and clarity around file paths

rynowak:
I noticed we were really undertesting all of the things that handle
paths and file names. I gave this some love and a little clean up where
we weren't doing the right thing in RazorSourceDocument.

Also changed the template engine tests to use the
FileSystemRazorProject. These tests are already using the files on disk
as inputs. I turned off checksums for these since they now have the full
file path, and that would not be portable.


src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
```diff
@@ -183,13 +183,20 @@ public static RazorSourceDocument ReadFrom(RazorProjectItem projectItem)
...
+                // Then fall back to the FilePath (yeah it's a bad name) which is like an MVC view engine path
+                // It's much better to have something than nothing.
                 filePath = projectItem.FilePath;
             }
 
             using (var stream = projectItem.Read())
             {
                 // Autodetect the encoding.
-                return new StreamSourceDocument(stream, null, new RazorSourceDocumentProperties(filePath, projectItem.FilePath));
+                return new StreamSourceDocument(stream, null, new RazorSourceDocumentProperties(filePath, projectItem.RelativePhysicalPath));
```
rynowak: This was supposed to be updated in a previous commit but it got missed somehow.

rynowak: This was causing some codegen issues around recompilation, which is how I found it.

rynowak: The codegen tests don't use the template engine, and already had these documents created the right way.

src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
```diff
@@ -183,13 +183,20 @@ public static RazorSourceDocument ReadFrom(RazorProjectItem projectItem)
             if (string.IsNullOrEmpty(filePath))
             {
                 // Fall back to the relative path only if necessary.
+                filePath = projectItem.RelativePhysicalPath;
+            }
+
+            if (string.IsNullOrEmpty(filePath))
+            {
+                // Then fall back to the FilePath (yeah it's a bad name) which is like an MVC view engine path
```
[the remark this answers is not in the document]
rynowak: Yah, other than this filename being crazy, I don't feel like RazorProject is quite so bad. I was down on it earlier this week, but I think we got a lot more right than wrong here.

PR #1924, opened 2018-01-10
Code dump

rynowak:
This builds, but isn't fully implemented yet.t

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
@@ -1,0 +1,91 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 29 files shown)


rynowak:
FYI only - doing a code dump of what I have so far into the repo so @ajaybhargavb can finish the implementation.

src/Microsoft.AspNetCore.Razor.Tools/Application.cs
```diff
@@ -0,0 +1,91 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Threading;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.AspNetCore.Razor.TagHelperTool
```
rynowak: These namespaces should all be changed to `Microsoft.AspNetCore.Razor.Tools`

Issue #1913, opened 2018-01-07
Flaky test: WorkspaceChanged_ProjectRemovedEvent_RemovesProject 


rynowak:
I saw this fail on appveyor, and it looks like virtually none of our code is on the stack. 

https://ci.appveyor.com/project/aspnetci/razor/build/1.0.3753

mkArtakMSFT:
@rynowak, any suggestion who should look into this?

rynowak:
@NTaylorMullen can probably start.  We should first do some googling and see if anyone else has hit something like this.

NTaylorMullen:
This is due to Roslyn's dependency on MEF. Looks like MEF has an issue when solutions/workspaces are created in parallel (we do this when creating/using adhoc workspaces). I'll build a helper method that locks around the creation of workspaces so we don't run into issues.

NTaylorMullen:
f025a27f042e13d0a753d559f2f2e9cec8ae0a1c

PR #1926, opened 2018-01-11
Unify how AdhocWorkspace is constructed.

NTaylorMullen:
- This fixes flakey tests that use AdhocWorkspace.
- Created a new `TestWorkspace` type to lock on `AdhocWorkspace` construction. This type can't be constructed/changed in parallel.
- Updated test usages of `AdhocWorkspace` to make use of `TestWorkspace`.
- Added a `Microsoft.CodeAnalysis.Razor.Workspaces.Test.Common` project to have the `TestWorkspace` type.

#1913

### DIFF
--- a/test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
+++ b/test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
@@ -12,7 +12,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
     {
         public DefaultProjectSnapshotManagerTest()
         {
-            Workspace = new AdhocWorkspace();
+            Workspace = TestWorkspace.Create();
             ProjectManager = new TestProjectSnapshotManager(Enumerable.Empty<ProjectSnapshotChangeTrigger>(), Workspace);
         }
 
--- a/test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorDocumentManagerTest.cs
+++ b/test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorDocumentManagerTest.cs
@@ -31,7 +31,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
 
         private ImportDocumentManager ImportDocumentManager => Mock.Of<ImportDocumentManager>();
 
-        private Workspace Workspace => new AdhocWorkspace();
+        private Workspace Workspace => TestWorkspace.Create();
 
         private TextBufferProjectService SupportedProjectService { get; } = Mock.Of<TextBufferProjectService>(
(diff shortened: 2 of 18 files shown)


test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
```diff
@@ -10,289 +10,317 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
...
+            Project project = null;
+            var workspace = TestWorkspace.Create(ws =>
+            {
+                project = ws.CurrentSolution.AddProject("Test", "Test", LanguageNames.CSharp);
+            });
+            var projectManager = new TestProjectSnapshotManager(Enumerable.Empty<ProjectSnapshotChangeTrigger>(), workspace);
 
-            Assert.True(ProjectManager.ListenersNotified);
-            Assert.False(ProjectManager.WorkerStarted);
+            return (workspace, project, projectManager);
```
rynowak: As an aside, I've come to really like the pattern of just setting properties on the class.

rynowak: BTW you'll now have to skip ALL of these tests on net46 because of the issue with value tuples that will cause it to fail on appveyor. Workaround is to not use valuetuple

NTaylorMullen: > As an aside, I've come to really like the pattern of just setting properties on the class.

Ya me too. I forgot why I had to move away from that here. I'll have to look at it again.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorDocumentManagerTest.cs
```diff
@@ -31,8 +31,6 @@ public class DefaultRazorDocumentManagerTest : ForegroundDispatcherTestBase
 
         private ImportDocumentManager ImportDocumentManager => Mock.Of<ImportDocumentManager>();
 
-        private Workspace Workspace => new AdhocWorkspace();
```
rynowak: You don't really need to rewrite all of this, just use your new class to create the workspace

test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
rynowak: But yeah, that tuple thing is really painful right now. I didn't look but I suspect that's why your appveyor is teh failz0r

rynowak:
@NTaylorMullen - whats the status if this? Lots of failure in appveyor 😢

PR #1929, opened 2018-01-12
Use FilePath if RelativePhysicalPath on RazorProjectItem is null

pranavkm:


src/Microsoft.AspNetCore.Razor.Language/RazorSourceDocument.cs
```diff
@@ -196,7 +196,8 @@ public static RazorSourceDocument ReadFrom(RazorProjectItem projectItem)
             using (var stream = projectItem.Read())
             {
                 // Autodetect the encoding.
-                return new StreamSourceDocument(stream, null, new RazorSourceDocumentProperties(filePath, projectItem.RelativePhysicalPath));
+                var relativePath = projectItem.RelativePhysicalPath ?? projectItem.FilePath;
```
ajaybhargavb: Should the check be `string.IsNullOrEmpty`? Looks like that is what we do above.

ajaybhargavb: Probably unrelated to this bug fix but should we also normalize this path? It is very likely for `RelativePhysicalPath` and `FilePath` to have different path separators. @rynowak thoughts?

rynowak: We normalize the path in the places where we use it for something important. I think this is fine.

PR #1933, opened 2018-01-14
Make Razor SDK support refs and views on publish

rynowak:
These are default behaviors for precompilation.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
@@ -31,6 +31,22 @@
     
     <!-- File name (without extension) of the assembly produced by Razor -->
     <RazorTargetName Condition="'$(RazorTargetName)'==''">$(TargetName).PrecompiledViews</RazorTargetName>
+
+    <!-- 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -31,6 +31,22 @@
...
     <!-- File name (without extension) of the assembly produced by Razor -->
     <RazorTargetName Condition="'$(RazorTargetName)'==''">$(TargetName).PrecompiledViews</RazorTargetName>
+
+    <!-- 
+      Set to true to copy RazorGenerate items to the publish directory (.cshtml) files.
+
+      Typically Razor files are not needed for a published application if they participate in compilation at build-time
+      or publish-time. By default, the Razor SDK will suppress the copying of RazorGenerate items to the publish directory.
+    -->
+    <CopyRazorGenerateFilesToPublishDirectory Condition="'$(CopyRazorGenerateFilesToPublishDirectory)'==''">false</CopyRazorGenerateFilesToPublishDirectory>
```
pranavkm: Shouldn't these go in the props file?

rynowak: I'm not actually sure.

I personally like putting these in the targets because then you can tell whether the user configured it or not.

PR #1934, opened 2018-01-15
Split up Razor SDK logic

rynowak:
Splitting the fragile parts of code generation into its own file. We're
possibly separating the Razor SDK into a part that ships in the dotnet
SDK and part that ships in a nupkg with the compiler.

This means defining more of a fixed lifecycle in the main file, and
treating our actual work as more like extensibility.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
@@ -31,7 +31,7 @@
     AfterTargets="ResolveRazorGenerateInputs"
     Condition="'@(RazorGenerate)'!=''">
 
-    <Hash ItemsToHash="@(RazorGenerate)" Condition="'@(RazorGenerate)'!=''">
+    <Hash ItemsToHash="@(RazorGenerate)">
       <Output TaskParameter="HashResult" PropertyName="_RazorGenerateInputsHash" />
     </Hash>
 
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
@@ -1,0 +1,126 @@
+<Project>
+  <!--
+    Targets used for Razor SDK code generation. Support for the RazorCoreGenerate target.
+  -->
+
+  <!-- 
+    Consider these properties to be private to this targets file. The main Razor SDK should define all of the properties
+    that we use to pass data back and forth. 
+  -->
+  <PropertyGroup>
+    <!-- Used for tag helper discovery -->
+    <_RazorTagHelperInputCache>$(IntermediateOutputPath)$(TargetName).TagHelpers.input.cache</_RazorTagHelperInputCache>
+    <_RazorTagHelperOutputCache>$(IntermediateOutputPath)$(TargetName).TagHelpers.output.cache</_RazorTagHelperOutputCache>
+    
+    <!-- Used to locate our tools -->
+    <_RazorGenerateToolAssembly>$(_RazorMSBuildRoot)tools\Microsoft.AspNetCore.Razor.GenerateTool.dll</_RazorGenerateToolAssembly>
+    <_RazorTagHelperToolAssembly>$(_RazorMSBuildRoot)tools\Microsoft.AspNetCore.Razor.TagHelperTool.dll</_RazorTagHelperToolAssembly>
+
+    <!-- Used to hash file inputs for RazorGenerate -->
+    <_RazorGenerateInputsHash></_RazorGenerateInputsHash>
+    <_RazorGenerateInputsHashFile>$(IntermediateOutputPath)$(MSBuildProjectName).RazorCoreGenerate.cache</_RazorGenerateInputsHashFile>
+  </PropertyGroup>
+
+  <!--
+    This target will only be called when we have some .cshtml files that are going to participate in code generation.
+
+    This is part of the chain of targets that are called once we've actually committed to generating code.
+  -->
+  <Target 
+    Name="_HashRazorGenerateInputs" 
+    AfterTargets="ResolveRazorGenerateInputs"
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.Compilation.targets
```diff
@@ -47,7 +47,6 @@
           Outputs="@(RazorIntermediateAssembly);
                    @(_RazorDebugSymbolsIntermediatePath);
                    $(NonExistentFile)"
-          DependsOnTargets="RazorCoreGenerate"
```
rynowak: I did a little general cleanup here. We don't really plan for a target like `RazorCoreCompile` to be directly invoked, so I'm not trying to make it look like it's supported.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -3,18 +3,6 @@
...
-    Razor defines two primary targets:
-      'RazorGenerate' - which updates generated code
-      'RazorCompile' - compiles an assembly from generated code
-
-    Use these properties to attach behavior to the corresponding target.
-  -->
-  <PropertyGroup>
-    <RazorGenerateDependsOn>RazorResolveGenerateInputs;RazorCoreGenerate</RazorGenerateDependsOn>
-    <RazorCompileDependsOn>RazorGenerate;RazorCoreCompile</RazorCompileDependsOn>
-  </PropertyGroup>
```
rynowak: Moved these into the .targets. That's the style that the .net targets follow. With Before/After targets nowadays, this isn't super meaningful for anyone but us anyways.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -81,130 +92,71 @@
...
+      </RazorGenerate>
     </ItemGroup>
   </Target>
 
   <!-- 
     Gathers input assemblies for Tag Helper discovery and compilation. Add items to @(ReferencePath)
   -->
   <Target
-    Name="_RazorResolveReferences"
+    Name="ResolveAssemblyReferenceRazorGenerateInputs"
```
rynowak: This target is SDK-like so it belongs here. We won't have much reason to change this if as we evolve Razor.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -81,130 +92,71 @@
...
+    <GetCopyToOutputDirectoryItemsDependsOn>$(GetCopyToOutputDirectoryItemsDependsOn);_RazorGetCopyToOutputDirectoryItems</GetCopyToOutputDirectoryItemsDependsOn>
+  </PropertyGroup>
 
+  <!--
+    Gathers input source files for code generation. This is a separate target so that we can avoid
+    lots of work when there are no inputs for code generation.
+
+    Notice there are plenty of Condition="'@(RazorGenerate)'!=''" to avoid running more expensive targets.
+  -->
+  <Target Name="ResolveRazorGenerateInputs">
```
rynowak: This target is SDK-like so it belongs here. We won't have much reason to change this if as we evolve Razor.

rynowak:
@pranavkm ping

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -0,0 +1,126 @@
...
+    This target will only be called when we have some .cshtml files that are going to participate in code generation.
+
+    This is part of the chain of targets that are called once we've actually committed to generating code.
+  -->
+  <Target 
+    Name="_HashRazorGenerateInputs" 
+    AfterTargets="ResolveRazorGenerateInputs"
+    Condition="'@(RazorGenerate)'!=''">
+
+    <Hash ItemsToHash="@(RazorGenerate)" Condition="'@(RazorGenerate)'!=''">
```
pranavkm: You can remove the `Condition` here since it's already being checked in the target.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -81,130 +92,71 @@
...
+  <PropertyGroup Condition="'$(RazorCompileOnBuild)'=='true'">
+    <PrepareForRunDependsOn>RazorCompile;$(PrepareForRunDependsOn);_RazorCopyFilesToOutputDirectory</PrepareForRunDependsOn>
+    <GetCopyToOutputDirectoryItemsDependsOn>$(GetCopyToOutputDirectoryItemsDependsOn);_RazorGetCopyToOutputDirectoryItems</GetCopyToOutputDirectoryItemsDependsOn>
+  </PropertyGroup>
 
+  <!--
+    Gathers input source files for code generation. This is a separate target so that we can avoid
+    lots of work when there are no inputs for code generation.
+
+    Notice there are plenty of Condition="'@(RazorGenerate)'!=''" to avoid running more expensive targets.
```
pranavkm: Comment seems out of place here.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
pranavkm: `_ResolveAssemblyReferenceRazorGenerateInputs`?  A user could attach to `PrepareForRazor` if they want to tweak the inputs

rynowak: I found this kind of detailed/specific extensibility very nice to use when I was figuring out the publish stuff. I also think this is unlikely to change significantly.

PR #1935, opened 2018-01-16
Attempt to fix Razor CI flakiness

rynowak:
We're seeing some test failures on the CI where the build has no output.
Sure enough, using Process.Exited is trap. It doesn't guarantee that all
of the output has been written when it's trigger.

This is a different approach that shouldn't suffer from the same
problem.


rynowak:
This is failing on appveyor and it seems to reproduce

rynowak:
oh NVM I am bad at refresh

rynowak:
Travis is having issues R/N so I'm going to go ahead and merge

PR #1936, opened 2018-01-16
Bring build infrastructure bits over to rel/vs15.6

NTaylorMullen:
- The rel/vs15.6 branch couldn't build on its own so I decided to look at the build infrastructure commits that were missing (compared to dev) and brought those over. Did some assembly reference checking in IL spy to ensure that we didn't introduce any unintended binary compatibility changes.


build/MPack.targets
```diff
@@ -1,12 +1,37 @@
 ﻿<Project>
   <PropertyGroup>
-    <PackageDependsOn Condition="'$(OS)'=='Windows_NT'">$(PackageDependsOn);GenerateMPack</PackageDependsOn>
```
NTaylorMullen: We're purposefully turning off mpack generation for this branch because it's for VS windows only and doesn't have all the product bits required to ship an mpack.

build/MPack.targets
```diff
@@ -1,12 +1,37 @@
...
     <AddinName>Microsoft.VisualStudio.Mac.RazorAddin</AddinName>
     <AddinDirectory>$(RepositoryRoot)tooling\$(AddinName)\</AddinDirectory>
+    <MPackArtifactCategory>shipoob</MPackArtifactCategory>
```
natemcmaster: noship?

build/MPack.targets
natemcmaster: Consider adding a boolean property like `BuildMPack` that defaults to `false` in this branch instead of deleting code. This makes merging between branches easier. Also, you can share this setting in multiple places.
 i.e. you also need to disable `GetMPackArtifactInfo`

build/MPack.targets
NTaylorMullen: Eh, gonna nuke the file entirely.

natemcmaster: Why not just nuke all of MPack code then?

Directory.Build.props
```diff
@@ -1,6 +1,7 @@
 ﻿<Project>
```
pranavkm: Does korebuild.json \ korebuild.lock.txt need to be updated too?

build/MPack.targets
NTaylorMullen: > Why not just nuke all of MPack code then?

That's what I meant 😄

build/MPack.targets
```diff
@@ -53,11 +65,11 @@
...
-    <Copy SourceFiles="$(LanguageServiceOutputPath)\%(LanguageServiceAssemblies.Identity)" DestinationFolder="$(MPackSourcesDir)" />
-    <Copy SourceFiles="$(AddinOutputPath)$(AddinName).dll" DestinationFolder="$(MPackSourcesDir)" />
+    <Copy SourceFiles="$(LanguageServiceOutputPath)\%(LanguageServiceAssemblies.Identity)" DestinationFolder="$(MPackIntermediateOutputPath)" />
+    <Copy SourceFiles="$(AddinOutputPath)$(AddinName).dll" DestinationFolder="$(MPackIntermediateOutputPath)" />
 
     <!-- We cannot use the ZipArchive task due to how it functions in CoreCLR. The archive it generates is unreadable by Visual Studio for Mac. -->
-    <Exec Command="powershell.exe -NonInteractive -command &quot;&amp; { Add-Type -AssemblyName System.IO.Compression.FileSystem; [System.IO.Compression.ZipFile]::CreateFromDirectory('$(MPackSourcesDir)', '$(MPackOutputPath)') } &quot;" />
+    <Exec Command="powershell.exe -NonInteractive -command &quot;&amp; { Add-Type -AssemblyName System.IO.Compression.FileSystem; [System.IO.Compression.ZipFile]::CreateFromDirectory('$(MPackIntermediateOutputPath)', '$(MPackOutputPath)') } &quot;" />
```
pranavkm: Unrelated, weren't you successfully able to use ZipArchive in Coherence-Signed?

build/MPack.targets
natemcmaster: seems like a potentially painful merging strategy, but if it is what your team wants, :shipit:

build/MPack.targets
NTaylorMullen: Yup, this is something we should consider moving to in the dev branch 😄

natemcmaster: ZipStream was computing CRC32 incorrectly for a while. Should be fixed now https://github.com/dotnet/corefx/issues/25803

NTaylorMullen: https://github.com/aspnet/Razor/issues/1937

Directory.Build.props
NTaylorMullen: korebuild-lock.txt was updated and I don't believe korebuild.json needs to be.

Issue #1927, opened 2018-01-11
Block.Length is expensive at design time.


NTaylorMullen:
Lots of perf issues regarding `Block.get_Length` in VS.

Opening MSN.cshtml in VS:
![image](https://user-images.githubusercontent.com/2008729/34846494-366be9da-f6cd-11e7-937a-93e403ea3b42.png)

rynowak:
What's the priority of this? Should we be looking at fixing this for 15.6?

NTaylorMullen:
> What's the priority of this? Should we be looking at fixing this for 15.6?

Figuring that out now in another email thread.

NTaylorMullen:
92111ea515504d80fedadeb23aeb748b6a946ca5

PR #1938, opened 2018-01-17
Make Block and Span cache Length.

NTaylorMullen:
- Part of caching length required the `Span`'s `ReplaceWith` method to propagate its changes to its parents so that they can all invalidate their length caches.
- Added Span tests to validate the interaction of caching.
- There is no benchmarking infrastructure on this branch.

#1927

From 4.2s for MSN.cshtml in an isolated test to 0s 😉 

## Before
image

## After
image

/cc @ToddGrun

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
@@ -11,6 +11,8 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal class Block : SyntaxTreeNode
     {
+        private int? _length;
+
         public Block(BlockBuilder source)
             : this(source.Type, source.Children, source.ChunkGenerator)
         {
@@ -58,7 +60,25 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             }
         }
 
-        public override int Length => Children.Sum(child => child.Length);
+        public override int Length
+        {
+            get
+            {
+                if (_length == null)
+                {
+                    var length = 0;
+                    for (var i = 0; i < Children.Count; i++)
+                    {
+                        length += Children[i].Length;
+                    }
+
+                    _length = length;
+                }
+
+                return _length.Value;
+            }
+        }
+
 
         public virtual IEnumerable<Span> Flatten()
         {
@@ -212,6 +232,14 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
         public override void Accept(ParserVisitor visitor)
         {
             visitor.VisitBlock(this);
+        }
+
+        internal void ChildChanged()
+        {
+            // A node in our graph has changed. We'll need to recompute our length the next time we're asked for it.
+            _length = null;
+
+            Parent?.ChildChanged();
         }
 
         private class EquivalenceComparer : IEqualityComparer<SyntaxTreeNode>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
```diff
@@ -214,6 +235,12 @@ public override void Accept(ParserVisitor visitor)
             visitor.VisitBlock(this);
         }
 
+        internal void ChildChanged()
+        {
+            // A span that's in our graph has changed. We'll need to recompute our length the next time we're asked for it.
+            _length = -1;
```
ToddGrun: This doesn't need to propogate up any farther?

src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal class Block : SyntaxTreeNode
     {
+        private int _length = -1;
```
rynowak: use `int?` it does the same thing but it's much more semantically meaningful

src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
ToddGrun: Nevermind, you do it in the span class. I would have thought this block would have been responsible for notifying it's parent.

NTaylorMullen: > responsible for notifying it's parent.

Hmmm, ya gonna go with having the block notify its parent. I like that better 😄

src/Microsoft.AspNetCore.Razor.Language/Legacy/Block.cs
ToddGrun: Is memory consumption a concern?

NTaylorMullen: Not that we've seen.

src/Microsoft.AspNetCore.Razor.Language/Legacy/Span.cs
```diff
@@ -32,7 +33,32 @@ public Span(SpanBuilder builder)
...
+                        length += Symbols[i].Content.Length;
+                    }
+                }
+                else
+                {
+                    length = _content.Length;
+                }
+
+                _length = length;
+                return length;
```
rynowak: This is kinda sloppy considering what this actually does.

This kinds of code is more more cleaner if written like:
```
if (_length == null)
{
   _length = ....
}

return _length;
```

If necessary turn your loop into a method.

rynowak: Also that will make it much more inlinable. Consider also sealing `Span`.

rynowak:
Regarding the perf... can we take a look at the remaining 1sec of time tomorrow? I realize that this just got 10x faster, but that turns a 5000ms delay into a 500ms delay.. That's still not amazing.

NTaylorMullen:
> Regarding the perf... can we take a look at the remaining 1sec of time tomorrow? I realize that this just got 10x faster, but that turns a 5000ms delay into a 500ms delay.. That's still not amazing.

Maybe I'm a perf result noob/read the perf results incorrectly but that 1s I thought was the delay between me hitting the "start profiling button" and entering a newline to begin actual calculations 😄 

I figured "OwnTime" was the important metric meaning the test took 6ms. I'm probably wrong though >_<

rynowak:
>I figured "OwnTime" was the important metric meaning the test took 6ms. I'm probably wrong though

Oh yeah, I read the table wrong too. It went from 10711ms to 49ms (Fast Beans)

PR #1939, opened 2018-01-17
Make RazorCompileOnPublish the default

rynowak:


rynowak:
We will need to make a change in the WebSDK related to this to make precompilation non-default there.

pranavkm:
> We will need to make a change in the WebSDK related to this to make precompilation non-default there.

Make it non-default conditionally?

pranavkm:
Tests are failing btw. Could you check what's up with that?

rynowak:
> Make it non-default conditionally?

Not sure I know what that means. I'm meeting with those people this morning to discuss.

pranavkm:
The issue as I understand it is that we want users targeting Asp.Net Core 2.0.0 to be able to continue getting precompilation when they use the 2.1.0 SDK.

rynowak:
Yes that's correct 😆 - now I get what you mean. We're meeting with those peeps this morning.

Issue #1921, opened 2018-01-09
Meta: Add Razor build server


rynowak:
Checklist:

- [x] Get tag helper discovery and code generation working end to end (off by default) (https://github.com/aspnet/Razor/issues/1943)
- [x] Resolve how many/what tools we have (https://github.com/aspnet/Razor/issues/1953)
- [x] Server timeout after a certain time
- [x] Server should support different Razor toolset versions at the same time (https://github.com/aspnet/Razor/pull/2058)
- [x] Include more data as part of the inputs to the server
- [x] Logging cleanup (https://github.com/aspnet/Razor/issues/1954)
- [x] Integrate in to the build
- [x] Turn it on by default when it is in good shape (https://github.com/aspnet/Razor/pull/2131)
- [x] Tests (https://github.com/aspnet/Razor/issues/1958)
- [x] Get automated perf measurements working

ajaybhargavb:
Perf data when the server is already running,
**Target Performance Summary:**
        0 ms  PrepareResourceNames                       1 calls
        0 ms  BeforeResGen                               1 calls
        0 ms  AfterBuild                                 1 calls
        0 ms  ResolvePackageDependenciesForBuild         1 calls
        0 ms  SetWin32ManifestProperties                 1 calls
        0 ms  AfterRebuild                               1 calls
        0 ms  BeforeBuild                                1 calls
        0 ms  Clean                                      1 calls
        0 ms  BeforeCompile                              1 calls
        0 ms  PrepareForRun                              1 calls
        0 ms  AfterClean                                 1 calls
        0 ms  PrepareProjectReferences                   1 calls
        0 ms  PrepareResources                           1 calls
        0 ms  CoreBuild                                  1 calls
        0 ms  GenerateAssemblyInfo                       1 calls
        0 ms  BeforeClean                                1 calls
        0 ms  GetFrameworkPaths                          1 calls
        0 ms  GetReferenceAssemblyPaths                  1 calls
        0 ms  RazorGenerate                              1 calls
        0 ms  AfterResolveReferences                     1 calls
        0 ms  BeforeRebuild                              1 calls
        0 ms  ResGen                                     1 calls
        0 ms  CreateCustomManifestResourceNames          1 calls
        0 ms  Compile                                    1 calls
        0 ms  _CopySourceItemsToOutputDirectory          1 calls
        0 ms  RazorCompile                               1 calls
        0 ms  CreateSatelliteAssemblies                  1 calls
[...]

PR #1941, opened 2018-01-17
Build server working end to end

ajaybhargavb:
#1943 
Meta issue - #1921 

The build server now takes care of both discovering tag helpers and generating Razor CSharp documents. This saves about 1.8s of build time when the server is warm. (Perf data: https://github.com/aspnet/Razor/issues/1921#issuecomment-358096928)

- Added a switch to enable/disable server compilation
- Limited PipeName length to workaround limitations of socket length in Unix systems
- Some more general cleanup

Things that need to be done in a future PR: (Checklist here https://github.com/aspnet/Razor/issues/1921#issue-286973192)
- Currently there is code duplication between `GenerateTool`, `TagHelperTool` and `Razor.Tools`. Need to unify them
- Figure out when/how server shutdown should happen
- More cleanup around `BuildProtocol`

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.TagHelperTool/Application.cs
```diff
@@ -20,8 +20,9 @@ public Application()
 
             HelpOption("-?|-h|--help");
 
-            TagHelperManifest = Option("-o", "output file", CommandOptionType.SingleValue);
             Assemblies = Argument("assemblies", "assemblies to search for tag helpers", multipleValues: true);
+            TagHelperManifest = Option("-o", "output file", CommandOptionType.SingleValue);
+            ProjectRoot = Option("-p", "project root directory", CommandOptionType.SingleValue);
```
ajaybhargavb: I understand we didn't want to change this. But this looked necessary as we don't always want to rely on the CurrentDirectory to be the our working directory.

rynowak: Agreed. We can't rely on `Environment.CurrentDirectory` unless we know we're running as a CLI

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+
+        protected virtual bool TryExecuteOnServer(string pathToTool, string responseFileCommands, string commandLineCommands, out int result)
+        {
+            CompilerServerLogger.Log("Server execution started.");
+            using (_razorServerCts = new CancellationTokenSource())
+            {
+                CompilerServerLogger.Log($"CommandLine = '{commandLineCommands}'");
+                CompilerServerLogger.Log($"BuildResponseFile = '{responseFileCommands}'");
+
+                var clientDir = Path.GetDirectoryName(ServerAssembly);
```
NTaylorMullen: Add a comment about this expectation that the client bits exist in the server bits (that we ship both together)

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+
+                var workingDir = CurrentDirectoryToUse();
+                var buildPaths = new BuildPathsAlt(
+                    clientDir: clientDir,
```
NTaylorMullen: no need for named param

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+                    // MSBuild doesn't need the .NET SDK directory
+                    sdkDir: null,
+                    workingDir: workingDir,
+                    tempDir: BuildServerConnection.GetTempPath(workingDir));
```
NTaylorMullen: pull this out into a variable for debugability.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+
+                var responseTask = BuildServerConnection.RunServerCompilation(
+                    Command,
+                    GetArguments(responseFileCommands),
+                    buildPaths,
+                    keepAlive: null,
```
NTaylorMullen: Having keepAlive = null here is that just saying that this isn't a keep alive request?

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+
+                responseTask.Wait(_razorServerCts.Token);
+
+                var response = responseTask.Result;
+                if (response.Type == BuildResponse.ResponseType.Completed &&
+                    response is CompletedBuildResponse completedResponse)
+                {
+                    CompilerServerLogger.Log("Server execution completed.");
+
+                    result = completedResponse.ReturnCode;
```
NTaylorMullen: if result is non-zero her we may want to log something. Otherwise we'll be failing to invoke the server without an error.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+        private string LibDirectoryToUse()
+        {
+            // First check the real environment.
+            var libDirectory = Environment.GetEnvironmentVariable("LIB");
+
+            // Now go through additional environment variables.
+            var additionalVariables = EnvironmentVariables;
+            if (additionalVariables != null)
+            {
+                foreach (var variable in EnvironmentVariables)
```
NTaylorMullen: `LastOrDefault(...) ?? Environment.GetEnvironmentVariable("LIB")`

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+
+        private struct SourceItem
+        {
+            public SourceItem(string fullPath, string viewEnginePath)
+            {
+                FullPath = fullPath;
+                ViewEnginePath = viewEnginePath;
+            }
+
+            public string FullPath { get; }
```
NTaylorMullen: You never use this. Can represent this entire object as a `string viewEnginePath` 😉

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        private OutputItem[] GenerateCode(RazorTemplateEngine templateEngine, IReadOnlyList<SourceItem> sources)
+        {
+            var outputs = new OutputItem[sources.Count];
+            Parallel.For(0, outputs.Length, new ParallelOptions() { MaxDegreeOfParallelism = 4 }, i =>
+            {
+                var source = sources[i];
+
+                var csharpDocument = templateEngine.GenerateCode(source.ViewEnginePath);
```
NTaylorMullen: If only this engine returned a code document you wouldn't need an output item at all. I'll make that possible with the project engine 😄

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+            var tagHelpers = GetTagHelpers(tagHelperManifest);
+
+            var engine = RazorEngine.Create(b =>
+            {
+                RazorExtensions.Register(b);
+
+                b.Features.Add(new StaticTagHelperFeature() { TagHelpers = tagHelpers, });
+            });
+
+            var templateEngine = new MvcRazorTemplateEngine(engine, RazorProject.Create(projectDirectory));
```
NTaylorMullen: We'll eventually need to abstract this to support multiple versions of Razor.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+                if (result.CSharpDocument.Diagnostics.Count > 0)
+                {
+                    success = false;
+                    foreach (var error in result.CSharpDocument.Diagnostics)
+                    {
+                        Console.Error.WriteLine(error.ToString());
+                    }
+                }
+
+                var outputFilePath = Path.Combine(outputDirectory, Path.ChangeExtension(result.ViewFileInfo.ViewEnginePath.Substring(1), ".cs"));
```
NTaylorMullen: lol make these into some variables bro, this is debug hell waiting to happen 😉

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+            if (!File.Exists(tagHelperManifest))
+            {
+                return Array.Empty<TagHelperDescriptor>();
+            }
+
+            using (var stream = File.OpenRead(tagHelperManifest))
+            {
+                var reader = new JsonTextReader(new StreamReader(stream));
+
+                var serializer = new JsonSerializer();
```
NTaylorMullen: We've done this so many times we might want to consider creating a `RazorJsonTextSerializer` that does all this.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
-        protected virtual bool TryExecuteOnServer(out bool result)
+        public override void Cancel()
         {
-            result = false;
+            base.Cancel();
+
+            _razorServerCts?.Cancel();
+        }
+
+        protected virtual bool TryExecuteOnServer(string pathToTool, string responseFileCommands, string commandLineCommands, out int result)
```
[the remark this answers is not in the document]
ajaybhargavb: `UseServer` applies to all commands. But with this, any command can just override this and return false if it doesn't want to be run on the server.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
ajaybhargavb: Added to the checklist 👍

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+        {
+            CompilerServerLogger.Log("Server execution started.");
```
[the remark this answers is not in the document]
ajaybhargavb: I'm not really sure what you mean here.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -62,12 +76,19 @@ public override bool Execute()
                 }
             }
 
-            if (TryExecuteOnServer(out var result))
+            return base.Execute();
+        }
+
+        protected override int ExecuteTool(string pathToTool, string responseFileCommands, string commandLineCommands)
```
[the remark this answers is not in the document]
ajaybhargavb: I assumed MsBuild takes care of providing errors when a required argument or option is missing. Otherwise we just want to pass it through to the compiler/tool.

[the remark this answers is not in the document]
rynowak: image

src/Microsoft.AspNetCore.Razor.TagHelperTool/RunCommand.cs
```diff
@@ -32,12 +32,15 @@ private int Execute(Application application)
             }
 
             return ExecuteCore(
+                projectDirectory: application.ProjectRoot.Value() ?? Environment.CurrentDirectory,
```
rynowak: Do this in validate arguments instead. You can manually set the arguments' value using `.Values.Add(...)`

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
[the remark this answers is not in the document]
rynowak: I don't think we need any of this logging here. We want this to use MSBuild's logging, and we should address that as part of the item we have for logging.

delete it or leave it for now is fine, but I don't want to block on it.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -82,10 +103,102 @@ protected override void LogToolCommand(string message)
...
+            {
+                workingDirectory = Directory.GetCurrentDirectory();
+            }
+            return workingDirectory;
+        }
+
+        /// <summary>
+        /// Get the "LIB" environment variable, or NULL if none.
+        /// </summary>
+        private string LibDirectoryToUse()
```
rynowak: I don't think we need any of this do we?

src/Microsoft.AspNetCore.Razor.Tools/DiscoverCommand.cs
```diff
@@ -51,13 +54,16 @@ protected override bool ValidateArguments()
         protected override Task<int> ExecuteCoreAsync()
         {
             var result = ExecuteCore(
+                projectDirectory: ProjectDirectory.Value() ?? Environment.CurrentDirectory,
```
rynowak: use `ValidateArguments` for this

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -0,0 +1,204 @@
...
+                sources: Sources.Values.ToArray());
+
+            return Task.FromResult(result);
+        }
+
+        protected override bool ValidateArguments()
+        {
+            if (string.IsNullOrEmpty(OutputDirectory.Value()))
+            {
+                Error.WriteLine($"{OutputDirectory.ValueName} not specified.");
```
rynowak: BTW when we run on the server, we need to replace these `TextWriters` with one hooked up to the server logger.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
rynowak: Yep. This data will need to flow through the SDK

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
rynowak: Or we could make them `ISerializable` and use `BinarySerializer`. I don't super want to use JSON.NET here

rynowak: We could also consider making these JSON serializers into a shared-source project so that it's easier to maintain

src/Microsoft.AspNetCore.Razor.Tools/Roslyn/BuildProtocol.cs
```diff
@@ -506,10 +506,11 @@ internal sealed class RejectedBuildResponse : BuildResponse
...
     // The id numbers below are just random. It's useful to use id numbers
     // that won't occur accidentally for debugging.
-    internal enum RequestLanguage
+    internal enum RequestCommand
     {
-        CSharpCompile = 0x44532521,
-        VisualBasicCompile = 0x44532522,
+        None = 0x44532621,
+        RazorTagHelper = 0x44532622,
+        RazorGenerate = 0x44532623,
```
rynowak: Nice

src/Microsoft.AspNetCore.Razor.Tools/Roslyn/CommandLineUtilities.cs
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Text;
+
+namespace Roslyn.Utilities
```
rynowak: It might be good to add a note here that we basically copied this verbatim. I don't think there's much value in us trying to refactor or change this code.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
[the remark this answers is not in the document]
ajaybhargavb: Leaving this here for now as it is nice to have for debugging.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
ajaybhargavb: Yup. https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/RequestDispatcher.cs#L295

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
ajaybhargavb: This was all copy paste. I am going to address all the comments in `GenerateCommand` as part of https://github.com/aspnet/Razor/issues/1953

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
ajaybhargavb: Will address as part of https://github.com/aspnet/Razor/issues/1953

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -62,12 +76,19 @@ public override bool Execute()
                 }
             }
 
-            if (TryExecuteOnServer(out var result))
+            return base.Execute();
+        }
+
+        protected override int ExecuteTool(string pathToTool, string responseFileCommands, string commandLineCommands)
+        {
+            if (UseServer &&
```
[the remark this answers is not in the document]
ajaybhargavb: Making `ServerAssembly` [Required]

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
ajaybhargavb: Will address in https://github.com/aspnet/Razor/issues/1954.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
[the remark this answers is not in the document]
ajaybhargavb: Will address in https://github.com/aspnet/Razor/issues/1954

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
ajaybhargavb: Will address as part of #1953

ajaybhargavb:
🆙 📅

PR #1944, opened 2018-01-18
Add a property so the web sdk can know that the Razor compiler is ava……ilable.

rynowak:


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -13,6 +13,12 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <!--
+      Used by the Web SDK to know if the Razor SDK is present. This needs to live in a nuget package (not in the SDK)
+      so that it only shows up in supported versions.
+    -->
+    <IsRazorCompilerReferenced>true<IsRazorCompilerReferenced>
```
rynowak: @vijayrkn - I went with a slightly different name. The nuget package is actually the thing that's significant here, not the SDK itself.

vijayrkn: will update my PR. Thanks!

PR #1947, opened 2018-01-19
Prepare for Web SDK changes

rynowak:
The Web SDK is going to set a new msbuild property to tell us that the
Razor SDK should be active. This hasn't been integrated into our build
system yet, so I'm temporarily hacking it until we get that change. At
that time I'll remove the special cases in these projects.


rynowak:
FYI only - I'm going to merge this once travis passes so it can flow through the build. The next set of changes will be focused on compatibility between Razor SDK and MvcPrecompilation, so I want to get this checked in first we can deal with the fallout.

Issue #1828, opened 2017-12-14
Meta: Replace MvcTemplateEngine


rynowak:
Living list of things a part of this issue.

- [x] Add contracts for RazorProjectEngine creation
- [ ] Make MVC registrations work with RazorProjectEngine
- [x] Lift RazorEngineBuilderExtension concepts onto RazorProjectEngineBuilder

rynowak:
Related to built-time-razor, because we're hard coding more MVC-isms in yet another place.

NTaylorMullen:
- [ ] Add contracts for `RazorProjectEngine` creation
- [ ] Make MVC registrations work with `RazorProjectEngine`
- [ ] Lift `RazorEngineBuilderExtension` concepts onto `RazorProjectEngineBuilder`

NTaylorMullen:
59f2cf8e66daec67c90c9b4294d2cf5ba7f7230a

PR #1949, opened 2018-01-19
Add key contracts and builders for RazorProjectEngine.

NTaylorMullen:
- These contracts introduce a new `RazorProjectEngine` concept which allows for users to configure 1 entity that's responsible for the RazorEngine and project.
- The `RazorProjectEngineBuilder` has a collection of features that are dispersed on the created `RazorEngine` and the `RazorProjectEngine`.
- Included a complete implementation of `RazorProjectEngine` it introduces the extension points for the project engine. The primary one includes the `IRazorImportFeature`. Allowing multiple imports to take part in import discovery will come later.
- Included a complete project engine builder implementation.

My goal for this design was to come up with a pattern that had 1 entry point to configure both the project engine and Razor engine resulting in a user (or framework) never having to create the RazorEngine manually. This allows for a framework to fully customize the Razor code generation process from finding imports to changing how your Razor turns into C#.

How this will play out in practice is you'll do:

```C#
var mvcEngine = RazorProjectEngine.Create(projectFileSystem, builder =>
{
    // Register all things MVC
    RazorExtensions.Register(builder);
});

var blazorEngine = RazorProjectEngine.Create(projectFileSystem, builder =>
{
    // Register all things Blazor
    BlazorExtensions.Register(builder);
});
```

Big take away is that `RazorProjectEngine` concepts and `RazorEngine` concepts are ambiguous in the `RazorProjectEngineBuilder`; that builder has the knowledge to split the features among the two engine/project engine properly.

#1828 

/cc @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class DefaultRazorProjectEngine : RazorProjectEngine
+    {
+        public DefaultRazorProjectEngine(
+            RazorEngine engine,
+            RazorProjectFileSystem fileSystem,
+            IReadOnlyList<IRazorProjectEngineFeature> features)
+        {
+            if (engine == null)
+            {
+                throw new ArgumentNullException(nameof(engine));
+            }
+
+            if (fileSystem == null)
+            {
+                throw new ArgumentNullException(nameof(fileSystem));
+            }
+
+            if (features == null)
+            {
+                throw new ArgumentNullException(nameof(features));
+            }
+
+            Engine = engine;
+            FileSystem = fileSystem;
+            Features = features;
+
+            for (var i = 0; i < features.Count; i++)
+            {
+                features[i].ProjectEngine = this;
+            }
+        }
+
+        public override RazorProjectFileSystem FileSystem { get; }
+
+        public override RazorEngine Engine { get; }
+
+        public override IReadOnlyList<IRazorProjectEngineFeature> Features { get; }
+
+        public override RazorCodeDocument Process(string filePath)
+        {
+            if (string.IsNullOrEmpty(filePath))
+            {
+                throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(filePath));
+            }
+
+            var projectItem = FileSystem.GetItem(filePath);
+            var sourceDocument = RazorSourceDocument.ReadFrom(projectItem);
+            var codeDocument = Process(sourceDocument);
+
+            return codeDocument;
+        }
+
+        public override RazorCodeDocument Process(RazorSourceDocument sourceDocument)
+        {
+            if (sourceDocument == null)
+            {
+                throw new ArgumentNullException(nameof(sourceDocument));
+            }
+
+            var importFeature = GetRequiredFeature<IRazorImportFeature>();
+            var imports = importFeature.GetImports(sourceDocument.FilePath);
+
+            var codeDocument = RazorCodeDocument.Create(sourceDocument, imports);
+
+            Engine.Process(codeDocument);
+
+            return codeDocument;
+        }
+
+        private TFeature GetRequiredFeature<TFeature>() where TFeature : IRazorProjectEngineFeature
+        {
+            var feature = Features.OfType<TFeature>().FirstOrDefault();
+            if (feature == null)
+            {
+                throw new InvalidOperationException(
+                    Resources.FormatRazorProjectEngineMissingFeatureDependency(
+                        typeof(RazorProjectEngine).FullName,
+                        typeof(TFeature).FullName));
+            }
+
+            return feature;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.Language/IRazorImportFeature.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/IRazorImportFeature.cs
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
+
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    public interface IRazorEngineFeature : IRazorFeature
+    public interface IRazorImportFeature : IRazorProjectEngineFeature
     {
-        RazorEngine Engine { get; set; }
+        IReadOnlyList<RazorSourceDocument> GetImports(string sourceFilePath);
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.Language.Extensions;
 
@@ -19,7 +21,7 @@ namespace Microsoft.AspNetCore.Razor.Language
         {
             var builder = new DefaultRazorEngineBuilder(designTime: false);
             AddDefaults(builder);
-            AddRuntimeDefaults(builder);
+            AddDefaultRuntimeFeatures(builder.Features);
             configure?.Invoke(builder);
             return builder.Build();
         }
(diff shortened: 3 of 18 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            AddDesignTimeDefaults(builder);
+            configure?.Invoke(builder);
+
+            return builder.Build();
+        }
+
+        private static void AddDesignTimeDefaults(RazorProjectEngineBuilder builder)
+        {
+            var defaultEngine = RazorEngine.CreateDesignTime();
```
NTaylorMullen: Yes, we're creating a dummy engine. This is valuable because we support creating `RazorEngine`s independently of a project engine; otherwise we'd have to add a new project engine specific call path to the RazorEngine creation code.

Ultimately we throw away the engine shell and take all of its features/phases.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectFileSystem.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public abstract class RazorProjectFileSystem : RazorProject
```
NTaylorMullen: Thought it was fine if we still have all the methods return `RazorProjectitem` instead of `RazorProjectFileSystemItem`. If we went with the later idea we'd have to re-implement all of RazorProject. It wouldn't be too awful but I felt that `RazorProjectItem` is sufficient enough.

NTaylorMullen:
Once this take of the design is signed off on i'll write tests.

src/Microsoft.AspNetCore.Razor.Language/IRazorImportFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public interface IRazorImportFeature : IRazorProjectEngineFeature
```
NTaylorMullen: In a future PR i'll introduce `IRazorImportExpander` to enable multiple import expanders to take part in import discovery. That will be an implementation detail of this.

src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
```diff
@@ -45,6 +45,13 @@ public static RazorEngine CreateEmpty(Action<IRazorEngineBuilder> configure)
             return builder.Build();
         }
 
+        public static RazorEngine CreateDesignTimeEmpty(Action<IRazorEngineBuilder> configure)
```
NTaylorMullen: We should have always had this given we have a runtime variant.

src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
```diff
@@ -40,8 +40,25 @@ public static RazorEngine CreateDesignTime(Action<IRazorEngineBuilder> configure
...
+                throw new ArgumentNullException(nameof(configure));
+            }
+
             var builder = new DefaultRazorEngineBuilder(designTime: false);
-            configure?.Invoke(builder);
+            configure.Invoke(builder);
+            return builder.Build();
+        }
+
+        public static RazorEngine CreateDesignTimeEmpty(Action<IRazorEngineBuilder> configure)
```
NTaylorMullen: Should have always had this since we have a runtime variant. Doesn't ever make sense to not provide a configure method to create empty because an empty razor engine doesn't function.

src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
```diff
@@ -40,8 +40,25 @@ public static RazorEngine CreateDesignTime(Action<IRazorEngineBuilder> configure
...
+                throw new ArgumentNullException(nameof(configure));
+            }
+
             var builder = new DefaultRazorEngineBuilder(designTime: false);
-            configure?.Invoke(builder);
+            configure(builder);
+            return builder.Build();
+        }
+
+        public static RazorEngine CreateDesignTimeEmpty(Action<IRazorEngineBuilder> configure)
```
NTaylorMullen: Should have always had this since we have a runtime variant. Doesn't ever make sense to not provide a configure method to create empty because an empty razor engine doesn't function.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+
+            var projectItem = ProjectFileSystem.GetItem(filePath);
+            var sourceDocument = RazorSourceDocument.ReadFrom(projectItem);
+            var codeDocument = Process(sourceDocument);
+
+            return codeDocument;
+        }
+
+        public override RazorCodeDocument Process(RazorSourceDocument sourceDocument)
```
rynowak: It's not clear to me that this will be useful in the long term. Who's the consumer of this? The editor?

Is the expectation that the file system reflects the on-disk state? or the in-memory state (IDE)? If it's the latter, is passing in the source document useful?

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -0,0 +1,93 @@
...
+            ProjectFileSystem = projectFileSystem;
+            Features = features;
+
+            for (var i = 0; i < features.Count; i++)
+            {
+                features[i].ProjectEngine = this;
+            }
+        }
+
+        public override RazorProjectFileSystem ProjectFileSystem { get; }
```
rynowak: Can we call this `FileSystem` everywhere?

src/Microsoft.AspNetCore.Razor.Language/IRazorImportFeature.cs
rynowak: I think you've lost me on the to-do here. Wouldn't MVC just supply its own implementation of this? WOuldn't the default implemenntation just noop?

src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
rynowak: Sure

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
rynowak: That's not valid. Features aren't expected to support reinitialization (setting `Engine` twice).

I want to understand more about why you think this is necessary.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectFileSystem.cs
rynowak: I think you made the right choice.

src/Microsoft.AspNetCore.Razor.Language/Resources.resx
```diff
@@ -533,4 +533,7 @@ Instead, wrap the contents of the block in "{{}}":
   <data name="PropertyMustNotBeNull" xml:space="preserve">
     <value>The '{0}.{1}' property must not be null.</value>
   </data>
+  <data name="MissingFeatureDependency" xml:space="preserve">
+    <value>The '{0}' is missing feature '{1}'.</value>
+  </data>
```
rynowak: This could use some love. Look at the error messages for RazorEngine and get inspired.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
NTaylorMullen: > That's not valid. Features aren't expected to support reinitialization (setting Engine twice).

Good point. Gonna have to think about what to do here.



> I want to understand more about why you think this is necessary.

When configuring the Razor system we want to show all `Engine` features to users who are customizing the `IRazorFeature` collection.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
NTaylorMullen: Sure

src/Microsoft.AspNetCore.Razor.Language/Resources.resx
NTaylorMullen: lol did. They're nearly identical 😉. I'll change the Resource name to be razor project engine specific. I think you're misunderstanding this error for a feature missing another feature dependency.

NTaylorMullen: Actually, going to change the direction on this. Not going to make the `IRazorImportFeature` a required feature. If the feature isn't available we'll assume 0 imports.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngineBuilderExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public static class RazorProjectEngineBuilderExtensions
+    {
+        public static void ReplaceFeature<TFeature>(this RazorProjectEngineBuilder builder, IRazorFeature feature) where TFeature : IRazorFeature
```
ajaybhargavb: Just discussed with @NTaylorMullen. The user shouldn't need to call this method to add a feature for the first time.

test/Microsoft.AspNetCore.Razor.Language.Test/DefaultRazorProjectEngineTest.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public class DefaultRazorProjectEngineTest
+    {
+        [Fact]
+        public void Process_AsksFileSystemForItems()
+        {
+            // Arrange
+            var razorProjectItem = new TestRazorProjectItem("/some/path.cshtml"); ;
```
ajaybhargavb: Extra semicolon

PR #1959, opened 2018-01-23
Add MVC support for RazorProjectEngine.

NTaylorMullen:
- Make `RazorProjectEngine` call paths for all feature registrations.
- Add `DefaultMvcImportFeature` for latest and 1.X MVC.
- Ported `AddTargetExtension` and `AddDirective` to `RazorProjectEngineBuilderExtensions`.
- Added tests and a test file system project type.

**Note**: In this PR there is a lot of duplicated code due to the complexities of having two public API paths for Razor (`IRazorEngineBuilder` and `RazorProjectEngineBuilder`). Tried initially factoring the code into two private paths that operate on feature collections but that wasn't straightforward due to all of our `AddDirective`, `AddCodeTargetExtension` expectations and of course `ICollection<T>` not being covariant.

#1828 

/cc @mkArtakMSFT


src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/InjectDirective.cs
```diff
@@ -26,6 +26,32 @@ public static class InjectDirective
...
         public static IRazorEngineBuilder Register(IRazorEngineBuilder builder)
         {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            // ---------------------------------------------------------------------------------------------
+            // When updating these registrations also update the RazorProjectEngineBuilder overload as well.
+            // ---------------------------------------------------------------------------------------------
```
rynowak: Can we obsolete this method? Who calls it?

NTaylorMullen:
🆙 📅 

> We discussed and plan to obsolete and freeze the existing methods that act on IRazorEngineBuilder.
> 
> This way there's one way to do it, and we don't need to maintain old code paths, they can just sit there forever and gather dust, which doesn't harm us

Once we have a fully working end-to-end solution with `RazorProjectEngine` i'll actually obsolete all the APIs. Logged an issue to track: https://github.com/aspnet/Razor/issues/1970. For now I've moved the obsolete impls to the bottom of each file and placed them in an obsolete region. That way its obvious we're letting them rot 😉

rynowak:
> Once we have a fully working end-to-end solution with RazorProjectEngine i'll actually obsolete all the APIs. 

Yeah, solid plan 👍 I like it.

Issue #1961, opened 2018-01-23
Flow RazorLanguageVersion to RazorEngine.


NTaylorMullen:
Today we instantiate our options flags always with the latest Razor version. We need to flow our version information to the RazorEngine so we can build an appropriate options feature.

The end-user behavior that you end up seeing is that parser option features are on when they shouldn't be, i.e. bound minimized boolean TagHelper attributes.

NTaylorMullen:
80f943caef2dcc4309743c8896eb6344dac3326a
c0cb8f009c34c518dc854fd9a1dc5c297b5162f4

PR #1962, opened 2018-01-24
Flow RazorLanguageVersion to RazorEngine.

NTaylorMullen:
- Restructured RazorLanguageVersion to be a sealed concrete type to enable things like `RazorLanguageVersion.Latest`; it also allows us to make broader changes in the future. Also, in the future if we want to add support for overriding operators to enable greater than comparisons we can as well.
- Removed version validity checks because we restrict who can construct a `RazorLanguageVersion` now. This way we don't have to check for valid versions all throughout our code.
- Added a simple `ProjectExtensibilityConfiguration` => `RazorLanguageVersion` method in the `DefaultProjectExtensibilityConfigurationFactory` to temporarily enable letting the system operate on the `RazorLanguageVersion`. Eventually that entire class will change.
- Also validated that in VS we now properly detect the RazorLanguageVersion and show appropriate IntelliSense.

#1961 

/cc @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
@@ -15,26 +15,35 @@ namespace Microsoft.AspNetCore.Razor.Language
             return Create(configure: null);
         }
 
-        public static RazorEngine Create(Action<IRazorEngineBuilder> configure)
-        {
-            var builder = new DefaultRazorEngineBuilder(designTime: false);
-            AddDefaults(builder);
-            AddRuntimeDefaults(builder);
-            configure?.Invoke(builder);
-            return builder.Build();
-        }
-
+        public static RazorEngine Create(Action<IRazorEngineBuilder> configure) => CreateCore(RazorConfiguration.DefaultRuntime, configure);
 
         public static RazorEngine CreateDesignTime()
         {
             return CreateDesignTime(configure: null);
         }
 
-        public static RazorEngine CreateDesignTime(Action<IRazorEngineBuilder> configure)
+        public static RazorEngine CreateDesignTime(Action<IRazorEngineBuilder> configure) => CreateCore(RazorConfiguration.DefaultDesignTime, configure);
+
+        // Internal since RazorEngine APIs are going to be obsolete.
+        internal static RazorEngine CreateCore(RazorConfiguration configuration, Action<IRazorEngineBuilder> configure)
         {
-            var builder = new DefaultRazorEngineBuilder(designTime: true);
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            var builder = new DefaultRazorEngineBuilder(configuration.DesignTime);
             AddDefaults(builder);
-            AddDesignTimeDefaults(builder);
+
+            if (configuration.DesignTime)
+            {
+                AddDesignTimeDefaults(configuration, builder);
+            }
+            else
+            {
+                AddRuntimeDefaults(configuration, builder);
+            }
+
             configure?.Invoke(builder);
             return builder.Build();
         }
@@ -99,10 +108,10 @@ namespace Microsoft.AspNetCore.Razor.Language
             builder.Features.Add(configurationFeature);
         }
 
-        internal static void AddRuntimeDefaults(IRazorEngineBuilder builder)
+        internal static void AddRuntimeDefaults(RazorConfiguration configuration, IRazorEngineBuilder builder)
         {
             // Configure options
-            builder.Features.Add(new DefaultRazorParserOptionsFeature(designTime: false, version: RazorParserOptions.LatestRazorLanguageVersion));
+            builder.Features.Add(new DefaultRazorParserOptionsFeature(designTime: false, version: configuration.LanguageVersion));
             builder.Features.Add(new DefaultRazorCodeGenerationOptionsFeature(designTime: false));
 
             // Intermediate Node Passes
@@ -113,10 +122,10 @@ namespace Microsoft.AspNetCore.Razor.Language
             builder.AddTargetExtension(new PreallocatedAttributeTargetExtension());
         }
 
-        internal static void AddDesignTimeDefaults(IRazorEngineBuilder builder)
+        internal static void AddDesignTimeDefaults(RazorConfiguration configuration, IRazorEngineBuilder builder)
         {
             // Configure options
-            builder.Features.Add(new DefaultRazorParserOptionsFeature(designTime: true, version: RazorParserOptions.LatestRazorLanguageVersion));
+            builder.Features.Add(new DefaultRazorParserOptionsFeature(designTime: true, version: configuration.LanguageVersion));
             builder.Features.Add(new DefaultRazorCodeGenerationOptionsFeature(designTime: true));
             builder.Features.Add(new SuppressChecksumOptionsFeature());
 
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
@@ -1,33 +1,50 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    public enum RazorLanguageVersion
+    public sealed class RazorLanguageVersion : IEquatable<RazorLanguageVersion>
     {
-        Version1_0 = 1,
+        public static readonly RazorLanguageVersion Version_1_0 = new RazorLanguageVersion(1, 0);
 
-        Version1_1 = 2,
+        public static readonly RazorLanguageVersion Version_1_1 = new RazorLanguageVersion(1, 1);
 
-        Version2_0 = 3,
+        public static readonly RazorLanguageVersion Version_2_0 = new RazorLanguageVersion(2, 0);
 
-        Version2_1 = 4,
-    }
+        public static readonly RazorLanguageVersion Version_2_1 = new RazorLanguageVersion(2, 1);
 
-    internal static class RazorLanguageVersionExtensions
-    {
-        internal static bool IsValid(this RazorLanguageVersion version)
+        public static readonly RazorLanguageVersion Latest = Version_2_1;
+
+        // Don't want anyone else constructing language versions.
+        private RazorLanguageVersion(int major, int minor)
         {
-            switch (version)
+            Major = major;
+            Minor = minor;
+        }
+
+        public int Major { get; }
+
+        public int Minor { get; }
+
+        public bool Equals(RazorLanguageVersion other)
+        {
+            if (other == null)
             {
-                case RazorLanguageVersion.Version1_0:
-                case RazorLanguageVersion.Version1_1:
-                case RazorLanguageVersion.Version2_0:
-                case RazorLanguageVersion.Version2_1:
-                    return true;
+                return false;
             }
 
-            return false;
+            // We're the only one who can create RazorLanguageVersions so reference equality is sufficient.
+            return ReferenceEquals(this, other);
         }
+
+        public override int GetHashCode()
+        {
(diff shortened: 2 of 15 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
```diff
@@ -15,26 +15,39 @@ public static RazorEngine Create()
             return Create(configure: null);
         }
 
-        public static RazorEngine Create(Action<IRazorEngineBuilder> configure)
+        public static RazorEngine Create(Action<IRazorEngineBuilder> configure) => Create(RazorLanguageVersion.Latest, configure);
+
+        public static RazorEngine Create(RazorLanguageVersion version, Action<IRazorEngineBuilder> configure)
```
NTaylorMullen: Thoughts on these being `public`? At first I had them `internal` but then I figured that providing the `RazorLanguageVersion` is pretty essential to building a `RazorEngine`. Given that we'll have a similar API for the `RazorProjectEngine` (but will exist on the builder) I wanted to keep a similar API available even if you can't detect it directly via the builder.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -1,33 +1,33 @@
...
+            Major = major;
+            Minor = minor;
         }
+
+        public int Major { get; }
+
+        public int Minor { get; }
+
+        public override string ToString() => $"Razor Language '{Major}.{Minor}'";
     }
```
NTaylorMullen: My vision for this is that when we eventually flow the MSBuild Razor Language version we'll add a method on here to determine or create a proper `RazorLanguageVersion`. Either way I wanted to avoid us having a potentially invalid `RazorLanguageVersion` in our system and therefore having to check `IsValid` in several locations (that's why this is the only class that can construct this class).

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
```
NTaylorMullen: This entire class will eventually be re-constructed to support input from MSBuild and build the project extensibility pieces accordingly. For now the changes here are to just get the flow going and setup for success when we move to MSBuild.

src/Microsoft.VisualStudio.Editor.Razor/DefaultTemplateEngineFactoryService.cs
```diff
@@ -15,6 +15,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
     internal class DefaultTemplateEngineFactoryService : RazorTemplateEngineFactoryService
     {
         private readonly static MvcExtensibilityConfiguration DefaultConfiguration = new MvcExtensibilityConfiguration(
+            RazorLanguageVersion.Version2_0,
```
NTaylorMullen: Going to do another pass in a different PR on all of the versions in this class and in `DefaultProjectExtensibilityConfigurationFactory` to update them to 2.1 (latest) since 15.7 aligns with 2.1

src/Microsoft.AspNetCore.Razor.Language/RazorEngine.cs
```diff
@@ -15,26 +15,39 @@ public static RazorEngine Create()
             return Create(configure: null);
         }
 
-        public static RazorEngine Create(Action<IRazorEngineBuilder> configure)
+        public static RazorEngine Create(Action<IRazorEngineBuilder> configure) => Create(RazorLanguageVersion.Latest, configure);
```
rynowak: I don't think we should provide this overload.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -1,33 +1,33 @@
...
 
+using System;
+
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    public enum RazorLanguageVersion
+    public sealed class RazorLanguageVersion
     {
-        Version1_0 = 1,
+        public static readonly RazorLanguageVersion Version1_0 = new RazorLanguageVersion(1, 0);
```
rynowak: Suggestion: name these like `Version_1_0` instead. That reads a little nicer and it's what I did in MVC for this kind of thing.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
rynowak: Yeah, this will eventually have a `TryParse` method.

src/Microsoft.AspNetCore.Razor.Language/RazorParserOptions.cs
```diff
@@ -8,15 +8,13 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     public abstract class RazorParserOptions
     {
-        internal static readonly RazorLanguageVersion LatestRazorLanguageVersion = RazorLanguageVersion.Version2_1;
-
         public static RazorParserOptions CreateDefault()
```
rynowak: We might end up obsoleting some of these things when the day is through.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MvcExtensibilityConfiguration.cs
```diff
@@ -51,10 +58,12 @@ public override bool Equals(ProjectExtensibilityConfiguration other)
...
-            return Enumerable.SequenceEqual(
-                Assemblies.OrderBy(a => a.Identity.Name).Select(a => a.Identity),
-                other.Assemblies.OrderBy(a => a.Identity.Name).Select(a => a.Identity),
-                AssemblyIdentityEqualityComparer.NameAndVersion);
+            return 
+                Enumerable.SequenceEqual(
+                    Assemblies.OrderBy(a => a.Identity.Name).Select(a => a.Identity),
+                    other.Assemblies.OrderBy(a => a.Identity.Name).Select(a => a.Identity),
+                    AssemblyIdentityEqualityComparer.NameAndVersion) &&
+                RazorLanguageVersion == other.RazorLanguageVersion;
```
rynowak: Was this just a bug?

NTaylorMullen: No, just formatted and added the `RazorLanguageVersion` check to the equality.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MvcExtensibilityConfiguration.cs
```diff
@@ -28,19 +30,24 @@ internal class MvcExtensibilityConfiguration : ProjectExtensibilityConfiguration
...
+        // MVC: '2.0.0' (fallback) | Razor Language '2.0.0'
+        // or 
+        // MVC: '2.1.3' | Razor Language '2.1.3'
+        public override string DisplayName => $"MVC: {MvcAssembly.Identity.Version.ToString(3)}" + (Kind == ProjectExtensibilityConfigurationKind.Fallback? " (fallback)" : string.Empty) + " | " + RazorLanguageVersion;
 
         public override ProjectExtensibilityConfigurationKind Kind { get; }
 
         public override ProjectExtensibilityAssembly RazorAssembly { get; }
 
+        public override RazorLanguageVersion RazorLanguageVersion { get; }
```
NTaylorMullen: Exposing `RazorLanguageVersion` here and not `RazorConfiguration` because I'm not convinced that we should tie projects to runtime/design time. Someone else should make that decision depending on how this extensibility configuration is consumed.

NTaylorMullen:
🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MvcExtensibilityConfiguration.cs
rynowak: I'm not sure I understand the rationale behind that, but this seems like a safe minimal change for now. We'll need to revisit this class again in the near future, so we'll eventually do the right thing anyway 😆. Let's discuss when we get back to this spot.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -1,33 +1,33 @@
...
-            return false;
+            Major = major;
+            Minor = minor;
         }
+
+        public int Major { get; }
+
+        public int Minor { get; }
+
+        public override string ToString() => $"Razor Language '{Major}.{Minor}'";
```
rynowak: Suggest taking `Language` out of this. Including `Language` makes it seem like it should be localized.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MvcExtensibilityConfiguration.cs
```diff
@@ -4,13 +4,15 @@
...
 using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 {
     internal class MvcExtensibilityConfiguration : ProjectExtensibilityConfiguration
     {
         public MvcExtensibilityConfiguration(
+            RazorLanguageVersion razorLanguageVersion,
```
rynowak: languageVersion

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MvcExtensibilityConfiguration.cs
rynowak: Ok gotcha

rynowak: nit: can you put the language version first? That should be very cheap to compare.

rynowak: another non-nit. Either make `RazorLanguageVersion` implement `IEquatable<>` or use object.ReferenceEquals(...). Using `==` with non-simple types is non-obvious what it does.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectExtensibilityConfiguration.cs
```diff
@@ -16,6 +17,8 @@ internal abstract class ProjectExtensibilityConfiguration : IEquatable<ProjectEx
 
         public abstract ProjectExtensibilityAssembly RazorAssembly { get; }
 
+        public abstract RazorLanguageVersion RazorLanguageVersion { get; }
```
rynowak: LanguageVersion

src/Microsoft.VisualStudio.Editor.Razor/DefaultTemplateEngineFactoryService.cs
```diff
@@ -41,11 +42,13 @@ public override RazorTemplateEngine Create(string projectPath, Action<IRazorEngi
...
             var configuration = (project?.Configuration as MvcExtensibilityConfiguration) ?? DefaultConfiguration;
+            var razorLanguageVersion = configuration.RazorLanguageVersion;
+            var razorConfiguration = new RazorConfiguration(razorLanguageVersion, designTime: true);
 
             RazorEngine engine;
-            if (configuration.RazorAssembly.Identity.Version.Major == 1)
+            if (razorLanguageVersion.Major == 1)
             {
-                engine = RazorEngine.CreateDesignTime(b =>
+                engine = RazorEngine.CreateCore(razorConfiguration, b =>
```
rynowak: Looks good. I assume this is the meat of it. The pieces come together pretty well.

PR #1964, opened 2018-01-24
Design: Razor SDK

pranavkm:


src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -83,14 +82,22 @@
...
-  <Target Name="_ResolveRazorGenerateOutputs" AfterTargets="PrepareForRazorGenerate">
+  <Target Name="_ResolveRazorGenerateOutputs">
     <ItemGroup>
       <_RazorGenerateOutput Include="%(RazorGenerate.GeneratedOutput)" Condition="'%(RazorGenerate.GeneratedOutput)'!=''"/>    
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <RazorCoreGenerateDependsOn>
+      _HashRazorGenerateInputs;
```
pranavkm: Made these two details of `RazorCoreGenerate` since they do not produce any global state that's useful in the SDK.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -117,7 +124,11 @@
     </ItemGroup>
   </Target>
 
-  <Target Name="_ResolveGeneratedRazorCompileInputs" BeforeTargets="ResolveRazorCompileInputs">
+  <PropertyGroup>
+    <ResolveRazorCompileInputsDependsOn>$(ResolveRazorCompileInputsDependsOn);_ResolveGeneratedRazorCompileInputs</ResolveRazorCompileInputsDependsOn>
```
pranavkm: Just my personal preference to use DependsOn properties over `Before` \ `After` Targets. It also makes the order of it's execution more deterministic.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
```diff
@@ -1,381 +1,3 @@
-﻿<Project>
```
pranavkm: I plan on moving CodeGeneration.targets in to this file. Kept them separate to make the diff easier to read.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
```diff
@@ -0,0 +1,20 @@
...
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved. 
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Import Project="$(_MicrosoftAspNetCoreRazorSdkCurrentVersionPath)" 
+    Condition="'$(_MicrosoftAspNetCoreRazorSdkCurrentVersionPath)' != '' AND Exists('$(_MicrosoftAspNetCoreRazorSdkCurrentVersionPath)')" />
```
pranavkm: My goal here was to have a way to override the targets in the SDK with the targets in the package. The props file in package sets up this property to point to the target next to it and we use it. Let me know if there's a better way to do this, I couldn't find good examples for this.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
```diff
@@ -0,0 +1,395 @@
+﻿<Project>
```
pranavkm: This is from Design.targets minus the reference to CodeGeneration. I'll point out the changes I made.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
```diff
@@ -0,0 +1,395 @@
...
+  <PropertyGroup>
+    <PrepareForRazorGenerateDependsOn>
+      ResolveRazorGenerateInputs;
+      ResolveAssemblyReferenceRazorGenerateInputs;
+      ResolveTagHelperRazorGenerateInputs
+    </PrepareForRazorGenerateDependsOn>
+
+    <RazorGenerateDependsOn>
+      PrepareForRazorGenerate;
+      _EnsureRazorCompilerReferenced;
```
pranavkm: Added this as a guard against the user invoking `/t:RazorCoreGenerate` without referencing the package.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
```diff
@@ -0,0 +1,395 @@
...
+    <ResolveRazorCompileInputsDependsOn>
+    </ResolveRazorCompileInputsDependsOn>
+  </PropertyGroup>
+
+  <!--
+    Gathers inputs to the RazorCoreCompile target into the @(RazorCompile) itemgroup.
+
+    This is marker target so that the code generation targets can attach.
+  -->
+  <Target Name="ResolveRazorCompileInputs" DependsOnTargets="$(ResolveRazorCompileInputsDependsOn)">
```
pranavkm: Added a `DependsOnTargets` here that we attach to.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
```diff
@@ -0,0 +1,395 @@
...
+    -->
+    <ItemGroup Condition="'$(CopyRefAssembliesToPublishDirectory)'=='false'">
+      <ResolvedFileToPublish 
+        Remove="%(ResolvedFileToPublish.Identity)"
+        Condition="'%(ResolvedFileToPublish.RelativePath)'=='$(RefAssembliesFolderName)\%(Filename)%(Extension)'"/>
+    </ItemGroup>
+
+  </Target>
+
+  <Target Name="_EnsureRazorCompilerReferenced" Condition="'$(IsRazorCompilerReferenced)' != 'true'">
```
pranavkm: The error message

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -98,6 +98,7 @@ void CreateDirectoryProps(string originalProjectRoot, string binariesRoot, strin
 
                 var text = $@"
 <Project>
+  <Import Project=""{originalProjectRoot}\..\..\..\src\Microsoft.AspNetCore.Razor.Sdk\SDK\Sdk.props""/>
```
pranavkm: Doing this before the `PropertyGroup` to mimic msbuid's import order.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
rynowak: I think this is fine. Since we're crossing a component boundary it's probably good that RazorCoreGenerate has a depends on.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -1,50 +1,13 @@
...
-  </PropertyGroup>
-
-  <!--
-    Razor also attaches itself by default to some of the standard .NET targets. Uses these properties to
-    configure this behaviour.
-  -->
-  <PropertyGroup>
-    <RazorGenerateDependsOn>RazorResolveGenerateInputs;RazorCoreGenerate</RazorGenerateDependsOn>
-    <RazorCompileDependsOn>RazorGenerate;RazorCoreCompile</RazorCompileDependsOn>
+    <IsRazorCompilerReferenced>true</IsRazorCompilerReferenced>
```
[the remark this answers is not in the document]
rynowak: Yes. The presence of this tells the WebSDK whether or not to assume it can use the SDK.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.props
```diff
@@ -0,0 +1,17 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved. 
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.CurrentVersion.props"
```
rynowak: Can we give this a filename that has Razor in it? `Sdk.CurrentVersion.props` vs Razor.Sdk.CurrentVersion.props`

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.props
```diff
@@ -0,0 +1,17 @@
...
+    Condition="Exists('$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.CurrentVersion.props')" />
```
rynowak: Should we really check for exists here? Isn't this bug if it doesn't exist?

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
rynowak: > My goal here was to have a way to override the targets in the SDK with the targets in the package

Outside of testing scenarios, I don't understand why this is a goal.

The intention is that the SDK provides 60% of the functionality, and the packages provide the other 40%.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
rynowak: The other thing you could do is define the target here to always report an error, and then override with imports.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Sdk.CurrentVersion.targets
rynowak: You could also have this file import the codegeneration targets (if defined). Then the SDK has precise control over the import order.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
rynowak: OK I see now that you're doing this everywhere. I just made the opposite choice when building all of this except for the 'top level' things. We should talk about this and make a decision.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
pranavkm: Testing - If we get an implicit reference to Razor.Sdk via Web.Sdk (❓) this would be a way to override it in our test projects.

OOB solutions - Akin to how Roslyn's Microsoft.NET.Compilers package, we could point users to this package if we want them to try our new SDK features or get bug fixes. 

Theoretically just adding the package reference would suffice since the targets from package would stomp over the targets from the SDK, this just seemed more predictable.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.props
pranavkm: I copied this from the websdk - https://github.com/aspnet/websdk/blob/master/src/ProjectSystem/Microsoft.NET.Sdk.Web.ProjectSystem.Targets/Sdk.props, but yes, does seem like an unncessary check.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
```diff
@@ -0,0 +1,20 @@
...
+
+  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.CurrentVersion.targets"
+    Condition="'$(MicrosoftAspNetCoreRazorSdkTargetsImported)' != 'true' AND Exists('$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.CurrentVersion.props')" />
```
NTaylorMullen: Shouldn't `MicrosoftAspNetCoreRazorSdkTargetsImported` be a private property? Seems like an implementation detail of ours. Also kind of odd that we then expect the currentversion.targets to set this to `true`.

pranavkm: There's some precedence to using a non-private property - https://github.com/Microsoft/msbuild/blob/master/src/Tasks/Microsoft.Common.props#L137.

pranavkm:
🆙 📅

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
rynowak: I think it's fine to make it public. If it's a contract between our package and our SDK, I think it should be public. We need to have this discipline to get these things right (most of the time).

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -1,50 +1,18 @@
...
-      Set to true to automatically include Razor (.cshtml) files in @(RazorGenerate) from @(Content).
-    -->
-    <EnableDefaultRazorGenerateItems Condition="'$(EnableDefaultRazorGenerateItems)'==''">true</EnableDefaultRazorGenerateItems>
-
-    <!-- 
-      Set to true to copy RazorGenerate items (.cshtml) to the publish directory.
-
-      Typically Razor files are not needed for a published application if they participate in compilation at build-time
-      or publish-time. By default, the Razor SDK will suppress the copying of RazorGenerate items to the publish directory.
+    Used by the Web SDK to know if the Razor SDK is present. This needs to live in a nuget package (not in the SDK)
```
rynowak: > if the Razor SDK is present

This is probably my fault :sad:

Can you change this to say "if the Razor SDK can be used for compilation"

rynowak: The Razor SDK is always there. 







(watching you)

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -1,50 +1,18 @@
...
-      Set to true to copy reference assembly items to the publish directory.
+    <IsRazorCompilerReferenced>true</IsRazorCompilerReferenced>
 
-      Typically reference assemblies are not needed for a published application if Razor compilation occurs at build-time
-      or publish-time. By default, the Razor SDK will suppress the copying of reference assemblies to the publish directory.
+    <!--
+    Location of the CodeGeneration targets. The SDK uses this to import the file ensuring deterministic import order.
     -->
-    <CopyRefAssembliesToPublishDirectory Condition="'$(CopyRefAssembliesToPublishDirectory)'==''">false</CopyRefAssembliesToPublishDirectory>
+    <MicrosoftAspNetCoreRazorCodeGenerationTargetsPath>$(MSBuildThisFileDirectory)Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets</MicrosoftAspNetCoreRazorCodeGenerationTargetsPath>
```
rynowak: We could simplify this name to something like `RazorCodeGenerationTargetsPath`.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Microsoft.AspNetCore.Razor.Sdk.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <Import Project="Sdk.CurrentVersion.props" />
+
+  <PropertyGroup>
+    <!-- Redirect the SDK to use the targets from the current package -->
+    <_MicrosoftAspNetCoreRazorSdkCurrentVersionPath>$(MSBuildFileDirectory)Sdk.CurrentVersion.targets</_MicrosoftAspNetCoreRazorSdkCurrentVersionPath>
```
rynowak: Think about just simplifying this name too, just like `RazorSdkCurrentVersionPath`

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -373,6 +381,11 @@
 
   </Target>
 
+  <Target Name="_EnsureRazorCompilerReferenced" Condition="'$(IsRazorCompilerReferenced)' != 'true'">
+    <Error Text="A reference to Microsoft.AspNetCore.Razor.Design is required for this target to succeed. See todo:fwlink for instructions on how to resolve this." />
```
rynowak: I don't want to block on this, but it would be cool to talk to @vijayrkn or someone else with more SDK experience about how we should make this actionable. Can you log an issue to do a polish pass on this error?

rynowak:
@pranavkm - this is done right? Should this PR be closed?

pranavkm:
Yup.

rynowak:
BTW it looks like a few things from the PR aren't addressed in what got checked in. Still has names like `MicrosoftAspNetCoreRazorCodeGenerationTargetsPath`

pranavkm:
I fixed those locally before pushing - https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props#L12.

rynowak:
Ah ok, there's still a `MicrosoftAspNetCoreRazorSdkTargetsImported` 

Maybe I forgot to comment on it... 😕

Issue #1966, opened 2018-01-24
Generate MPacks when building on non-windows.


NTaylorMullen:
We used to disable MPack generation on non-windows systems because we would use powershell to zip mpacks. Now that the `ZipArchive` task is fixed we should go back to building MPacks on non-windows.

NTaylorMullen:
5c6ff4366f8f0b9e414bd9632fd50ed438f44ebb

NTaylorMullen:
@ToddGrun @KirillOsenkov you guys can now build MPacks on non-windows systems.

KirillOsenkov:
:+1: cc @davidkarlas

rynowak:
We still need to solve portable PDBs it sounds like, but this is a start

KirillOsenkov:
Filed https://github.com/aspnet/Razor/issues/1968 to track this.

rynowak:
Thanks!

PR #1967, opened 2018-01-24
Make it so MPacks generate in non-windows build.

NTaylorMullen:
We used to disable MPack generation on non-windows systems because we would use powershell to zip mpacks. Now that the ZipArchive task is fixed we can go back to building MPacks on non-windows.

/cc @ToddGrun @KirillOsenkov

#1966


NTaylorMullen:
Also verified mpack is generated and installable on my Mac 👍

Issue #1958, opened 2018-01-23
Build server: Add tests


ajaybhargavb:
Meta issue - #1921

ajaybhargavb:
004ff204aa8a781e148534113f48ab33e83271b4

PR #1969, opened 2018-01-24
Added tests for BuildProtocol and RequestDispatcher

ajaybhargavb:
Issue - #1958 

More tests coming soon 

FYi @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Tools/RequestDispatcher.cs
```diff
@@ -34,427 +34,434 @@ public static RequestDispatcher Create(ConnectionHost connectionHost, CompilerHo
         public static readonly TimeSpan GCTimeout = TimeSpan.FromSeconds(30);
 
         public abstract void Run();
+    }
 
-        private enum State
+    // Internal for testing
+    internal class DefaultRequestDispatcher : RequestDispatcher
```
rynowak: If this is going to be internal and non-nested pull it out into it's own file

rynowak: BTW because you changed the nesting, everything in this class shows up in the diff, so it's pretty hard to review.

ajaybhargavb: Sorry about that. This should help https://github.com/aspnet/Razor/pull/1969/files?w=1

src/Microsoft.AspNetCore.Razor.Tools/Roslyn/CompilerServerLogger.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.CodeAnalysis.CommandLine
     internal class CompilerServerLogger
     {
         // Environment variable, if set, to enable logging and set the file to log to.
-        private const string environmentVariable = "RoslynCommandLineLogFile";
+        private const string environmentVariable = "RazorBuildServerCommandLineLogFile";
```
rynowak: `EnvironmentVariable`

rynowak: Suggest `RAZORBUILDSERVER_LOG`

test/Microsoft.AspNetCore.Razor.Tools.Test/RequestDispatcherTest.cs
```diff
@@ -0,0 +1,552 @@
...
+    {
+        private static BuildRequest EmptyBuildRequest => new BuildRequest(1, Array.Empty<BuildRequest.Argument>());
+
+        private static BuildResponse EmptyBuildResponse => new CompletedBuildResponse(
+            returnCode: 0,
+            utf8output: false,
+            output: string.Empty);
+
+        [Fact]
+        public async Task ReadFailure()
```
rynowak: These need better names across the board. It's not really clear what's the expected outcome and why.

rynowak: A better name would be something like `Dispatcher_WhenReadingRequestFails_ClosesConnection`

test/Microsoft.AspNetCore.Razor.Tools.Test/RequestDispatcherTest.cs
```diff
@@ -0,0 +1,552 @@
...
+            await dispatcherTask;
+
+            // Assert
+            Assert.Equal(totalCount, eventBus.CompletedCount);
+            Assert.True(eventBus.LastProcessedTime.HasValue);
+            Assert.True(eventBus.HitKeepAliveTimeout);
+        }
+
+        [Fact]
+        public void ClientExceptionShouldBeginShutdown()
```
rynowak: This seems like a pretty good start. I want to spend a little time thinking about the scenarios covered here to see if we cover everything.

test/Microsoft.AspNetCore.Razor.Tools.Test/RequestDispatcherTest.cs
```diff
@@ -0,0 +1,552 @@
...
+            var dispatcher = new DefaultRequestDispatcher(connectionHost, compilerHost, CancellationToken.None, eventBus, keepAlive);
+
+            // Act
+            dispatcher.Run();
+
+            // Assert
+            Assert.True(eventBus.HasDetectedBadConnection);
+            Assert.True(listenCancellationToken.IsCancellationRequested);
+        }
+
```
rynowak: Is there a test for the happy path? Considering we have two commands, we should at least exercise the serialization functionality for both of those.

ajaybhargavb: You are right. Those tests will be there in an upcoming PR

test/Microsoft.AspNetCore.Razor.Tools.Test/Roslyn/BuildProtocolTest.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.CommandLine;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    public class BuildProtocolTest
```
rynowak: What's BuildProtocol?

test/Microsoft.AspNetCore.Razor.Tools.Test/Roslyn/BuildProtocolTest.cs
```diff
@@ -0,0 +1,129 @@
...
+    {
+        [Fact]
+        public async Task ReadWriteCompleted()
+        {
+            // Arrange
+            var response = new CompletedBuildResponse(42, utf8output: false, output: "a string");
```
rynowak: We should think about renaming the request and response classes. We don't really 'build', we do other things.

test/Microsoft.AspNetCore.Razor.Tools.Test/Roslyn/BuildProtocolTest.cs
ajaybhargavb: The mega class that contains `BuildRequest`, `BuildResponse` and their implementations.

test/Microsoft.AspNetCore.Razor.Tools.Test/Roslyn/BuildProtocolTest.cs
ajaybhargavb: I was actually thinking of renaming these to `ServerProtocol`, `ServerRequest` and `ServerResponse`. But that will be in a separate cleanup PR.

test/Microsoft.AspNetCore.Razor.Tools.Test/Roslyn/BuildProtocolTest.cs
rynowak: ewww mega class.

If you're introducing tests for it now, why not also include the split/cleanup. You'll just have more to cleanup this way.

ajaybhargavb: Sure yea that makes sense 👍

test/Microsoft.AspNetCore.Razor.Tools.Test/RequestDispatcherTest.cs
```diff
@@ -0,0 +1,552 @@
...
+        [Fact]
+        public async Task ReadFailure()
+        {
+            // Arrange
+            var stream = new Mock<Stream>();
+            var compilerHost = CreateCompilerHost();
+            var connectionHost = CreateConnectionHost();
+            var dispatcher = new DefaultRequestDispatcher(connectionHost, compilerHost, CancellationToken.None);
+            var connection = CreateConnection(stream.Object);
```
NTaylorMullen: Nice little tip for one-off mocks that are only used to fulfill parameters is to use `Mock.Of<T>`: 

`var connection = CreateConnection(Mock.Of<Stream>())`

ajaybhargavb:
🆙 📅 

- Added more tests
- Cleanup and renames
This should be all the tests that we want to port from Roslyn. We should add more tests to cover the scenarios we care about. 

**Tests yet to be added**
- Unit tests for `DiscoverCommand` and `GenerateCommand`
- Tests for code that creates the server and sends the request (client side code in `DotnetToolTask` and `BuildServerConnection`)
- Integration tests that use the server and test the fallback scenario

@rynowak @NTaylorMullen, please add to this if I missed anything.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -13,10 +14,13 @@ public ServerCommand(Application parent)
             : base(parent, "server")
         {
             Pipe = Option("-p|--pipe", "name of named pipe", CommandOptionType.SingleValue);
+            KeepAlive = Option("-k|--keep-alive", "sets the default idle timeout for the server in seconds", CommandOptionType.SingleValue);
```
NTaylorMullen: Does `timeout`/`idleTimeout` make more sense here?

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -58,25 +75,55 @@ internal abstract class Client : IDisposable
...
+                        await Stream.ReadAsync(Array.Empty<byte>(), 0, 0, cancellationToken);
+                        CompilerServerLogger.Log($"After poking pipe {Identifier}.");
+                    }
+                    catch (OperationCanceledException)
+                    {
+                    }
+                    catch (Exception e)
+                    {
+                        // It is okay for this call to fail.  Errors will be reflected in the
+                        // IsConnected property which will be read on the next iteration of the
```
NTaylorMullen: incomplete comment. Left me on the edge of my seat.

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -12,6 +12,23 @@ namespace Microsoft.AspNetCore.Razor.Tools
 {
     internal abstract class Client : IDisposable
     {
+        private static int counter;
```
NTaylorMullen: lol super hack. Why isnt a guid (or even getting the current instances hash code) sufficient?

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -58,25 +75,55 @@ internal abstract class Client : IDisposable
...
+                Identifier = identifier;
             }
 
             public override Stream Stream { get; }
 
+            public override string Identifier { get; }
+
+            public async override Task WaitForDisconnectAsync(CancellationToken cancellationToken)
+            {
+                if (!(Stream is PipeStream pipeStream))
```
NTaylorMullen: You do this check later as well. When is the Stream not a PipeStream?

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
ajaybhargavb: We do the same thing here as well https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/ConnectionHost.cs#L20. This is only used for logging purposes. I think the reason for the counter was to log the order in which the clients get connected/disconnected.

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
ajaybhargavb: Yes, We always back it with a PipeStream. But since `Client` is an abstraction and `NamedPipeClient` is only its implementation, I didn't want to assume that it always is a PipeStream.

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
NTaylorMullen: Fair enough 👍

PR #1972, opened 2018-01-25
Add RazorLanguageVersion flow to preview1.

NTaylorMullen:
**The changes here were already [signed off on](https://github.com/aspnet/Razor/pull/1962), this PR is just a pre-measure to changing our preview1 branch so I can get sign off on these changes making it in**

- Given that this change impacts public API it's important that we get this into the previews prior to us ever shipping public API that we don't intend for end-users to consume.
- This fix doesn't have any functional impact on runtime code.
- This fix will be making its way to VS 15.7 to enable the editor to turn on / off parser features based off of language 

@mkArtakMSFT is it cool if I get this into preview1? Not sure of the procedure for getting changes in.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Properties/Resources.Designer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Properties/Resources.Designer.cs
@@ -669,20 +669,6 @@ namespace Microsoft.AspNetCore.Razor.Language
             => string.Format(CultureInfo.CurrentCulture, GetString("InvalidOperation_SpanIsNotChangeOwner"), p0, p1);
 
         /// <summary>
-        /// Provided value for razor language version is unsupported or invalid: '{0}'.
-        /// </summary>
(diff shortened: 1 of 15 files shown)


rynowak:
> Not sure of the procedure for getting changes in.

Send email to engineering leads

NTaylorMullen:
@rynowak during standup you mentioned that you were thinking this should go into preview2. If we did that we'd be breaking API between preview1 and preview2, do you feel strongly?

rynowak:
> If we did that we'd be breaking API between preview1 and preview2

This isn't a concern. I don't feel strongly about any of this, just trying to avoid extra work trying to get this checked in somewhere that it isn't really used.

NTaylorMullen:
> This isn't a concern. I don't feel strongly about any of this, just trying to avoid extra work trying to get this checked in somewhere that it isn't really used.

ok sure, will get this into dev then.

NTaylorMullen:
This branch going away. Not going into preview1.

Issue #1912, opened 2018-01-06
RazorGenerate and RazorCompile can't be invoked directly with multi-targeting


rynowak:
You can't call these targets at the command line using `dotnet msbuild /t:RazorGenerate` if the project does multi-targeting

rynowak:
Add `/p:TargetFramework=....` to workaround.

pranavkm:
Do we want to consider taking this for preview1? For the release, this involves removing https://github.com/aspnet/Razor/blob/release/2.1/src/Microsoft.AspNetCore.Razor.Design/buildMultiTargeting/Microsoft.AspNetCore.Razor.Design.targets.

rynowak:
It doesn't hurt to ask. Send mail to PMs and eng leads.

pranavkm:
Preview1 - addressed in https://github.com/aspnet/Razor/commit/766862390b52db5fd4e6e8e502cf2949e7d74eb6
Preview2 - addressed in https://github.com/aspnet/Razor/commit/42c3102cd46413b4caeb711fc904f50920fd8861

PR #1975, opened 2018-01-26
Don't import the SDK when cross-targeting

pranavkm:
When we invoke `dotnet msbuild /t:RazorGenerate` on a cross-targeting project, the build fails saying it cannot find the target `ResolveReferences`. This is because the target is only defined for inner builds. The change in essence follows the same pattern as the CSharp targets where targets like `Compile` are only available in the inner build.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
+++ b/src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
@@ -12,9 +12,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Import Project="$(_RazorSdkCurrentVersionPath)"
-    Condition="'$(_RazorSdkCurrentVersionPath)' != '' AND Exists('$(_RazorSdkCurrentVersionPath)')" />
+    Condition="'$(IsCrossTargetingBuild)' != 'true' AND '$(_RazorSdkCurrentVersionPath)' != '' AND Exists('$(_RazorSdkCurrentVersionPath)')" />
 
   <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Razor.Sdk.CurrentVersion.targets"
-    Condition="'$(MicrosoftAspNetCoreRazorSdkTargetsImported)' != 'true'" />
+    Condition="'$(IsCrossTargetingBuild)' != 'true' AND '$(MicrosoftAspNetCoreRazorSdkTargetsImported)' != 'true'" />
 
 </Project>
--- a/src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Microsoft.AspNetCore.Razor.Sdk.props
+++ b/src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Microsoft.AspNetCore.Razor.Sdk.props
@@ -1,8 +1,19 @@
+<!--
+***********************************************************************************************
+Microsoft.AspNetCore.Razor.Sdk.props
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
 <Project>
-  <Import Project="Sdk.CurrentVersion.props" />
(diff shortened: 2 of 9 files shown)


rynowak:
So is this the fix for https://github.com/aspnet/Razor/issues/1912?

What's the experience if you try this?

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
```diff
@@ -12,9 +12,9 @@ Copyright (c) .NET Foundation. All rights reserved.
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Import Project="$(_RazorSdkCurrentVersionPath)"
-    Condition="'$(_RazorSdkCurrentVersionPath)' != '' AND Exists('$(_RazorSdkCurrentVersionPath)')" />
+    Condition="'$(IsCrossTargetingBuild)' != 'true' AND '$(_RazorSdkCurrentVersionPath)' != '' AND Exists('$(_RazorSdkCurrentVersionPath)')" />
 
   <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Razor.Sdk.CurrentVersion.targets"
-    Condition="'$(MicrosoftAspNetCoreRazorSdkTargetsImported)' != 'true'" />
+    Condition="'$(IsCrossTargetingBuild)' != 'true' AND '$(MicrosoftAspNetCoreRazorSdkTargetsImported)' != 'true'" />
```
rynowak: Can we rename `MicrosoftAspNetCoreRazorSdkTargetsImported` to just `RazorSdkTargetsImported`

pranavkm:
We get the same behavior as calling `Compile` when you run it in a cross targeted project without specifying a TFM i.e. msbuild fails saying the target does not exist.

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
pranavkm: Ahh, I missed that one. Sure.

rynowak:
Could we do what publish does then? They add a dummy target that tells you that you need to specify a TFM

pranavkm:
🆙 📅

src/Microsoft.AspNetCore.Razor.Sdk/SDK/Sdk.targets
```diff
@@ -9,12 +9,13 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
...
+<Project>
 
-  <Import Project="$(_RazorSdkCurrentVersionPath)"
-    Condition="'$(_RazorSdkCurrentVersionPath)' != '' AND Exists('$(_RazorSdkCurrentVersionPath)')" />
+  <PropertyGroup Condition="'$(RazorSdkTargetsPath)' == ''">
+    <RazorSdkTargetsPath Condition="'$(IsCrossTargetingBuild)' == 'true'">$(MSBuildThisFileDirectory)..\buildMultiTargeting\Razor.Sdk.CurrentVersion.MultiTargeting.targets</RazorSdkTargetsPath>
+    <RazorSdkTargetsPath Condition="'$(IsCrossTargetingBuild)' != 'true'">$(MSBuildThisFileDirectory)..\build\netstandard2.0\Razor.Sdk.CurrentVersion.targets</RazorSdkTargetsPath>
+  </PropertyGroup>
 
-  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Razor.Sdk.CurrentVersion.targets"
```
pranavkm: I removed `MicrosoftAspNetCoreRazorSdkTargetsImported` and made `RazorSdkTargetsPath`. Sort of what Microsoft.CSharp.Targets does

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Microsoft.AspNetCore.Razor.Sdk.props
```diff
@@ -1,8 +1,19 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
 <Project>
-  <Import Project="Sdk.CurrentVersion.props" />
+  <Import Project="Razor.Sdk.CurrentVersion.props" />
```
pranavkm: Missed updating this earlier.

PR #1977, opened 2018-01-26
Merge vs15.7 into dev

NTaylorMullen:
This adds support for `RazorConfiguration` / `RazorLanguageVersion`. These changes have already been signed off on so this is really just an FYI but they don't currently build on their own in this branch. I'll be sending a follow up PR that fixes those breaks (it revolves around updating `RazorProjectEngine`)

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
@@ -1,0 +1,28 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public sealed class RazorConfiguration
+    {
+        public static readonly RazorConfiguration DefaultRuntime = new RazorConfiguration(RazorLanguageVersion.Latest, designTime: false);
+        public static readonly RazorConfiguration DefaultDesignTime = new RazorConfiguration(RazorLanguageVersion.Latest, designTime: true);
+
+        public RazorConfiguration(RazorLanguageVersion languageVersion, bool designTime)
+        {
+            if (languageVersion == null)
+            {
+                throw new ArgumentNullException(nameof(languageVersion));
+            }
+
+            LanguageVersion = languageVersion;
+            DesignTime = designTime;
+        }
+
+        public RazorLanguageVersion LanguageVersion { get; }
+
+        public bool DesignTime { get; }
+    }
+}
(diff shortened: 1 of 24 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public sealed class RazorConfiguration
+    {
+        public static readonly RazorConfiguration DefaultRuntime = new RazorConfiguration(RazorLanguageVersion.Latest, designTime: false);
+        public static readonly RazorConfiguration DefaultDesignTime = new RazorConfiguration(RazorLanguageVersion.Latest, designTime: true);
```
rynowak: I'm really not sure this is a thing. I don't know exactly what a default `RazorConfiguration` means

NTaylorMullen: Good point, in RazorEngine land it makes sense because we have separate Runtime/DesignTIme methods that are used to create goodness but in the new RazorProjectEngine world you'll have to go out of your way to have a design time configuration; at which point you might as well construct the configuration from the ground up. I'll address this in the coming PR.

NTaylorMullen: Going to leave a runtime default as `RazorConfiguration.Default` as a convenience for tests.

rynowak: OK, but even for runtime I'm not sure that we'll ever have a **GLOBAL DEFAULT CONFIGURATION**

NTaylorMullen: > OK, but even for runtime I'm not sure that we'll ever have a GLOBAL DEFAULT CONFIGURATION

Why wouldn't RazorLanguageVersion.Latest + designTime: false not be a default config?

rynowak: Because what code does it generate?

rynowak: Myabe I'm geting too far inside my own head here.

NTaylorMullen: > Because what code does it generate?

Code compliant with the latest Razor version. I think you and I might see different futures for the `RazorConfiguration` class. I don't envision `RazorConfiguration` being the super config class that also encompasses Razor extensibility. I view a higher layer containing that logic (like we have today in `ProjectExtensibilityConfiguration`

rynowak: Ok, let's go with your idea for now then. We can revisit - I think it will be easy to figure out when we've written more codez.

PR #1978, opened 2018-01-26
Update RazorProjectEngine to use RazorConfiguration.

NTaylorMullen:
- In this PR we do away with `CreateDesignTime` on the `RazorProjectEngine`. Instead we now have an overload that takes in a configuration and does the right thing.
- Updated `RazorProjectEngineBuilder` to have a configuration.
- Updated `RazorConfiguration` to only have a `Default`. Setting up a razor configuration for design time now requires calling code to construct the configuration manually.

/cc @mkArtakMSFT

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
@@ -20,74 +20,37 @@ namespace Microsoft.AspNetCore.Razor.Language
 
         public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem) => Create(fileSystem, configure: null);
 
-        public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+        public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure) => Create(fileSystem, RazorConfiguration.Default, configure);
+
+        public static RazorProjectEngine Create(
+            RazorProjectFileSystem fileSystem,
+            RazorConfiguration configuration,
+            Action<RazorProjectEngineBuilder> configure)
         {
             if (fileSystem == null)
             {
                 throw new ArgumentNullException(nameof(fileSystem));
             }
 
-            var builder = new DefaultRazorProjectEngineBuilder(designTime: false, fileSystem: fileSystem);
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            var builder = new DefaultRazorProjectEngineBuilder(configuration, fileSystem);
 
             AddDefaults(builder);
-            AddRuntimeDefaults(builder);
+
+            if (configuration.DesignTime)
+            {
+                AddDesignTimeDefaults(builder);
+            }
+            else
+            {
+                AddRuntimeDefaults(builder);
+            }
+
             configure?.Invoke(builder);
-
-            return builder.Build();
-        }
-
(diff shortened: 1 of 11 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -20,60 +20,45 @@ public abstract class RazorProjectEngine
 
         public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem) => Create(fileSystem, configure: null);
 
-        public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+        public static RazorProjectEngine Create(RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure) => Create(fileSystem, configure, RazorConfiguration.Default);
```
rynowak: What would you think about just providing the third-arg version of this?

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -20,60 +20,45 @@ public abstract class RazorProjectEngine
...
+
+        public static RazorProjectEngine Create(
+            RazorProjectFileSystem fileSystem,
+            Action<RazorProjectEngineBuilder> configure,
+            RazorConfiguration configuration)
```
rynowak: Can we put the `Action<>` last? That will look much cleaner to use

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -20,60 +20,45 @@ public abstract class RazorProjectEngine
...
-            var builder = new DefaultRazorProjectEngineBuilder(designTime: false, fileSystem: fileSystem);
-
-            configure(builder);
+            configure?.Invoke(builder);
 
             return builder.Build();
         }
 
-        public static RazorProjectEngine CreateDesignTimeEmpty(RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+        public static RazorProjectEngine CreateEmpty(
```
rynowak: Do we have a strong use case for this as a public API?

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
NTaylorMullen: sure.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
NTaylorMullen: Not today. Can always add later, i'll nuke it.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
NTaylorMullen: ehhh, not a fan. You'll rarely use the third overload outside of Razor so I'd prefer not to force any extensibility pieces to use that.

NTaylorMullen:
🆙 🗡

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
rynowak: > force

You keep using this word 😆  It's not a burden to be explicit about something if it's like 10 characters (the configuration).  I also super don't understand who would use this without providing **any** setup for the builder.

I don't want to block your PR on this issues, let's resolve to talk about it next week. I think we're repeating some mistakes made in the first version of this code.

Issue #1979, opened 2018-01-26
Add non-workspace specific foreground thread validator for MEF services


NTaylorMullen:
This is a part of https://github.com/aspnet/Razor/issues/1973 but felt that it deserved its own issue since it's a key part of the Razor tooling system.

So the basic issue is that today we have a `ForegroundDispatcher` which is a workspace service; given that it's a workspace service this means that we need to have a workspace in order to get a foreground dispatcher. In the case of MEF services there are instances when we wont know what's the right workspace in VS for Mac (Such as the `RazorTextViewConnectionListener`), primarily because we'll be at a point when we haven't been given any text buffer or text view.

~Now you might be thinking ok, well lets break the connection between `ForegroundDispatcher` and workspaces; however, today our foreground dispatcher indirectly uses the `Workspaces` lifetime / instantiation threads to determine foreground/background thread task schedulers so we can dispatch between them. This is highly important. So we can't just make the dispatcher assume foreground/background schedulers (we need to rely on the workspaces lifetime to do that).~

~This issue tracks creating a `VisualStudioThreadValidator` (internal for now) that only has a `IsForegroundThread` property. Today we don't need to dispatch between foreground/background threads in the IDE's context.~

~I also want to mention in VS windows context the `VisualStudioThreadValidator` will functionally act exactly like the `ForegroundDispatcher.IsForegroundThread` since there's only a single workspace.~~

Edit: See below comment

rynowak:
I'm interested to see if we can fix this without rewriting too much stuff. I really don't think we need new concepts.

If we make the `ForegroupDispatcher` a general-MEF thing and then add an extension method to get it given the workspacehostservices, that should solve this with very little churn.

NTaylorMullen:
> I'm interested to see if we can fix this without rewriting too much stuff. I really don't think we need new concepts.
> 
> If we make the ForegroupDispatcher a general-MEF thing and then add an extension method to get it given the workspacehostservices, that should solve this with very little churn.

Actually, debugging through VS a bit more I realized that I made a bad assumption that our workspace service was constructed when the workspace was constructed. That means our foreground dispatcher today 100% depends on whoever is constructing it to be on the foreground thread. Indirectly that ends up being WTE. Given that finding I think we can make `ForegroundDispatcher` 100% MEF and have all of our workspace factories take that in (if the type they're constructing needs it).

rynowak:
[...]

PR #1981, opened 2018-01-26
Change ForegroundDispatcher to be a MEF service.

NTaylorMullen:
- The `ForegroundDispatcher` needed to be accessible by services without requiring a workspace; given that it doesn't have any ties to the `Workspace` other than being a service of one I was able to move it into a MEF service.
- Updated all workspace inclusions of the dispatcher to use importing constructors instead.
- Updated the Mac + Windows implementations to be exported as MEF pieces.

#1979

**Note:** Working on testing all these bits in VS with WTEs latest bits but have to do a build and install in their code base which takes a decent amount of time 😉 

/cc @mkArtakMSFT


NTaylorMullen:
Just verified this with WTE's bits installed on a local VS. 👍 👍 👍

rynowak:
Cool beans

Issue #1982, opened 2018-01-26
Change EditorSettingsManager's lifetime to not be per-workspace.


NTaylorMullen:
Today `EditorSettingsManager` is exposed as a MEF service but ultimately is bound to a single workspace. The MEF service today delegates all of its state tracking and event handling down to the single workspace its bound to. It functions this way because when it was initially created we thought that the workspace was lived as long as the IDE. Now that we know that's not the case we need to change how the editor settings manager works without breaking public API. We can do this by lifting its state / event handling functionality (`EditorSettingsManagerInternal`) into its MEF implementation. Next we can re-purpose `EditorSettingsManagerInternal` into a less-API heavy middleman for any workspace services concerned with editor settings to consume.

Functionally this will work 100% the same in VS windows and will make it possible to work over multiple workspaces in VS Mac.

NTaylorMullen:
99010aaea45dbcc821e96c58fa9b71590069560c

PR #1983, opened 2018-01-27
Change EditorSettingsManager to not be per-workspace.

NTaylorMullen:
- Lifted `DefaultEditorSettingsManagerInternal`s state / event handling functionality into its `DefaultEditorSettingsManager`.
- Re-purposed `EditorSettingsManagerInternal` into a less-API heavy middleman for any workspace services concerned with editor settings to consume.
- Moved `DefaultEditorSettingsManagerInternal` into the Editor.Razor assembly since it needed the `EditorSettingsManager` API to function properly. The contract still exists at the `CodeANalysis.Razor.Workspace` level.
- Made the new `DefaultEditorSettingsManager` clean-up dereferenced event handlers since we don't know when workspaces go away and don't want to leak memory and/or trigger bad OnChanged code.
- Added tests to validate new garbage collection functionality and the separation between editorsettingsmanager and the editorsettingmanger internal.

#1982

/cc @mkArtakMSFT


src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
```diff
@@ -12,26 +13,95 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
+            for (var i = 0; i < _changedHandlers.Count; i++)
+            {
+                if (_changedHandlers[i].TryGetTarget(out var handler))
+                {
+                    handlersToInvoke.Add(handler);
+                }
+                else
+                {
+                    // Handler was dereferenced, clean it up.
+                    _changedHandlers.RemoveAt(i--);
```
rynowak: There it folks... The Taylor.

rynowak: /cc @Eilon

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
```diff
@@ -12,26 +13,95 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
+            }
+            remove
+            {
+                _foregroundDispatcher.AssertForegroundThread();
+
+                for (var i = 0; i < _changedHandlers.Count; i++)
+                {
+                    if (_changedHandlers[i].TryGetTarget(out var target) && target == value)
+                    {
+                        _changedHandlers.RemoveAt(i);
```
rynowak: This has a bug. You need to use #TheTaylor

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
```diff
@@ -12,26 +13,95 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
+            _foregroundDispatcher = foregroundDispatcher;
+            _settings = EditorSettings.Default;
+            _changedHandlers = new List<WeakReference<EventHandler<EditorSettingsChangedEventArgs>>>();
         }
 
         public override event EventHandler<EditorSettingsChangedEventArgs> Changed
         {
-            add => _editorSettingsManager.Changed += value;
-            remove => _editorSettingsManager.Changed -= value;
+            add
```
rynowak: What made you decide on weakreference here? Because it's hard to know when to detatch?

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManagerInternal.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.CodeAnalysis.Razor.Editor;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class DefaultEditorSettingsManagerInternal : EditorSettingsManagerInternal
```
rynowak: Consider renaming this hierarchy to be called `WorkspaceEditorSettings`. That's a better name and makes the different between these two things clear.

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManagerInternal.cs
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            if (editorSettingsManager == null)
+            {
+                throw new ArgumentNullException(nameof(editorSettingsManager));
+            }
+
+            _foregroundDispatcher = foregroundDispatcher;
+            _editorSettingsManager = editorSettingsManager;
+            _editorSettingsManager.Changed += OnChanged;
```
rynowak: It's probably worth a comment here that we don't unsubscribe, and we're expecting ESM to use weakevents.

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
```diff
@@ -12,26 +13,95 @@ namespace Microsoft.VisualStudio.Editor.Razor
...
+        }
+
+        private void OnChanged()
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            var handlersToInvoke = new List<EventHandler<EditorSettingsChangedEventArgs>>();
+            for (var i = 0; i < _changedHandlers.Count; i++)
+            {
+                if (_changedHandlers[i].TryGetTarget(out var handler))
```
rynowak: I'm not super stoked on the idea, but we could patternize this into a reusable weakevent class. It's a fairly common pattern thanks to WPF, but we don't want to depend on that.

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManagerInternal.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public override EditorSettings Current => _editorSettingsManager.Current;
+
+        // Internal for testing
+        internal void OnChanged(object sender, EditorSettingsChangedEventArgs e)
```
rynowak: This usage here has a very serious bug. If you're using weakevents like this, you need to keep a reference to the **delegate**.

That looks like:
```
private readonly EventHandler<EditorSettingsChangedEventArgs> _handler;
...

_handler = OnChanged;
_editorSettingsManager.Changed += _handler;
```

The reason is that the thing you have a weakreference to is an `EventHandler<EditorSettingsChangedEventArgs>`. Method-group-to-delegate-conversion will always return a new instance, so you need to manually cache it.

Without manually caching `_handler` in this case the only reference is the weakreference, and it's immediately eligible for GC.

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
NTaylorMullen: Purposefully using #TheTaylor here... Try writing it backwards you have to insert(0) to maintain handler invocation order, at which point that code needs a comment about un-reversing the handler order adding unnecessary complexity. In this case I actually undid the reverse loop, this is far clearer no matter how much you dislike i--

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
NTaylorMullen: Nope, we return immediately no #TheTaylor necessary

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
NTaylorMullen: Yes

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManagerInternal.cs
NTaylorMullen: Yup you're totally right

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
rynowak: Ah I see, you're only removing one at a time.

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManager.cs
rynowak: You could just skip the temporary `handlersToInvoke` here, and invoke handlers while doing #TheTaylor then.

NTaylorMullen: > You could just skip the temporary handlersToInvoke here, and invoke handlers while doing #TheTaylor then.

Want to capture the handlers prior to invoking them one by one because it's possible for them to remove themselves/add themselves when they're being invoked. That would break the loop entirely.

NTaylorMullen: > if (!_changedHandlers[i].TryGetTarget(out var target))
>     {
>         _changedHandlers.RemoveAt(i);
>         continue;
>     }

lol you really don't like i--

rynowak: Ah I see, that's valuable then. My two small reasonable requests:
1. Document in a comment why it's complicated. You just gave me a good reason.
2. Use a while loop when you're going to manipulate `i`

NTaylorMullen: Will do

src/Microsoft.VisualStudio.Editor.Razor/DefaultEditorSettingsManagerInternal.cs
NTaylorMullen: Thought of an awesome solution. Going to have the `EditorSettingsManagerInternal` track its changed listener. When it has 0 listeners it'll detatch from `EditorSettingsManager`. Basically we're forwarding our lifetime problems to anyone listening to `EditorSettingsManagerInternal`. In this case the document tracker listens to the settings manager internal and it does know when things are tearing down.

This way we can do away with weak references entirely on the `EditorSettingsManager` and we'll still be garbage collected.

rynowak: OK, that sounds like a good solution 👍

src/Microsoft.VisualStudio.Editor.Razor/DefaultWorkspaceEditorSettings.cs
```diff
@@ -0,0 +1,94 @@
...
+                {
+                    // We detatch from the editor settings manager when no one is listening to allow us to be garbage
+                    // collected in the case that the workspace is tearing down.
+                    DetachFromEditorSettingsManager();
+                }
+            }
+        }
+
+        // Internal for testing
+        internal virtual void AttachToEditorSettingsManager()
```
NTaylorMullen: can't wait till we can use `private protected`

rynowak: Last I recall, @anurse was looking into why that isn't the default everywhere for us. I want this feature too, but I don't super want Razor to deviate from the default.

anurse: VSCode doesn't support it yet. Omnisharp does, but there hasn't been a release of the C# extension with it.

anurse: https://github.com/aspnet/BuildTools/issues/566

PR #1985, opened 2018-01-29
Print an error when Razor targets are invoked from a multitargeting build

pranavkm:


pranavkm:
Ported from https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Sdk/buildMultiTargeting/Razor.Sdk.CurrentVersion.MultiTargeting.targets

PR #1988, opened 2018-01-29
Some pre requisites for adding build server integration tests

ajaybhargavb:
#1958 
Meta - #1921 

This PR adds a test fixture to handle build server shutdown after all tests have finished executing.
- Also added a way to pass in pipe name for the server in our tasks
- Added a sample build server integration test. (Will add more soon)
- Renamed a few test infrastructure pieces to make them more generic

FYI @mkArtakMSFT


build/repo.targets
```diff
@@ -13,6 +13,7 @@
     <RazorDesignTestProject>$(RepositoryRoot)test\Microsoft.AspNetCore.Razor.Design.Test\</RazorDesignTestProject>
     <MSBuildLocationFileTemplate>$(RazorDesignTestProject)BuildVariables.cs.template</MSBuildLocationFileTemplate>
     <MSBuildLocationFileOutput>$(RazorDesignTestProject)obj\BuildVariables.generated.cs</MSBuildLocationFileOutput>
+    <RazorToolAssembly>$(RepositoryRoot)src\Microsoft.AspNetCore.Razor.Design\bin\$(Configuration)\netstandard2.0\tools\rzc.dll</RazorToolAssembly>
```
ajaybhargavb: Is there a better way to get the tool assembly path in tests?

rynowak: Take a look at how the test projects are set up. The build creates the folder structure of `Razor.Design` as it would appear after unpacking the nupkg. I'm not sure this is needed at all.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessKind.cs
```diff
@@ -9,12 +9,15 @@
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
-    internal enum MSBuildProcessKind
+    internal enum DotnetProcessKind
```
rynowak: Why change this? I don't get it.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessManager.cs
```diff
@@ -9,40 +9,51 @@
...
+            else if (dotnetProcessKind == DotnetProcessKind.DotnetMsBuild)
             {
                 processStartInfo.FileName = "dotnet";
                 processStartInfo.Arguments = $"msbuild {arguments}";
             }
+            else
+            {
+                processStartInfo.FileName = "dotnet";
+                processStartInfo.Arguments = arguments;
+            }
```
rynowak: Can we separate all of this from the code path that runs MSBuild? Are you sure that running the shutdown command has the same requirments as the MSBuild junk?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessKind.cs
ajaybhargavb: Because I'm reusing these for the shutdown command as well and that has nothing to do with MSBuild

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProcessResult.cs
```diff
@@ -3,9 +3,9 @@
 
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
-    internal class MSBuildResult
+    internal class ProcessResult
```
rynowak: I don't think we should change this. This isn't supposed to be general purpose, it's use in a set of assert methods that target MSBuild explicitly.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessManager.cs
ajaybhargavb: It doesn't need the project directory but it uses the same code to create a process and all that stuff. I thought reusing these made sense

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -103,6 +103,7 @@ void CreateDirectoryProps(string originalProjectRoot, string binariesRoot, strin
     <OriginalProjectRoot>{originalProjectRoot}</OriginalProjectRoot>
     <BinariesRoot>{binariesRoot}</BinariesRoot>
     <_RazorMSBuildRoot>$(OriginalProjectRoot)\..\..\..\src\Microsoft.AspNetCore.Razor.Design\bin\{configuration}\netstandard2.0\</_RazorMSBuildRoot>
+    <_RazorBuildServerPipeName>{BuildServerTestFixture.PipeName}</_RazorBuildServerPipeName>
```
rynowak: Think about just passing this in the command line args to MSBuild. That's much more first class and simple than this.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessKind.cs
rynowak: > that has nothing to do with MSBuild

That's exactly my point

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessManager.cs
rynowak: What parts does it need? I'd like to see this refactored in a more thoughtful way.

ajaybhargavb: > What parts does it need?

Everything in the method past this line. I'll try to refactor it better.

build/repo.targets
ajaybhargavb: I'll take closer look. Just to be clear I want this value to be accessible in C# in my tests so I can execute shutdown command (Not just setting this value and let MSBuild access it).

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
+{
+    public class BuildServerTestFixture : IDisposable
+    {
+        private static readonly TimeSpan _defaultShutdownTimeout = TimeSpan.FromSeconds(60);
+        private static readonly string _pipeName = Guid.NewGuid().ToString();
```
NTaylorMullen: I assume our build server tests don't run in parallel?

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -74,6 +71,7 @@
       DebugTool="$(_RazorDebugTagHelperTool)"
       ToolAssembly="$(_RazorToolAssembly)"
       UseServer="$(UseRazorBuildServer)"
+      PipeName="$(RazorBuildServerPipeName)"
```
rynowak: Can you make this guy private? eg `_RazorToolAssembly`

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.props
```diff
@@ -20,6 +20,9 @@
     <_RazorTaskFolder Condition=" '$(MSBuildRuntimeType)' != 'Core' ">net46</_RazorTaskFolder>
     
     <_RazorTaskAssembly Condition="'$(_RazorTaskAssembly)'==''">$(_RazorMSBuildRoot)\tasks\$(_RazorTaskFolder)\Microsoft.AspNetCore.Razor.Tasks.dll</_RazorTaskAssembly>
+
+    <!-- Used to locate our tools -->
+    <_RazorToolAssembly Condition="'$(_RazorToolAssembly)'==''">$(_RazorMSBuildRoot)tools\rzc.dll</_RazorToolAssembly>
```
rynowak: 👍

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -72,8 +72,10 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!-- Default to on if MvcRazorCompileOnPublish isn't set for some reason -->
     <RazorCompileOnPublish Condition="'$(RazorCompileOnPublish)'==''">true</RazorCompileOnPublish>
-    
-    <UseRazorBuildServer Condition="'$(UseRazorBuildServer)'==''">false</UseRazorBuildServer>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(UseRazorBuildServer)'==''">
+    <UseRazorBuildServer>false</UseRazorBuildServer>
```
rynowak: revert this, it was better bfore

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -0,0 +1,46 @@
...
+                $"/p:RazorCompileOnBuild=true /p:UseRazorBuildServer=true /p:RazorBuildServerPipeName={_pipeName}",
+                msBuildProcessKind: msBuildProcessKind);
+
+            Assert.BuildPassed(result);
+            Assert.FileExists(result, OutputPath, "SimpleMvc.dll");
+            Assert.FileExists(result, OutputPath, "SimpleMvc.pdb");
+            Assert.FileExists(result, OutputPath, "SimpleMvc.PrecompiledViews.dll");
+            Assert.FileExists(result, OutputPath, "SimpleMvc.PrecompiledViews.pdb");
+
+            if (RuntimeEnvironment.OperatingSystemPlatform != Platform.Darwin)
```
rynowak: You can simplify this verification, we don't need to test everything everywhere. It's probably a good simplification to remove the check for the console output

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            // Shutdown the build server.
+            using (var cts = new CancellationTokenSource(_defaultShutdownTimeout))
+            {
+                cts.Token.Register(() =>
+                {
+                    throw new TimeoutException("Shutting down the build server took longer than expected.");
+                });
+
+                var application = new Application(cts.Token);
```
rynowak: AH, running the shutdown command in process. Now that's using your noggin!

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
ajaybhargavb: Really? It was under a property group with `Condition="'$(RazorCompileOnPublish)'==''`. What about the cases when it isn't satisfied?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
+{
+    public class BuildServerTestFixture : IDisposable
+    {
+        private static readonly TimeSpan _defaultShutdownTimeout = TimeSpan.FromSeconds(60);
+
+        public BuildServerTestFixture()
+        {
+            PipeName = Guid.NewGuid().ToString();
```
pranavkm: Would there be benefit in having the name be deterministic name? In the event you want to query the server to debug?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/DotnetProcessKind.cs
```diff
@@ -1,20 +1,14 @@
...
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
-
 namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
 {
     internal enum MSBuildProcessKind
     {
-        /// <summary>dotnet msbuild</summary>
+        /// <summary>dotnet msbuild {args}</summary>
```
pranavkm: Huh?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs
```diff
@@ -110,4 +110,4 @@ void Process_OutputDataReceived(object sender, DataReceivedEventArgs e)
             }
         }
     }
-}
+}
```
pranavkm: New line at EOF

test/Microsoft.AspNetCore.Razor.Design.Test/Microsoft.AspNetCore.Razor.Design.Test.csproj
```diff
@@ -42,12 +42,11 @@
     <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Razor.Design\Microsoft.AspNetCore.Razor.Design.csproj">
       <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
     </ProjectReference>
+    <ProjectReference Include="..\..\src\Microsoft.AspNetCore.Razor.Tools\Microsoft.AspNetCore.Razor.Tools.csproj" />
   </ItemGroup>
 
   <Target Name="EnsureBuildVariablesGeneratedFile">
-    <Error
```
pranavkm: Why?

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
pranavkm: Btw, why doesn't this live in the props file?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
```
NTaylorMullen: To make sure I understand this fixture correctly. The fixture relies on a test booting the rzc server and then the fixture tears it down yes? 

If that's the case I'd suggest booting the server as part of the fixture to not have variance in the tests.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
ajaybhargavb: I was just following the same pattern as `RazorCompileOnBuild` which is also set in the same file.

test/Microsoft.AspNetCore.Razor.Design.Test/Microsoft.AspNetCore.Razor.Design.Test.csproj
ajaybhargavb: VS automatically does this everytime I open this csproj.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
ajaybhargavb: I'm not sure I see the advantage in that. This is only generated once per fixture. And we use the same pipe name in all our tests. Could you give me an example for when that would be useful?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
ajaybhargavb: > To make sure I understand this fixture correctly. The fixture relies on a test booting the rzc server and then the fixture tears it down yes?
Yes

I thought about that. But it seemed better to let the tests do it as it is a real scenario and testing it is also useful. There  won't be much variance between tests other than the time taken.

NTaylorMullen: Testing bootup of rzc is definitely valuable but I'd split that into its own test which doesn't use the fixture.

The problem with relying on a random test booting rzc is if there's issues in booting rzc you'll end up getting false failures for tests. 

Also, by having the boot logic in the fixture it makes the fixture more understandable for future readers.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
rynowak: Ah gotcha. Yeah the right fix for this is to go in the .props file like:
```
<PropertyGroup>
<UseRazorBuildServer Condition="'$(UseRazorBuildServer)'==''">false</UseRazorBuildServer>
</PropertyGroup>
```

You can put it inside another propertygroup as long as that propertygroup doesn't have a condition

Avoid putting conditions on propertygroups unless you have a real need for it.

ajaybhargavb:
🆙 📅 @pranavkm @NTaylorMullen

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -14,6 +15,20 @@ public class BuildServerTestFixture : IDisposable
...
+
+#if DEBUG
+            var configuration = "Debug";
+#elif RELEASE
+            var configuration = "Release";
+#else
+#error Unknown Configuration
+#endif
+
+            var rzcDirectory = Path.GetFullPath($"../../../../../src/Microsoft.AspNetCore.Razor.Design/bin/{configuration}/netstandard2.0/tools/");
```
NTaylorMullen: Just use the sln location while loop that we do elsewhere instead of being super relative

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerTestFixture.cs
```diff
@@ -14,6 +15,20 @@ public class BuildServerTestFixture : IDisposable
...
+            if (!ServerConnection.TryCreateServerCore(rzcDirectory, PipeName))
```
NTaylorMullen: <3

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -0,0 +1,39 @@
...
+        private readonly string _pipeName;
+
+        public BuildServerIntegrationTest(BuildServerTestFixture fixture)
+        {
+            _pipeName = fixture.PipeName;
+        }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public Task Build_SimpleMvc_WithServer_UsingDotnetMSBuild_CanBuildSuccessfully()
```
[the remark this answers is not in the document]
ajaybhargavb: Spoke offline. This will be cleaned up as I add more tests in a new PR.

Issue #1847, opened 2017-12-16
Move path munging in Microsoft.AspNetCore.Razor.GenerateTool into MSBuild 


rynowak:
Right now `Microsoft.AspNetCore.Razor.GenerateTool` does a bunch of path manipulation to figure out what view engine paths and generated file names should be used. This should all live in MSBuild so it can be reused by a future design time build from Razor.

Issue #1829, opened 2017-12-14
Razor MSBuild needs to handle linked files properly


rynowak:
Creating an item to track this because we don't do this correctly today for precompilation. 

This is fundamentally a problem because precompilation looks at files on disk, rather than where a file is linked into the project. 

Links 'just work' when you use runtime compilation, we need to make sure this works well at build-time and in VS.

pranavkm:
Add to this cshtml files that are carried as content inside NuGet packages.

PR #1992, opened 2018-01-30
Use result of AssignTargetPath to calculate output paths

pranavkm:
Instead of performing path mangling, we'll use the established convention of using AssignTargetPath to calculate relative paths \ output paths.

This resolves https://github.com/aspnet/Razor/issues/1847 https://github.com/aspnet/Razor/issues/1829


src/Microsoft.AspNetCore.Razor.Language/RazorTemplateEngine.cs
```diff
@@ -100,6 +95,38 @@ public RazorCSharpDocument GenerateCode(RazorProjectItem projectItem)
             return GenerateCode(codeDocument);
         }
 
+        public virtual RazorCSharpDocument GenerateCode(
+            Func<Stream> streamFactory, 
+            string relativePhysicalPath,
```
pranavkm: Consider `<Content Include="Dir1\Dir2\Index.cshtml" Link="Dir3\NotIndex.cshtml" />`. The intent is to treat the file as if it is was located at `Dir3\NotIndex.cshtml` but read it from it's source location `Dir1\Dir2\Index.cshtml`. i.e. we need to be able to specify `filePath` from the task. We can figure out where this method needs to live, but I was just setting up my expectation

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -185,10 +185,29 @@ Copyright (c) .NET Foundation. All rights reserved.
     </ItemGroup>
 
     <ItemGroup>
-      <RazorGenerate Update="@(RazorGenerate)" Condtion="'%(RazorGenerate.GeneratedOutput)'==''">
```
pranavkm: It took me 20 minutes to figure out there was a typo. Bad @rynowak

test/testapps/SimpleMvc/SimpleMvc.csproj
```diff
@@ -12,4 +12,10 @@
     <!-- Workaround until we get a new SDK build -->
     <ResolvedRazorCompileToolset Condition="'$(RazorCompileToolset)'==''">RazorSDK</ResolvedRazorCompileToolset>
   </PropertyGroup>
+
+  <ItemGroup>
```
pranavkm: Will clean this up. Using this for local testing for now.

src/Microsoft.AspNetCore.Razor.Language/RazorTemplateEngine.cs
NTaylorMullen: I don't think I'd change this class to add support for linked files. This should be part of the RazorProject default implementation. That implementation should be able to have an internal understanding of `FilePath => ProjectItem` and then the ProjectItem itself can have an understanding of if it should read content from a different location.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -116,10 +116,9 @@
       DebugTool="$(_RazorDebugGenerateCodeTool)"
       ToolAssembly="$(_RazorToolAssembly)"
       UseServer="$(UseRazorBuildServer)"
-      Sources="@(RazorGenerate)"
+      Inputs="@(RazorGenerate)"
```
rynowak: Can we keep this called Sources?

.gitignore
```diff
@@ -36,3 +36,5 @@ global.json
 BenchmarkDotNet.Artifacts/
 Microsoft.VisualStudio.RazorExtension.nuget.props
 Microsoft.VisualStudio.RazorExtension.nuget.targets
+msbuild.binlog
+msbuild.log
```
rynowak: 👏

src/Microsoft.AspNetCore.Razor.Language/RazorTemplateEngine.cs
```diff
@@ -53,14 +54,8 @@ public class RazorTemplateEngine
...
-            get
-            {
-                return _options;
-            }
-            set
-            {
-                _options = value ?? throw new ArgumentNullException(nameof(value));
-            }
+            get => _options;
+            set => _options = value ?? throw new ArgumentNullException(nameof(value));
```
rynowak: gross, but I'll allow it.

src/Microsoft.AspNetCore.Razor.Language/RazorTemplateEngine.cs
rynowak: Agree with @NTaylorMullen. I anticipated this, and it needs to be part of the filesystem.

rynowak: We will also need to flow the same information in through the Razor project system into the IDE.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
rynowak: lel

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -185,10 +185,29 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      </RazorGenerate>
+
+      <RazorGenerate Condition="'%(RazorGenerate.GeneratedOutput)'==''">
         <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)%(RelativeDir)%(Filename).cs</GeneratedOutput>
       </RazorGenerate>
     </ItemGroup>
+
+    <ItemGroup>
+      <RazorGenerate Condition="'%(RazorGenerate.ViewEnginePath)' == '' AND '%(RazorGenerate.Link)' != ''">
+        <ViewEnginePath>%(RazorGenerate.Link)</ViewEnginePath>
```
rynowak: I want to come up with another name for this. ViewEngine is an MVC-ism. Ideas?

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -79,10 +92,19 @@ private int ExecuteCore(string projectDirectory, string outputDirectory, string
                 b.Features.Add(new StaticTagHelperFeature() { TagHelpers = tagHelpers, });
             });
 
-            var templateEngine = new MvcRazorTemplateEngine(engine, RazorProject.Create(projectDirectory));
+            
+            var inputItems = GetInputItems(projectDirectory, sources, outputs, relativePaths);
+            var memoryProject = GetMemoryProject(inputItems);
+            var compositeProject = new CompositeRazorProject(
```
pranavkm: ❓

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
pranavkm: `RelativePath` \ `PhysicalRelativePath` \ `RootRelativePath`?

src/Microsoft.AspNetCore.Razor.Language/CompositeRazorProject.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public class CompositeRazorProject : RazorProject
```
rynowak: These should be internal and subclass `RazorProjectSystem`

src/Microsoft.AspNetCore.Razor.Language/MemoryRazorProject.cs
```diff
@@ -0,0 +1,201 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public class MemoryRazorProject : RazorProject
```
rynowak: VirualProjectFileSystem

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
rynowak: `RelativePath` or `RelativePhysicalPath`. I like `RelativePath` a bit better

src/Microsoft.AspNetCore.Razor.Language/MemoryRazorProject.cs
NTaylorMullen: `internal` also think this should probably be in the Razor.Tools project.

src/Microsoft.AspNetCore.Razor.Language/CompositeRazorProject.cs
NTaylorMullen: Should also live in Razor.Tools until we have a business need to bring it over to Razor.Language

src/Microsoft.AspNetCore.Razor.Language/MemoryRazorProject.cs
```diff
@@ -0,0 +1,201 @@
...
+    {
+        private readonly DirectoryNode RootNode = new DirectoryNode("/", "/");
```
NTaylorMullen: I'd yank this guy out of this class and require the `VirtualProjectFileSystem` to be constructed with one. Also would rename it to `VirtualFileSystemHierarchy`

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -97,16 +119,26 @@ private int ExecuteCore(string projectDirectory, string outputDirectory, string
...
 
             return success ? 0 : -1;
         }
 
+        private MemoryRazorProject GetMemoryProject(InputItem[] inputItems)
+        {
+            var project = new MemoryRazorProject();
+            foreach (var item in inputItems)
+            {
+                var projectItem = MemoryRazorProject.CreateProjectItem(item.CreateReadStream, item.PhysicalRelativePath, item.SourcePath, item.FilePath);
```
NTaylorMullen: Could just simplify InputItem to have a `RazorProjectItem` member. It basicccally has all the same information except for the `OutputPath`

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -162,37 +188,43 @@ private OutputItem[] GenerateCode(RazorTemplateEngine templateEngine, IReadOnlyL
...
 
-            public string ViewEnginePath { get; }
+            public string FilePath { get; }
 
             public Stream CreateReadStream()
             {
                 // We are setting buffer size to 1 to prevent FileStream from allocating it's internal buffer
                 // 0 causes constructor to throw
                 var bufferSize = 1;
                 return new FileStream(
```
NTaylorMullen: `VirtualRazorProjectItem` can do all of this instead of taking a stream factory. (heck given the signature here it doesn't need to take any stream argument)

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -18,15 +19,18 @@ internal class GenerateCommand : CommandBase
...
             ProjectDirectory = Option("-p", "project root directory", CommandOptionType.SingleValue);
-            OutputDirectory = Option("-o", "output directory", CommandOptionType.SingleValue);
             TagHelperManifest = Option("-t", "tag helper manifest file", CommandOptionType.SingleValue);
         }
 
-        public CommandArgument Sources { get; }
+        public CommandOption Sources { get; }
 
-        public CommandOption OutputDirectory { get; }
+        public CommandOption Outputs { get; }
```
ajaybhargavb: I might be missing something but why do we need separate output file paths instead of just a directory?

src/Microsoft.AspNetCore.Razor.Tasks/RazorGenerate.cs
```diff
@@ -3,47 +3,74 @@
...
             var builder = new StringBuilder();
 
             builder.AppendLine(Command);
 
-            for (var i = 0; i < Sources.Length; i++)
+            for (var i = 0; i < Inputs.Length; i++)
             {
-                builder.AppendLine(Sources[i]);
+                var input = Inputs[i];
+                builder.AppendLine("-s");
```
ajaybhargavb: In case of CommandOptionType.MultipleValue, do you need to specify this option with every value?

pranavkm:
🆙 📅

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -181,11 +183,17 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ItemGroup Condition="'$(EnableDefaultRazorGenerateItems)'=='true'">
       <RazorGenerate Include="@(Content)" Condition="'%(Content.Extension)'=='.cshtml'" />
     </ItemGroup>
+  </Target>
+
+  <Target Name="AssignRazorGenerateTargetPaths" Condition="'@(RazorGenerate)' != ''">
+    <AssignTargetPath Files="@(RazorGenerate)" RootFolder="$(MSBuildProjectDirectory)">
+      <Output TaskParameter="AssignedFiles" ItemName="RazorGenerateWithTargetPath" />
```
pranavkm: How about we just use `TargetPath` instead of `RelativePath`? The two end up being the same value.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -209,6 +217,21 @@ Copyright (c) .NET Foundation. All rights reserved.
   <Target Name="ResolveRazorCompileInputs" DependsOnTargets="$(ResolveRazorCompileInputsDependsOn)">
   </Target>
 
+  <Target Name="ResolveRazorEmbeddedResources" Condition="'$(EmbedRazorGenerateSources)'=='true'">
```
pranavkm: Moved this from code generation to SDK since it doesn't really require anything from code generation.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
pranavkm: We want MSBuild to tell us where individual files go

src/Microsoft.AspNetCore.Razor.Tools/CompositeRazorProjectFileSystem.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    internal class CompositeRazorProjectFileSystem : RazorProjectFileSystem
```
pranavkm: Need to add unit tests for this

src/Microsoft.AspNetCore.Razor.Tasks/RazorGenerate.cs
```diff
@@ -1,28 +1,44 @@
...
 
 namespace Microsoft.AspNetCore.Razor.Tasks
 {
     public class RazorGenerate : DotNetToolTask
     {
-        [Required]
-        public string[] Sources { get; set; }
+        private const string GeneratedOutput = nameof(GeneratedOutput);
+        private const string TargetPath = nameof(TargetPath);
+        private const string FullPath = nameof(FullPath);
```
rynowak: Not a fan of this usage of nameof. The strings **can't** change. nameof does the opposite of what you want

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
rynowak: We want MSBuild to control all of the path computations. It's the ideal tool for that

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -79,10 +98,18 @@ private int ExecuteCore(string projectDirectory, string outputDirectory, string
...
-            var templateEngine = new MvcRazorTemplateEngine(engine, RazorProject.Create(projectDirectory));
 
-            var sourceItems = GetRazorFiles(projectDirectory, sources);
-            var results = GenerateCode(templateEngine, sourceItems);
+            var inputItems = GetInputItems(projectDirectory, sources, outputs, relativePaths);
+            var compositeProject = new CompositeRazorProjectFileSystem(
+                new[]
+                {
+                    GetVirtualRazorProjectSystem(inputItems),
+                    (RazorProjectFileSystem)RazorProject.Create(projectDirectory),
```
rynowak: Can you add this as static method to `RazorProjectFileSystem`

ajaybhargavb:
Build seems to be failing though

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
```diff
@@ -0,0 +1,200 @@
...
+            return directory?.EnumerateItems() ?? Enumerable.Empty<RazorProjectItem>();
+        }
+
+        public override RazorProjectItem GetItem(string path)
+        {
+            path = NormalizeAndEnsureValidPath(path);
+            return Root.GetItem(path) ?? new NotFoundProjectItem(string.Empty, path);
+        }
+
+        public void Add(RazorProjectItem projectItem)
```
NTaylorMullen: Just to note (in regards to other comments I've left): if you were to lift `DirectoryNode` into its own class as `VirtualFileSystemHierarchy` you could easily test each piece individually in true unit test fashion 😄.

NTaylorMullen: null check project item.

test/Microsoft.AspNetCore.Razor.Language.Test/VirtualRazorProjectFileSystemTest.cs
```diff
@@ -0,0 +1,193 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public class VirtualRazorProjectFileSystemTest
```
NTaylorMullen: Going through these tests a lot of them are testing logic controlled by `DirectoryNode`, aka more integration tests than unit tests. I'm concerned that we're missing unit test coverage / testability of `VirtualRazorProjectFileSystem` inner workings.

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            var filePath = NormalizeAndEnsureValidPath(projectItem.FilePath);
+            Root.AddItem(new FileNode(filePath, projectItem));
+        }
+
+        private class DirectoryNode
```
NTaylorMullen: You should break this out into its own type (or expose internal if you don't want to do that) and test it separately. It has a lot of logic which I'm uncomfortable relying on integration level tests to verify.

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
pranavkm: I'd rather leave `DirectoryNode` an implementation detail for `VirtualRazorProjectFileSystem`. We can still test these regardless.

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
pranavkm: Sure

test/Microsoft.AspNetCore.Razor.Language.Test/VirtualRazorProjectFileSystemTest.cs
pranavkm: Sure

pranavkm:
🆙 📅 

I'll fix up the xplat tests after this test run

pranavkm:
Tests fixed on Travis

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
```diff
@@ -0,0 +1,214 @@
...
+            {
+                throw new ArgumentNullException(nameof(projectItem));
+            }
+
+            var filePath = NormalizeAndEnsureValidPath(projectItem.FilePath);
+            Root.AddFile(new FileNode(filePath, projectItem));
+        }
+
+        [DebuggerDisplay("{Path}")]
+        internal class DirectoryNode
```
NTaylorMullen: leave comment that it's internal for testing.

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
```diff
@@ -0,0 +1,214 @@
...
+                        return currentDirectory;
+                    }
+                }
+
+                return null;
+            }
+        }
+
+        [DebuggerDisplay("{Path}")]
+        internal struct FileNode
```
NTaylorMullen: ditto

Issue #1991, opened 2018-01-30
Shutdown test in ServerLifecycleTest takes forever to run on linux/mac


ajaybhargavb:
https://travis-ci.org/aspnet/Razor/jobs/334963674#L1145

ajaybhargavb:
Need to revert 399216730a4351dd614a690bdeb948ad0b37e8fe after fixing this.

ajaybhargavb:
7e18efbf1fdfdf5c30672cb396218ebea9b406b3

PR #1994, opened 2018-01-31
Use ManualResetEventSlim instead of TaskCompletionSource

ajaybhargavb:
https://github.com/aspnet/Razor/issues/1991

Doing this fixes the flakiness in two tests.

Better to review using https://github.com/aspnet/Razor/pull/1994/files?w=1

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
@@ -116,82 +116,86 @@ namespace Microsoft.AspNetCore.Razor.Tools
         /// A shutdown request should not abort an existing compilation.  It should be allowed to run to 
         /// completion.
         /// </summary>
-        // Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991
-        [ConditionalFact]
-        [OSSkipCondition(OperatingSystems.Linux)]
-        [OSSkipCondition(OperatingSystems.MacOSX)]
+        [Fact]
         public async Task ServerRunning_ShutdownRequest_DoesNotAbortCompilation()
         {
             // Arrange
-            var completionSource = new TaskCompletionSource<bool>();
-            var host = CreateCompilerHost(c => c.ExecuteFunc = (req, ct) =>
+            using (var startedMre = new ManualResetEventSlim(initialState: false))
+            using (var finishedMre = new ManualResetEventSlim(initialState: false))
             {
-                // We want this to keep running even after the shutdown is seen.
-                completionSource.Task.Wait();
-                return EmptyServerResponse;
-            });
+                var host = CreateCompilerHost(c => c.ExecuteFunc = (req, ct) =>
+                {
+                    // We want this to keep running even after the shutdown is seen.
+                    startedMre.Set();
+                    finishedMre.Wait();
+                    return EmptyServerResponse;
+                });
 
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -116,82 +116,86 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
...
             {
-                // We want this to keep running even after the shutdown is seen.
-                completionSource.Task.Wait();
-                return EmptyServerResponse;
-            });
+                var host = CreateCompilerHost(c => c.ExecuteFunc = (req, ct) =>
+                {
+                    // We want this to keep running even after the shutdown is seen.
+                    startedMre.Set();
+                    finishedMre.Wait();
```
pranavkm: This should have some reasonable timeout

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -116,82 +116,86 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
...
-                    Assert.Equal(Process.GetCurrentProcess().Id, processId);
-                    Assert.False(compileTask.IsCompleted);
-                }
+                    var compileTask = ServerUtilities.Send(serverData.PipeName, EmptyServerRequest);
+                    startedMre.Wait();
 
-                // Now let the task complete.
-                completionSource.SetResult(true);
+                    // Act
+                    for (var i = 0; i < 10; i++)
```
pranavkm: Why is this in a loop?

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -116,82 +116,86 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
...
+                    {
+                        // The compilation is now in progress, send the shutdown.
+                        var processId = await ServerUtilities.SendShutdown(serverData.PipeName);
+                        Assert.Equal(Process.GetCurrentProcess().Id, processId);
```
pranavkm: Out of curiosity, what's the significance of this assertion?

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
ajaybhargavb: This is a test for testing multiple shutdown requests.

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
ajaybhargavb: I think this is to make sure we shutdown the same process that we started and not some unrelated server that was running. Fyi these tests were ported from https://github.com/dotnet/roslyn/blob/master/src/Compilers/Server/VBCSCompilerTests/CompilerServerApiTest.cs#L387-L513

pranavkm: Could you add a comment to this effect? In the event this ever fails, it'll be more obvious what it is that failed.

ajaybhargavb:
Closing this as the correct fix is here https://github.com/aspnet/Razor/pull/2128

Issue #1989, opened 2018-01-30
Add WorkspaceProvider as a means to go from ITextBuffer => Workspace


NTaylorMullen:
We have several API's which today utilize `VisualStudioWorkspaceAccessor` to get Workspace services. These API's can't use that type anymore given that we're moving away from assuming 1 workspace. Instead we should have a workspace provider which allows us to go from text buffer to workspace and then dynamically grab any services we need off of a workspace retrieved via TextBuffer.

NTaylorMullen:
6d2460ae7fe6631b956750c31d080950ab77d6dd
fff1c01aa06d905fa1e9f36dd5375061fac167ab

PR #1996, opened 2018-01-31
Update DefaultRazorEditorFactoryService to be per-workspace (in Windows)

NTaylorMullen:
- Added a `WorkspaceProvider` API to enable the factory to retrieve a workspace given a text buffer.
- Added a way to add test services to `AdhocWorkspace` so we can test against services being retrieved from a `Workspace`. This will be much more common once we rely on services coming from `TextBuffer`s in our other tooling pieces.
- Added tests for the default workspace provider.

**Note:** This PR lays the groundwork for enabling the remaining Razor pieces to not use `VisualStudioWorkspaceAccessor` and instead rely on getting workspaces given TextBuffers and the like.

#1989 

/cc @mkArtakMSFT

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorEditorFactoryService.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorEditorFactoryService.cs
@@ -14,10 +14,7 @@ namespace Microsoft.VisualStudio.Editor.Razor
     internal class DefaultRazorEditorFactoryService : RazorEditorFactoryService
     {
         private static readonly object RazorTextBufferInitializationKey = new object();
-
-        private readonly VisualStudioDocumentTrackerFactory _documentTrackerFactory;
-        private readonly VisualStudioRazorParserFactory _parserFactory;
-        private readonly BraceSmartIndenterFactory _braceSmartIndenterFactory;
+        private readonly VisualStudioWorkspaceAccessor _workspaceAccessor;
 
         [ImportingConstructor]
         public DefaultRazorEditorFactoryService(VisualStudioWorkspaceAccessor workspaceAccessor)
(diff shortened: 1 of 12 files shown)


NTaylorMullen:
This code ended up being pretty significant so getting out the Windows version now before it gets too unreviewable.

src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorEditorFactoryService.cs
```diff
@@ -72,7 +40,12 @@ public override bool TryGetDocumentTracker(ITextBuffer textBuffer, out VisualStu
                 return false;
             }
 
-            EnsureTextBufferInitialized(textBuffer);
+            var textBufferInitialized = EnsureTextBufferInitialized(textBuffer);
```
rynowak: Can you call this `TryInitializeTestBuffer` - it's not really an 'Ensure' if it sometimes fails

NTaylorMullen: Can do.

src/Microsoft.VisualStudio.LanguageServices.Razor/DefaultWorkspaceProvider.cs
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+            return true;
+        }
+
+        // Internal virtual for testing
+        internal virtual bool TryGetWorkspaceFromHierarchy(ITextBuffer textBuffer, out Workspace workspace)
+        {
+            // If there's no document we can't find the FilePath, or look for a matching hierarchy.
+            if (!_documentFactory.TryGetTextDocument(textBuffer, out var textDocument))
```
NTaylorMullen: Instead of do all this magic here, lets use the TextBufferProjectService.

PR #2000, opened 2018-02-01
Fix #1999

### LINKED ISSUE #1999: Error on publish with .cshtml content files from a .nupkg
Reported to me via email by @vijayrkn 

-----------------

We are gettng the following error from Razor sdk while trying to publish an ASP.NET Core 2.1 (.NET framework) application.  Full log attached

C:\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.aspnetcore.razor.design\2.1.0-preview1-28181\build\netstandard2.0\Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets(102,5): Error MSB3191: Unable to create directory "obj\Release\net461\Razor\C:\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.aspnetcore.identity.ui\2.1.0-preview1-28181\contentFiles\any\netstandard2.0\Areas\Identity\Pages\Account\". The given path's format is not supported.

Repro steps:
1. File > New Project >C# > Web >ASP.NET Core Web Application >.Net Framework, ASP.NET Core 2.1,Web Application > Change Authentication to Individual User Accounts > OK.
2. Right click the project, rebuild.
3. Publish

----------------------------

There's a bug in this build of identity that cause all of the .cshtml files to be included in the .nupkg. Because of that we've had the opportunity to find a bug in the Razor build support.

We don't handle files correctly here: https://github.com/aspnet/Razor/blob/release/2.1/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets#L86

MvcPrecompilation ignores files that aren't under the project's root directory. We planned to have the same behavior for 2.1.0-preview1 and add proper support for this scenario in 2.1.0-preview2 but it looks like we're not very good at ignoring these files.

rynowak:
The fix for this for preview1 is to ignore any files with an absolute path. MvcPrecompilation
ignores files outside the project root, and we're aiming for parity.

This will have a proper fix in preview2

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.targets
@@ -167,8 +167,16 @@
       <RazorGenerate Include="@(Content)" Condition="'%(Content.Extension)'=='.cshtml'" />
     </ItemGroup>
 
-    <ItemGroup>
-      <RazorGenerate Update="@(RazorGenerate)" Condtion="'%(RazorGenerate.GeneratedOutput)'==''">
+    <!--
+      For preview1 we only support files that are within the project's directory hierarchy. 
+
+      This is similar to to a limitation in MvcPrecompilation. This will be improved for preview2,
+      tracked by https://github.com/aspnet/Razor/issues/1829
+    -->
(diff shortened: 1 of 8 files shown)


pranavkm:
When merging to dev, can you do `git merge -s ours` and keep the work item open? We don't need any of this code in preview2 and we'll just write tests to make sure the scenario works.

rynowak:
I'm going to merge the test to dev, but then skip it and undo the product changes

Issue #2001, opened 2018-02-01
Once we support multiple Workspaces remove the concept of VisualStudioWorkspaceAccessor.Workspace


NTaylorMullen:
This is mostly to ensure we don't forget to do this. We've gone through all sorts of work to make our services not expect a "primary" workspace and I'd like to ensure we don't make that mistake again. Therefore, we should remove `VisualStudioWorkspaceAccessor.Workspace` once #1973 is completed.

NTaylorMullen:
ca15840d05994521101a86f7d364832a15bfd9dd

PR #2002, opened 2018-02-01
Add Mac implementation of VisualStudioWorkspaceAccessor new APIs.

NTaylorMullen:
- The Mac implementation does not have fallback logic like the windows variant because Workspaces in VS mac world are spun up / torn down regularly. We don't want to be tracking down bugs that involve us adding services or Razor logic to Workspaces that have nothing to do with Razor.
- Added a test for the `DefaultVisualStudioWorkspaceAccessor`. Could not add other tests due to limitations of MonoDevelop's abstractions.

#1989

### DIFF
--- a/src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultVisualStudioWorkspaceAccessor.cs
+++ b/src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultVisualStudioWorkspaceAccessor.cs
@@ -1,11 +1,13 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.ComponentModel.Composition;
-using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/DefaultVisualStudioWorkspaceAccessor.cs
```diff
@@ -1,28 +1,77 @@
...
+            if (projectService == null)
+            {
+                throw new ArgumentNullException(nameof(projectService));
+            }
+
+            _projectService = projectService;
         }
 
-        public override Workspace Workspace { get; }
+        public override Workspace Workspace => TypeSystemService.Workspace;
```
rynowak: Just for clarity, I'm assuming you plan to delete this property once you remove all of the uses of it right?

NTaylorMullen: Yup, that's what [this](https://github.com/aspnet/Razor/issues/2001) workitem tracks.

Issue #2004, opened 2018-02-01
Change DefaultTagHelperFactsService to not expect a single "primary" workspace.


NTaylorMullen:
Today the `DefaultTagHelperFactsService` consumes the `VisualStudioWorkspace` in an effort to pull out 
 a `TagHelperFactsServiceInternal`. Given the APIs the `TagHelperFactsService` has it doesn't currently have a way to get a `Workspace` given a request; we'll need to pursue other routes of having the same API but not relying on a primary workspace for service dependencies.

NTaylorMullen:
4eea2f6992d9e19524e786c1261bb28dde53ee38

PR #2005, opened 2018-02-02
Change DefaultTagHelperFactsService to not expect a primary workspace.

NTaylorMullen:
- Removed `TagHelperFactsServiceInternal` since we had no way of retrieving the internal Workspace service given the exposed public API. Not to mention I think the `TagHelperFactsService` made more sense as a MEF service anyhow.
- Moved `TagHelperFactsServiceInternal` tests to `Microsoft.VisualStudio.Editor.Razor.Test` project and changed them to utilize the non-"internal" version.
- Updated completion service API to rely on non-`TagHelperFactsServiceInternal` pieces.
- Tested the updated bits in this PR in VS.

#2004 

/cc @mkArtakMSFT


src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperFactsServiceInternal.cs
```diff
@@ -1,20 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Collections.Generic;
-using Microsoft.AspNetCore.Razor.Language;
-using Microsoft.CodeAnalysis.Host;
-
-namespace Microsoft.CodeAnalysis.Razor
-{
-    internal abstract class TagHelperFactsServiceInternal : ILanguageService
```
rynowak: I'm wonder if we will eventually pivot back the other way on where the implementation of these types live, but I'm going to trust you on this for now.

PR #2012, opened 2018-02-02
Add prelimianry support for extensions to Razor

rynowak:
This PR adds MSBuild insfrastructure to the SDK that can understand
concepts we need to expose to the project, code generator and runtime
like:
- Language version
- Configuration
- Extensions (plugins)

As an example of how this works, I've done the wireup for MVC. This will
now generate assembly attributes in your application that can act as a
source-of-truth for what should be included in runtime compilation, and
it's all based on the project-file. This means that it can be delivered
and configured by packages.

The next step here is to implement a loader for RazorProjectEngine based
on these primitives, and then use it in our CLI tools and MVC.

The next step after that is to expose it in VS and VS4Mac through the
project system.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
@@ -25,8 +25,8 @@
 
   <ItemGroup>
     <RazorExtension Include="MVC-2.1">
-      <AssemblyName>Microsoft.AspNetCore.Razor.Extensions</AssemblyName>
-      <AssemblyFilePath>$(MSBuildThisFileDirectory)..\..\lib\netstandard2.0\Microsoft.AspNetCore.Razor.Extensions.dll</AssemblyFilePath>
+      <AssemblyName>Microsoft.AspNetCore.Mvc.Razor.Extensions</AssemblyName>
+      <AssemblyFilePath>$(MSBuildThisFileDirectory)..\..\lib\netstandard2.0\Microsoft.AspNetCore.Mvc.Razor.Extensions.dll</AssemblyFilePath>
     </RazorExtension>
   </ItemGroup>
 </Project>
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
@@ -1,0 +1,32 @@
+<Project>
+  <!-- 
+      MSBuild support for Razor code generation that targets ASP.NET Core MVC 2.X
+
+      The properties and items here are designed to be read by CPS so they should be just simple evaluation-time values
+      and should not require targets to initialize. 
+  -->
+  <PropertyGroup>
+    <!-- 
+      Set the primary configuration supported by this pacakge as the default configuration for Razor. 
+    -->
+    <RazorDefaultConfiguration Condition="'$(RazorDefaultConfiguration)'==''">MVC-2.1</RazorDefaultConfiguration>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--
+      While technically the assembly in this package can provide support for the MVC-2.0 configuration, don't declare
+      it here. The IDE is hardcoded to inject 2.0 support when needed. The settings flowing through MSBuild should reflect
+      the project's runtime.
+    -->
+    <RazorConfiguration Include="MVC-2.1">
+      <Extensions>MVC-2.1;$(CustomRazorExtension)</Extensions>
+    </RazorConfiguration>
+  </ItemGroup>
+
+  <ItemGroup>
+    <RazorExtension Include="MVC-2.1">
+      <AssemblyName>Microsoft.AspNetCore.Razor.Extensions</AssemblyName>
+      <AssemblyFilePath>$(MSBuildThisFileDirectory)..\..\lib\netstandard2.0\Microsoft.AspNetCore.Razor.Extensions.dll</AssemblyFilePath>
+    </RazorExtension>
+  </ItemGroup>
+</Project>
--- a/src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
+++ b/src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
@@ -9,8 +9,4 @@
     <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsHashCodeCombinerSourcesPackageVersion)" />
   </ItemGroup>
 
-  <ItemGroup>
-    <Folder Include="Hosting\" />
-  </ItemGroup>
-
 </Project>
--- a/src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
+++ b/src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
@@ -9,4 +9,8 @@
     <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsHashCodeCombinerSourcesPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Hosting\" />
+  </ItemGroup>
+
 </Project>
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
@@ -2,23 +2,57 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
     public sealed class RazorConfiguration
     {
-        public static readonly RazorConfiguration Default = new RazorConfiguration(RazorLanguageVersion.Latest, designTime: false);
+        public static readonly RazorConfiguration Default = new RazorConfiguration(
+            RazorLanguageVersion.Latest, 
+            "unnamed",
+            Array.Empty<RazorExtension>(),
+            designTime: false);
 
-        public RazorConfiguration(RazorLanguageVersion languageVersion, bool designTime)
+        // This is used only in some back-compat scenarios. We don't expose it because there's no
+        // use case for anyone else to use it.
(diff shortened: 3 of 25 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
```diff
@@ -2,24 +2,58 @@
...
+            designTime: false);
 
-        public RazorConfiguration(RazorLanguageVersion languageVersion, bool designTime)
+        // This is used only in some back-compat scenarios. We don't expose it because there's no
+        // use case for anyone else to use it.
+        internal static readonly RazorConfiguration DefaultDesignTime = new RazorConfiguration(
+            RazorLanguageVersion.Latest,
+            "unnamed",
+            Array.Empty<RazorExtension>(),
+            designTime: true);
```
rynowak: @NTaylorMullen - design time being a part of the configuration is making me sad.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -2,32 +2,107 @@
...
+            {
+                version = Version_1_0;
+                return true;
+            }
+
+            version = null;
+            return false;
+        }
+
+        public static RazorLanguageVersion Parse(string languageVersion)
```
rynowak: @NTaylorMullen so aspects of your original design turned out to be problematic once parsing is introduced. Your ToString() doesn't roundtrip and if latest == 2.1 then that's a hard problem to solve.

I went a different way, open to feedback.

In this model, you use IComparable rather than equals to tell if the version supports a feature. So equality still is just reference equality.

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -45,6 +120,9 @@ public override int GetHashCode()
             return base.GetHashCode();
         }
 
-        public override string ToString() => $"Razor '{Major}.{Minor}'";
+        // We need this to round-trip with TryParse
+        public override string ToString() => Text;
+
+        private string DebuggerToString() => $"Razor 'Text'";
```
rynowak: OOps lol will fix.

test/testapps/AppWithP2PReference/AppWithP2PReference.csproj
```diff
@@ -1,10 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+<Project>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk.Web" />
```
rynowak: @pranavkm I ran into some problems when I started messing with the import order. It got a lot more complicated once I was no longer using Directory.props to import things.

However what's here now basically will reflect the final state once we're integrated with the CLI.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
```diff
@@ -0,0 +1,32 @@
...
+      the project's runtime.
+    -->
+    <RazorConfiguration Include="MVC-2.1">
+      <Extensions>MVC-2.1;$(CustomRazorExtension)</Extensions>
+    </RazorConfiguration>
+  </ItemGroup>
+
+  <ItemGroup>
+    <RazorExtension Include="MVC-2.1">
+      <AssemblyName>Microsoft.AspNetCore.Razor.Extensions</AssemblyName>
```
NTaylorMullen: Is this a new assembly that doesn't exist yet or should it be Mvc.Extensions? Same q below about assembly file path

src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
```diff
@@ -9,4 +9,8 @@
     <PackageReference Include="Microsoft.Extensions.HashCodeCombiner.Sources" PrivateAssets="All" Version="$(MicrosoftExtensionsHashCodeCombinerSourcesPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Hosting\" />
```
NTaylorMullen: Why do we need this?

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
NTaylorMullen: Hmm should we just do away with "Latest" then? We could always force people to specify the version. The only piece I don't like is that `RazorLanguageVersion.Latest != RazorLanguagVersion.Version_2_1`. I guess we could always override the equality operator.

src/Microsoft.AspNetCore.Razor.Language/RazorExtension.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public abstract class RazorExtension
+    {
+        public abstract string ExtensionName { get; }
```
NTaylorMullen: I imagine this would eventually have a way to configure a passed in RazorProjectEngine yes?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
rynowak: Oh yeah this is totes wrong. Will fix it

src/Microsoft.AspNetCore.Razor.Language/Microsoft.AspNetCore.Razor.Language.csproj
rynowak: Nope. VS did this

src/Microsoft.AspNetCore.Razor.Language/RazorExtension.cs
rynowak: perhaps

src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
rynowak: I want to make sure that Latest can be parsed. If you're OK with not roundtripping latest maybe your original design is OK. I might just revert all of this, let me think about it a bit more.

pranavkm: This seems good. We're consistent with how Roslyn does this while being better in that you don't have to call `GetEffectiveVersion`

src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorConfigurationNameAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+    public sealed class RazorConfigurationNameAttribute : Attribute
+    {
+        /// <summary>
+        /// Creates a new instance of <see cref="RazorConfigurationNameAttribute"/>.
+        /// </summary>
+        /// <param name="configurationName">The name of the Razor configuration.</param>
+        public RazorConfigurationNameAttribute(string configurationName)
+        {
+            if (configurationName == null)
```
pranavkm: Throwing from an attribute ctor is generally bad. It ends up throwing when you construct it e.g. GetAttributes() as opposed to when you consume it.

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
```diff
@@ -161,6 +161,46 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+    Name="RazorGetAssemblyAttributes" 
+    AfterTargets="GetAssemblyAttributes"
+    Condition="'$(GenerateRazorAssemblyInfo)'=='true' and '$(RazorDefaultConfiguration)'!=''">
+
+    <ItemGroup>
+      <_ResolvedRazorConfiguration Include="@(RazorConfiguration)" Condition="'%(RazorConfiguration.Identity)'=='$(RazorDefaultConfiguration)'" />
+      <_ResolvedRazorExtensionName Include="%(_ResolvedRazorConfiguration.Extensions)"/>
+    </ItemGroup>
+
+    <FindInList List="@(RazorExtension)" ItemSpecToFind="%(_ResolvedRazorExtensionName.Identity)">
```
pranavkm: Inline it - `<FindInList List="@(RazorExtension)" ItemSpecToFind="%(_ResolvedRazorConfiguration.Identity)">`?

test/testapps/Directory.Build.targets
```diff
@@ -1,3 +1,2 @@
 <Project>
```
pranavkm: Hmm, we're missing picking up the root Directory.Buid.targets (which overrides the runtime framework version). Can you have this import `SolutionRoot\Directory.Build.targets`?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
```diff
@@ -0,0 +1,32 @@
+<Project>
```
pranavkm: `ToolsVersion`

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props
```diff
@@ -0,0 +1,32 @@
...
+    <RazorDefaultConfiguration Condition="'$(RazorDefaultConfiguration)'==''">MVC-2.1</RazorDefaultConfiguration>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!--
+      While technically the assembly in this package can provide support for the MVC-2.0 configuration, don't declare
+      it here. The IDE is hardcoded to inject 2.0 support when needed. The settings flowing through MSBuild should reflect
+      the project's runtime.
+    -->
+    <RazorConfiguration Include="MVC-2.1">
```
pranavkm: Is this required in a cross-targeting project?

src/Microsoft.AspNetCore.Razor.Sdk/build/netstandard2.0/Razor.Sdk.CurrentVersion.targets
rynowak: This is intended. `_ResolvedRazorConfiguration` is a list that gets created from a single string.

test/testapps/Directory.Build.targets
rynowak: I added this but then had to take it out. It's causing some test failures and it's unrelated to my work. 

It's specifically a non-goal to make these project stay "up to date" with out build system.

PR #2020, opened 2018-02-04
Add HostProject and update ProjectSnapshotManager

rynowak:
This is a somewhat complex addition to the ProjectSnapshotManager. Now
that we accept updates from the underlying IDE project system we need to
coordinate those with the Workspace.

This means that ProjectSnapshot itself now also has a version concept.


build/dependencies.props
```diff
@@ -45,6 +45,7 @@
     <VSIX_MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
     <VSIX_MicrosoftCodeAnalysisEditorFeaturesTextPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisEditorFeaturesTextPackageVersion>
     <VSIX_MicrosoftCodeAnalysisRemoteRazorServiceHubPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisRemoteRazorServiceHubPackageVersion>
+    <VSIX_MicrosoftCodeAnalysisVisualBasicWorkspacesPackageVersion>2.6.0-beta1-62023-02</VSIX_MicrosoftCodeAnalysisVisualBasicWorkspacesPackageVersion>
```
rynowak: Added this so we could test the case of using a VB project. This was missing from our tests.

test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/WorkspaceProjectSnapshotChangeTriggerTest.cs
```diff
@@ -17,12 +17,38 @@ public WorkspaceProjectSnapshotChangeTriggerTest()
...
+            var projectId3 = ProjectId.CreateNewId("Three");
+
+            SolutionWithTwoProjects = Workspace.CurrentSolution
+                .AddProject(ProjectInfo.Create(
+                    projectId1,
+                    VersionStamp.Default,
+                    "One",
+                    "One",
+                    LanguageNames.CSharp,
+                    filePath: "One.csproj"))
```
rynowak: There's a lot of churn in the tests because we need file paths now.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -7,6 +7,18 @@
...
+    // immutable view of the underlying project systems.
+    //
+    // The HostProject support all of the configuration that the Razor SDK exposes via the project system
+    // (language version, extensions, named configuration).
+    //
+    // The WorkspaceProject is needed to support our use of Roslyn Compilations for Tag Helpers and other
+    // C# based constructs.
+    //
+    // The somewhat complex task performed here is to unify these two concepts (both independently stateful)
+    // and also to provide a facility for computed state based on these two concepts.
```
rynowak: Think of this PR as a rewrite of this class. The state that this class has to manage is much more sophisticated now.

rynowak:
Note that this PR on it's own breaks the VS experience. We need something that will actually initialize the `HostProject` or else tag helpers are broken. That will come next, but I wanted to try and keep the size managable.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -15,32 +15,84 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
...
+
+        public DefaultProjectSnapshot(Project workspaceProject, VersionStamp? version = null)
+        {
+            if (workspaceProject == null)
+            {
+                throw new ArgumentNullException(nameof(workspaceProject));
+            }
+
+            WorkspaceProject = workspaceProject;
+            FilePath = workspaceProject.FilePath;
```
NTaylorMullen: Can we ever get into a state were the `FilePath` on `workspaceProject` is `null`? Asking because I'm trying to determine if it'd be valuable to solidify our system to understand snapshots that have a `null` file path (no host project or workspace project).

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -57,7 +69,8 @@ internal class DefaultProjectSnapshotManager : ProjectSnapshotManagerBase
             _triggers = triggers.ToArray();
             Workspace = workspace;
 
-            _projects = new Dictionary<ProjectId, DefaultProjectSnapshot>();
+            _projects = new Dictionary<string, DefaultProjectSnapshot>();
```
NTaylorMullen: add comparer

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -74,119 +87,234 @@ public override IReadOnlyList<ProjectSnapshot> Projects
...
-            if (underlyingProject == null)
+            if (hostProject == null)
             {
-                throw new ArgumentNullException(nameof(underlyingProject));
+                throw new ArgumentNullException(nameof(hostProject));
             }
 
-            var snapshot = new DefaultProjectSnapshot(underlyingProject);
-            _projects[underlyingProject.Id] = snapshot;
+            // It's possible that the workspace project was already initialized, and the host project is showing up
```
NTaylorMullen: Good to be extra cautious but out of curiosity when would this be possible? From my understanding Roslyn reacts to the pieces being added to the host project system.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: Definitely super sophisiticated I agree. Was a skeptical at first but the more I think about/see what you did, I'm liking it a lot.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -74,119 +87,234 @@ public override IReadOnlyList<ProjectSnapshot> Projects
...
+
+            if (!IsSupportedWorkspaceProject(workspaceProject))
             {
-                _projects.Remove(underlyingProject.Id);
+                return;
+            }
 
-                // We need to notify listeners about every project removal.
-                NotifyListeners(new ProjectChangeEventArgs(snapshot, ProjectChangeKind.Removed));
+            // We also need to check the projectId here. If this is a multi-targeting project then we are only interested
```
NTaylorMullen: Hmm, definitely not necessary yet BUT given that we don't have a drop down selection of what Razor TFM to use we should probably eventually be picking the TFM that matches the one in the editor.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostProject.cs
```diff
@@ -0,0 +1,30 @@
...
+                throw new ArgumentNullException(nameof(languageVersion));
+            }
+
+            FilePath = projectFilePath;
+            LanguageVersion = languageVersion;
+        }
+
+        public string FilePath { get; }
+
+        public string LanguageVersion { get; }
```
NTaylorMullen: Should this be `RazorLanguageVersion`?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotWorkerQueue.cs
```diff
@@ -40,7 +39,7 @@ public ProjectSnapshotWorkerQueue(ForegroundDispatcher foregroundDispatcher, Def
             _projectManager = projectManager;
             _projectWorker = projectWorker;
 
-            _projects = new Dictionary<ProjectId, Project>();
+            _projects = new Dictionary<string, ProjectSnapshotUpdateContext>();
```
NTaylorMullen: comparer

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/WorkspaceProjectSnapshotChangeTrigger.cs
```diff
@@ -23,51 +23,60 @@ private void InitializeSolution(Solution solution)
...
             {
                 case WorkspaceChangeKind.ProjectAdded:
                     {
-                        underlyingProject = e.NewSolution.GetProject(e.ProjectId);
-                        Debug.Assert(underlyingProject != null);
+                        project = e.NewSolution.GetProject(e.ProjectId);
+                        Debug.Assert(project != null);
 
-                        if (underlyingProject.Language == LanguageNames.CSharp)
+                        if (DefaultProjectSnapshotManager.IsSupportedWorkspaceProject(project))
```
NTaylorMullen: The snapshot manager checks this anyways. I'd un-static the method and leave that checking to the manager. Same elsewhere

rynowak:
> No massive comments. However, could we leave this PR open (unmerged) until the VS experience is working again? I frequently use local Razor bits in VS to test our pieces

My plan is to merge things into vs15.7 when they are in good working order. 

Once that's done you and I should stage a merge to dev that includes the VS4Mac changes.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
rynowak: This is all in flux. Right now the VS contained language infrastructure doesn't support it.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
rynowak: You can deffo have projects in the workspace that don't have a path, but we don't support them.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
rynowak: Another design idea that I had was to make it so that you **always** have a HostProject when you have a snapshot. That simplifies things quite a bit.
 
Basically you may or may not have a Workspace Project, but you always have a HostProject. Thoughts?

NTaylorMullen:
> My plan is to merge things into vs15.7 when they are in good working order.

Oh derp, didn't realize 157. Ya go for it.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: > Basically you may or may not have a Workspace Project, but you always have a HostProject. Thoughts?

Yup, that's what I was getting at 😄

rynowak:
Feel free to take a look at the other PRs here as well. They are super relevant.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
rynowak: @NTaylorMullen - so let's make a decision then.. Should we discuss this tomorrow when I get into the office or are you convinced?

Mainly I want to have confidence that building the VS4Mac version of this is not going to change the design **again**. Can you take a gander at what you have to work with in VS4Mac and make a judegement call?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
rynowak: I also considered us associating multiple workspace projects with a single HostProject.

rynowak: However, all of these ideas add substantial complexity and it's not clear what they solve. If the contained language service only supports a single TFM that's an absolute blocker.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostProject.cs
rynowak: Nah, i think it's apparant what it means.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/WorkspaceProjectSnapshotChangeTrigger.cs
rynowak: Sure

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: > Should we discuss this tomorrow when I get into the office or are you convinced?

I'm convinced. I'm just curious how you want to do it is all. All I know is that I don't want to bring back `CreateDesignTime` on `RazorEngine` / `RazorProjectEngine` 😄

rynowak:
My plan is to leave this PR open until we're ready to do a merge and insertion to VS 15.7 so that we can keep the 15.7 branch clean. I'll be merging my other PRs into this branch so that I can use it as a base.

PR #2021, opened 2018-02-04
Add prelimianry support for extensions to Razor (#2012)

rynowak:
* Add prelimianry support for extensions to Razor

This PR adds MSBuild insfrastructure to the SDK that can understand
concepts we need to expose to the project, code generator and runtime
like:
- Language version
- Configuration
- Extensions (plugins)

As an example of how this works, I've done the wireup for MVC. This will
now generate assembly attributes in your application that can act as a
source-of-truth for what should be included in runtime compilation, and
it's all based on the project-file. This means that it can be delivered
and configured by packages.

The next step here is to implement a loader for RazorProjectEngine based
on these primitives, and then use it in our CLI tools and MVC.

The next step after that is to expose it in VS and VS4Mac through the
project system.

(cherry picked from commit 5b28c06d6453696e98fb7bf32abe3aef09a5c26b)

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Properties/Resources.Designer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Properties/Resources.Designer.cs
@@ -822,6 +822,20 @@ namespace Microsoft.AspNetCore.Razor.Language
         internal static string FormatSectionDirective_NameToken_Name()
             => GetString("SectionDirective_NameToken_Name");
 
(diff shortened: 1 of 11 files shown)


rynowak:
FYI only. This is a cherry-pick to the vs15.7 branch for our 2.1 tooling work.

rynowak:
Also, I can't spell preliminary. But I'm leaving it.

PR #2022, opened 2018-02-05
Implement a project system using CPS

rynowak:


src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+                await _commonServices.ThreadingService.SwitchToUIThread();
+                var projectManager = GetProjectManager();
+
+                var razorProperties = await _commonServices.ActiveConfiguredProjectRazorProperties.GetRazorGeneralPropertiesAsync().ConfigureAwait(false);
+                var languageVersion = await razorProperties.RazorLangVersion.GetEvaluatedValueAtEndAsync().ConfigureAwait(false);
+
+                if (string.IsNullOrEmpty(languageVersion))
+                {
+                    languageVersion = "2.1";
+                }
```
rynowak: We don't currently handle the case where a project file hasn't set the language version

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/Rules/RazorGeneral.cs
```diff
@@ -0,0 +1,212 @@
+//------------------------------------------------------------------------------
```
rynowak: We have to check this in because the task that creates it doesn't support Core MSBuild 😢

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+        private readonly AsyncSemaphore _lock;
+
+        private ProjectSnapshotManagerBase _projectManager;
+        private HostProject _current;
+        private IDisposable _subscription;
+
+        [ImportingConstructor]
+        public RazorProjectHost(
+            IUnconfiguredProjectCommonServices commonServices,
+            [Import(typeof(VisualStudioWorkspace))] Workspace workspace)
```
NTaylorMullen: Hmmm, does the Roslyn project system end up doing this? Only reason why I ask is because of all the work we just did to not use the VisualStudioWorkspace 😄.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+            // Don't try to evaluate any properties here since the project is still loading and we require access
+            // to the UI thread to push our updates.
+            //
+            // Just subscribe and handle the notification later.
+            var receiver = new ActionBlock<IProjectVersionedValue<IProjectSubscriptionUpdate>>(OnProjectChanged);
+            _subscription = _commonServices.ActiveConfiguredProjectSubscription.JointRuleSource.SourceBlock.LinkTo(
+                receiver,
+                initialDataAsNew: true,
+                suppressVersionOnlyUpdates: true,
+                ruleNames: new string[] { RazorGeneral.SchemaName, });
```
NTaylorMullen: I know the RazorGeneral schema was auto generated but for clarity, what does it mean to provide a rule name here? In this case what does our razor general rule do? only subscribe us to projects with Razor files?

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+
+            return Task.CompletedTask;
+        }
+
+        protected override async Task DisposeCoreAsync(bool initialized)
+        {
+            _subscription.Dispose();
```
NTaylorMullen: I assume InitializeCoreAsycn is guaranteed to have run yes? I only ask because it's not our code.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+
+                if (string.IsNullOrEmpty(languageVersion))
+                {
+                    languageVersion = "2.1";
+                }
+
+                var hostProject = new HostProject(_commonServices.UnconfiguredProject.FullPath, languageVersion);
+                if (_current == null)
+                {
+                    _current = hostProject;
```
NTaylorMullen: so i'm a little confused about the lifetimes here. This class is exported as a MEF service which makes me believe there's 1 per IDE yet we have a _current which represents the current host project which has a file path. 

I thought `HostProject` represented VS' version of csproj were there can be > 1? I think I'm missing something 😄

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
```diff
@@ -11,6 +11,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Build" Version="$(MicrosoftBuildPackageVersion)" />
```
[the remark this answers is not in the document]
rynowak: Yes. You still need this to build.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
rynowak: I think this is fine. We only care about the main VS workspace in VS for windows. 

The other workspaces don't ever have real projects.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -0,0 +1,182 @@
...
+        [ProjectAutoLoad(ProjectLoadCheckpoint.ProjectFactoryCompleted)]
+        public Task LoadAsync()
+        {
+            return InitializeAsync();
+        }
+
+        // Should only be called from the UI thread.
+        private ProjectSnapshotManagerBase GetProjectManager()
+        {
+            _commonServices.ThreadingService.VerifyOnUIThread();
```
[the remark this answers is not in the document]
rynowak: Not materially. I'm following the CPS conventions in the CPS code.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
rynowak: It means that we'll recieve updates for all of the properties defined by `RazorGeneral`

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
rynowak: I can make this more defensive.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
rynowak: Nah, CPS works differently. 

You get one of these per project in the solution.

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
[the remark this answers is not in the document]
NTaylorMullen: > Yes. You still need this to build.

I'm asking if they can be private refs, aka do we need them at runtime or is the auto-generated pieces enough?

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
NTaylorMullen: > You get one of these per project in the solution.

ah beautiful, ok!

src/Microsoft.VisualStudio.LanguageServices.Razor/Microsoft.VisualStudio.LanguageServices.Razor.csproj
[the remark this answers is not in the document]
rynowak: Yes, we reference types defined in these assemblies. I ran into problems running tests without these.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
rynowak: Yeah, there's a pattern for this.. I'm supposed to look at `initialized` - lol

NTaylorMullen: Glad I asked 😄

Issue #1968, opened 2018-01-24
Generate portable .pds for the .mpack


KirillOsenkov:
Ideally if we can we should just add the portable .pdbs for all .dlls directly into the .mpack.

NTaylorMullen:
> Ideally if we can we should just add the portable .pdbs for all .dlls directly into the .mpack.

Is adding the pdbs to the mpack the expected way to ship PDBs for addins?

KirillOsenkov:
I guess it's the most comfortable one for the debugging experience - debugging will just work and you won't have to worry about adding any additional symbol servers or any other actions. 

@mhutch

mhutch:
Yes, that's currently the only way to get symbolicated stack traces for a deployed extension

rynowak:
That sounds fine to me if you're all happy with it 😆.

NTaylorMullen:
@mhutch @KirillOsenkov is there anything we need to update/add in the manifest/addininfo to account for PDBs or will their presence in the mpack be sufficient?

KirillOsenkov:
Yeah as long as they're next to their corresponding .dll that should be enough. Thanks!

NTaylorMullen:
83c31f611f12e375ebea419d264bc094af6ce832

PR #2025, opened 2018-02-05
Update GenerateMPACK to capture PDBs.

NTaylorMullen:
- VS Mac expects pdbs to sit side-by-side.

@natemcmaster does our GenerateMPACK target run prior to the Portable => Full windows pdb conversion? We need the MPack PDBs to be portable is why I ask.

#1968 

/cc @KirillOsenkov

### DIFF
--- a/build/MPack.targets
+++ b/build/MPack.targets
@@ -44,13 +44,12 @@
       3. All language service binaries
     -->
 
-  
     <MakeDir Directories="$(MPackIntermediateOutputPath)" Condition="!Exists('$(MPackIntermediateOutputPath)')" />
     <Delete Files="$(MPackOutputPath)" Condition="Exists('$(MPackOutputPath)')" />
 
     <!-- We need to resolve the language service assemblies to generate an addin.info for the mpack -->
     <XmlPeek XmlInputPath="$(MPackManifest)" Query="/ExtensionModel/Runtime/Import/@assembly">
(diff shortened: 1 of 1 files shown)


build/MPack.targets
```diff
@@ -70,10 +69,17 @@
       <AddinInfoLines Include="&lt;/Addin&gt;" />
     </ItemGroup>
 
+    <ItemGroup>
+      <LanguageServiceAssemblies Include="$(LanguageServiceOutputPath)%(LanguageServiceAssemblyNames.Identity)" />
+      <LanguageServicePDBs Include="%(LanguageServiceAssemblies.RootDir)%(Directory)%(FileName).pdb" Condition="Exists('%(LanguageServiceAssemblies.RootDir)%(Directory)%(FileName).pdb')" />
```
rynowak: Generally MSBuild item names aren't plural

natemcmaster:
> does our GenerateMPACK target run prior to the Portable => Full windows pdb conversion? 

Yes. Full PDB conversion doesn't happen in a local repo build, only in the Coherence-Signed step.

build/MPack.targets
NTaylorMullen: ooo good to know. I'll update all the names in here.

NTaylorMullen:
🆙 📅  (two commits, second commit changes plural item group names)

PR #2026, opened 2018-02-06
Split arguments for the build server properly

ajaybhargavb:
Before this change, we were not splitting the response file arguments properly. E.g. `C:\Program Files\dotnet\sdk\NuGetFallbackFolder\microsoft.netcore.app\2.1.0-preview1-26115-03\ref\netcoreapp2.1\System.Security.SecureString.dll` was being split into `C:\Program` and the rest.

I'm now splitting the arguments by Newline instead of whitespace.

FYI @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -170,9 +170,8 @@ private string CurrentDirectoryToUse()
 
         private List<string> GetArguments(string responseFileCommands)
         {
-            var responseFileArguments =
-                CommandLineUtilities.SplitCommandLineIntoArguments(responseFileCommands, removeHashComments: true);
```
ajaybhargavb: We don't really need this because we don't use the `commandLineArguments` and we format `responseFileArguments`  ourselves and we don't add hash comments. I feel the below code is much simpler and less bug prone. Correct me if I am missing something.

test/testapps/Whitespace in path/Models/ErrorViewModel.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Whitespace_in_path.Models
```
ajaybhargavb: This app is just a copy-paste and rename of `SimpleMvc`.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -170,9 +170,8 @@ private string CurrentDirectoryToUse()
 
         private List<string> GetArguments(string responseFileCommands)
```
pranavkm: Could this be changed to `IList`? That would let you remove the `ToList` call

test/testapps/Whitespace in path/Models/ErrorViewModel.cs
pranavkm: We could consider modifying `ProjectDirectory` to accept a target path and deploy SimpleMvc to a path with spaces. Might let us test a variant where the project name itself doesn't have spaces but the path does e.g. `%temp%\razor\path with spaces\SimpleMvc`

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/InitializeTestProjectAttribute.cs
```diff
@@ -10,11 +10,13 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
     public class InitializeTestProjectAttribute : BeforeAfterTestAttribute
     {
         private readonly string _projectName;
+        private readonly string _baseDirectory;
         private readonly string[] _additionalProjects;
 
-        public InitializeTestProjectAttribute(string projectName, params string[] additionalProjects)
+        public InitializeTestProjectAttribute(string projectName, string baseDirectory = "", params string[] additionalProjects)
```
pranavkm: Doesn't this break tests that specify `additionalProjects` today?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/ProjectDirectory.cs
```diff
@@ -18,9 +18,9 @@ internal class ProjectDirectory : IDisposable
         public bool PreserveWorkingDirectory { get; set; }
 #endif
 
-        public static ProjectDirectory Create(string projectName, string[] additionalProjects)
+        public static ProjectDirectory Create(string projectName, string baseDirectory, string[] additionalProjects)
         {
-            var destinationPath = Path.Combine(Path.GetTempPath(), "Razor", Path.GetRandomFileName());
+            var destinationPath = Path.Combine(Path.GetTempPath(), "Razor", baseDirectory, Path.GetRandomFileName());
```
pranavkm: This does miss the scenario of a project file with spaces. We can do some work to cover that scenario in a test since that's also an interesting scenario to cover.

ajaybhargavb: Can we just rename `SimpleMvc` to `Simple Mvc`? I don't see any reason why not.

ajaybhargavb:
Got a verbal signoff from @pranavkm

PR #2027, opened 2018-02-06
Flow RazorConfiguration through ProjectSnapshot

rynowak:
Updates the ProjectSnapshot (and everything that consumes it) to use the
RazorConfiguration that comes from the new project system.

Also some enhancements to the Razor Info window:

image


src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
```diff
@@ -7,24 +7,24 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    public sealed class RazorConfiguration
+    public abstract class RazorConfiguration
```
rynowak: It ended up just feeling a little better to me to take this abstract. Then if we need to flow more data in the IDE cases we can.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectExtensibilityConfigurationFactory.cs
```diff
@@ -1,142 +0,0 @@
...
-using System;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Razor.Language;
-
-namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
-{
-    // This is hardcoded for now. A more complete design would fan out to a list of providers.
-    internal class DefaultProjectExtensibilityConfigurationFactory : ProjectExtensibilityConfigurationFactory
```
rynowak: All of this is gone and handled by the project system now. This means we no longer have to asynchronously compute the version.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -118,7 +116,7 @@ private DefaultProjectSnapshot(ProjectSnapshotUpdateContext update, DefaultProje
             Version = other.Version;
         }
 
-        public override ProjectExtensibilityConfiguration Configuration { get; }
+        public override RazorConfiguration Configuration => HostProject?.Configuration;
```
rynowak: The churn here is due to the fact that we don't need to compute this in the background anymore, we always get to know up front what the version is.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -204,14 +202,16 @@ public DefaultProjectSnapshot WithComputedUpdate(ProjectSnapshotUpdateContext up
...
         {
             if (original == null)
             {
                 throw new ArgumentNullException(nameof(original));
             }
 
-            return !object.Equals(Configuration, original.Configuration);
+            // We don't have any computed state right now, so treat all background updates as
+            // significant.
+            return !object.Equals(ComputedVersion, original.ComputedVersion);
```
rynowak: I decided to leave in the concept of computed state and the background worker code to compute it. I think we'll find more things to use it for (tag helpers). For now it seems fine to leave it and more destabilizing to delete it.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotWorker.cs
```diff
@@ -46,14 +37,15 @@ public override Task ProcessUpdateAsync(ProjectSnapshotUpdateContext update, Can
             return ProjectUpdatesCoreAsync(update);
         }
 
-        private async Task ProjectUpdatesCoreAsync(object state)
+        protected virtual void OnProcessingUpdate()
         {
-            var update = (ProjectSnapshotUpdateContext)state;
+        }
```
rynowak: Added this for testability.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
```diff
@@ -0,0 +1,82 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    internal class FallbackRazorConfiguration : RazorConfiguration
```
rynowak: Moving this up a level simplifies a lot of things. Now we have a static list of supported legacy configurations. In addition since this is a different type now, we can tell at a glance if the configuration come from the project system or if it was a fallback.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
```diff
@@ -0,0 +1,82 @@
...
+            if (extensions == null)
+            {
+                throw new ArgumentNullException(nameof(extensions));
+            }
+
+            LanguageVersion = languageVersion;
+            ConfigurationName = configurationName;
+            Extensions = extensions;
+
+            DesignTime = true;
```
rynowak: Design time being a part of the configuration is still making me sad.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorExtension.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    internal class FallbackRazorExtension : RazorExtension
```
rynowak: This will get fleshed out more soon, just relax.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/IUnconfiguredProjectCommonServices.cs
```diff
@@ -13,7 +13,7 @@ internal class UnconfiguredProjectCommonServices : IUnconfiguredProjectCommonSer
     {
         private readonly ActiveConfiguredProject<ConfiguredProject> _activeConfiguredProject;
         private readonly ActiveConfiguredProject<IAssemblyReferencesService> _activeConfiguredProjectAssemblyReferences;
-        private readonly ActiveConfiguredProject<RazorProjectProperties> _activeConfiguredProjectProperties;
+        private readonly ActiveConfiguredProject<Rules.RazorProjectProperties> _activeConfiguredProjectProperties;
```
rynowak: Note that I changed the namespace of the generated Rules classes so it wouldn't conflict with our other types. You just don't ever add a using for `.Rules`

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -116,7 +118,7 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
                 receiver,
                 initialDataAsNew: true,
                 suppressVersionOnlyUpdates: true,
-                ruleNames: new string[] { RazorGeneral.SchemaName, });
+                ruleNames: new string[] { Rules.RazorGeneral.SchemaName, Rules.RazorConfiguration.SchemaName, Rules.RazorExtension.SchemaName });
```
rynowak: Now we're listening for changes to the `RazorExtension` and `RazorConfiguration` itemgroups as well.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -151,18 +153,45 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
...
+                        var includedExtensions = c.Value[Rules.RazorConfiguration.ExtensionsProperty]
+                            .Split(';')
+                            .Select(name => extensions.Where(e => e.ExtensionName == name).FirstOrDefault())
+                            .Where(e => e != null)
+                            .ToArray();
+
+                        return new ProjectSystemRazorConfiguration(parsedVersion, c.Key, includedExtensions);
+                    }).ToArray();
+
+                    configuration = configurations.Where(c => c.ConfigurationName == defaultConfiguration).FirstOrDefault();
```
rynowak: This code turns this https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.props

into a RazorConfiguration

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -188,7 +217,7 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
             });
         }
 
-        private async Task<string> DetectFallbackLanguageVersion()
+        private async Task<RazorConfiguration> DetectFallbackLanguageConfiguration()
```
rynowak: This code only works right now when the project has already been restored. Currently trying to figure out how to deal with that. I was hoping that `GetResolvedReferencesAsync` would wait for a design-time build, but alas it does not.

This is only an issue for fallback cases.

test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
```diff
@@ -342,13 +328,12 @@ public void WorkspaceProjectChanged_BackgroundUpdate_StillDirty_WithSignificantC
...
             snapshot = ProjectManager.GetSnapshot(project);
             Assert.True(snapshot.IsDirty);
-            Assert.Same(configuration, snapshot.Configuration);
 
             Assert.True(ProjectManager.ListenersNotified);
             Assert.True(ProjectManager.WorkerStarted);
         }
 
-        [Fact]
+        [Fact(Skip = "We no longer have any background-computed state")]
```
rynowak: I had to skip two tests here because they test the case where a background update has occurred but didn't result in a state change. Since we don't have any background updates (yet) we don't have the facility to test this scenario.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultTemplateEngineFactoryServiceTest.cs
```diff
@@ -128,38 +107,6 @@ public void Create_DoesNotSupportViewComponentTagHelpers_ForVersion1_0()
             Assert.Empty(engine.Engine.Features.OfType<Mvc1_X.ViewComponentTagHelperPass>());
         }
 
-        [Fact]
-        public void Create_HigherMvcVersion_UsesLatest()
```
rynowak: This scenario doesn't need to be handled by this class anymore. The project system deals with unsupported versions now.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/ProjectSystem/TestProjectChangeDescription.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information
+
+using Microsoft.VisualStudio.ProjectSystem.Properties;
+
+namespace Microsoft.VisualStudio.ProjectSystem
+{
+    internal class TestProjectChangeDescription : IProjectChangeDescription
```
rynowak: Since the RazorProjectHost now looks at the updates instead of the current state, I had to add a bunch of plumbing to provide different state.

tooling/Microsoft.VisualStudio.RazorExtension/RazorInfo/RazorInfoToolWindow.cs
```diff
@@ -53,28 +58,69 @@ protected override void Dispose(bool disposing)
 
             if (disposing)
             {
-                _workspace.WorkspaceChanged -= Workspace_WorkspaceChanged;
+                _projectManager.Changed -= ProjectManager_Changed;
             }
         }
 
-        private void Reset(Solution solution)
```
rynowak: This code was based on the solution before, but now it's based on our project system.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
NTaylorMullen: I agree, lets talk about that one in person.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -128,7 +126,7 @@ private DefaultProjectSnapshot(ProjectSnapshotUpdateContext update, DefaultProje
 
         public override bool IsUnloaded => HostProject == null && WorkspaceProject == null;
 
-        public override string LanguageVersion => HostProject?.LanguageVersion;
+        public override string LanguageVersion => HostProject?.Configuration.LanguageVersion.ToString();
```
NTaylorMullen: Might as well just nuke this. Don't want to have a ton of proxy properties until we see a need for them.

src/Microsoft.AspNetCore.Razor.Language/RazorConfiguration.cs
```diff
@@ -44,18 +44,38 @@ public sealed class RazorConfiguration
...
+            {
+                LanguageVersion = languageVersion;
+                ConfigurationName = configurationName;
+                Extensions = extensions;
+                DesignTime = designTime;
+            }
+
+            public override string ConfigurationName { get; }
+
+            public override IReadOnlyList<RazorExtension> Extensions { get; }
```
NTaylorMullen: Interesting. When you first added RazorExtension I didn't see that it was in the Razor.Language assembly. Gots me curious 😄

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
NTaylorMullen: How is it simpler than just being in Mvc.Razor.Extensions?

rynowak: These assemblies aren't going to reference that when I'm done 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
rynowak: Sure, that makes sense

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
rynowak: I should mention that supports introduces loose coupling based on the configuration/extension names. So VS will be able to find `MVC-1.1` and configure it, but without anything in between having a direct reference.

The change here is that the workspaces assembly reflects our history. It hardcodes the notion of MVC's old configurations but doesn't hardcode the implementation.

rynowak:
Updated and rebased this

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/FallbackRazorConfiguration.cs
NTaylorMullen: I see I see I see. Ok 👍

rynowak:
@NTaylorMullen - do you you have feedback about this PR?

PR #2028, opened 2018-02-06
Use assembly references as a fallback for version

rynowak:
This implements version detection as a fallback when the language
version is not found. This happens for all Razor/MVC versions prior to
2.1.

This is similar to what we do today inside the language service but this
is a more approriate way to do it.


src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -170,6 +187,25 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
             });
         }
 
+        private async Task<string> DetectFallbackLanguageVersion()
```
rynowak: This is a much simplified version of what we had before... because we **only** care about the old supported scenarios of MVC1.X and MVC2.0.

Everything we care about in the future will use the MSBuild approach

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHost.cs
```diff
@@ -170,6 +187,25 @@ protected override Task InitializeCoreAsync(CancellationToken cancellationToken)
...
+        {
+            var references = await _commonServices.ActiveConfiguredProjectAssemblyReferences.GetResolvedReferencesAsync().ConfigureAwait(false);
+            foreach (var reference in references)
+            {
+                var fullPath = await reference.GetFullPathAsync().ConfigureAwait(false);
+                if (fullPath.EndsWith(MvcAssemblyFileName, StringComparison.OrdinalIgnoreCase))
+                {
+                    var assemblyName = await reference.GetAssemblyNameAsync().ConfigureAwait(false);
```
pranavkm: What happens when a reference is unresolved? Does it return null?

rynowak: In that case the `reference` itself wouldn't be included in the collection.

rynowak:
Hold tight. I've done a massive rewrite of this codes because this doesn't handle the case where the packages haven't restored yet. I basically had to go back to the drawring board.

rynowak:
@NTaylorMullen @pranavkm updated

This solves the issues with ordering.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
```diff
@@ -0,0 +1,91 @@
...
+using Microsoft.VisualStudio.LanguageServices;
+using Microsoft.VisualStudio.ProjectSystem;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    // Somewhat similar to https://github.com/dotnet/project-system/blob/fa074d228dcff6dae9e48ce43dd4a3a5aa22e8f0/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+    //
+    // This class is responsible for intializing the Razor ProjectSnapshotManager for cases where
+    // MSBuild provides configuration support (>= 2.1).
+    [AppliesTo("DotNetCoreRazor & DotNetCoreRazorConfiguration")]
```
rynowak: I've added a capability to describe the new behaviour. Then we can use switch with loader gets used dynamically.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+
+                var languageVersion = update.Value.CurrentState[RazorGeneral.SchemaName].Properties[RazorGeneral.RazorLangVersionProperty];
+
+                if (string.IsNullOrEmpty(languageVersion))
+                {
+                    // Ok we can't find a language version. Let's assume this project isn't using Razor then.
+                    await UpdateProjectUnsafeAsync(null).ConfigureAwait(false);
+                    return;
+                }
```
rynowak: I switched to this approach since it's what we'll need to do in the next step and it's just a bit simpler. It also only requires the UI thread when we need to actually push a change. The old approach did lots of hopping around.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/FallbackRazorProjectHost.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.VisualStudio.ProjectSystem;
+using ResolvedCompilationReference = Microsoft.CodeAnalysis.Razor.ProjectSystem.ManageProjectSystemSchema.ResolvedCompilationReference;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    // Somewhat similar to https://github.com/dotnet/project-system/blob/fa074d228dcff6dae9e48ce43dd4a3a5aa22e8f0/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
+    //
+    // This class is responsible for intializing the Razor ProjectSnapshotManager for cases where
+    // MSBuild does not provides configuration support (< 2.1).
+    [AppliesTo("DotNetCoreRazor & !DotNetCoreRazorConfiguration")]
```
rynowak: Notice the negation of `DotNetCoreRazorConfiguration`

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/FallbackRazorProjectHost.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                using (var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite | FileShare.Delete))
+                using (var reader = new PEReader(stream))
+                {
+                    var metadataReader = reader.GetMetadataReader();
+
+                    var assemblyDefinition = metadataReader.GetAssemblyDefinition();
+                    return assemblyDefinition.Version;
+                }
+            }
```
rynowak: We're going to disk now to read the assembly version. Using the service to do it still has an ordering problem. Since we're listening to evaluation changes from MSBuild we're still running to early to ask the assembly service for its version.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/FallbackRazorProjectHost.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            await ExecuteWithLock(async () =>
+            {
+                if (IsDisposing || IsDisposed)
+                {
+                    return;
+                }
+
+                string mvcReferenceFullPath = null;
+                var references = update.Value.CurrentState[ResolvedCompilationReference.SchemaName].Items;
```
NTaylorMullen: Wondering if we should add an API that has this "is valid host project" logic. When it comes to VS4Mac we'll have to introduce these concepts all over again in a non-CPS way.

Not saying this is necessary for this PR but more food for thought on the final forms of these APIs.

rynowak: I'd be interested to know what you're thinking about. 

Right now I'm thinking about the project system and the data flow therein, so I'm not trying to work very hard to address the editor's needs. 

Here's what I'm thinking of:
- **every** open dotnet core razor document open in the editor should belong to a project
- if a dotnet core razor document needs to open and doesn't have a project to belong to, the snapshot manager will create an *ephemeral* project
- the snapshot manager can remove an *ephemeral* project when all documents are closed
- the snapshot manager can *reify* the *ephemeral* project by getting an update from the underlying project system

This means that the project system:
A. Needs no knowledge about editors and open documents
B. Only needs to reflect the underlying truth of the host's project system

This means that the editor:
A. Can always have a project to work with
B. Can get 'canned' answers about tag helpers and such when the project is *ephemeral*
C. Doesn't really care if the project is real or not

All of the heavy lifting is in the snapshot manager/document manager, of which we only have one implementation.

Thoughts?

rynowak:
Do you have any concerns about this? I consider this done.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/FallbackRazorProjectHost.cs
NTaylorMullen: > All of the heavy lifting is in the snapshot manager/document manager, of which we only have one implementation.
> 
> Thoughts?

Love the separation you're bringing up. Depending on how much logic ends up sitting on the `HostProject` side of the world I might want to consider building a cross-platform pattern way of orchestrating `HostProject` => snapshot/document manager interaction. 

Right now I think we have a minimal set of things going on in HostProject land which is awesome; however, things like falling back to older versions of MVC is something that every host project system will have to understand.

NTaylorMullen:
> Do you have any concerns about this? I consider this done.

nah broseph doooo it

Issue #1954, opened 2018-01-20
Build server: Logging cleanup


ajaybhargavb:
Meta issue - #1921 

Currently we log all over the place with different types of loggers (MSBuild logging, CompilerServerLogger). We need to clean it up and make it better.

ajaybhargavb:
8b78b4f42158723b7f2d60f17bf4c8a44c27168a

PR #2031, opened 2018-02-07
Cleaned up logging in build server

ajaybhargavb:
Issue - #1954 
Meta - #1921 

Before this we weren't passing the output from the different commands in the build server to the task that is waiting for the result. Included some general cleanup as well.


src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
```diff
@@ -26,13 +27,20 @@ public override ServerResponse Execute(ServerRequest request, CancellationToken
...
+                    {
+                        Out = writer,
+                        Error = writer,
+                    };
+                    var commandArgs = parsed.args.ToArray();
 
-                var exitCode = app.Execute(commandArgs);
-                var output = app.Out.ToString() ?? string.Empty;
+                    var exitCode = app.Execute(commandArgs);
+                    var output = writer.ToString() ?? string.Empty;
```
rynowak: Would it be possible to avoid the stringwriter + tostring if the logging is turned off? This could be a significant amount of text in some cases and could cause LOH issues.

src/Microsoft.AspNetCore.Razor.Tools/Roslyn/CompilerServerLogger.cs
```diff
@@ -13,10 +12,10 @@ namespace Microsoft.CodeAnalysis.CommandLine
...
     /// Class for logging information about what happens in the server and client parts of the 
-    /// Roslyn command line compiler and build tasks. Useful for debugging what is going on.
+    /// Razor command line compiler and build tasks. Useful for debugging what is going on.
     /// </summary>
     /// <remarks>
-    /// To use the logging, set the environment variable RoslynCommandLineLogFile to the name
+    /// To use the logging, set the environment variable RAZORBUILDSERVER_LOG to the name
     /// of a file to log to. This file is logged to by both client and server components.
     /// </remarks>
     internal class CompilerServerLogger
```
rynowak: Can you rename and move out of this folder if you are happy with it? Ultimately we want to clean out anything in the Roslyn folder.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -115,11 +112,11 @@ public override void Cancel()
...
             out int result)
         {
-            CompilerServerLogger.Log("Server execution started.");
+            Log.LogMessage("Server execution started.");
             using (_razorServerCts = new CancellationTokenSource())
             {
-                CompilerServerLogger.Log($"CommandLine = '{commandLineCommands}'");
-                CompilerServerLogger.Log($"ServerResponseFile = '{responseFileCommands}'");
+                Log.LogMessage($"CommandLine = '{commandLineCommands}'");
+                Log.LogMessage($"ServerResponseFile = '{responseFileCommands}'");
```
rynowak: Does this show up in the msbuild log by default? we want this to only show up if you've asked for debug level logging

ajaybhargavb: It shouldn't. We set the importance conditionally here https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs#L37

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
ajaybhargavb: Hmm not sure. I'll look into it. Would it still be an issue if we use a StreamWriter with a MemoryStream and read from it?

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -118,6 +118,7 @@
       DebugTool="$(_RazorDebugGenerateCodeTool)"
       ToolAssembly="$(_RazorToolAssembly)"
       UseServer="$(UseRazorBuildServer)"
+      DisableFallback="$(_RazorDisableBuildServerFallback)"
```
rynowak: Call this `ForceServer`

rynowak: Or consider making UseServer a string

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -115,11 +115,11 @@ public override void Cancel()
             string commandLineCommands,
             out int result)
         {
-            CompilerServerLogger.Log("Server execution started.");
+            Log.LogMessage("Server execution started.");
```
rynowak: This should be Low, it's just tracing

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -141,14 +141,30 @@ public override void Cancel()
                 {
                     result = completedResponse.ReturnCode;
 
-                    CompilerServerLogger.Log($"Server execution completed with return code {result}.");
+                    if (result == 0)
+                    {
+                        Log.LogMessage($"Server execution completed with return code {result}.");
```
rynowak: This should be low

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -141,14 +141,30 @@ public override void Cancel()
...
+                        return true;
+                    }
+                    else
+                    {
+                        Log.LogWarning($"Server execution completed with return code {result}. For more info, check the server log file in the location specified by the RAZORBUILDSERVER_LOG environment variable.");
+                    }
+                }
+                else
+                {
+                    Log.LogWarning($"Server execution failed with response {response.Type}. For more info, check the server log file in the location specified by the RAZORBUILDSERVER_LOG environment variable.");
```
rynowak: There messages aren't user-actionable, these should be Low

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -141,14 +141,30 @@ public override void Cancel()
...
 
+                if (DisableFallback)
+                {
                     return true;
                 }
-            }
 
-            CompilerServerLogger.Log("Server execution failed.");
-            result = -1;
+                Log.LogWarning("Fallback to in-process execution.");
```
rynowak: There messages aren't user-actionable, these should be Low

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -170,7 +186,7 @@ private string CurrentDirectoryToUse()
 
         private IList<string> GetArguments(string responseFileCommands)
         {
-            var list = responseFileCommands.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            var list = responseFileCommands.Split(new[] { ".dotnet" }, StringSplitOptions.RemoveEmptyEntries);
```
rynowak: ???

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
```diff
@@ -28,11 +28,22 @@ public override ServerResponse Execute(ServerRequest request, CancellationToken
...
+                    using (var writer = new StreamWriter(ServerLogger.UnderlyingStream))
+                    {
+                        app.Out = writer;
+                        app.Error = writer;
+                        exitCode = app.Execute(commandArgs);
+                    }
+                }
+                else
+                {
+                    exitCode = app.Execute(commandArgs);
```
rynowak: Set app.Out and app.Error to Stream.Null

src/Microsoft.AspNetCore.Razor.Tools/Microsoft.AspNetCore.Razor.Tools.csproj
```diff
@@ -12,6 +12,12 @@
     <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Roslyn\**" />
+    <EmbeddedResource Remove="Roslyn\**" />
+    <None Remove="Roslyn\**" />
+  </ItemGroup>
```
rynowak: I think this can be removed

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
ajaybhargavb: Just found out that the StandardOutputLoggingImportance is Low. So going to leave these as is.

rynowak: What does that mean? Does that mean that `Log.LogMessage(...)` applies a default of Low?

If that's the case then either remove Low where we specify it, or specify it everywhere. When you do something other than the default, I assume you have a good reason 😆

ajaybhargavb: That is the plan. I realized this after sending this update. I'll make it consistent.

rynowak: BTW the default is Normal https://github.com/Microsoft/msbuild/blob/master/src/Shared/TaskLoggingHelper.cs#L292

There are two things that matter:
1. What is the level of the **message**
2. What is the level of the **filter** applied to messages

1 is controlled where you do logging
2 is controlled by `StandardOutputLoggingImportance`

The correct level for the kind of messages that appear here is Low. They are diagnostic tracing and not for end users to see.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
ajaybhargavb: Btw I thought about making `UseServer` a string but we are doing this only for tests and `UseServer` is a "public" switch and we don't want to make it too complicated to use. I am going with `ForceServer` approach

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
ajaybhargavb: Wow I didn't expect them to not use https://github.com/Microsoft/msbuild/blob/master/src/Utilities/ToolTask.cs#L376. Okay I'll be explicit.

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
```diff
@@ -26,13 +26,33 @@ public override ServerResponse Execute(ServerRequest request, CancellationToken
...
+                var exitCode = 0;
+                var output = string.Empty;
                 var app = new Application(cancellationToken);
                 var commandArgs = parsed.args.ToArray();
 
-                var exitCode = app.Execute(commandArgs);
-                var output = app.Out.ToString() ?? string.Empty;
+                if (ServerLogger.IsLoggingEnabled)
+                {
+                    using (var writer = new StringWriter())
```
ajaybhargavb: Had to move away from passing the file stream directly. It started throwing "another process is using this file" exception. So delegating it to `ServerLogger` is best. Refer https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/Roslyn/CompilerServerLogger.cs#L116

PR #2033, opened 2018-02-07
Decouple tooling from MVC assemblies

rynowak:
As a proof-of-concept, this drops all hardcoding between the Razor tooling code and the MVC functionality for VCTH and code generation. There are still some shims that are required, and where to place them is a bit tricky for layering reasons.


rynowak:
I had to hack up a bunch of stuff to get this going, but it works!

src/Microsoft.VisualStudio.Editor.Razor/LegacyTemplateEngineFactory_1_0.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.Reflection;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Razor;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    [ExportCustomTemplateEngineFactory("MVC-1.0", SupportsSerialization = true)]
+    internal class LegacyTemplateEngineFactory_1_0 : ICustomTemplateEngineFactory
```
rynowak: This seems like a *good enough* way to allow a VSIX to hijack all of our functionality. 

We need to keep coming up with more polish in this area, but this will achieve our immediate goals.

rynowak: For layering reasons this shim is defined in the Razor.Editor dll. We can likely resolve the issues that led me to believe that this is the best solution in time.

For the proof of concept, I used reflection here instead of direct references. I wanted to prove out that this kind of indirection will work using service hub and it does 😆 👍

src/Microsoft.CodeAnalysis.Remote.Razor/RazorLanguageService.cs
```diff
@@ -19,25 +21,69 @@ internal class RazorLanguageService : ServiceHubServiceBase
...
 
             var solution = await GetSolutionAsync(cancellationToken).ConfigureAwait(false);
             var project = solution.GetProject(projectId);
 
-            var resolver = new DefaultTagHelperResolver(designTime: true);
-            var result = await resolver.GetTagHelpersAsync(project, cancellationToken).ConfigureAwait(false);
+            ((SerializedProjectSnapshot)snapshot).InitializeWorkspaceProject(project);
+
+            RazorTemplateEngine templateEngine = null;
+            if (factoryTypeName != null)
```
rynowak: We're being a bit sneaky. 

The VS side decides which factory to use to create the template engine, and then serializes the AQFN over to service hub. 

Then we rehydrate the factory on the service hub side. This relies on the assembly containing the factory to exist in the Load context, which we already do, and anyone shipping a VSIX can easily do as well.

Then we use the factory to configure the crap out of a new engine.

Then we find the tag helper providers in there and use them.

rynowak: We will do something a bit different when we want to ship this kind of extensibility in a nukpg. This is good enough for anyone shipping a VSIX to get started... ahem..

ajaybhargavb:
This is looking really cool! 🍾 👏

src/Microsoft.CodeAnalysis.Remote.Razor/RazorLanguageService.cs
```diff
@@ -19,25 +21,69 @@ internal class RazorLanguageService : ServiceHubServiceBase
...
+            {
+                var factoryType = Type.GetType(factoryTypeName, throwOnError: true);
+                var factory = (ICustomTemplateEngineFactory)Activator.CreateInstance(factoryType);
+
+                templateEngine = factory.Create(snapshot.Configuration, EmptyProject.Instance, (b) =>
```
NTaylorMullen: Could always get the project path from the project snapshot instead of needing an empty project instance.

rynowak: We need a project to create a template engine.

NTaylorMullen: Ya you can do a `RazorProject.Create(...)` from the project snapshots file path.

rynowak: Ah I see what you mean! Yeah I did the empty thing on purpose because I don't want any code running here to look at the files.

NTaylorMullen: > I don't want any code running here to look at the files.

O rly? Why's that?

rynowak: Because we don't have a clear need for it.

rynowak:
Updated and removed design label

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/RazorExtensions.cs
```diff
@@ -37,6 +37,7 @@ public static void RegisterViewComponentTagHelpers(IRazorEngineBuilder builder)
         {
             EnsureDesignTime(builder);
 
+            builder.Features.Add(new ViewComponentTagHelperDescriptorProvider());
```
rynowak: We have a kind of a wierd thing going on where these features don't get registered exception manually. This was hardcoded in a bad way. There's no hard in us registering these features even if they go unused.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/ViewComponentTypeVisitor.cs
```diff
@@ -24,10 +24,9 @@ public static ViewComponentTypeVisitor Create(Compilation compilation, List<INam
                 var symbol = compilation.GetAssemblyOrModuleSymbol(reference) as IAssemblySymbol;
                 if (symbol != null)
                 {
-                    if (string.Equals(symbol.Identity.Name, ViewComponentTypes.Assembly, StringComparison.Ordinal) &&
-                        symbol.Identity.Version > ViewComponentTypes.AssemblyVersion)
+                    if (string.Equals(symbol.Identity.Name, ViewComponentTypes.Assembly, StringComparison.Ordinal))
                     {
-                        enabled = true;
+                        enabled = symbol.Identity.Version >= ViewComponentTypes.AssemblyVersion;
```
rynowak: This was busto. As a result VCTH wasn't working in a 1.1 app.

PR #2036, opened 2018-02-08
Move to RazorProjectFileSystem.

NTaylorMullen:
- Changed all RazorProject usage to use `RazorProjectFileSystem`. This was possible because `RazorProjectFileSystem` inherits from RazorProject.
- Renamed `FileSystemRazorProject` to `DefaultRazorProjectFileSystem`.
- Renamed FileSystemRazorProjectItem` to `DefaultRazorProjectItem`.
- Obsoleted `RazorProject.Create`

#1828

/cc @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectItem.cs
```diff
@@ -5,16 +5,16 @@
...
         /// <summary>
-        /// Initializes a new instance of <see cref="FileSystemRazorProjectItem"/>.
+        /// Initializes a new instance of <see cref="DefaultRazorProjectItem"/>.
         /// </summary>
         /// <param name="basePath">The base path.</param>
         /// <param name="relativePhysicalPath">The physical path of the base path.</param>
         /// <param name="filePath">The path.</param>
         /// <param name="file">The <see cref="FileInfo"/>.</param>
-        public FileSystemRazorProjectItem(string basePath, string filePath, string relativePhysicalPath, FileInfo file)
+        public DefaultRazorProjectItem(string basePath, string filePath, string relativePhysicalPath, FileInfo file)
```
NTaylorMullen: It being based off of the file system is a given.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectFileSystem.cs
```diff
@@ -8,9 +8,9 @@
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
-    internal class FileSystemRazorProject : RazorProjectFileSystem
+    internal class DefaultRazorProjectFileSystem : RazorProjectFileSystem
```
NTaylorMullen: FileSystem is implied

src/Microsoft.AspNetCore.Razor.Language/RazorProjectFileSystem.cs
```diff
@@ -1,18 +1,25 @@
...
     {
         /// <summary>
-        /// Create a Razor project based on a physical file system.
+        /// Create a Razor project file system based off of a root directory.
         /// </summary>
         /// <param name="rootDirectoryPath">The directory to root the file system at.</param>
-        /// <returns>A <see cref="RazorProject"/></returns>
-        public static new RazorProjectFileSystem Create(string rootDirectoryPath)
+        /// <returns>A <see cref="RazorProjectFileSystem"/></returns>
+        public new static RazorProjectFileSystem Create(string rootDirectoryPath)
```
rynowak: :clap:

test/Microsoft.AspNetCore.Razor.Language.Test/RazorTemplateEngineTest.cs
```diff
@@ -245,7 +245,7 @@ public void CreateCodeDocument_NullImportFileName_IncludesDefaultImportIfNotNull
         {
             // Arrange
             var projectItem = new TestRazorProjectItem("/Views/Home/Index.cshtml");
-            var project = new TestRazorProject(new[] { projectItem });
+            var project = new TestRazorProjectFileSystem(new[] { projectItem });
```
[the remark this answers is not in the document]
NTaylorMullen: Of course, it wouldn't build otherwise 😄

toddams:
May I ask why Razor is becoming more and more tightly coupled with FileSystem?

What about custom Razor sources?

NTaylorMullen:
@toddams FileSystem in this PR is our own type, not the `IFileSystem` type used in most core projects. Basically our Razor variant of file system can ready something from anywhere; it just depends on items in the file system having a hierarchy.

toddams:
I mean, for example to generate template code I need RazorEngine.
Previously I used RazorEngine.Create() and it worked file, but now I need to use RazorProjectEngine.Create()
which accepts RazorProjectFileSystem as a second parameter. Does that mean I have to pass some dummy implementation of IFileSystem in order to get RazorEngine from RazorEngineProject?

NTaylorMullen:
Short answer is yes. We changed the requirements to parse a Razor file because the ways we were flowing information into RazorEngine before had a lot of "gotcha" cases since we allowed basically anything to flow through our system.

We should probably expose https://github.com/aspnet/Razor/blob/41324bd37a860753d4e678318b6ba1d367b6249d/src/Microsoft.AspNetCore.Razor.Language/RazorProjectFileSystem.cs#L10 but in the interim you can copy our `EmptyProjectFileSystem`.

toddams:
I realize Razor itself is mostly used in ASP.NET MVC and all the development is performed with this thing in mind. 

Introducing **RazorProject** approach was like a breath of a fresh air comparing to old razor and it became much easier to abstract from the filesystem and MVC-style structure. But now it feels like it's a step backwards.
What I wanted to say, is that if there's a chance to make Razor less locked over ASP.NET MVC - that would be super amazing :)

Issue #2037, opened 2018-02-08
Include VSIX pdbs in build output


NTaylorMullen:
Today our RazorExtension VSIX build symbols are included inside of the VSIX; however, we need to include them in the build output so they can be published properly.

NTaylorMullen:
6c75f294190e2f65cd58c1c5e1ae698958bf3374

PR #2038, opened 2018-02-08
Include VSIX symbols in build output.

NTaylorMullen:
#2037

/cc @mkArtakMSFT


tooling/Microsoft.VisualStudio.RazorExtension/Microsoft.VisualStudio.RazorExtension.csproj
```diff
@@ -45,7 +45,7 @@
     <IncludeDebugSymbolsInVSIXContainer>true</IncludeDebugSymbolsInVSIXContainer>
```
natemcmaster: False?

NTaylorMullen: Pfff you saw nothing; totally didn't just force push that away... I promise

Issue #2049, opened 2018-02-10
Move RazorProjectEngine to use RazorProjectItem as its primary input


NTaylorMullen:
Today RazorProjectEngine allows `string` and `SourceDocument` however, the problem with allowing general `SouceDocument`s is that the `FileSystem` that the `RazorProjectEngine` is based off of can only handle certain paths based off its implementation. I.E. if the `FileSystem` is queried with absolute paths and the implementation happens to be the composite or `VirtualRazorProjectFileSystem` it'll explode. Therefore, when using `SourceDocument`s we don't know whether to pass in the source documents `FilePath` or `RelativeFilePath`. This usually shows up in not finding imports for given Razor files. In the past this was worked around because int the `RazorTemplateEngine` we used black magic to call each `RazorTemplateEngine` method in a super meticulous order with the expectation that the underlying implementations did not call into paths that they'd normally call into.

Moving the `RazorProjectEngine` to consume project items enforces the idea that a user should provide items from the file system or at least items that conform to it.

NTaylorMullen:
3375fc8b998e5e40d5080abd6f3c1cedd18892c9

PR #2050, opened 2018-02-10
Change RazorProjectEngine to operate on RazorProjectItems.

NTaylorMullen:
- Removed the `Process(string)` overload to make it extra clear that you must operate on project items. This way we also don't need to worry about the various formats of paths that can flow through the system.
- Updated tests to use the new project item format.
- Did a few formatting fixes on unrealted files.

#2049 

/cc @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngineFeatureBase.cs
```diff
@@ -9,7 +9,7 @@ public abstract class RazorProjectEngineFeatureBase : IRazorProjectEngineFeature
     {
         private RazorProjectEngine _projectEngine;
 
-        public RazorProjectEngine ProjectEngine
+        public virtual RazorProjectEngine ProjectEngine
```
NTaylorMullen: Unrelated, just should have been done before.

src/Microsoft.AspNetCore.Razor.Language/VirtualRazorProjectFileSystem.cs
```diff
@@ -10,20 +10,19 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     internal class VirtualRazorProjectFileSystem : RazorProjectFileSystem
     {
-        private readonly DirectoryNode Root = new DirectoryNode("/");
+        private readonly DirectoryNode _root = new DirectoryNode("/");
```
NTaylorMullen: Unrelated.

PR #2058, opened 2018-02-12
Decouple Razor tools from MVC

rynowak:
Adds a loader (with shadow copying in server mode) based on the Roslyn
Analyzer loader design.

Adds some targets to the Razor SDK that we can use to compute the
configuration and extensions.

Passes all of the metadata through to the command line tools so they can
deal with extensions.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Properties/AssemblyInfo.cs
@@ -2,6 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Mvc.Razor.Extensions;
+using Microsoft.AspNetCore.Razor.Language;
+
+[assembly: ProvideRazorExtensionInitializer("MVC-2.0", typeof(ExtensionInitializer))]
+[assembly: ProvideRazorExtensionInitializer("MVC-2.1", typeof(ExtensionInitializer))]
 
 [assembly: InternalsVisibleTo("Microsoft.AspNetCore.Mvc.Razor.Extensions.Test, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
 [assembly: InternalsVisibleTo("Microsoft.VisualStudio.Editor.Razor, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs
@@ -25,6 +25,8 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             InheritsDirective.Register(builder);
             SectionDirective.Register(builder);
 
+            builder.Features.Add(new ViewComponentTagHelperDescriptorProvider());
+
             builder.AddTargetExtension(new ViewComponentTagHelperTargetExtension());
             builder.AddTargetExtension(new TemplateTargetExtension()
             {
--- a/src/Microsoft.AspNetCore.Razor.Language/AssemblyExtension.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/AssemblyExtension.cs
@@ -1,0 +1,31 @@
+﻿// Copyright(c) .NET Foundation.All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class AssemblyExtension : RazorExtension
+    {
+        public AssemblyExtension(string extensionName, Assembly assembly)
+        {
+            if (extensionName == null)
+            {
+                throw new ArgumentNullException(nameof(extensionName));
+            }
+
+            if (assembly == null)
+            {
+                throw new ArgumentNullException(nameof(assembly));
+            }
+
+            ExtensionName = extensionName;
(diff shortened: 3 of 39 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorExtensions.cs
```diff
@@ -25,6 +25,8 @@ public static void Register(RazorProjectEngineBuilder builder)
             InheritsDirective.Register(builder);
             SectionDirective.Register(builder);
 
+            builder.Features.Add(new ViewComponentTagHelperDescriptorProvider());
```
rynowak: We should be adding this all of the time. The fact that this code is selective about when we add it creates all kinds of complications.

src/Microsoft.AspNetCore.Razor.Language/AssemblyExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            ExtensionName = extensionName;
+            Assembly = assembly;
+        }
+
+        public override string ExtensionName { get; }
+
+        public Assembly Assembly { get; }
+
+        public void AddExtension(RazorProjectEngineBuilder builder)
```
rynowak: I might change how this works. For instance I kinda want to put this logic inside the `RazorProjectEngine.Create` method, thoughts?

src/Microsoft.AspNetCore.Razor.Language/RazorProjectFileSystem.cs
```diff
@@ -21,5 +24,19 @@ public new static RazorProjectFileSystem Create(string rootDirectoryPath)
 
             return new DefaultRazorProjectFileSystem(rootDirectoryPath);
         }
+
+
+        private class EmptyProjectFileSystem : RazorProjectFileSystem
```
rynowak: On reflection this isn't actually correct. Will clean this up.

src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
```diff
@@ -51,6 +64,21 @@ protected override string GenerateResponseFileCommands()
...
+            builder.AppendLine("-c");
+            builder.AppendLine(Configuration[0].GetMetadata(Identity));
+
+            for (var i = 0; i < Extensions.Length; i++)
+            {
+                builder.AppendLine("-n");
+                builder.AppendLine(Extensions[i].GetMetadata(Identity));
+
+                builder.AppendLine("-e");
+                builder.AppendLine(Path.GetFullPath(Extensions[i].GetMetadata(AssemblyFilePath)));
```
rynowak: Passing the name + assembly is mostly for clarity and diagnostics. It doesn't currently influence any behavior, but it feels wrong to me to leave it out.

src/Microsoft.AspNetCore.Razor.Tools/DefaultExtensionDependencyChecker.cs
```diff
@@ -0,0 +1,155 @@
...
+    {
+        // These are treated as prefixes. So `Microsoft.CodeAnalysis.Razor` would be assumed to work.
+        private static readonly string[] DefaultIgnoredAssemblies = new string[]
+        {
+            "mscorlib",
+            "netstandard",
+            "System",
+            "Microsoft.CodeAnalysis",
+            "Microsoft.AspNetCore.Razor.Language",
+        };
```
rynowak: This is admittedly kinda gross, but it seems good enough for now.

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerLogger.cs
```diff
@@ -139,7 +139,33 @@ private static int GetCurrentThreadId()
         /// </summary>
         private static string GetLoggingPrefix()
         {
-            return string.Format("{0} PID={1} TID={2} Ticks={3}: ", s_prefix, GetCurrentProcessId(), GetCurrentThreadId(), Environment.TickCount);
+            return string.Format("{0} PID={1} TID={2} Ticks={3}: ", _prefix, GetCurrentProcessId(), GetCurrentThreadId(), Environment.TickCount);
+        }
+
+        private class LogWriter : TextWriter
```
rynowak: I don't think I ended up using this for anything. I plan to revert this file.

rynowak:
@pranavkm if you could, please take a look at the MSBuild parts of this PR. I want to get these in ASAP so I can get them into the SDK and minimize the breaking change.

src/Microsoft.AspNetCore.Razor.Language/RazorExtensionInitializer.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    public abstract class RazorExtensionInitializer
+    {
+        public abstract void Initialize(RazorProjectEngineBuilder builder);
```
pranavkm: Should we add an empty ctor to this since we expect it to be initializable without args?

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -181,6 +182,24 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <Target 
+    Name="ResolveRazorConfiguration"
+    Condition="'$(RazorDefaultConfiguration)'!=''">
+
+    <ItemGroup>
+      <ResolvedRazorConfiguration Include="@(RazorConfiguration)" Condition="'%(RazorConfiguration.Identity)'=='$(RazorDefaultConfiguration)'" />
+      <_ResolvedRazorExtensionName Include="%(ResolvedRazorConfiguration.Extensions)"/>
+    </ItemGroup>
+
+    <FindInList List="@(RazorExtension)" ItemSpecToFind="%(_ResolvedRazorExtensionName.Identity)">
```
pranavkm: `<FindInList List="@(RazorExtension)" ItemSpecToFind="@(ResolvedRazorConfiguration->Metadata('Extensions'))">`

src/Microsoft.AspNetCore.Razor.Language/RazorExtensionInitializer.cs
rynowak: I thought about that but it doesn't really do anything. Anyone who subclasses can just define another constructor. I can't think of a way to force a derived class to do the right thing.

rynowak:

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Properties/AssemblyInfo.cs
```diff
@@ -2,6 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Mvc.Razor.Extensions;
+using Microsoft.AspNetCore.Razor.Language;
+
+[assembly: ProvideRazorExtensionInitializer(typeof(ExtensionInitializer))]
```
rynowak: This should require an extension name

src/Microsoft.AspNetCore.Razor.Language/AssemblyExtension.cs
rynowak: I agree with this feedback. I'll give this a try and see if it's small enough to cram into this PR.

rynowak:
@NTaylorMullen updated

rynowak:
I'm going to go ahead with this since we reviewed it in person. If there are any follow up comments I'll address them in a separate checkin.

PR #2061, opened 2018-02-13
Ignoring HTML comments in tag helper's body

mkArtakMSFT:
This change addresses https://github.com/aspnet/Razor/issues/2062 issue.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
@@ -19,7 +19,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
         // Special
         Comment = 8,
         Tag = 9,
-
         HtmlComment = 10
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
@@ -18,6 +18,8 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
         // Special
         Comment = 8,
-        Tag = 9
+        Tag = 9,
+
+        HtmlComment = 10
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
@@ -207,6 +207,10 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             if (last != null)
             {
                 Accept(last);
+                if (At(HtmlSymbolType.OpenAngle) && last.Type == HtmlSymbolType.Text)
+                {
+                    Output(SpanKindInternal.Markup);
+                }
             }
         }
 
@@ -494,26 +498,34 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             {
                 if (CurrentSymbol.Type == HtmlSymbolType.DoubleHyphen)
                 {
-                    AcceptAndMoveNext();
-
-                    Span.EditHandler.AcceptedCharacters = AcceptedCharactersInternal.Any;
-                    while (!EndOfFile)
+                    using (Context.Builder.StartBlock(BlockKindInternal.HtmlComment))
                     {
-                        SkipToAndParseCode(HtmlSymbolType.DoubleHyphen);
-                        if (At(HtmlSymbolType.DoubleHyphen))
+                        AcceptAndMoveNext();
+
+                        Span.EditHandler.AcceptedCharacters = AcceptedCharactersInternal.Any;
+                        while (!EndOfFile)
                         {
-                            AcceptWhile(HtmlSymbolType.DoubleHyphen);
-
-                            if (At(HtmlSymbolType.Text) &&
-                                string.Equals(CurrentSymbol.Content, "-", StringComparison.Ordinal))
+                            SkipToAndParseCode(HtmlSymbolType.DoubleHyphen);
+                            if (At(HtmlSymbolType.DoubleHyphen))
                             {
-                                AcceptAndMoveNext();
-                            }
+                                AcceptWhile(HtmlSymbolType.DoubleHyphen);
 
-                            if (At(HtmlSymbolType.CloseAngle))
-                            {
-                                AcceptAndMoveNext();
-                                return true;
+                                if (At(HtmlSymbolType.Text) &&
+                                    string.Equals(CurrentSymbol.Content, "-", StringComparison.Ordinal))
+                                {
+                                    AcceptAndMoveNext();
+                                }
+
+                                if (At(HtmlSymbolType.CloseAngle))
+                                {
+                                    // This is the end of a comment block
+                                    Accept(this.CurrentSymbol);
+                                    Output(SpanKindInternal.Markup);
+
+                                    NextToken();
+                                    //AcceptAndMoveNext();
+                                    return true;
+                                }
                             }
                         }
                     }
@@ -1476,6 +1488,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                     {
                         AcceptAndMoveNext(); // Accept '<'
                         BangTag();
+
                         return;
                     }
 
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
@@ -613,7 +613,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                     if (NextIs(HtmlSymbolType.CloseAngle))
                     {
                         // Check condition 2.3: We're at the end of a comment. Check to make sure the text ending is allowed.
-                        isValidComment = !SymbolSequenceEndsWithItems(p, HtmlSymbolType.OpenAngle, HtmlSymbolType.Bang, HtmlSymbolType.DoubleHyphen);
+                        isValidComment = !IsCommentContentDisallowed(p);
                         return true;
                     }
                     else if (NextIs(ns => IsDashSymbol(ns) && NextIs(HtmlSymbolType.CloseAngle)))
@@ -643,21 +643,28 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             return isValidComment;
         }
 
-        internal static bool SymbolSequenceEndsWithItems(IEnumerable<HtmlSymbol> sequence, params HtmlSymbolType[] items)
+        /// <summary>
+        /// Verifies, that the sequence doesn't end with the "&lt;!-" HtmlSymbols. Note, the first symbol is an opening bracket symbol
+        /// </summary>
+        internal static bool IsCommentContentDisallowed(IEnumerable<HtmlSymbol> sequence)
         {
-            int index = items.Length;
-            foreach (var previousSymbol in sequence)
+            var reversedSequence = sequence.Reverse();
+            var disallowEnding = new[] { new HtmlSymbol("-", HtmlSymbolType.Text), new HtmlSymbol("!", HtmlSymbolType.Bang), new HtmlSymbol("<", HtmlSymbolType.OpenAngle) };
+            var index = 0;
+            foreach (var item in reversedSequence)
             {
-                if (index == 0)
+                if (!item.Equals(disallowEnding[index++]))
                 {
-                    break;
+                    return false;
                 }
 
-                if (items[--index] != previousSymbol.Type)
-                    return false;
+                if (index == disallowEnding.Length)
+                {
+                    return true;
+                }
             }
 
-            return index == 0;
+            return false;
         }
 
         private bool CData()
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
@@ -614,8 +614,8 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                 {
                     if (NextIs(HtmlSymbolType.CloseAngle))
                     {
-                        // We're at the end of a comment. check the condition 2.3 to make sure the text ending is allowed.
-                        isValidComment = !EndsWithSymbolsSequence(p, HtmlSymbolType.OpenAngle, HtmlSymbolType.Bang, HtmlSymbolType.DoubleHyphen);
+                        // Check condition 2.3: We're at the end of a comment. Check to make sure the text ending is allowed.
+                        isValidComment = !SymbolSequenceEndsWithItems(p, HtmlSymbolType.OpenAngle, HtmlSymbolType.Bang, HtmlSymbolType.DoubleHyphen);
                         breakLookahead = true;
                     }
                     else if (NextIs(ns => IsDashSymbol(ns) && NextIs(HtmlSymbolType.CloseAngle)))
@@ -645,17 +645,17 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             return isValidComment;
         }
 
-        private bool EndsWithSymbolsSequence(IEnumerable<HtmlSymbol> symbols, params HtmlSymbolType[] sequenceToMatchWith)
+        internal static bool SymbolSequenceEndsWithItems(IEnumerable<HtmlSymbol> sequence, params HtmlSymbolType[] items)
         {
-            int index = sequenceToMatchWith.Length;
-            foreach (var previousSymbol in symbols)
+            int index = items.Length;
+            foreach (var previousSymbol in sequence)
             {
                 if (index == 0)
                 {
                     break;
                 }
 
-                if (sequenceToMatchWith[--index] != previousSymbol.Type)
+                if (items[--index] != previousSymbol.Type)
                     return false;
             }
 
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
@@ -492,7 +492,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
             if (AcceptAndMoveNext())
             {
-                if (CurrentSymbol.Type == HtmlSymbolType.DoubleHyphen)
+                if (IsHtmlCommentAhead())
                 {
                     using (Context.Builder.StartBlock(BlockKindInternal.HtmlComment))
                     {
@@ -505,32 +505,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                             SkipToAndParseCode(HtmlSymbolType.DoubleHyphen);
                             if (At(HtmlSymbolType.DoubleHyphen))
                             {
-                                var lastDoubleHyphen = CurrentSymbol;
-                                AcceptWhile(s =>
-                                {
-                                    if (NextIs(HtmlSymbolType.DoubleHyphen))
-                                    {
-                                        lastDoubleHyphen = s;
-                                        return true;
-                                    }
-
-                                    NextToken();
-                                    EnsureCurrent();
-                                    return false;
-                                });
-
-                                if (At(HtmlSymbolType.Text) &&
-                                    string.Equals(CurrentSymbol.Content, "-", StringComparison.Ordinal))
-                                {
-                                    // Doing this here to maintain the order of symbols
-                                    if (!NextIs(HtmlSymbolType.CloseAngle))
-                                    {
-                                        Accept(lastDoubleHyphen);
-                                        lastDoubleHyphen = null;
-                                    }
-
-                                    AcceptAndMoveNext();
-                                }
+                                var lastDoubleHyphen = AcceptAllButLastDoubleHypens();
 
                                 if (At(HtmlSymbolType.CloseAngle))
                                 {
@@ -541,7 +516,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                                     Accept(lastDoubleHyphen);
                                     AcceptAndMoveNext();
                                     Output(SpanKindInternal.Markup, AcceptedCharactersInternal.None);
-
                                     return true;
                                 }
                                 else if (lastDoubleHyphen != null)
@@ -551,8 +525,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                             }
                         }
                     }
-
-                    return false;
                 }
                 else if (CurrentSymbol.Type == HtmlSymbolType.LeftBracket)
                 {
@@ -569,6 +541,125 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             }
 
             return false;
+        }
+
+        private HtmlSymbol AcceptAllButLastDoubleHypens()
+        {
+            var lastDoubleHyphen = CurrentSymbol;
+            AcceptWhile(s =>
+            {
+                if (NextIs(HtmlSymbolType.DoubleHyphen))
+                {
+                    lastDoubleHyphen = s;
+                    return true;
+                }
+
+                NextToken();
+                EnsureCurrent();
+                return false;
+            });
+
+            if (At(HtmlSymbolType.Text) && IsDashSymbol(CurrentSymbol))
+            {
+                // Doing this here to maintain the order of symbols
+                if (!NextIs(HtmlSymbolType.CloseAngle))
+                {
+                    Accept(lastDoubleHyphen);
+                    lastDoubleHyphen = null;
+                }
+
+                AcceptAndMoveNext();
+            }
+
+            return lastDoubleHyphen;
+        }
+
+        private static bool IsDashSymbol(HtmlSymbol symbol)
+        {
+            return string.Equals(symbol.Content, "-", StringComparison.Ordinal);
+        }
+
+        private bool IsHtmlCommentAhead()
+        {
+            /*
+             * From HTML5 Specification, available at http://www.w3.org/TR/html52/syntax.html#comments
+             * 
+             * Comments must have the following format:
+             * 1. The string "<!--"
+             * 2. Optionally, text, with the additional restriction that the text
+             *      2.1 must not start with the string ">" nor start with the string "->"
+             *      2.2 nor contain the strings "<!--", "-->", or "--!>"
+             *      2.3 nor end with the string "<!-".
+             * 3. The string "-->"
+             * 
+             * */
+
+            if (CurrentSymbol.Type != HtmlSymbolType.DoubleHyphen)
+            {
+                return false;
+            }
+
+            // Check condition 2.1
+            if (NextIs(HtmlSymbolType.CloseAngle) || NextIs(next => IsDashSymbol(next) && NextIs(HtmlSymbolType.CloseAngle)))
+            {
+                return false;
+            }
+
+            // Check condition 2.2
+            bool isValidComment = false;
+            LookaheadUntil((s, p) =>
+            {
+                bool breakLookahead = false;
+                if (s.Type == HtmlSymbolType.DoubleHyphen)
+                {
+                    if (NextIs(HtmlSymbolType.CloseAngle))
+                    {
+                        // We're at the end of a comment. check the condition 2.3 to make sure the text ending is allowed.
+                        isValidComment = !EndsWithSymbolsSequence(p, HtmlSymbolType.OpenAngle, HtmlSymbolType.Bang, HtmlSymbolType.DoubleHyphen);
+                        breakLookahead = true;
+                    }
+                    else if (NextIs(ns => IsDashSymbol(ns) && NextIs(HtmlSymbolType.CloseAngle)))
+                    {
+                        // This is also a valid closing comment case, as the dashes lookup is treated with DoubleHyphen symbols first.
+                        isValidComment = true;
+                        breakLookahead = true;
+                    }
+                    else if (NextIs(ns => ns.Type == HtmlSymbolType.Bang && NextIs(HtmlSymbolType.CloseAngle)))
+                    {
+                        isValidComment = false;
+                        breakLookahead = true;
+                    }
+                }
+                else if (s.Type == HtmlSymbolType.OpenAngle)
+                {
+                    if (NextIs(ns => ns.Type == HtmlSymbolType.Bang && NextIs(HtmlSymbolType.DoubleHyphen)))
+                    {
+                        isValidComment = false;
+                        breakLookahead = true;
+                    }
+                }
+
+                return breakLookahead;
+            });
+
+            return isValidComment;
+        }
+
+        private bool EndsWithSymbolsSequence(IEnumerable<HtmlSymbol> symbols, params HtmlSymbolType[] sequenceToMatchWith)
+        {
+            int index = sequenceToMatchWith.Length;
+            foreach (var previousSymbol in symbols)
+            {
+                if (index == 0)
+                {
+                    break;
+                }
+
+                if (sequenceToMatchWith[--index] != previousSymbol.Type)
+                    return false;
+            }
+
+            return index == 0;
         }
 
         private bool CData()
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
@@ -496,6 +496,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                 {
                     using (Context.Builder.StartBlock(BlockKindInternal.HtmlComment))
                     {
+                        // Accept the double-hyphen symbol at the beginning of the comment block.
                         AcceptAndMoveNext();
                         Output(SpanKindInternal.Markup, AcceptedCharactersInternal.None);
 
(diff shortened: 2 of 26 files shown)


mkArtakMSFT:
@NTaylorMullen, @rynowak this is my first attempt. although the test passes, I would like to get your opinion about the approach.

mkArtakMSFT:
Thanks @grahamehorner. I'll try to address that too...

grahamehorner:
Your email has been reported as SPAM and your email domain will be blocked if you continue to send unsolicited emails repeatedly.

iot-cloud


Gabriel Cloud System Limited
4 St. Gabriel Ave, Riverside Village,
Hebburn, Tyne & Wear, England NE31 1AG
CAUTION:This email and any attachments transmitted with it contain information that is CONFIDENTIAL and may be privileged or otherwise protected from disclosure. If you are not the intended recipient please notify us immediately and delete all copies of this email message and any attachments transmitted with it from your system. If you are not the intended recipient, any use, disclosure or copying of this email message and any attachments transmitted with it is strictly prohibited. This email message and any attachments transmitted with it cannot be guaranteed to be secure or error free as information could be intercepted, corrupted, lost, destroyed or incomplete. Although we routinely screen for viruses, it makes no representation or warranty as to the absence of viruses in this email or any attachments. The views expressed in this email are those of the sender and not necessarily those of Gabriel Cloud Systems LTD; if you require clarification please contact the sender.
proud sponsors of
Hebburn Town Juniors FC
(Junior Madrid)

Get Outlook<https://aka.ms/sdimjr> for iOS
________________________________

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -817,6 +817,11 @@ private static bool IsEndTag(Block tagBlock)
             return relevantSymbol.Type == HtmlSymbolType.ForwardSlash;
         }
 
+        private static bool IsCommentTag(Span span)
+        {
+            return span.Content.StartsWith("<!--");
```
ajaybhargavb: I think we should verify `Span.Kind` here instead. We have a `Comment` kind.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -116,7 +116,7 @@ private void RewriteTags(Block input, ErrorSink errorSink, int depth)
                         continue;
                     }
                 }
-                else
+                else if (!IsCommentTag((Span)child))
```
ajaybhargavb: Also suggest putting this check inside the Validate method so we have all validation in one place.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
mkArtakMSFT: HTML Comments are treated as `Markup`, not `Comment`

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -817,6 +817,14 @@ private static bool IsEndTag(Block tagBlock)
             return relevantSymbol.Type == HtmlSymbolType.ForwardSlash;
         }
 
+        private static bool IsComment(Span span)
+        {
+            bool isHtmlComment = span.Content?.StartsWith("<!--") == true;
```
NTaylorMullen: compare symbols. Using `.Content` is actually quite expensive. Can also simplify the execution path:

```C#
if (IBeHtmlCommentLol)
{
    // HTML comment <!-- ... -->
    return true;
}

if (IBeRazorCommentLol)
{
    // Razor comment @* .. *@
    return true;
}

return false
```

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -817,6 +817,14 @@ private static bool IsEndTag(Block tagBlock)
             return relevantSymbol.Type == HtmlSymbolType.ForwardSlash;
         }
 
+        private static bool IsComment(Span span)
```
NTaylorMullen: Mark this internal with a `// Internal for testing` and unit test

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
NTaylorMullen: Also to note, we use `var` instead of the type names

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperParseTreeRewriterTest.cs
```diff
@@ -1108,6 +1108,91 @@ public static TheoryData AllowedChildrenData
             EvaluateData(descriptors, documentContent, (MarkupBlock)expectedOutput, (RazorDiagnostic[])expectedErrors);
         }
 
+        [Fact]
+        public void Rewrite_AllowsHtmlCommentsAsChildren()
+        {
+            // Arrangestring documentContent,
```
NTaylorMullen: comment goodness 😄

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperParseTreeRewriterTest.cs
```diff
@@ -1108,6 +1108,91 @@ public static TheoryData AllowedChildrenData
...
+            IEnumerable<string> allowedChildren = new List<string> { "b" };
+            string literal = "asdf";
+            string commentOutput = $"<!--Hello World-->";
```
NTaylorMullen: Given that this is an integration test I'd make this a little more complex: `<!-- Hello @world -->`.

Also, no need for `$`

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/TagHelperParseTreeRewriterTest.cs
```diff
@@ -1108,6 +1108,91 @@ public static TheoryData AllowedChildrenData
             EvaluateData(descriptors, documentContent, (MarkupBlock)expectedOutput, (RazorDiagnostic[])expectedErrors);
         }
 
+        [Fact]
+        public void Rewrite_AllowsHtmlCommentsAsChildren()
```
NTaylorMullen: vars in this file, can apply general comments in this method to the one below as well.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
mkArtakMSFT: Fixed

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
NTaylorMullen: > Otherwise would have to do the same for the other methods too, which is out of scope from this change.

Don't need to make it consistent. This is a new method on a legacy class that we should definitely make sure works on its own. In an ideal world we would have done that for all of the private methods in this class but at the time we chose to rely on integration tests to verify all of this works. We've definitely had to pay for that decision in test maintenance 😢

mkArtakMSFT:
Ok folks. I've finished updating all the tests and everything is passing now. @NTaylorMullen, there is one issue I've noticed, which I would like to change, but wanted to run this by you first. When there is non-closed comment tag, the actual markup after the opening comment tag is not part of the comment block, but a new markup block. While this is a corner case scenario, I think it should still be part of the comment block, so I'd be happy to fix that too, if you agree.

rynowak:
Does this need a feature/version flag? This is a change in what the language accepts?

NTaylorMullen:
> When there is non-closed comment tag, the actual markup after the opening comment tag is not part of the comment block, but a new markup block. While this is a corner case scenario, I think it should still be part of the comment block, so I'd be happy to fix that too, if you agree.

Hmm, so we need to abide by the HTML5 spec. It has rules for classifying what an HTML comment is and is not.



> Does this need a feature/version flag? This is a change in what the language accepts?

Yes, for older versions of the runtime we need to still disallow HTML comments.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+                    }
+                }
+
+                return breakLookahead;
+            });
+
+            return isValidComment;
+        }
+
+        private bool EndsWithSymbolsSequence(IEnumerable<HtmlSymbol> symbols, params HtmlSymbolType[] sequenceToMatchWith)
```
mkArtakMSFT: Turning this to static, as this has no state dependency

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+                    lastDoubleHyphen = s;
+                    return true;
+                }
+
+                NextToken();
+                EnsureCurrent();
+                return false;
+            });
+
+            if (At(HtmlSymbolType.Text) && IsDashSymbol(CurrentSymbol))
```
NTaylorMullen: Not sure I follow the purpose of this block, could you explain?

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+            }
+
+            // Check condition 2.1
+            if (NextIs(HtmlSymbolType.CloseAngle) || NextIs(next => IsDashSymbol(next) && NextIs(HtmlSymbolType.CloseAngle)))
+            {
+                return false;
+            }
+
+            // Check condition 2.2
+            bool isValidComment = false;
```
NTaylorMullen: var is your friend 😉, we don't explicitly call out variable types anywhere.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+            LookaheadUntil((s, p) =>
+            {
+                bool breakLookahead = false;
```
NTaylorMullen: just return inline, no reason to have this variable.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+
+            // Check condition 2.1
+            if (NextIs(HtmlSymbolType.CloseAngle) || NextIs(next => IsDashSymbol(next) && NextIs(HtmlSymbolType.CloseAngle)))
+            {
+                return false;
+            }
+
+            // Check condition 2.2
+            bool isValidComment = false;
+            LookaheadUntil((s, p) =>
```
NTaylorMullen: Use full names for the parameters here since you use them a lot.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+            {
+                bool breakLookahead = false;
+                if (s.Type == HtmlSymbolType.DoubleHyphen)
+                {
+                    if (NextIs(HtmlSymbolType.CloseAngle))
+                    {
```
NTaylorMullen: I'd add comments for all these blocks based on what they account for. I.e. this block accounts for `<!-->`

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+                    }
+                }
+
+                return breakLookahead;
+            });
+
+            return isValidComment;
+        }
+
+        internal static bool SymbolSequenceEndsWithItems(IEnumerable<HtmlSymbol> sequence, params HtmlSymbolType[] items)
```
NTaylorMullen: no need for params since you only ever use it as an array.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -1474,8 +1599,14 @@ private void ScanTagInDocumentContext()
                     // Checking to see if we meet the conditions of a special '!' tag: <!DOCTYPE, <![CDATA[, <!--.
                     if (!IsBangEscape(lookahead: 1))
                     {
+                        if (Lookahead(2)?.Type == HtmlSymbolType.DoubleHyphen)
```
NTaylorMullen: Why this?

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -488,7 +488,13 @@ private bool HasAllowedChildren()
 
         private void ValidateParentAllowsContent(Span child, ErrorSink errorSink)
         {
-            if (HasAllowedChildren())
+            var isDisallowedContent = true;
+            if (_featureFlags.AllowHtmlCommentsInTagHelpers)
```
NTaylorMullen: Do this bit inside the `HasAllowedChildren()` block to avoid doing the work if unneeded.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -109,6 +109,53 @@ protected TSymbol Lookahead(int count)
...
+
+            while (true)
+            {
+                if (!NextToken())
+                {
+                    break;
+                }
+
+                symbols.Add(CurrentSymbol);
+                if (condition(CurrentSymbol, symbols.Reverse<TSymbol>()))
```
NTaylorMullen: Instead of constantly reversing the symbols i'd just `Insert(0)` when building the symbols list.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+
+            return lastDoubleHyphen;
+        }
+
+        private static bool IsDashSymbol(HtmlSymbol symbol)
+        {
+            return string.Equals(symbol.Content, "-", StringComparison.Ordinal);
+        }
+
+        private bool IsHtmlCommentAhead()
```
NTaylorMullen: Should be unit testing this guy.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
             return false;
         }
 
+        private HtmlSymbol AcceptAllButLastDoubleHypens()
```
NTaylorMullen: Should be unit testing.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +543,125 @@ private bool BangTag()
...
+                    lastDoubleHyphen = null;
+                }
+
+                AcceptAndMoveNext();
+            }
+
+            return lastDoubleHyphen;
+        }
+
+        private static bool IsDashSymbol(HtmlSymbol symbol)
```
NTaylorMullen: Should be unit testing.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: fixed

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: fixed

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
mkArtakMSFT: There is logical difference between those two checks. HasAllowedChildren has a different meaning, so it doesn't make sense to combine those. I'll just move that into an if block.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
mkArtakMSFT: That'll be way too slower. I just created a test to compare those two and for 100K operations current approach takes ~5ms, whereas with the proposed approach - ~700ms. Leaving as is.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: Think about the following case: "--->" = "--" + "->" = DoubleHyphen + "-" + ClosingAngle
The tokenizer will start parsing the long dashes from the beginning, accepting doubleHyphen symbols.
First symbol will be processed, and the leftover part will be "->". This, by itself isn't really a comment closing. As that logic is purely based on our approach, we should make sure we process that case in swapped order. So here I approach it:
"--->" = "--" + "-" + ">" = "-" + "--" + ">"

So here I don't accept the last double hyphen symbol but keep it in memory just in case there'll be a single dash there. If so, I first accept the dash, so I have double hyphen and closing angle nicely stacked up. Hope this makes sense

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -492,33 +492,39 @@ private bool BangTag()
...
+                        AcceptAndMoveNext();
+                        Output(SpanKindInternal.Markup, AcceptedCharactersInternal.None);
 
-                            if (At(HtmlSymbolType.Text) &&
-                                string.Equals(CurrentSymbol.Content, "-", StringComparison.Ordinal))
+                        Span.EditHandler.AcceptedCharacters = AcceptedCharactersInternal.WhiteSpace;
+                        while (!EndOfFile)
+                        {
+                            SkipToAndParseCode(HtmlSymbolType.DoubleHyphen);
+                            if (At(HtmlSymbolType.DoubleHyphen))
```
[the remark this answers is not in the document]
mkArtakMSFT: I followed the usage pattern in other places, where the things was checked again after.
Bug now as we already have looked ahead and confirmed this is a valid HTML comment, doing so is unnecessary. Removed the check.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -492,33 +492,39 @@ private bool BangTag()
...
-
-                    Span.EditHandler.AcceptedCharacters = AcceptedCharactersInternal.Any;
-                    while (!EndOfFile)
+                    using (Context.Builder.StartBlock(BlockKindInternal.HtmlComment))
                     {
-                        SkipToAndParseCode(HtmlSymbolType.DoubleHyphen);
-                        if (At(HtmlSymbolType.DoubleHyphen))
-                        {
-                            AcceptWhile(HtmlSymbolType.DoubleHyphen);
+                        AcceptAndMoveNext();
```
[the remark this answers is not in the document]
mkArtakMSFT: correct. Added the comment.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: We want to make sure that the '<' will not be part of whatever content was before it, in case of a comment

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
NTaylorMullen: O_o super shocked. 👍

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
NTaylorMullen: In the case of a triple dash, doesnt the last `-` get re-ordered?

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: done

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: done

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: done

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: I kind of was in search for better approach, and finally came up with something better.

mkArtakMSFT:
@ajaybhargavb , @NTaylorMullen please look through one more time. Thanks!

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+
+            // Check condition 2.1
+            if (NextIs(HtmlSymbolType.CloseAngle) || NextIs(next => IsDashSymbol(next) && NextIs(HtmlSymbolType.CloseAngle)))
+            {
+                return false;
+            }
+
+            // Check condition 2.2
+            var isValidComment = false;
+            LookaheadUntil((s, p) =>
```
ajaybhargavb: Suggest using full names here to make it easier to understand. It's not easy to figure out what `p` is here.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+            {
+                if (s.Type == HtmlSymbolType.DoubleHyphen)
+                {
+                    if (NextIs(HtmlSymbolType.CloseAngle))
+                    {
+                        // Check condition 2.3: We're at the end of a comment. Check to make sure the text ending is allowed.
+                        isValidComment = !IsCommentContentDisallowed(p);
```
ajaybhargavb: Unnecessary double negative here. Better to name this method `EnsureValidCommentContent` or `IsValidCommentContent` and invert the checks.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+            return isValidComment;
+        }
+
+        /// <summary>
+        /// Verifies, that the sequence doesn't end with the "&lt;!-" HtmlSymbols. Note, the first symbol is an opening bracket symbol
+        /// </summary>
+        internal static bool IsCommentContentDisallowed(IEnumerable<HtmlSymbol> sequence)
+        {
+            var reversedSequence = sequence.Reverse();
+            var disallowEnding = new[] { new HtmlSymbol("-", HtmlSymbolType.Text), new HtmlSymbol("!", HtmlSymbolType.Bang), new HtmlSymbol("<", HtmlSymbolType.OpenAngle) };
```
NTaylorMullen: Throw this into a static field so its not reconstructed every time.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+                    if (NextIs(HtmlSymbolType.CloseAngle))
+                    {
+                        // Check condition 2.3: We're at the end of a comment. Check to make sure the text ending is allowed.
+                        isValidComment = !IsCommentContentDisallowed(p);
+                        return true;
+                    }
+                    else if (NextIs(ns => IsDashSymbol(ns) && NextIs(HtmlSymbolType.CloseAngle)))
+                    {
+                        // This is also a valid closing comment case, as the dashes lookup is treated with DoubleHyphen symbols first.
+                        isValidComment = true;
```
ajaybhargavb: why is there no check here for the comment content? `isValidComment = !IsCommentContentDisallowed(p);`
Same question for the `else if` below.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+                        return true;
+                    }
+                    else if (NextIs(ns => ns.Type == HtmlSymbolType.Bang && NextIs(HtmlSymbolType.CloseAngle)))
+                    {
+                        isValidComment = false;
```
ajaybhargavb: Looks like this is checking if `--!>` is present in the content. Shouldn't this check be part of `isCommentContentDisallowed()`. That way all the checks about the comment content stay within this method.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/HtmlMarkupParserTests.cs
```diff
@@ -0,0 +1,206 @@
...
+        {
+            // Arrange
+            var sut = CreateTestParserForContent("-- not closed comment");
+
+            // Act & Assert
+            Assert.False(sut.IsHtmlCommentAhead());
+        }
+
+        [Fact]
+        public void IsHtmlCommentAhead_ReturnsFalseForCommentWithoutLastClosingAngle()
```
NTaylorMullen: Add a case for `@DateTime.Now` inside the comment.

test/Microsoft.AspNetCore.Razor.Language.Test/TagHelperParseTreeRewriterTests.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Language.Test
+{
+    public class TagHelperParseTreeRewriterTests
```
NTaylorMullen: Add a few more test cases:

- Razor comments since you added support for that.
- Your parser feature flag being turned on/off

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: Sure, I thought about that actually, but earlier decided not to do that, as in case where there are no comments, it wouldn't be used.

ajaybhargavb: > 2.2 nor contain the strings "", or "--!>"

I don't see anything that disallows `<!--` within the comment content.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/HtmlMarkupParserTests.cs
mkArtakMSFT: Done

src/Microsoft.AspNetCore.Razor.Language/Legacy/BlockKindInternal.cs
```diff
@@ -18,6 +18,8 @@ internal enum BlockKindInternal
 
         // Special
         Comment = 8,
-        Tag = 9
+        Tag = 9,
+
```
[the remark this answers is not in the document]
mkArtakMSFT: Fixed

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+                {
+                    lastDoubleHyphen = s;
+                    return true;
+                }
+
+                return false;
+            });
+
+            NextToken();
+            EnsureCurrent();
```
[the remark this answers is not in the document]
mkArtakMSFT: Although I found no explicit call to EnsureCurrent() inside NextToken, but the tests pass, so am removing it. Thanks!

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
             return false;
         }
 
+        protected HtmlSymbol AcceptAllButLastDoubleHypens()
```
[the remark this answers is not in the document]
mkArtakMSFT: Fixed

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -537,6 +541,132 @@ private bool BangTag()
...
+                    lastDoubleHyphen = null;
+                }
+
+                AcceptAndMoveNext();
+            }
+
+            return lastDoubleHyphen;
+        }
+
+        internal static bool IsDashSymbol(HtmlSymbol symbol)
```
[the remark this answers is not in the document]
mkArtakMSFT: Done

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: Done

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: I've been thinkin about this, but the code looks more understandable this way inside the method. Sorry, won't change this.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: I see that the naming is not ideal here and it causes confusion. I'm going to rename the "IsCommentContentDisallowed" to something better to indicate that it checks only for invalid ending for comment content.

grahamehorner: IsCommentTerminatorValid ?

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: It's actually handle. Added more comments to make it simpler to understand

grahamehorner:
looking good to me, thanks for all your work on this guys

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
mkArtakMSFT: I've added comment explaining that. It's a bit complicated to think about all the possibilities, but tests do cover all the scenarios, so I hope it can serve as a good confirmation.

mkArtakMSFT:
@ajaybhargavb, I've applied your feedback. The algorithm looks good, I've added cases and clarifying comments to show that all the cases are covered.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: I read your comment. I'm still not fully convinced about this. Maybe I am still missing something. I see you have a test with 
`-- Some comment content in here <!--->` which is invalid. But is the following valid(Should it be allowed and does the current code allow it)?
`-- Some comment content in here <!---->` (Notice the extra dash at the end)

test/Microsoft.AspNetCore.Razor.Language.Test/TagHelperParseTreeRewriterTests.cs
mkArtakMSFT: Done

mkArtakMSFT:
Thanks for the feedback, guys! The build is constantly failing on Mac, but succeeding on Linux. The failure is unrelated to the code change, so I'm going to finalize this now.

PR #2064, opened 2018-02-13
Make imports work without warnings or errors when the Sdk is referenc…

pranavkm:
…ed via MsBuild and package.

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.props
+++ b/src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.props
@@ -10,7 +10,5 @@ Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
 <Project ToolsVersion="14.0">
-
-  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.Razor.CurrentVersion.props" />
-
+  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.Razor.CurrentVersion.props" Condition="'$(_RazorSdkPackageReferenced)' != 'true'" />
 </Project>
--- a/src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.targets
+++ b/src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.targets
@@ -11,11 +11,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project ToolsVersion="14.0">
 
-  <PropertyGroup Condition="'$(RazorSdkTargetsPath)' == ''">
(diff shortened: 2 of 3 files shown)


src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.props
```diff
@@ -10,7 +10,5 @@ Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
 <Project ToolsVersion="14.0">
-
-  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.Razor.CurrentVersion.props" />
-
+  <Import Project="$(MSBuildThisFileDirectory)..\build\netstandard2.0\Sdk.Razor.CurrentVersion.props" Condition="'$(_RazorSdkPackageReferenced)' != 'true'" />
```
pranavkm: Re-importing this file twice was kinda fine in the past since we only set up `PropertyGroup` things. Now that we add to `ItemGroup` this becomes problematic since it results in duplicate entries.

src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.targets
```diff
@@ -11,11 +11,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 -->
 <Project ToolsVersion="14.0">
 
-  <PropertyGroup Condition="'$(RazorSdkTargetsPath)' == ''">
```
pranavkm: The target in the package expects the Sdk to import this file. It affects a corner scenario where you only reference the package (but not the Sdk). This just leaves it up to Sdk\ package to import their own targets.

src/Microsoft.NET.Sdk.Razor/buildMultiTargeting/Microsoft.NET.Sdk.Razor.props
```diff
@@ -0,0 +1,16 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0">
+  <PropertyGroup>
+    <_RazorSdkPackageReferenced>true</_RazorSdkPackageReferenced>
```
pranavkm: Keeping it consistent with the non multi-targeting scenario.

src/Microsoft.NET.Sdk.Razor/Sdk/Sdk.targets
pranavkm: It ended up being used in regular package references scenarios too, not just test. We don't particularly use this anymore, so removing it.

PR #2051, opened 2018-02-10
Move to RazorProjectEngine.

NTaylorMullen:
- Instead of using Razor/Mvc TemplateEngine use `RazorProjectEngine`. This involved changing several locations (each of which used `RazorTemplateEngine` in an entirely different way) to use the RazorProjectEngine's two Process methods.
- Changed an unused public API `VisualStudioRazorParser.TemplateEngine` to `VisualStudioRazorParser.RazorProjectEngine`.
- Ported the remainder of `RazorEngineBuilder`'s extension methods over to `RazorProjectEngineBuilder`. These were used in tests and our `RazorGenerate` tool.
- Added a few test helper methods/classes to enable simple testing of the `RazorProjectEngine`.
- Resolved several test hacks that were working around little discrepancies each of the `RazorTemplateEngine` APIs.
- Changed the template engine factory service to be a project engine factory service.

**Note:** All tests pass in Razor, still need to do the manual labor of verifying this works in MVC (once converted) and in tooling.

#1828

PR #2066, opened 2018-02-13
Get rid of RazorConfiguration.DesignTime

rynowak:
This change makes it so that we no longer create 'design time' engines.
The choice of design time or runtime is made when we initiate a code
generation operation.

Options instances are now created as part of the CodeDocument
initialization. Our existing code can still be created using a
RazorEngine so our passes that initialize the options still support the
old code path.


src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/ModelDirective.cs
```diff
@@ -79,7 +72,7 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument, DocumentInte
                 var visitor = new Visitor();
                 var modelType = GetModelType(documentNode, visitor);
 
-                if (_designTime)
+                if (documentNode.Options.DesignTime)
```
rynowak: There aren't really that many places where we need to know up front if something is design time or not. This was probably the simplest part of the PR.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcImportProjectFeature.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X
 {
-    internal class DefaultMvcImportFeature : RazorProjectEngineFeatureBase, IRazorImportFeature
+    internal class MvcImportProjectFeature : RazorProjectEngineFeatureBase, IImportProjectFeature
```
rynowak: I did some naming cleanup on these classes. They are all *XyzProjectFeature* now.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -31,7 +32,7 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
             var document = new DocumentIntermediateNode();
             var builder = IntermediateNodeBuilder.Create(document);
 
-            document.Options = _optionsFeature.GetOptions();
+            document.Options = codeDocument.GetCodeGenerationOptions() ?? _optionsFeature.GetOptions();
```
rynowak: This phase can still be used with the 2.0 code path, so it needs to support the old way of creating the options.

src/Microsoft.AspNetCore.Razor.Language/RazorCodeGenerationOptionsBuilder.cs
```diff
@@ -5,7 +5,9 @@ namespace Microsoft.AspNetCore.Razor.Language
 {
     public abstract class RazorCodeGenerationOptionsBuilder
     {
-        public abstract bool DesignTime { get; }
+        public virtual RazorConfiguration Configuration => null;
+
+        public abstract bool DesignTime { get; set; }
```
rynowak: @Eilon - how do you feel about a small breaking change like this?? The alternative would be to make the setter a method call, which would be gross but non-breaking.

This is a class that someone is very unlikely to have inherited from.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -3,26 +3,63 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language.Extensions;
```
rynowak: Take a beat to look at what I did here, because I expanded the surface of this class significantly, but it's mostly convenience, clarity, and surfacing things we already have.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -3,26 +3,63 @@
...
+                throw new ArgumentNullException(nameof(projectItem));
+            }
+
+            return ProcessCore(projectItem, ConfigureDesignTimeParserOptions, ConfigureDesignTimeCodeGenerationOptions);
+        }
+
+        protected abstract RazorCodeDocument ProcessCore(
+            RazorProjectItem projectItem,
+            Action<RazorParserOptionsBuilder> configureParser,
+            Action<RazorCodeGenerationOptionsBuilder> configureCodeGeneration);
```
rynowak: This is the main spirit of this change. You configure the options - including design time - once per code document.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -37,54 +74,90 @@ public abstract class RazorProjectEngine
...
+            AddDefaultsFeatures(builder.Features);
 
             configure?.Invoke(builder);
 
             return builder.Build();
         }
-
-        private static void AddDefaults(RazorProjectEngineBuilder builder)
+        
+        private static void AddDefaultsFeatures(ICollection<IRazorFeature> features)
```
rynowak: I decided to copy the existing code since this was diverging from the `RazorEngine` path more and more.

src/Microsoft.AspNetCore.Razor.Language/RazorCodeGenerationOptionsBuilder.cs
rynowak: The other Idea that I have is to add `IsDesignTime` and obsolete this one. I don't like that idea as much because there's a bunch of other places with a `DesignTime` property and this one would stick out.

Eilon: It does appear that no one in the universe at all is using this API outside of us. Nothing on the WORLD-WIDE web, nothing on GitHub.

I'd still be more comfortable if we didn't break it, though. I'd like us to be more serious about not breaking things whenever reasonably possible. Adding a method for setting is a bit gross, but hardly a big deal for us to do.

rynowak: K

rynowak:
Updated and removed breaking change

rynowak:
Updated and resolved taylor's feedback.

I can take over your PR https://github.com/aspnet/Razor/pull/2051 and rebase on top of this once it's in.

Issue #2068, opened 2018-02-14
Change IImportProjectFeature to return a list of project items


NTaylorMullen:
Today we return a list of source documents but those source documents don't provide enough pathing information. Instead, we should return project items and then convert them into source documents.

NTaylorMullen:
e200b69511e10d3c3c367a7ea96a3d2465c8dd7c

PR #2072, opened 2018-02-15
Change IImportProjectFeature to consume RazorProjectItems.

NTaylorMullen:
- Updated all implementations of `IImportProjectFeature`; for MVC I went ahead and made a single project item that's always returned for MVC scenarios. That project item is smart about returning its content in a light-weight stream fashion.
- Had to add a `RazorProjectItem` => `RazorSourceDocument` conversion mechanic into `DefaultRazorProjectEngine`.
- Added tests for `DefaultRazorProjectItem.ConvertToSourceDocument`.
- Removed the `ProjectEngine` API from `VisualStudioRazorParser`. This was unrelated but was missed feedback.

#2068

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcImportProjectFeature.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcImportProjectFeature.cs
@@ -14,14 +14,14 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X
     {
         private const string ImportsFileName = "_ViewImports.cshtml";
 
-        public IReadOnlyList<RazorSourceDocument> GetImports(RazorProjectItem projectItem)
+        public IReadOnlyList<RazorProjectItem> GetImports(RazorProjectItem projectItem)
         {
             if (projectItem == null)
             {
                 throw new ArgumentNullException(nameof(projectItem));
             }
 
-            var imports = new List<RazorSourceDocument>();
+            var imports = new List<RazorProjectItem>();
             AddDefaultDirectivesImport(imports);
 
             // We add hierarchical imports second so any default directive imports can be overridden.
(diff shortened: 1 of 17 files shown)


src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
```diff
@@ -11,8 +11,6 @@ public abstract class VisualStudioRazorParser
     {
         public abstract event EventHandler<DocumentStructureChangedEventArgs> DocumentStructureChanged;
 
-        public abstract RazorProjectEngine ProjectEngine { get; }
```
NTaylorMullen: Unrelated, just missed feedback from the previous PR.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcImportProjectFeature.cs
```diff
@@ -31,54 +31,59 @@ public IReadOnlyList<RazorSourceDocument> GetImports(RazorProjectItem projectIte
...
+                contentBytes.CopyTo(_defaultImportBytes, preamble.Length);
             }
+
+            public override string BasePath => null;
+
+            public override string FilePath => null;
+
+            public override string PhysicalPath => null;
+
+            public override bool Exists => false;
```
rynowak: Should be true YO

NTaylorMullen: Solid catch, should be that way in the non-1_X as well 👍

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -133,5 +136,25 @@ private void ConfigureDesignTimeCodeGenerationOptions(RazorCodeGenerationOptions
...
+        // Internal for testing
+        internal static RazorSourceDocument ConvertToSourceDocument(RazorProjectItem importItem)
+        {
+            if (importItem.Exists)
+            {
+                // Normal import, has file paths, content etc.
+                return RazorSourceDocument.ReadFrom(importItem);
+            }
+
+            if (importItem.FilePath == null)
```
rynowak: if exists is true for global (non-file) imports, then you don't need this case.

rynowak:
Waiting for y'all to merge this so I can integrate it to 15.7 👍

src/Microsoft.VisualStudio.Editor.Razor/VisualStudioRazorParser.cs
rynowak: OK cool beans

Issue #2075, opened 2018-02-15
ResolveTagHelperRazorGenerateInputs target gets invoked even when no inputs change


ajaybhargavb:
This was causing the `RazorTagHelper` task to run even when there were no changes to any csharp files.

ajaybhargavb:
c2932674216e0dd72b60991fa303fbfd8867d3b1

PR #2076, opened 2018-02-15
Fixed unnecessary invocation of TagHelper target

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2075

MsBuild was still deleting the `SimpleMvc.TagHelpers.input.cache` file when we add it to FileWrites with `ItemName=FileWrites`.

cc' @pranavkm who has more context.

Also added a test.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
+++ b/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
@@ -60,11 +60,11 @@
       -->
     <Touch
       Files="$(_RazorTagHelperInputCache)"
-      AlwaysCreate="true">
-      <Output
-        TaskParameter="TouchedFiles"
-        ItemName="FileWrites" />
-    </Touch>
+      AlwaysCreate="true" />
+
+    <ItemGroup>
+      <FileWrites Include="$(_RazorTagHelperInputCache)" />
+    </ItemGroup>
 
     <RazorTagHelper
       Debug="$(_RazorDebugTagHelperTask)"
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
@@ -1,0 +1,60 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
+{
+    public class BuildIncrementalismTest : MSBuildIntegrationTestBase
+    {
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task BuildIncremental_SimpleMvc_PersistsTargetInputFile()
+        {
+            // Arrange
+            var thumbprintLookup = new Dictionary<string, FileThumbPrint>();
+
+            // Act 1
+            var result = await DotnetMSBuild("Build", $"/p:RazorCompileOnBuild=true");
+
+            var directoryPath = Path.Combine(result.Project.DirectoryPath, IntermediateOutputPath);
+            var filesToIgnore = new[]
+            {
+                // These files are generated on every build.
+                Path.Combine(directoryPath, "SimpleMvc.csproj.CopyComplete"),
+                Path.Combine(directoryPath, "SimpleMvc.csproj.FileListAbsolute.txt"),
+            };
+            var files = Directory.GetFiles(directoryPath).Where(p => !filesToIgnore.Contains(p));
+            foreach (var file in files)
+            {
+                var thumbprint = GetThumbPrint(file);
+                thumbprintLookup[file] = thumbprint;
+            }
+
+            // Assert 1
+            Assert.BuildPassed(result);
+
+            // Act & Assert 2
+            for (var i = 0; i < 2; i++)
+            {
+                // We want to make sure nothing changed between multiple incremental builds.
+                using (var razorGenDirectoryLock = LockDirectory(RazorIntermediateOutputPath))
+                {
+                    result = await DotnetMSBuild("Build", $"/p:RazorCompileOnBuild=true");
+                }
+
+                Assert.BuildPassed(result);
+                foreach (var file in files)
+                {
+                    var thumbprint = GetThumbPrint(file);
+                    Assert.Equal(thumbprintLookup[file], thumbprint);
+                }
+            }
+        }
+    }
+}


test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task BuildIncremental_SimpleMvc_PersistsTagHelperTargetInputFile()
+        {
+            // Build 1
+            var result = await DotnetMSBuild(
+                "Build",
+                $"/p:RazorCompileOnBuild=true /p:UseRazorBuildServer=true /p:_RazorBuildServerPipeName={_pipeName} /p:_RazorForceBuildServer=true",
```
ajaybhargavb: Using the server here because we eventually want all these tests to run on the server.

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
```diff
@@ -60,11 +60,11 @@
...
     <Touch
       Files="$(_RazorTagHelperInputCache)"
-      AlwaysCreate="true">
-      <Output
-        TaskParameter="TouchedFiles"
-        ItemName="FileWrites" />
-    </Touch>
+      AlwaysCreate="true" />
+
+    <ItemGroup>
```
pranavkm: This is a pretty funky MSBuild gotcha. When running an incremental build on an MSBuild target, it skips individual tasks inside the target, but continues to evaluate `PropertyGroup` \ `ItemGroup` inside the target. Here's a simple example:

```xml
<Target Name="TestDep"
    Inputs="@(Compile)"
    Outputs="$(MSBuildProjectFullPath)">
    
    <Message Text="I'm not incremental" />
    
    <ItemGroup>
        <Stuff Include="Hello world" />
    </ItemGroup>
    
  </Target>
  
  <Target Name="Test" DependsOnTargets="TestDep">
    <Message Text="@(Stuff)" Importance="High" />
  </Target>
```

Here's the output from the 2nd (incremental) run
image

MSBuild skips the `Message` task, but evaluates the itemgroup. 

In conjunction with this behavior, Microsoft.Common.targets has a clean up target where it deletes any files that are present in the output directories but are not specified in `FileWrites`. Consequently every incremental build resulted in the cache file from being deleted without us recording a write. This fixes the behavior.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
pranavkm: Could we instead test the expectation ie. rebuilding twice results in incremental twice? It would cover any other files that affect incrementalism rather than this specific case

src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets
rynowak: Eww that's wierd. TIL I guess

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
rynowak: Yeah I agree with @pranavkm - I don't fundamentally accept that it's a goal that all tests use the server

ajaybhargavb: > I don't fundamentally accept that it's a goal that all tests use the server

I misspoke. I didn't mean all the tests. I meant most tests because we'll have build server turned on by default.

rynowak: But this still doesn't need to be hardcoded here.

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
```diff
@@ -0,0 +1,84 @@
...
+                $"/p:RazorCompileOnBuild=true /p:UseRazorBuildServer=true /p:_RazorBuildServerPipeName={_pipeName} /p:_RazorForceBuildServer=true");
+
+            var directoryPath = Path.Combine(result.Project.DirectoryPath, IntermediateOutputPath);
+            var filesToIgnore = new[]
+            {
+                // These files are generated on every build.
+                Path.Combine(directoryPath, "SimpleMvc.csproj.CopyComplete"),
+                Path.Combine(directoryPath, "SimpleMvc.csproj.FileListAbsolute.txt"),
+            };
+            var files = Directory.EnumerateFiles(directoryPath).Where(p => !filesToIgnore.Contains(p));
```
pranavkm: Use `GetFiles` or do a `ToArray`. This should tell if a file went missing between builds.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
```diff
@@ -0,0 +1,84 @@
...
+            foreach (var file in files)
+            {
+                var thumbprint = GetThumbPrint(file);
+                thumbprintLookup[file] = thumbprint;
+            }
+
+            // Assert 1
+            Assert.BuildPassed(result);
+
+            // Act 2
```
pranavkm: Consider putting this in a loop. Avoids the trouble of changing things if we fix up stuff in the future.  Also would be useful to add a comment that says why we're testing this i.e. we want to make sure all files on disk remain identical across multiple incremental builds.

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
ajaybhargavb: @rynowak I agree this is ugly. I plan to clean this up as part of a separate cleanup. We have other tests that have the hardcoding.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
```diff
@@ -0,0 +1,71 @@
...
+        [InitializeTestProject("SimpleMvc")]
+        public async Task BuildIncremental_SimpleMvc_PersistsTargetInputFile()
+        {
+            // Arrange
+            var thumbprintLookup = new Dictionary<string, FileThumbPrint>();
+
+            // Act 1
+            var result = await DotnetMSBuild(
+                "Build",
+                $"/p:RazorCompileOnBuild=true /p:UseRazorBuildServer=true /p:_RazorBuildServerPipeName={_pipeName} /p:_RazorForceBuildServer=true");
```
rynowak: The tests that **specifically target the server** should use look like this. The tests that don't require the server shouldn't hardcode it.

ajaybhargavb: Got it. Will remove.

PR #2091, opened 2018-02-20
Make RazorCompileOnBuild=true

### LINKED ISSUE #1986: Enable RazorCompileOnBuild=true by default
Due to how our factoring is being done in 2.1-preview1 - the necessary properties to activate Razor  compilation (using RazorSDK) during publish are part of the Web SDK. This means that a class library project won't do precompilation during publishing by default.

The workaround is to define a property that the WebSDK sets.
```
  <PropertyGroup>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <ResolvedRazorCompileToolset>RazorSDK</ResolvedRazorCompileToolset>
  </PropertyGroup>
```

We plan to take ownership of these definitions from the WebSDK in 2.1-preview2 so this should just get solved at that point.

pranavkm:
Fixes #1986


rynowak:
This looks like the right change. **But** we need to wait on this until we have buy-in to make the change since this affects inner loop perf.

pranavkm:
🆙 📅

Issue #2080, opened 2018-02-16
RazorProjectEngineBuilder.AddDefaultImports should take a source document or a string


rynowak:
We merged this from our tests into the product, but it accepts params of `RazorProjectItem` which isn't a very useful type for extenders. This should be cleaned up to take strings or source documents.

NTaylorMullen:
How would this work with our `IImportProjectFeature` (returns project items) and the various APIs that consume those imports?

If we were to go this route I would suggest enforcing a null file path on the default import source documents.

rynowak:
The infrastructure would create the project item for you. It's a pain to have to stub out this class just so you can add some usings 😆

NTaylorMullen:
c5e83c61f89037647c13169983a42d9d342ea850

PR #2092, opened 2018-02-21
Changed AddDefaultImports to take strings instead of project items.

NTaylorMullen:
- Strings here was important because any import added to the system dynamically needs to eventually make its way back to being a project item. With strings we can state that they do exist (have content) but do not have any file paths associated.
- Updated all call sites to use the new AddDefaultImports string based api.

#2080

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngineBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngineBuilderExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngineBuilderExtensions.cs
```diff
@@ -229,6 +230,36 @@ public IReadOnlyList<RazorProjectItem> GetImports(RazorProjectItem projectItem)
 
                 return imports;
             }
+
+            private class DynamicProjectItem : RazorProjectItem
```
ajaybhargavb: Suggest calling this `InMemoryProjectItem` to make it clear

PR #2102, opened 2018-02-22
Change final code documents to not contain non-existent imports.

NTaylorMullen:
- Existent imports are imports that have content that contribute to the processing of a Razor document. Prior to this we had a legacy expectation that code documents had empty markers in them for all of their import locations. This proved troublesome when cross-referencing files that had file paths and were supposed to be existent but weren't in metadata. Now that we have a project engine with a de-coupled import feature we can rely on the import feature for finding all locations of important files and then strip out any non-existent items.

This solves an issue where when we're trying to determine if a precompiled view is dirty we know if a `_ViewImports.cshtml` was deleted (and not just a marker to begin with).

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
@@ -65,7 +65,7 @@ namespace Microsoft.AspNetCore.Razor.Language
 
             var importFeature = GetRequiredFeature<IImportProjectFeature>();
             var importItems = importFeature.GetImports(projectItem);
-            var importSourceDocuments = importItems.Select(ConvertToSourceDocument);
+            var importSourceDocuments = GetImportSourceDocuments(importItems);
 
             var parserOptions = GetRequiredFeature<IRazorParserOptionsFactoryProjectFeature>().Create(ConfigureParserOptions);
             var codeGenerationOptions = GetRequiredFeature<IRazorCodeGenerationOptionsFactoryProjectFeature>().Create(ConfigureCodeGenerationOptions);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -138,17 +137,21 @@ private void ConfigureDesignTimeCodeGenerationOptions(RazorCodeGenerationOptions
         }
 
         // Internal for testing
-        internal static RazorSourceDocument ConvertToSourceDocument(RazorProjectItem importItem)
+        internal static IEnumerable<RazorSourceDocument> GetSourceDocumentImports(IReadOnlyList<RazorProjectItem> importItems)
```
pranavkm: `IReadOnlyList<RazorSourceDocument> GetImportSourceDocuments`

NTaylorMullen:
> My memory says that the list of contributing sources we put on the compiled item is intended to include non-existent files and that the algorithm in MVC is supposed to do the right thing.

So that goal came from our old system only having one mechanism for determining which imports applied to which document. Basically, it needed to have enough data there to be used for two purposes:

1. What potential files impact the current Razor document so we can reparse if any of those files change?
2. What document content should be included when parsing the primary Razor document?

Now that we satisfy both of these requirements at different stages of processing a Razor document,
 1 with our import feature and 2 with our codeDocument.Imports we don't need to overload the code documents imports representation with ambiguous information.

tl;dr; I'm trying to solve having [RazorSourceChecksumAttribute](https://github.com/aspnet/Razor/blob/60e5818884470c923bcac4271dffb369c50aa89d/src/Microsoft.AspNetCore.Razor.Runtime/Hosting/RazorSourceChecksumAttribute.cs)'s decorating pre-compiled views always represent actual _ViewImports content. Otherwise that attribute will need to have a notion of an import being a marker vs not being a marker import. This is all to determine if a precompiled view is dirty but is a general fix for problems like this.

rynowak:
Ok. discussed this more. Get a signoff from @pranavkm

NTaylorMullen:
@pranavkm just confirming this fixed your MVC issue yes?

pranavkm:
Verifying it now.

pranavkm:
Done. Fixes the issue :shipit:

PR #2103, opened 2018-02-22
Perf: Cache assembly references in build server DiscoverCommand

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2111
This is an effort to reduce the perf impact of `MetadataReference.CreateFromFile()`. This is mostly inspired from Roslyn.

Before:
image

After:
image

Update: Latest perf data https://github.com/aspnet/Razor/pull/2103#issuecomment-368134121

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
@@ -6,17 +6,19 @@ using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
 using System.Threading;
+using Microsoft.CodeAnalysis;
 using Microsoft.Extensions.CommandLineUtils;
 
 namespace Microsoft.AspNetCore.Razor.Tools
 {
     internal class Application : CommandLineApplication
     {
-        public Application(CancellationToken cancellationToken, ExtensionAssemblyLoader loader, ExtensionDependencyChecker checker)
+        public Application(CancellationToken cancellationToken, ExtensionAssemblyLoader loader, ExtensionDependencyChecker checker, Func<string, MetadataReferenceProperties, PortableExecutableReference> assemblyReferenceProvider)
         {
             CancellationToken = cancellationToken;
             Checker = checker;
             Loader = loader;
+            AssemblyReferenceProvider = assemblyReferenceProvider;
 
             Name = "rzc";
             FullName = "Microsoft ASP.NET Core Razor CLI tool";
@@ -36,6 +38,8 @@ namespace Microsoft.AspNetCore.Razor.Tools
         public ExtensionAssemblyLoader Loader { get; }
 
         public ExtensionDependencyChecker Checker { get; }
+
+        public Func<string, MetadataReferenceProperties, PortableExecutableReference> AssemblyReferenceProvider { get; }
 
         public new int Execute(params string[] args)
         {
--- a/src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using Microsoft.CodeAnalysis;
 
 namespace Microsoft.AspNetCore.Razor.Tools
 {
@@ -30,7 +32,11 @@ namespace Microsoft.AspNetCore.Razor.Tools
                 // consistently reject a request that doesn't specify everything it needs. Otherwise the request
                 // could succeed sometimes if it relies on transient state.
                 Loader = new DefaultExtensionAssemblyLoader(Path.Combine(Path.GetTempPath(), "Razor-Server"));
+
+                AssemblyReferenceProvider = (path, properties) => new CachingMetadataReference(path, properties);
             }
+
+            public Func<string, MetadataReferenceProperties, PortableExecutableReference> AssemblyReferenceProvider { get; }
 
             public ExtensionAssemblyLoader Loader { get; }
 
@@ -48,7 +54,7 @@ namespace Microsoft.AspNetCore.Razor.Tools
                 var writer = ServerLogger.IsLoggingEnabled ? new StringWriter() : TextWriter.Null;
 
                 var checker = new DefaultExtensionDependencyChecker(Loader, writer);
-                var app = new Application(cancellationToken, Loader, checker)
+                var app = new Application(cancellationToken, Loader, checker, AssemblyReferenceProvider)
                 {
                     Out = writer,
                     Error = writer,
--- a/src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
@@ -1,0 +1,207 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    /// <summary>
+    /// Cache with a fixed size that evicts the least recently used members.
+    /// Thread-safe.
+    /// This was taken from https://github.com/dotnet/roslyn/blob/749c0ec135d7d080658dc1aa794d15229c3d10d2/src/Compilers/Core/Portable/InternalUtilities/ConcurrentLruCache.cs.
+    /// </summary>
+    internal class ConcurrentLruCache<TKey, TValue>
+    {
+        private readonly int _capacity;
+
+        private readonly Dictionary<TKey, CacheValue> _cache;
+        private readonly LinkedList<TKey> _nodeList;
+        // This is a naive course-grained lock, it can probably be optimized
+        private readonly object _lockObject = new object();
+
+        public ConcurrentLruCache(int capacity)
+            : this (capacity, EqualityComparer<TKey>.Default)
+        {
+        }
+
+        public ConcurrentLruCache(int capacity, IEqualityComparer<TKey> comparer)
+        {
+            if (capacity <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(capacity));
+            }
+            _capacity = capacity;
+            _cache = new Dictionary<TKey, CacheValue>(capacity, comparer);
+            _nodeList = new LinkedList<TKey>();
+        }
+
+        /// <summary>
+        /// Create cache from an array. The cache capacity will be the size
+        /// of the array. All elements of the array will be added to the 
+        /// cache. If any duplicate keys are found in the array a
+        /// <see cref="ArgumentException"/> will be thrown.
+        /// </summary>
+        public ConcurrentLruCache(KeyValuePair<TKey, TValue>[] array)
+            : this(array.Length)
+        {
+            foreach (var kvp in array)
+            {
+                UnsafeAdd(kvp.Key, kvp.Value);
+            }
+        }
+
+        public int Count
+        {
+            get
+            {
+                lock (_lockObject)
+                {
+                    return _cache.Count;
+                }
+            }
+        }
+
+        public void Add(TKey key, TValue value)
+        {
+            lock (_lockObject)
+            {
+                UnsafeAdd(key, value);
+            }
+        }
+
+        public TValue GetOrAdd(TKey key, TValue value)
+        {
+            lock (_lockObject)
+            {
+                if (UnsafeTryGetValue(key, out var result))
+                {
+                    return result;
+                }
+                else
+                {
+                    UnsafeAdd(key, value);
+                    return value;
+                }
+            }
+        }
+
+        public bool TryGetValue(TKey key, out TValue value)
+        {
+            lock (_lockObject)
+            {
+                return UnsafeTryGetValue(key, out value);
+            }
+        }
+
+        public bool Remove(TKey key)
+        {
+            lock (_lockObject)
+            {
+                return UnsafeRemove(key);
+            }
+        }
+
+        /// <summary>
+        /// For testing. Very expensive.
+        /// </summary>
+        internal IEnumerable<KeyValuePair<TKey, TValue>> TestingEnumerable
+        {
+            get
+            {
+                lock (_lockObject)
+                {
+                    foreach (var key in _nodeList)
+                    {
+                        var kvp = new KeyValuePair<TKey, TValue>(key, _cache[key].Value);
+                        yield return kvp;
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Doesn't lock.
+        /// </summary>
+        private bool UnsafeTryGetValue(TKey key, out TValue value)
+        {
+            if (_cache.TryGetValue(key, out var result))
+            {
+                MoveNodeToTop(result.Node);
+                value = result.Value;
+                return true;
+            }
+            else
+            {
+                value = default(TValue);
+                return false;
+            }
+        }
+
+        private void MoveNodeToTop(LinkedListNode<TKey> node)
+        {
+            if (!object.ReferenceEquals(_nodeList.First, node))
+            {
+                _nodeList.Remove(node);
+                _nodeList.AddFirst(node);
+            }
+        }
+
+        /// <summary>
+        /// Expects non-empty cache. Does not lock.
+        /// </summary>
+        private void UnsafeEvictLastNode()
+        {
+            Debug.Assert(_capacity > 0);
+            var lastNode = _nodeList.Last;
+            _nodeList.Remove(lastNode);
+            _cache.Remove(lastNode.Value);
+        }
+
+        private void UnsafeAddNodeToTop(TKey key, TValue value)
+        {
+            var node = new LinkedListNode<TKey>(key);
+            _cache.Add(key, new CacheValue(value, node));
+            _nodeList.AddFirst(node);
+        }
+
+        /// <summary>
+        /// Doesn't lock.
+        /// </summary>
+        private void UnsafeAdd(TKey key, TValue value)
+        {
+            if (_cache.TryGetValue(key, out var result))
+            {
+                throw new ArgumentException("Key already exists", nameof(key));
+            }
+            else
+            {
+                if (_cache.Count == _capacity)
+                {
+                    UnsafeEvictLastNode();
+                }
+                UnsafeAddNodeToTop(key, value);
+            }
+        }
+
+        private bool UnsafeRemove(TKey key)
+        {
+            _nodeList.Remove(key);
+            return _cache.Remove(key);
+        }
+
+        private struct CacheValue
+        {
+            public CacheValue(TValue value, LinkedListNode<TKey> node)
+            {
+                Value = value;
+                Node = node;
+            }
+
+            public TValue Value { get; }
+
+            public LinkedListNode<TKey> Node { get; }
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
@@ -1,0 +1,86 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection.PortableExecutable;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    internal class MetadataCache
+    {
+        // Store 1000 entries -- arbitrary number
+        private const int CacheSize = 1000;
+        private readonly ConcurrentLruCache<string, MetadataCacheEntry> _metadataCache =
+            new ConcurrentLruCache<string, MetadataCacheEntry>(CacheSize, StringComparer.OrdinalIgnoreCase);
+
+        // For testing purposes only.
+        internal ConcurrentLruCache<string, MetadataCacheEntry> Cache => _metadataCache;
+
+        internal Metadata GetMetadata(string fullPath)
+        {
+            var timestamp = GetFileTimeStamp(fullPath);
+
+            // Check if we have an entry in the dictionary.
+            if (_metadataCache.TryGetValue(fullPath, out var entry))
+            {
+                if (timestamp.HasValue && timestamp.Value == entry.Timestamp)
+                {
+                    // The file has not changed since we cached it. Return the cached entry.
+                    return entry.Metadata;
+                }
+                else
+                {
+                    // The file has changed recently. Remove the cache entry.
+                    _metadataCache.Remove(fullPath);
+                }
+            }
+
+            Metadata metadata;
+            using (var fileStream = File.OpenRead(fullPath))
+            {
+                metadata = AssemblyMetadata.CreateFromStream(fileStream, PEStreamOptions.PrefetchMetadata);
+            }
+
+            _metadataCache.GetOrAdd(fullPath, new MetadataCacheEntry(timestamp.Value, metadata));
+
+            return metadata;
+        }
+
+        private static DateTime? GetFileTimeStamp(string fullPath)
+        {
+            try
+            {
+                Debug.Assert(Path.IsPathRooted(fullPath));
+
+                return File.GetLastWriteTimeUtc(fullPath);
+            }
+            catch (Exception e)
+            {
+                // There are several exceptions that can occur here: NotSupportedException or PathTooLongException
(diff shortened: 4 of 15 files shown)


rynowak:
Are the results here based on wall-clock time or CPU time? 

Also what about https://github.com/dotnet/roslyn/blob/master/src/Compilers/Server/ServerShared/MetadataCache.cs#L109?

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
```diff
@@ -19,6 +19,9 @@ public static CompilerHost Create()
 
         private class DefaultCompilerHost : CompilerHost
         {
+            // Store 100 entries -- arbitrary number
+            private const int CacheSize = 100;
```
rynowak: unused?

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    internal class AssemblyReferenceCache
+    {
+        // Store 1000 entries -- arbitrary number
+        private const int CacheSize = 1000;
```
rynowak: In our primary scenarios, is 1000 big enough?

ajaybhargavb:
> Are the results here based on wall-clock time or CPU time?

CPU time. From what I can see if I drill into DiscoverCommand, both Wall clock and CPU time show pretty much the same results.

> Also what about https://github.com/dotnet/roslyn/blob/master/src/Compilers/Server/ServerShared/MetadataCache.cs#L109?

The main difference between Roslyn's code and what I have here is this. They cache the result of `GetMetadata` which is called from `CSharpCompiler`. I am caching the result of `MetadataReference.CreateFromFile` which is used by our TagHelperDescriptorFactories.

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
ajaybhargavb: I only tested this with `SimpleMvc` in our testapps. And the count for that is 147.

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
ajaybhargavb: Will remove.

rynowak:
so if 64 ms is the wall clock time of the discover command - what's the time spend in the Task that calls it? I thought the discovery task took about 300ms total (wall clock)

ajaybhargavb:
So the 300ms was from the VM that I was running `WebSmallCore` on. The wall clock time for the TagHelperTask when I profiled on my dev machine is  84 ms. As the data shows DiscoveryCommand 60% of this time. So I expect the time to proportionately go down when we run on different VMs.

ajaybhargavb:
Also note this data is for the Razor testapp `SimpleMvc` not `WebSmallCore`.

rynowak:
> Also note this data is for the Razor testapp SimpleMvc not WebSmallCore.

AH

Yeah the data is going to be really different on an actual app. You're going to have significantly more assemblies.

ajaybhargavb:
Results from WebSmallCore:

No cache:
image

Cache MetadataReference.CreateFromFile:
image

Cache result of GetMetadata (First commit of this PR):
image

From what I can see from this, 
"Caching results of `GetMetadata`" > "Caching results of `MetadataReference.CreateFromFile`" > "No cache"

I'll update the PR

rynowak:
Great! That's more like what I expected. So it looks like we could take about 100ms off the incremental build cost. 

What is the impact on the memory usage of the server process? IE what does task manager report for private working set? If it's less than 20mb I don't think I'm concerned. If it's hundreds then something is wrong 😆 


/cc @mikeharder for FYI

ajaybhargavb:
> What is the impact on the memory usage of the server process?

It's around 45mb

rynowak:
> It's around 45mb

Hmm. Keep working on this for now. 

We should get more buy in before really turning this on by default.

ajaybhargavb:
> That's higher than I'd like. We should get more buy in before really turning this on by default.

Do you mean the caching? Or the entire build server?

rynowak:
caching

src/Microsoft.AspNetCore.Razor.Tools/FileKey.cs
```diff
@@ -0,0 +1,88 @@
...
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    // From:
+    // https://github.com/dotnet/roslyn/blob/749c0ec135d7d080658dc1aa794d15229c3d10d2/src/Compilers/Core/Portable/FileKey.cs
+    internal struct FileKey : IEquatable<FileKey>
+    {
+        /// <summary>
+        /// Full case-insensitive path.
+        /// </summary>
+        public readonly string FullPath;
```
pranavkm: Properties instead of fields?

src/Microsoft.AspNetCore.Razor.Tools/FileKey.cs
```diff
@@ -0,0 +1,88 @@
...
+            var hashCodeCombiner = HashCodeCombiner.Start();
+            hashCodeCombiner.Add(FullPath, StringComparer.OrdinalIgnoreCase);
+            hashCodeCombiner.Add(Timestamp);
+
+            return hashCodeCombiner.CombinedHash;
+        }
+
+        public override bool Equals(object obj)
+        {
+            return obj is FileKey && Equals((FileKey)obj);
```
pranavkm: `return obj is FileKey fileKey && Equals(fileKey)`

src/Microsoft.AspNetCore.Razor.Tools/FileKey.cs
```diff
@@ -0,0 +1,88 @@
...
+        public FileKey(string fullPath, DateTime timestamp)
+        {
+            Debug.Assert(Path.IsPathRooted(fullPath));
+            Debug.Assert(timestamp.Kind == DateTimeKind.Utc);
+
+            FullPath = fullPath;
+            Timestamp = timestamp;
+        }
+
+        /// <exception cref="IOException"/>
```
pranavkm: ?

src/Microsoft.AspNetCore.Razor.Tools/FileKey.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                return File.GetLastWriteTimeUtc(fullPath);
+            }
+            catch (IOException)
+            {
+                throw;
+            }
+            catch (Exception e)
+            {
+                throw new IOException(e.Message, e);
```
pranavkm: This kinda weird. What's the reason to do this?

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    /// <summary>
+    /// Cache with a fixed size that evicts the least recently used members.
+    /// Thread-safe.
+    /// This was taken from https://github.com/dotnet/roslyn/blob/749c0ec135d7d080658dc1aa794d15229c3d10d2/src/Compilers/Core/Portable/InternalUtilities/ConcurrentLruCache.cs.
+    /// </summary>
+    internal class ConcurrentLruCache<K, V>
```
pranavkm: `ConcurrentLruCache<TKey, TValue>`

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        private readonly int _capacity;
+
+        private struct CacheValue
```
pranavkm: This can go at the end of the file

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+        {
+            public V Value;
```
pranavkm: readonly properties

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// of the array. All elements of the array will be added to the 
+        /// cache. If any duplicate keys are found in the array a
+        /// <see cref="ArgumentException"/> will be thrown.
+        /// </summary>
+        public ConcurrentLruCache(KeyValuePair<K, V>[] array)
+            : this(array.Length)
+        {
+            foreach (var kvp in array)
+            {
+                this.UnsafeAdd(kvp.Key, kvp.Value, true);
```
pranavkm: Remove `this.`

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                _nodeList.Remove(node);
+                _nodeList.AddFirst(node);
+            }
+        }
+
+        /// <summary>
+        /// Expects non-empty cache. Does not lock.
+        /// </summary>
+        private void UnsafeEvictLastNode()
```
pranavkm: `Unsafe` seems like a poor choice to indicate not-locked.

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// For testing. Very expensive.
+        /// </summary>
+        internal IEnumerable<KeyValuePair<K, V>> TestingEnumerable
+        {
+            get
+            {
+                lock (_lockObject)
+                {
+                    var copy = new KeyValuePair<K, V>[_cache.Count];
+                    int index = 0;
```
pranavkm: `var`

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+                    foreach (K key in _nodeList)
+                    {
+                        copy[index++] = new KeyValuePair<K, V>(key,
```
pranavkm: Can you just `yield return` this?

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// </summary>
+        internal IEnumerable<KeyValuePair<K, V>> TestingEnumerable
+        {
+            get
+            {
+                lock (_lockObject)
+                {
+                    var copy = new KeyValuePair<K, V>[_cache.Count];
+                    int index = 0;
+                    foreach (K key in _nodeList)
```
pranavkm: `var`

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                if (_cache.Count == _capacity)
+                {
+                    UnsafeEvictLastNode();
+                }
+                UnsafeAddNodeToTop(key, value);
+            }
+        }
+
+        public V this[K key]
```
pranavkm: It doesn't look like we use this method from the caller - all I saw was a `GetOrAdd`. Could we remove unused code - reduces the burden of testing \ maintaining the type?

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
ajaybhargavb: This is what we do in other places in our code as well. Eg. https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/DefaultExtensionAssemblyLoader.cs#L125

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
ajaybhargavb: Using this in tests now. Removed other unused methods though.

ajaybhargavb:
🆙 📅  No longer in design

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -116,7 +116,9 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
         /// A shutdown request should not abort an existing compilation.  It should be allowed to run to 
         /// completion.
         /// </summary>
-        [Fact(Skip = "Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991")]
+        // Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991
+        [ConditionalFact]
```
ajaybhargavb: I disabled these to debug a CI flakiness issue a while ago but it turned out these were not the cause. I forgot to re-enable these.

src/Microsoft.AspNetCore.Razor.Tools/Application.cs
```diff
@@ -37,6 +39,8 @@ public Application(CancellationToken cancellationToken, ExtensionAssemblyLoader
 
         public ExtensionDependencyChecker Checker { get; }
 
+        public Func<string, MetadataReferenceProperties, PortableExecutableReference> AssemblyReferenceProvider { get; set; }
```
rynowak: Does this need to be get/set and passed in the constructor?

ajaybhargavb: Will remove the set.

test/Microsoft.AspNetCore.Razor.Tools.Test/MetadataCacheTest.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class MetadataCacheTest
+    {
+        [Fact]
+        public void GetMetadata_AddsToCache()
+        {
+            using (var directory = TempDirectory.Create())
+            {
+                // Arrange
+                var metadataCache = new MetadataCache();
+                var assemblyFilePath = LoaderTestResources.Delta.WriteToFile(directory.DirectoryPath, "Delta.dll");
```
pranavkm: What's `LoaderTestResources`?

test/Microsoft.AspNetCore.Razor.Tools.Test/MetadataCacheTest.cs
```diff
@@ -0,0 +1,79 @@
...
+                var result = metadataCache.GetMetadata(assemblyFilePath);
+
+                // Assert
+                Assert.NotNull(result);
+                Assert.Equal(1, metadataCache.Cache.Count);
+            }
+        }
+
+        [Fact]
+        public void GetMetadata_UsesCache()
```
pranavkm: Missing a test - `GetMetadata_ReturnsDifferentResultsForDifferentFiles`

test/Microsoft.AspNetCore.Razor.Tools.Test/MetadataCacheTest.cs
ajaybhargavb: It's something @rynowak added to test the AssemblyLoader and ExtensionDependencyChecker.

test/Microsoft.AspNetCore.Razor.Tools.Test/ConcurrentLruCacheTest.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            Assert.Equal(expected, cache.TestingEnumerable);
+        }
+
+        [Fact]
+        public void ConcurrentLruCache_ReadLastNode_EvictsSecondOldestNode()
+        {
+            var cache = new OrderedTestDictionary<int, int>(3)
+                { {1, 1}, {2, 2}, {3, 3} }.MakeCache();
+            var oneVal = cache[1];
```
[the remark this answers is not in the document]
ajaybhargavb: Yes. But this is needed so it is accessed the latest. To make the LRU cache retain it.

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            get
+            {
+                lock (_lockObject)
+                {
+                    return _cache.Count;
+                }
+            }
+        }
+
+        public void Add(TKey key, TValue value)
```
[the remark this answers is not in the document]
ajaybhargavb: It is still used in tests.

test/Microsoft.AspNetCore.Razor.Tools.Test/ConcurrentLruCacheTest.cs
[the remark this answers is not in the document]
pranavkm: Ahh ok. Could you also move this below test arrangement? And maybe make a comment to say, we're making `{2, 2}`, the MRU item

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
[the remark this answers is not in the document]
pranavkm: Can we get tests to use `GetOrAdd` instead? Just trying to reduce the amount of code in here we don't need to maintain

test/Microsoft.AspNetCore.Razor.Tools.Test/ConcurrentLruCacheTest.cs
```diff
@@ -0,0 +1,101 @@
...
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    public class ConcurrentLruCacheTest
+    {
+        /// <summary>
+        /// Dictionary for testing ConcurrentLruCache.
+        /// Like OrderedDictionary in the BCL, doesn't sort elements,
+        /// but rather stores them in the order added.
+        /// </summary>
+        private class OrderedTestDictionary<TKey, TValue> : IEnumerable<KeyValuePair<TKey, TValue>>
```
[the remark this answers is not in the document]
ajaybhargavb: I agree but this does make the tests look a lot simpler and it was basically a copy paste from Roslyn. I can not use this if you think it'll lead to complications later.

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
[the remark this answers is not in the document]
ajaybhargavb: Sure!

test/Microsoft.AspNetCore.Razor.Tools.Test/ConcurrentLruCacheTest.cs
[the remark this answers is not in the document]
pranavkm: private KeyValuePair<int, int>[] GetInputs() => Enumerable.Range(1, 4).Select(v => new KeyValuePair(v, v)).ToArray()

There FTFY

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,219 @@
...
+            _nodeList = new LinkedList<TKey>();
+        }
+
+        /// <summary>
+        /// Create cache from an array. The cache capacity will be the size
+        /// of the array. All elements of the array will be added to the 
+        /// cache. If any duplicate keys are found in the array a
+        /// <see cref="ArgumentException"/> will be thrown.
+        /// </summary>
+        public ConcurrentLruCache(KeyValuePair<TKey, TValue>[] array)
```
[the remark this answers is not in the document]
ajaybhargavb: Needed in tests.

test/Microsoft.AspNetCore.Razor.Tools.Test/MetadataCacheTest.cs
ajaybhargavb: Added

ajaybhargavb:
@pranavkm Updated.

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
```diff
@@ -0,0 +1,83 @@
...
+        internal ConcurrentLruCache<string, MetadataCacheEntry> Cache => _metadataCache;
+
+        internal Metadata GetMetadata(string fullPath)
+        {
+            var timestamp = GetFileTimeStamp(fullPath);
+
+            // Check if we have an entry in the dictionary.
+            if (_metadataCache.TryGetValue(fullPath, out var entry))
+            {
+                if (timestamp.HasValue && timestamp.Value == entry.Timestamp)
```
pranavkm: Should this allow the cache entry to have a newer timestamp?

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    _metadataCache.Remove(fullPath);
+                }
+            }
+
+            using (var fileStream = File.OpenRead(fullPath))
+            {
+                var metadata = AssemblyMetadata.CreateFromStream(fileStream, PEStreamOptions.PrefetchMetadata);
+
+                var result = _metadataCache.GetOrAdd(fullPath, new MetadataCacheEntry(timestamp.Value, metadata));
```
pranavkm: Let's move this outside the `using` block. `GetOrAdd` has a very coarse lock so you might have end up keeping the open for long periods

test/Microsoft.AspNetCore.Razor.Tools.Test/ConcurrentLruCacheTest.cs
```diff
@@ -0,0 +1,120 @@
...
+            cache.GetOrAdd(1, 1); // Read to make this MRU
+            cache.Add(4, 4); // Add a new node
+
+            // Assert
+            Assert.Equal(expected, cache.TestingEnumerable);
+        }
+
+        private KeyValuePair<int, int>[] GetKeyValueArray(IEnumerable<int> inputArray)
+        {
+            return inputArray.Select(v => new KeyValuePair<int, int>(v, v)).ToArray();
```
pranavkm: Glorious!

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -152,7 +154,9 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
         /// <summary>
         /// Multiple clients should be able to send shutdown requests to the server.
         /// </summary>
-        [Fact(Skip = "Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991")]
+        // Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991
+        [ConditionalFact]
```
pranavkm: Can you leave the Skip message on `ConditionalFact`?

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
pranavkm: This block of code would benefit from a code comment.

src/Microsoft.AspNetCore.Razor.Tools/ConcurrentLruCache.cs
```diff
@@ -0,0 +1,208 @@
...
+                {
+                    UnsafeEvictLastNode();
+                }
+                UnsafeAddNodeToTop(key, value);
+            }
+        }
+
+        private bool UnsafeRemove(TKey key)
+        {
+            Debug.Assert(_capacity > 0);
```
pranavkm: Odd thing to check

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
pranavkm: I think we should just the created `metadata` here and ignore `result`. There's a possible race here where `result` could be older than `metadata`(T1 -> Calls remove -> Makes metadata -> Does GetOAdd. Meanwhile T2 -> Fails TryGetValue and lands here. GetOrAddr returns results from T1 which might be older). The cache entry should get updated in the next call to this method, so we don't super need to worry about fixing that.

src/Microsoft.AspNetCore.Razor.Tools/MetadataCache.cs
ajaybhargavb: Thought about it. We don't super care if the file has a newer/older timestamp. We just want to get metadata for the file currently on that path. If someone replaced a file with another, there is no point in having the old one in cache regardless of the timestamp.
I'll add a comment.

pranavkm: Hmm, looking at this - the check looks kinda incorrect. It always the cache entry if it's up to date, doesn't it?

ajaybhargavb: I don't follow. Could you clarify?

pranavkm: Nevermind. I was looking at some weird code snippet.

Issue #2106, opened 2018-02-23
VS Crashes if it cannot read _ViewImports from disk.


NTaylorMullen:

rynowak:
Note that we made some changes to how we read files in general from Razor in 15.7+. That said, it seems fine for us to do some try/catching around viewimports specifically. 

We already ignore a viewimports that isn't present, so it seems fine for us to ignore one we can't read.

NTaylorMullen:
ca8077737b47fbc3114ddfaaf9ce1a8981b7449c

PR #2108, opened 2018-02-23
Protect against unreadable _ViewImports at design time.

NTaylorMullen:
- This is a safe fix for the VS15.6 release to protect against VS crashing when it cannot read a `_ViewImports.cshtml`  associated with a Razor document.
- In the case that we can't resolve the imports for a Razor document we fall back to using the default imports.

#2106

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/BackgroundParser.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/BackgroundParser.cs
@@ -345,10 +345,35 @@ namespace Microsoft.VisualStudio.Editor.Razor
 
             private RazorCodeDocument ParseChange(ITextSnapshot snapshot, CancellationToken token)
             {
+                const int MaxRetryCount = 5;
+
                 EnsureOnThread();
 
                 var sourceDocument = new TextSnapshotSourceDocument(snapshot, _filePath);
-                var imports = _templateEngine.GetImports(_filePath);
+
+                IEnumerable<RazorSourceDocument> imports = null;
+                var retries = 0;
+                while (retries++ < MaxRetryCount)
+                {
+                    try
+                    {
+                        imports = _templateEngine.GetImports(_filePath);
+                    }
+                    catch (IOException ex) when (ex is DirectoryNotFoundException || ex is FileNotFoundException || ex is PathTooLongException)
+                    {
+                        // No need to retry, we can't read the file.
+                        break;
+                    }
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editor.Razor/BackgroundParser.cs
```diff
@@ -345,10 +345,35 @@ private void WorkerLoop()
...
+                const int MaxRetryCount = 5;
+
                 EnsureOnThread();
 
                 var sourceDocument = new TextSnapshotSourceDocument(snapshot, _filePath);
-                var imports = _templateEngine.GetImports(_filePath);
+
+                IEnumerable<RazorSourceDocument> imports = null;
+                var retries = 0;
+                while (retries++ < MaxRetryCount)
```
NTaylorMullen: This is a tactical fix since it's for 15.6 RTM. In 15.7 + we'll do a more complete fix that logs errors if it can't access an import etc.

NTaylorMullen:
Installed these bits locally on 15.6 and verified that VS doesn't crash after trying to read inaccessible imports.

src/Microsoft.VisualStudio.Editor.Razor/BackgroundParser.cs
```diff
@@ -345,10 +345,35 @@ private void WorkerLoop()
...
+                    catch (IOException ex) when (ex is DirectoryNotFoundException || ex is FileNotFoundException || ex is PathTooLongException)
+                    {
+                        // No need to retry, we can't read the file.
+                        break;
+                    }
+                    catch (SystemException ex) when (ex is IOException || ex is UnauthorizedAccessException)
+                    {
+                        // File can't currently be read. Another process might be locking it or affecting its permissions.
+                        // We have less context of what's causing the failure so lets retry.
+                        Thread.Sleep(100);
```
NTaylorMullen: Note: This is isolated to the background thread that this class operates on (it `while(true)`s)

rynowak: Ok cool.

rynowak:
I think this is a good tactical fix. 

It would also be good to include the changes to `DefaultProjectItem.Read` that we made for 15.7 because that also helps. We had some problems with file-locking while working on the server and @ajaybhargavb did the fix for that, which was to use a different set of options to create the file stream. 

That's also a very safe and simple change.

NTaylorMullen:
> It would also be good to include the changes to DefaultProjectItem.Read that we made for 15.7

Sure I can do that.

PR #2118, opened 2018-02-27
Add RazorGenerate items to UpToDateCheckInput

### LINKED ISSUE #2048: Razor SDK project build-up-to-date check is wrong
VS has its own "up to date" logic that doesn't take Razor scenarios into account when using `RazorCompileOnBuild=true` which will become the default in 2.1.0-preview2.

The effect of this is that a change in a .cshtml file won't mark the project as 'out of date' as hoped. You have to trigger a rebuild of all dependent projects for things to function correctly.

We need to do some project system work to make this happen.

pranavkm:
Fixes #2048

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
+++ b/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
@@ -158,6 +158,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RazorIntermediateAssembly Condition="'$(RazorIntermediateAssembly)'==''" Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
     <!-- Used in Compilation.targets -->   
     <_RazorDebugSymbolsIntermediatePath Condition="'$(_RazorDebugSymbolsProduced)'=='true'" Include="$(IntermediateOutputPath)$(RazorTargetName).pdb" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -158,6 +158,8 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RazorIntermediateAssembly Condition="'$(RazorIntermediateAssembly)'==''" Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
     <!-- Used in Compilation.targets -->   
     <_RazorDebugSymbolsIntermediatePath Condition="'$(_RazorDebugSymbolsProduced)'=='true'" Include="$(IntermediateOutputPath)$(RazorTargetName).pdb" />
+    <!-- Add all cshtml files to UpToDateCheckInput - a collection used by the IDE's project system to determine if a project needs to be rebuilt -->
+    <UpToDateCheckInput Condition="'$(RazorCompileOnBuild)'=='true'" Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
```
pranavkm: This needs to be defined globally, not in a target, to work correctly. We can't use `RazorGenerate` for this since that's calculated inside a target. There might be some edge cases we might miss, like adding a RazorGenerate item that's not a Content, but users should be able to remedy it.

rynowak: Yeah I think this is fine.

PR #2124, opened 2018-02-28
Fix #2041 - Add static constructor to DocumentWriter

### LINKED ISSUE #2041: DocumentWriter.Create should be static
See: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DocumentWriter.cs#L11

This method was supposed to be static so that callers could create a writer and use it. Unfortunately we missed this 😢. The workaround is to do some [pretty awful stuff](https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor.Build/Core/RazorCompilation/Engine/BlazorLoweringPhase.cs#L39)

Suggestion:
- Make a new method to avoid confusion `CreateWriter(...)` and make it static
- `[Obsolete]` and `EditorBrowsableNever` the old one

Yeah, this is bad enough that **I** even think that it calls for EdtitorBrowsableNever

/cc @pranavkm for EditorBrowsableNever

rynowak:
The instance Create method was always supposed to be static.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DocumentWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DocumentWriter.cs
@@ -2,12 +2,30 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.ComponentModel;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
 
 namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 {
     public abstract class DocumentWriter
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DocumentWriter.cs
```diff
@@ -2,12 +2,30 @@
...
+
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            return new DefaultDocumentWriter(codeTarget, options);
+        }
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
pranavkm: ❤️

Issue #2018, opened 2018-02-02
Flaky test: Dispatcher_ProcessSimultaneousConnections_HitsKeepAliveTimeout


ryanbrandenburg:
I set `Dispatcher_ProcessSimultaneousConnections_HitsKeepAliveTimeout` to ignore for now because it [fails](http://aspnetci/viewLog.html?buildId=399383&tab=buildResultsDiv&buildTypeId=Lite_MvcTest#testNameId-5931557005942489049) sometimes. It should be investigated and re-enabled once fixed.

mkArtakMSFT:
@jbagga, can you please investigate this?

rynowak:
Just so there's no confusion, this is the Razor build server

mkArtakMSFT:
Thanks @rynowak. Reassigning to @ajaybhargavb.

ajaybhargavb:
0f1bc0e0bde2a16951ef1687a38e2379d823f385

PR #2126, opened 2018-03-01
Fix flaky dispatcher test

ajaybhargavb:
#2018 

The problem here was there was nothing ensuring Compilations are fully completed and written to the response before we send the client disconnect. Although I couldn't reproduce the exact failure from the CI, I am fairly certain this is the issue. I was able to make it fail by adding a delay in the compile task (which shouldn't happen).
The fix was to add a callback that would trigger when a compilation is fully completed and then wait on it before disconnecting.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Tools.Test/DefaultRequestDispatcherTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Tools.Test/DefaultRequestDispatcherTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Moq;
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Razor.Tools.Test/DefaultRequestDispatcherTest.cs
```diff
@@ -370,24 +372,44 @@ public void Dispatcher_ProcessMultipleConnections_HitsKeepAliveTimeout()
                 };
             });
 
-            var keepAlive = TimeSpan.FromSeconds(1);
             var eventBus = new TestableEventBus();
+            var completedCompilations = 0;
+            var allCompilationsComplete = new TaskCompletionSource<bool>();
+            eventBus.CompilationComplete += (obj, args) =>
+            {
+                if (++completedCompilations < totalCount)
```
NTaylorMullen:
```C#
if (++completedCompilations == totalCount)
{
    allCompilationsComplete.SetResult(true)
}
```

PR #2128, opened 2018-03-02
Fixed flaky shutdown tests

ajaybhargavb:
https://github.com/aspnet/Razor/issues/1991

The flakiness was caused by the race between the compilation and the shutdown requests. There was nothing guaranteeing that compilation was in progress before shutdown is seen.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
@@ -116,17 +116,19 @@ namespace Microsoft.AspNetCore.Razor.Tools
         /// A shutdown request should not abort an existing compilation.  It should be allowed to run to 
         /// completion.
         /// </summary>
-        // Skipping temporarily on non-windows. https://github.com/aspnet/Razor/issues/1991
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Razor.Tools.Test/ServerLifecycleTest.cs
```diff
@@ -215,16 +223,18 @@ public void ServerStartup_SuccessfullyAcquiredMutex()
...
             using (var serverData = ServerUtilities.CreateServer(compilerHost: host))
             {
-                var tasks = new List<Task<ServerResponse>>();
+                // Send all the requests.
+                var clients = new List<Client>();
                 for (var i = 0; i < requestCount; i++)
                 {
-                    var task = ServerUtilities.Send(serverData.PipeName, EmptyServerRequest);
-                    tasks.Add(task);
+                    var client = await Client.ConnectAsync(serverData.PipeName, timeout: null, cancellationToken: default);
```
ajaybhargavb: This is kinda similar to the previous two tests. We want to make sure we send all the requests before we send the cancellation and read the response. Before this, we might try to read the response before the cancellation goes through.

Issue #2081, opened 2018-02-16
Add VS4Mac implementation of our host based project system.


NTaylorMullen:
We have a windows project system which associated host projects with Roslyn projects; we also need this in VS4Mac.

NTaylorMullen:
1d602d12057296b1cd265e836c1eb3ea6531520f

PR #2137, opened 2018-03-06
Add host project system for VS4Mac.

NTaylorMullen:
- Tied into VS4Macs ProjectExtensions in order to bootstrap our Razor world.
- We currently watch all DotNet projects with the expectation that they're the only ones that can potentially turn into Razor compatible projects.
- Added a fallback Razor project host which is used for pre-Razor SDK Razor versions (< 2.1).
- Added a default Razor project host which consumes all MSBuild data from the users packages and sets up the Razor world accordingly.
- Had to modify some existing contracts to work better with new expectations. one of these was the VS4Mac specific Workspace accessor; essentially we needed to be able to lookup a workspace from a solution.
- Some of our previous expectations about addins were wrong (not being able to directly reference your libraries). To avoid using reflection to bootstrap our types I tried out directly referencing our libraries and all worked fine.

#2081


src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/Editor/DefaultTextBufferProjectService.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor.Editor
     [Export(typeof(TextBufferProjectService))]
     internal class DefaultTextBufferProjectService : TextBufferProjectService
     {
+        private const string DotNetCoreRazorCapability = "DotNetCoreRazor | AspNetCore";
```
NTaylorMullen: These are the base capabilities defined for current and old Razor in the MonoDevelop world.

tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
```diff
@@ -42,15 +42,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <!-- 
-      The extension project can not have a direct reference to its language service pieces. They are included via the manifest above.
```
NTaylorMullen: Some of our previous expectations about addins were wrong (not being able to directly reference your libraries). To avoid using reflection to bootstrap our types I tried out directly referencing our libraries and all worked fine.

tooling/Microsoft.VisualStudio.Mac.RazorAddin/RazorProjectExtension.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly ForegroundDispatcher _foregroundDispatcher;
+
+        public RazorProjectExtension()
+        {
+            _foregroundDispatcher = CompositionManager.GetExportedValue<ForegroundDispatcher>();
+        }
+
+        protected override void OnBoundToSolution()
+        {
+            if (!(Project is DotNetProject dotNetProject))
```
NTaylorMullen: This is the entry point for this entire PR so a good part to start at when reviewing.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/Editor/DefaultTextBufferProjectService.cs
rynowak: Not `DotNetCoreWeb`?

Can you provide a little more information about this?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultDotNetProjectHost.cs
```diff
@@ -0,0 +1,135 @@
...
+
+namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor.ProjectSystem
+{
+    internal class DefaultDotNetProjectHost : DotNetProjectHost
+    {
+        private const string ExplicitRazorConfigurationCapability = "DotNetCoreRazorConfiguration";
+
+        private readonly DotNetProject _project;
+        private readonly ForegroundDispatcher _foregroundDispatcher;
+        private readonly VisualStudioMacWorkspaceAccessor _workspaceAccessor;
```
rynowak: Is there a good reason why this isn't `VisualStudioWorkspaceAccessor`?

rynowak: So is `DotNetProjectHost` a mono type?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultDotNetProjectHost.cs
```diff
@@ -0,0 +1,135 @@
...
+            DetatchCurrentRazorProjectHost();
+        }
+
+        private void Project_ProjectCapabilitiesChanged(object sender, EventArgs e) => UpdateRazorHostProject();
+
+        private void UpdateRazorHostProject()
+        {
+            _foregroundDispatcher.AssertForegroundThread();
+
+            DetatchCurrentRazorProjectHost();
```
rynowak: Just a thought here, this seems like it will result in the project getting removed and re-created. That's not super desirable.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultDotNetProjectHost.cs
```diff
@@ -0,0 +1,135 @@
...
+                return false;
+            }
+
+            var languageService = workspace.Services.GetLanguageServices(RazorLanguage.Name);
+            var projectSnapshotManager = languageService.GetRequiredService<ProjectSnapshotManager>();
+
+            if (!(projectSnapshotManager is ProjectSnapshotManagerBase projectSnapshotManagerImpl))
+            {
+                projectSnapshotManagerBase = null;
+                return false;
```
rynowak: Just do a hard cast here. This won't ever fail.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
```diff
@@ -0,0 +1,75 @@
...
+            DotNetProject project,
+            ForegroundDispatcher foregroundDispatcher,
+            ProjectSnapshotManagerBase projectSnapshotManager)
+            : base(project, foregroundDispatcher, projectSnapshotManager)
+        {
+        }
+
+        protected override async Task OnProjectChangedAsync()
+        {
+            ForegroundDispatcher.AssertBackgroundThread();
```
rynowak: Maybe I haven't seen the complete picture here but what guarantees this is called on a background thread?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        protected async Task ExecuteWithLockAsync(Func<Task> func)
+        {
+            await _projectChangedSemaphore.WaitAsync().ConfigureAwait(false);
+
+            await func().ConfigureAwait(false);
+
+            _projectChangedSemaphore.Release();
```
rynowak: This needs a try/finally

rynowak: Is it possible to just use the VS Threading stuff here too?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/Editor/DefaultTextBufferProjectService.cs
NTaylorMullen: Haven't seen a DotNetCoreWeb. Todd was told to use these capabilities when determining "IsRazorCompatible"

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,166 @@
...
+        protected async Task UpdateHostProjectUnsafeAsync(HostProject newHostProject)
+        {
+            if (_foregroundDispatcher.IsForegroundThread)
+            {
+                UpdateHostProjectForeground(newHostProject);
+            }
+            else
+            {
+                await Task.Factory.StartNew(UpdateHostProjectForeground, newHostProject, CancellationToken.None, TaskCreationOptions.None, ForegroundDispatcher.ForegroundScheduler);
+            }
```
rynowak: This seems a little more complicated than just requiring this to be foreground only

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
NTaylorMullen: We invoke it ourselves via the foregorund dispatchers background task scheduler

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            var projectChanged = args.Any(arg => string.Equals(arg.Hint, ProjectChangedHint, StringComparison.Ordinal));
+            if (projectChanged)
+            {
+                // This method can be spammed for tons of project change events but all we really care about is "are we dirty?". 
+                // Therefore, we re-dispatch here to allow any remaining project change events to fire and to then only have 1 host
+                // project change trigger; this way we don't spam our own system with re-configure calls.
+                _batchingProjectChanges = true;
+                Task.Factory.StartNew(ProjectChangedBackgroundAsync, null, CancellationToken.None, TaskCreationOptions.None, ForegroundDispatcher.BackgroundScheduler);
```
rynowak: This is smart, I like it.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,166 @@
...
+        private async Task ProjectChangedBackgroundAsync(object state)
+        {
+            ForegroundDispatcher.AssertBackgroundThread();
+
+            _batchingProjectChanges = false;
+
+            await OnProjectChangedAsync();
+        }
+
+        private void DotNetProject_Modified(object sender, SolutionItemModifiedEventArgs args)
```
rynowak: Do they have a batching construct or do they just call this a million times?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultDotNetProjectHost.cs
NTaylorMullen: `DotNetProjectHost` is our own type that's VS4Mac specific. We can't use the normal VSWA because it doesn't allow us to take in a MonoDevelop solution and then retrieve the workspace for that solution.

test/Microsoft.VisualStudio.Mac.LanguageServices.Razor.Test/DefaultVisualStudioWorkspaceAccessorTest.cs
```diff
@@ -14,7 +14,7 @@ public class DefaultVisualStudioWorkspaceAccessorTest
         public void TryGetWorkspace_NoHostProject_ReturnsFalse()
         {
             // Arrange
-            var workspaceAccessor = new DefaultVisualStudioWorkspaceAccessor(Mock.Of<TextBufferProjectService>());
+            var workspaceAccessor = new DefaultVisualStudioMacWorkspaceAccessor(Mock.Of<TextBufferProjectService>());
```
rynowak: Test file needs a rename too

tooling/Microsoft.VisualStudio.Mac.RazorAddin/Microsoft.VisualStudio.Mac.RazorAddin.csproj
rynowak: Cool 👍

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DefaultDotNetProjectHost.cs
NTaylorMullen: You're correct it will. It'll only ever happen when the capabilities of the project change though. We could be more conditional about re-creating the project based on the capabilities if we really want but the capabilities changing happens so infrequently in VS4Mac world.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: Eh, `UpdateHostProjectUnsafeAsync` is always called on the background thread initially (we do time consuming work on the bg thread). I just added the if statement for the just-in-case situation. I could make it background thread only

tooling/Microsoft.VisualStudio.Mac.RazorAddin/RazorProjectExtension.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.VisualStudio.Mac.LanguageServices.Razor.ProjectSystem;
+using MonoDevelop.Ide.Composition;
+using MonoDevelop.Ide.TypeSystem;
+using MonoDevelop.Projects;
+
+namespace Microsoft.VisualStudio.Mac.RazorAddin
+{
+    internal class RazorProjectExtension : ProjectExtension
+    {
+        private readonly object ProjectPropertyLock = new object();
```
rynowak: naming, just call this `_lock`

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: They call a bajillion times

tooling/Microsoft.VisualStudio.Mac.RazorAddin/RazorProjectExtension.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Mac.LanguageServices.Razor.ProjectSystem;
+using MonoDevelop.Ide.Composition;
+using MonoDevelop.Ide.TypeSystem;
+using MonoDevelop.Projects;
+
+namespace Microsoft.VisualStudio.Mac.RazorAddin
+{
+    internal class RazorProjectExtension : ProjectExtension
```
rynowak: Consider combining this class with DefaultDotNetProjectHost.

Is there a good reason this has to be in Addin itself?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: nm, will just require callers to use the scheduler for the foreground thread.

tooling/Microsoft.VisualStudio.Mac.RazorAddin/RazorProjectExtension.cs
NTaylorMullen: I'd rather leave them separate. My reasoning is I like the separation of the bootsrapping code vs. the code that actually impacts what Razor hosts get created 😄

NTaylorMullen:
🆙 📅 working on tests now.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
NTaylorMullen: - tl;dr; I'd recommend reviewing this class as an entire file instead of looking at the diff.

Had to refactor this class in VS4Mac so to keep both worlds consistent I refactored the windows one as well. In some places I'm extra verbose/protective to stay consistent across platforms. In the end it came out a lot more clear and in was able to unit test 99% of it. Ported the VS4Mac unit tests on this refactored guy as well and left the existing tests to ensure everything worked 100% the same.

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/DotNetProjectHost.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using MonoDevelop.Projects;
+
+namespace Microsoft.VisualStudio.Mac.LanguageServices.Razor.ProjectSystem
+{
+    internal abstract class DotNetProjectHost
```
NTaylorMullen: Could technically get rid of this guy and the factory sine we can't really test the types anyways but figured I'd leave them around for consistency sake (that's what we do normally). If you feel differently I don't have a problem removing them.

build/dependencies.props
```diff
@@ -36,7 +36,7 @@
     <MicrosoftVisualStudioShellInterop90PackageVersion>9.0.30729</MicrosoftVisualStudioShellInterop90PackageVersion>
     <MicrosoftVisualStudioShellInteropPackageVersion>7.10.6071</MicrosoftVisualStudioShellInteropPackageVersion>
     <MicrosoftVisualStudioTextUIPackageVersion>15.6.161-preview</MicrosoftVisualStudioTextUIPackageVersion>
-    <MonoAddinsPackageVersion>1.3.7</MonoAddinsPackageVersion>
+    <MonoAddinsPackageVersion>1.3.8</MonoAddinsPackageVersion>
```
NTaylorMullen: Will do a PR to Universe to update this there as well.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
rynowak: This looks good.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                await func().ConfigureAwait(false);
+            }
+        }
+
+        private async Task ProjectChangedBackgroundAsync(object state)
+        {
+            ForegroundDispatcher.AssertBackgroundThread();
+
+            _batchingProjectChanges = false;
```
rynowak: BTW you don't have any thread-safety guarantees when updating this on the background thread. There is no guarantee that another thread will see this update in a timely manner. So if you need a real 'happens before' relationship here then you need something stronger than a bool.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                throw new ArgumentNullException(nameof(args));
+            }
+
+            ForegroundDispatcher.AssertForegroundThread();
+
+            if (_batchingProjectChanges)
+            {
+                // Already waiting to recompute host project, no need to do any more work to determine if we're dirty.
+                return;
```
rynowak: Yeah this kind of usage is a problem. You don't have a guarantee that the FG thread will see the 'false' value . You need a lock around that bool.

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: Hmmm, maybe I'm missing something but I don't allow more than one background thread to be running based off of this bool. Do you see something different?

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: That's ok though. We aren't passing around a stateful object. We re-compute the state at the time of `OnProjectChangedAsync`. So in the case that the background thread sets the `_batchingProjectChanges` to false (lets assume this if block has been entered) it'll already be queuing up changes based on the computed foreground state; otherwise this will have re-triggered another background computation. 

My intent here was for this to be a safe race. Maybe your race condition eyes are seeing something different though so feel free to call me out😄

src/Microsoft.VisualStudio.Mac.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: Ooooo I do see that we don't ensure ordering howeverrr, we do need to lock around this `OnProjectChangedAsync` call.

Issue #2134, opened 2018-03-06
Investigate and consider increasing build server shutdown timeout in tests


ajaybhargavb:
https://travis-ci.org/aspnet/Razor/jobs/349921394#L2732

Now that we are spawning multiple build servers at the same time during tests, shutting them down is taking longer than 1 minute. Need to investigate exactly why and maybe increase this timeout to prevent this failure.

cc @pranavkm

ajaybhargavb:
e085441ad3e9233b096e0e485b33cf1a07a390ef

PR #2140, opened 2018-03-07
Don't limit parallel threads for Razor.Design tests

ajaybhargavb:
#2134 

The build server tests try to shutdown multiple servers in parallel which needs a lot of threads. So in slower machines they just timeout.


pranavkm:
OSX tests failing due to different issue that @ryanbrandenburg is investigating. :shipit:

PR #2143, opened 2018-03-08
Rynowak/tag helpers design time

rynowak:


src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
```diff
@@ -9,6 +9,10 @@ namespace Microsoft.CodeAnalysis.Razor
 {
     public sealed class DefaultTagHelperDescriptorProvider : RazorEngineFeatureBase, ITagHelperDescriptorProvider
     {
+        [Obsolete(
+            "This property is obsolete will not be honored. Documentation will be included if " +
+            "TagHelperDescriptorProviderContext.IncludeDocumentation is set to true. Hidden tag helpers will" +
```
NTaylorMullen: use `nameof` for the type/property names. Also will need to file an issue for this one 😉

rynowak: I super don't care about this. nameof isn't helpful because we can't change the names.

Issue #2139, opened 2018-03-06
TagHelperDescriptor serializer doesn't roundtrip 'Kind'


rynowak:
The 'Kind' property doesn't get roundtripped. Which means that view component tag helpers and blazor components might not work when using the CLI toolset.

I think the problem is here: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.VisualStudio.LanguageServices.Razor/TagHelperDescriptorJsonConverter.cs#L86 - since Kind is read  only I don't think it is serialized by the default serializer.

mkArtakMSFT:
@kichalla, can you please look into this? Thanks!

ajaybhargavb:
@mkArtakMSFT @kichalla, I'm taking this issue because this is needed to fix https://github.com/aspnet/Razor/issues/2150.

ajaybhargavb:
fcd4484542733d9666aa981ba8709bb22ce3de4c

PR #2151, opened 2018-03-12
Fix TagHelperDescriptor.Kind serialization

ajaybhargavb:
#2139 

- We were never passing in the `TagHelperDescriptorKind` to the builder when we reconstruct the descriptor.
- Added a test.

### DIFF
--- a/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Serialization/TagHelperDescriptorSerializationTest.cs
+++ b/test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Serialization/TagHelperDescriptorSerializationTest.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


test/Microsoft.VisualStudio.LanguageServices.Razor.Test/Serialization/TagHelperDescriptorSerializationTest.cs
```diff
@@ -18,6 +18,51 @@ public void TagHelperDescriptor_RoundTripsProperly()
         {
             // Arrange
             var expectedDescriptor = CreateTagHelperDescriptor(
+                kind: TagHelperConventions.DefaultKind,
```
rynowak: Can we use something other than the default kind in the test? This test would pass even if you never fixed the bug

rynowak: Oh NVM I see you added a bonus test

Issue #2079, opened 2018-02-16
Razor build tasks should validate Configuration


rynowak:
The tag helper task needs some parameter validation https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs

And the generate task needs to validate configuration  https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tasks/RazorGenerate.cs#L39

mkArtakMSFT:
@ajaybhargavb , can you please look into this? Thanks!

ajaybhargavb:
d8e363d99c07d3afc5991dadb67a3352bb19c106

PR #2152, opened 2018-03-12
Added parameter validation for RazorTagHelper and RazorGenerate tasks

ajaybhargavb:
#2079 

- Added validation for `ProjectRoot` in RazorGenerate
- Added validation for RazorTagHelper (Mostly copy-paste from RazorGenerate)
- We never had any unit tests for the tasks but the integration tests should cover this.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
@@ -47,6 +47,42 @@ namespace Microsoft.AspNetCore.Razor.Tasks
             return false;
         }
 
+        protected override bool ValidateParameters()
+        {
+            if (!Directory.Exists(ProjectRoot))
+            {
+                Log.LogError("The specified project root directory {0} doesn't exist.", ProjectRoot);
+                return false;
+            }
+
+            if (Configuration.Length == 0)
+            {
+                Log.LogError("The project {0} must provide a value for {1}.", ProjectRoot, nameof(Configuration));
+                return false;
+            }
+
+            for (var i = 0; i < Assemblies.Length; i++)
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Tasks/RazorTagHelper.cs
```diff
@@ -47,6 +47,33 @@ protected override bool SkipTaskExecution()
...
+            for (var i = 0; i < Extensions.Length; i++)
+            {
+                if (!EnsureRequiredMetadata(Extensions[i], Identity) ||
+                    !EnsureRequiredMetadata(Extensions[i], AssemblyName) ||
+                    !EnsureRequiredMetadata(Extensions[i], AssemblyFilePath))
+                {
+                    return false;
+                }
+            }
+
```
rynowak: We should also validate that the references are full-qualified paths. The build server requires this.

ajaybhargavb: Added.

ajaybhargavb:
🆙 📅

rynowak:
Just as a general note on this, if you find any more *kinds* of input that will be ignored by the build server or will cause the build server to crash, they need to be validated here. The error reporting from the build server and CLI tool is very poor, we want to catch these things up front in MSBuild.

ajaybhargavb:
The other thing left to verify is `Version` but we can't do that reliably here because we don't have access to `RazorLanguageVersion`. So an invalid version will still go through and fail when trying to compile on the server.

ajaybhargavb:
I could source include that file alone. Going to do that unless you have any objections @rynowak

rynowak:
Sounds right to me.

ajaybhargavb:
Discussed offline. Not going to verify Version at this level.

PR #2153, opened 2018-03-13
[VSfM] Fix code completion disappearing/reseting because C# part of .cshtml in regenerated

DavidKarlas:
Razor AddIn generation of C# projection content inside .cshtml depends on code completion being visible hence it needs to know state.
Since we are using Roslyn code completion API which doesn't have method to check if code completion is visible and also it's internal, only other way would be to make some public static field inside CSharpBinding which indicates visibility of code completion, hence it's better to just store this value inside textView.Properties. We should be able to remove this in 15.8 when we switch to new VSEditor code completion API.


KirillOsenkov:
@NTaylorMullen @rynowak @ToddGrun

KirillOsenkov:
:+1:

rynowak:
@NTaylorMullen does this need to be ported to 15.7?

PR #2154, opened 2018-03-13
Avoid generating RelatedAssemblyAttribute when RazorSdk isn't in play

pranavkm:


test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIntegrationTest.cs
```diff
@@ -367,6 +367,51 @@ public Task Build_SimpleMvc_UsingDesktopMSBuildAndWithoutBuildServer_CanBuildSuc
...
+
+            Assert.FileExists(result, assemblyInfo);
+            Assert.FileDoesNotContain(result, assemblyInfo, "RelatedAssemblyAttribute");
+
+            Assert.FileDoesNotExist(result, IntermediateOutputPath, "SimpleMvc.RazorAssemblyInfo.cs");
+        }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task Build_AddsRelatedAssemblyPart_IfCompileOnPublishIsAllowed()
```
pranavkm: This scenario forces us in being defensive with loading related assembly parts i.e. no-op when the assembly cannot be found.

PR #2155, opened 2018-03-13
Fix typo in Razor.props

pranavkm:


pranavkm:
FYI. Spotted this when building Mvc referencing the package.

pranavkm:
I'm adding a reference to this in Mvc. Also added this to track this since it's a test gap we have - https://github.com/aspnet/Razor/issues/2156

PR #2160, opened 2018-03-14
Added Merge method to tag helper attribute list

Eonasdan:
Based  on a helper method I [created here](https://gist.github.com/Eonasdan/afe99a3c349ba37e1462fe20ccd75fee).

The goal is to allow a dev to easily merge attributes like `class` for instance:

    output.Attributes.Merge("class", "col-md-6");
    //class = "col-md-6"
    output.Attributes.Merge("class", "col-lg-6");
    //class = "col-lg-7 col-md-6"


Eonasdan:
@rynowak or @ajaybhargavb looks like travis dies in the mono osx build. Doesn't look like that's in my control though. It builds successfully from build.cmd locally.

Eonasdan:
Hello? Bueller? Bueller? Bueller?

@rynowak or @ajaybhargavb

rynowak:
Hi @Eonasdan - let's turn this into a discussion first. 

Any API additions that we make live **forever** so we try to but a lot of care into what we accept.

@DamianEdwards @dougbu - any thoughts on this?

dougbu:
Does any attribute other than `class` and perhaps `style` (which isn't used much except in early development) merit merging?

We have extension methods in MVC e.g. https://github.com/aspnet/Mvc/blob/2bc4c40485b799418bb50fbafa156a9487b372c1/src/Microsoft.AspNetCore.Mvc.TagHelpers/TagHelperOutputExtensions.cs#L166-L169
which handle the `class` attribute pretty well. Are these useful in non-MVC Razor scenarios? If yes, I suggest moving `TagHelperOutputExtensions` into Razor rather than coming up with a new API.

NTaylorMullen:
> @DamianEdwards @dougbu @NTaylorMullen - any thoughts on this?

Ya this would be valuable. It's hard to manage `TagHelperAttributeList`s.  Howevverr, the merging of individual attributes here will have to be a little different. We can't assume the values are strings.

Eonasdan:
@rynowak acknowledge. Thanks for getting back to me.

@dougbu I could easily see this in the case of build dynamic id's.

@NTaylorMullen : I wasn't really fond of assuming that value was a string, however, there was some place else I saw (can't remember where right off) that ultimately assumed that the value could be "tostring'ed"

mkArtakMSFT:
Thanks for the effort, @Eonasdan. After looking through this it seems it'll be better to have a higher perspective view to this rather than adding just a single extension method to support one functionality only.

Eilon:
@NTaylorMullen / @dougbu / @rynowak / @mkArtakMSFT - any further action planned on this PR?

dougbu:
> any further action planned on this PR?

Not from my perspective.

But, do we have an issue somewhere about providing something like `TagHelperOutputExtensions` more generally e.g. in Razor instead of MVC?

NTaylorMullen:
This all moved to the aspnet/AspNetCore-Tooling repo.

PR #2165, opened 2018-03-15
Display errors from build server as part of MsBuild output

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2175

Fixed a few things with Razor MsBuild tasks and its logging,
- When we get a `CompletedServerResponse` from the build server,  we should **not** fallback to cli execution. This is what [Roslyn](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/MSBuildTask/ManagedCompiler.cs#L605) does.
- Capturing `Out` and `Error` from the server separately. This helps with displaying just the error from the server without all the other noise.
- Setting the `Out` and `Error` property wasn't actually working because it was creating a new instance of `CommandLineApplication` for each command. Added a way to tunnel it through.
- Added a test

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
@@ -148,9 +148,18 @@ namespace Microsoft.AspNetCore.Razor.Tasks
                     }
                     else
                     {
-                        Log.LogMessage(
-                            StandardOutputLoggingImportance,
-                            $"Server execution completed with return code {result}. For more info, check the server log file in the location specified by the RAZORBUILDSERVER_LOG environment variable.");
+                        Log.LogMessage(StandardOutputLoggingImportance, $"Server execution completed with return code {result}. For more info, check the server log file in the location specified by the RAZORBUILDSERVER_LOG environment variable.");
+
+                        if (LogStandardErrorAsError)
+                        {
+                            LogErrors(completedResponse.ErrorOutput);
+                        }
+                        else
+                        {
+                            LogMessages(completedResponse.ErrorOutput, StandardErrorLoggingImportance);
+                        }
+
+                        return true;
                     }
                 }
                 else
@@ -158,9 +167,9 @@ namespace Microsoft.AspNetCore.Razor.Tasks
                     Log.LogMessage(
                         StandardOutputLoggingImportance,
                         $"Server execution failed with response {response.Type}. For more info, check the server log file in the location specified by the RAZORBUILDSERVER_LOG environment variable.");
-                }
 
-                result = -1;
+                    result = -1;
+                }
 
                 if (ForceServer)
                 {
@@ -172,6 +181,32 @@ namespace Microsoft.AspNetCore.Razor.Tasks
             }
 
             return false;
+        }
+
+        private void LogMessages(string output, MessageImportance messageImportance)
+        {
+            var lines = output.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (var line in lines)
+            {
+                if (!string.IsNullOrWhiteSpace(line))
+                {
+                    var trimmedMessage = line.Trim();
+                    Log.LogMessageFromText(trimmedMessage, messageImportance);
+                }
+            }
+        }
+
+        private void LogErrors(string output)
+        {
+            var lines = output.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (var line in lines)
+            {
+                if (!string.IsNullOrWhiteSpace(line))
+                {
+                    var trimmedMessage = line.Trim();
+                    Log.LogError(trimmedMessage);
+                }
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/Application.cs
@@ -13,12 +13,20 @@ namespace Microsoft.AspNetCore.Razor.Tools
 {
     internal class Application : CommandLineApplication
     {
-        public Application(CancellationToken cancellationToken, ExtensionAssemblyLoader loader, ExtensionDependencyChecker checker, Func<string, MetadataReferenceProperties, PortableExecutableReference> assemblyReferenceProvider)
+        public Application(
+            CancellationToken cancellationToken,
+            ExtensionAssemblyLoader loader,
+            ExtensionDependencyChecker checker,
+            Func<string, MetadataReferenceProperties, PortableExecutableReference> assemblyReferenceProvider,
+            TextWriter output = null,
+            TextWriter error = null)
         {
             CancellationToken = cancellationToken;
             Checker = checker;
             Loader = loader;
             AssemblyReferenceProvider = assemblyReferenceProvider;
+            Out = output ?? Out;
+            Error = error ?? Error;
 
             Name = "rzc";
             FullName = "Microsoft ASP.NET Core Razor CLI tool";
--- a/src/Microsoft.AspNetCore.Razor.Tools/CommandBase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/CommandBase.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.CommandLineUtils;
@@ -20,6 +21,8 @@ namespace Microsoft.AspNetCore.Razor.Tools
 
             base.Parent = parent;
             Name = name;
+            Out = parent.Out ?? Out;
+            Error = parent.Error ?? Error;
 
             Help = HelpOption("-?|-h|--help");
             OnExecute((Func<Task<int>>)ExecuteAsync);
--- a/src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
@@ -48,27 +48,27 @@ namespace Microsoft.AspNetCore.Razor.Tools
                 }
 
                 var exitCode = 0;
-                var output = string.Empty;
                 var commandArgs = parsed.args.ToArray();
 
-                var writer = ServerLogger.IsLoggingEnabled ? new StringWriter() : TextWriter.Null;
+                var outputWriter = new StringWriter();
+                var errorWriter = new StringWriter();
 
-                var checker = new DefaultExtensionDependencyChecker(Loader, writer, writer);
-                var app = new Application(cancellationToken, Loader, checker, AssemblyReferenceProvider)
-                {
-                    Out = writer,
-                    Error = writer,
-                };
+                var checker = new DefaultExtensionDependencyChecker(Loader, outputWriter, errorWriter);
+                var app = new Application(cancellationToken, Loader, checker, AssemblyReferenceProvider, outputWriter, errorWriter);
 
                 exitCode = app.Execute(commandArgs);
 
-                if (ServerLogger.IsLoggingEnabled)
-                {
-                    output = writer.ToString();
-                    ServerLogger.Log(output);
-                }
+                var output = outputWriter.ToString();
+                var error = errorWriter.ToString();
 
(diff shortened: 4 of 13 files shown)


src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -148,19 +148,28 @@ public override void Cancel()
...
+                        if (LogStandardErrorAsError)
+                        {
+                            Log.LogError(completedResponse.ErrorOutput);
+                        }
+                        else
+                        {
+                            Log.LogMessage(StandardErrorLoggingImportance, completedResponse.ErrorOutput);
+                        }
+
+                        return true;
```
ajaybhargavb: Doing the same thing as [Roslyn](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/MSBuildTask/ManagedCompiler.cs#L605) here.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -101,7 +101,7 @@ protected override bool ValidateArguments()
             }
             else if (!RazorLanguageVersion.TryParse(Version.Value(), out _))
             {
-                Error.WriteLine($"{Version.ValueName} is not a valid language version.");
+                Error.WriteLine($"{Version.Value()} is not a valid language version.");
```
ajaybhargavb: `Version.ValueName` is always empty.

rynowak:
> percolate 

Remind me to tell you a story about a percolator tomorrow.

src/Microsoft.AspNetCore.Razor.Tools/CommandBase.cs
```diff
@@ -20,6 +21,8 @@ protected CommandBase(Application parent, string name)
 
             base.Parent = parent;
             Name = name;
+            Out = output ?? Out;
+            Error = error ?? Error;
```
rynowak: I really don't get how all of this rejiggering of properties vs parameters helps. The properties are still there, and they are still mutable, but in some cases they are ignored?

src/Microsoft.AspNetCore.Razor.Tools/Application.cs
```diff
@@ -27,10 +35,10 @@ public Application(CancellationToken cancellationToken, ExtensionAssemblyLoader
 
             HelpOption("-?|-h|--help");
 
-            Commands.Add(new ServerCommand(this));
-            Commands.Add(new ShutdownCommand(this));
-            Commands.Add(new DiscoverCommand(this));
-            Commands.Add(new GenerateCommand(this));
+            Commands.Add(new ServerCommand(this, Out, Error));
+            Commands.Add(new ShutdownCommand(this, Out, Error));
+            Commands.Add(new DiscoverCommand(this, Out, Error));
```
pranavkm: lol

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
```diff
@@ -49,26 +50,25 @@ public override ServerResponse Execute(ServerRequest request, CancellationToken
...
 
                 exitCode = app.Execute(commandArgs);
 
-                if (ServerLogger.IsLoggingEnabled)
-                {
-                    output = writer.ToString();
-                    ServerLogger.Log(output);
-                }
+                output = outputWriter.ToString();
+                error = errorWriter.ToString();
```
rynowak: We did work **specifically** to avoid created a string builder and logging lots of output when it doesn't need to go anywhere. I think we should talk more about why this is important.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -84,5 +84,21 @@ public Task Build_SimpleMvc_WithServer_UsingDesktopMSBuild_CanBuildSuccessfully(
...
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task Build_ErrorInServer_DisplaysErrorInMsBuildOutput()
+        {
+            var result = await DotnetMSBuild(
+                "Build",
+                "/p:_RazorForceBuildServer=true /p:RazorLangVersion=5.0");
+
+            Assert.BuildFailed(result);
+            Assert.BuildOutputContainsLine(result, "5.0 is not a valid language version.");
```
rynowak: What's the full line look like coming from MSBuild?

src/Microsoft.AspNetCore.Razor.Tools/Application.cs
pranavkm: I wonder if we should just fix this in the source for `CommandLineApplication`. Seems like a common concern.

src/Microsoft.AspNetCore.Razor.Tools/CommandBase.cs
rynowak: How much output is there going to be?

pranavkm: We set the value on the top level `CommadLineApplication` but that doesn't affect any new Commands that are created. This makes them both use the same inputs and outputs

src/Microsoft.AspNetCore.Razor.Tools/CompilerHost.cs
ajaybhargavb: The main problem I am trying to fix here is this. 
Assume there is a compilation error in one of the views. 
It will try to compile it in the server. That will return an error code of -1 because `RazorGenerate` failed. The MsBuild task would fallback to in process compilation and repeat the same process and fail again. (This is the error we are currently seeing in output)
Instead, we could have reported the error from server without having to fallback to the cli. For that I need to create the StringWriter for capturing the `Error` stream atleast. We can talk tomorrow if `Output` stream is really needed.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
ajaybhargavb: This is the entire line.

src/Microsoft.AspNetCore.Razor.Tools/CommandBase.cs
ajaybhargavb: > How much output is there going to be?

Output will just be whatever the CLI tool would show (`DiscoverCommand` and `GenerateCommand`). All the other server specific logging like "Request received" etc will not show up here.

src/Microsoft.AspNetCore.Razor.Tools/Application.cs
ajaybhargavb: This definitely looks like a common problem. It's atleast worth filing a bug.

ajaybhargavb: Filed https://github.com/aspnet/Home/issues/2963

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Tools/LazyStringWriter.cs
```diff
@@ -0,0 +1,95 @@
...
+using System;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    internal class LazyStringWriter : TextWriter
+    {
+        // Arbitrary value
+        private const int DefaultMaxCapacity = 10000;
```
ajaybhargavb: @rynowak, do you think this is enough characters?

src/Microsoft.AspNetCore.Razor.Tools/LazyStringWriter.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    internal class LazyStringWriter : TextWriter
```
rynowak: This seems like kinda a big/complicated hammer - do you dislike something about my suggestion to only log 100 errors from inside the tool?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
rynowak: Is this still the entire line?

pranavkm: Plus it probably needs to say `Razor` somewhere in

src/Microsoft.AspNetCore.Razor.Tools/LazyStringWriter.cs
ajaybhargavb: Nothing I dislike about that. But this is a more general fix that covers all the commands. Plus it also only creates the StringWriter when needed. It is actually not complicated at all. If you still feel strongly about this I can go back to your suggestion.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
ajaybhargavb: > Is this still the entire line?

Yes, but there is also another line after this that says `rzc generate exited with code -1`.

> Plus it probably needs to say Razor somewhere in

👍

ajaybhargavb:
🆙 📅

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -84,5 +85,23 @@ public Task Build_SimpleMvc_WithServer_UsingDesktopMSBuild_CanBuildSuccessfully(
...
+        public async Task Build_ErrorInServer_DisplaysErrorInMsBuildOutput()
+        {
+            var result = await DotnetMSBuild(
+                "Build",
+                "/p:_RazorForceBuildServer=true /p:RazorLangVersion=5.0");
+
+            Assert.BuildFailed(result);
+            Assert.BuildOutputContainsLine(
+                result,
+                $"Invalid option 5.0 for Razor language version --version; must be Latest or a valid version in range {RazorLanguageVersion.Version_1_0} to {RazorLanguageVersion.Latest}.");
```
pranavkm: Can you just the verbatim string? `in range 1.0 to Latest`?

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -174,6 +183,32 @@ public override void Cancel()
             return false;
         }
 
+        private void LogMessages(string output, MessageImportance messageImportance)
+        {
+            var lines = output.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (var line in lines)
+            {
+                var trimmedMessage = line.Trim();
```
pranavkm: `string.IsNullorWhitespace`

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -166,9 +166,12 @@ protected override bool ValidateArguments()
...
                     success = false;
-                    foreach (var error in result.CSharpDocument.Diagnostics)
+
+                    // Only show the first 100 errors to prevent massive string allocations.
+                    for (var i = 0; i < result.CSharpDocument.Diagnostics.Count && i < 100; i++)
                     {
-                        Console.Error.WriteLine(error.ToString());
+                        var error = result.CSharpDocument.Diagnostics[i];
+                        Error.WriteLine(error.ToString());
                     }
```
rynowak: Can you add an additional error message if there are over 100 errors? Its important that this experience be polished.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
ajaybhargavb: Spoke offline

ajaybhargavb:
🆙 📅

PR #2176, opened 2018-03-16
Update BuildTools to fix dotnet-ci builds

pranavkm:


pranavkm:
Windows build failed with:

> Invalid option '7.2' for /langversion; must be ISO-1, ISO-2, Default, Latest or a valid version in range 1 to 7.1.

@mmitche, what version of VS do the Windows agents have? We set LangVersion = 7.2 across our projects and that requires VS 15.5 or later.

PR #2179, opened 2018-03-16
Fix #2149 give generated files distinct extension

### LINKED ISSUE #2149: Give Razor generated files a distinct file extension
Right now we use `.cs`, but based on some VS extensibility points we should probably choose a unique file extension.

If I don't hear any other great suggestions, I'm going to go with  `.g.cshtml.cs`

rynowak:
Gives our generated a files an extension that isn't used for any other
purpose (that we know of). This is handy for tooling to be able to
quickly know if a file is 'ours'. This comes up in places like
IVsSymbolicNavigationNotify (go to definition).

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
+++ b/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
@@ -300,7 +300,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup>
       <RazorGenerateWithTargetPath Condition="'%(RazorGenerateWithTargetPath.GeneratedOutput)' == ''">
-        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)$([System.IO.Path]::ChangeExtension('%(RazorGenerateWithTargetPath.TargetPath)', '.cs'))</GeneratedOutput>
+        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)$([System.IO.Path]::ChangeExtension('%(RazorGenerateWithTargetPath.TargetPath)', '$(RazorGenerateOutputFileExtension)'))</GeneratedOutput>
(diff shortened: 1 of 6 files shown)


src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -300,7 +300,7 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <ItemGroup>
       <RazorGenerateWithTargetPath Condition="'%(RazorGenerateWithTargetPath.GeneratedOutput)' == ''">
-        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)$([System.IO.Path]::ChangeExtension('%(RazorGenerateWithTargetPath.TargetPath)', '.cs'))</GeneratedOutput>
+        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)$([System.IO.Path]::ChangeExtension('%(RazorGenerateWithTargetPath.TargetPath)', '$(RazorGenerateOutputFileExtension)'))</GeneratedOutput>
```
pranavkm: Should we also use this for the generated assembly info: https://github.com/aspnet/Razor/blob/dev/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.GenerateAssemblyInfo.targets#L21?

pranavkm: Sorry, by do this, I mean make it `.g.cs`?

rynowak: We should do whatever the .NET SDK does. If they use `.g.cs` we should as well. That said, there is no magic here.

pranavkm: They don't.

PR #2180, opened 2018-03-17
Show error if page directive is not at the top of file

ajaybhargavb:
aspnet/Mvc#6633

Once we know that an `@page` exists in the document, we can process the same document again with `ParseLeadingDirectives=true` and make sure if the `@page` exists in the resulting code document. If not, log an error.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
@@ -13,6 +13,25 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
         public static readonly string RazorPageDocumentKind = "mvc.1.0.razor-page";
         public static readonly string RouteTemplateKey = "RouteTemplate";
 
+        private static readonly RazorProjectEngine LeadingDirectiveParsingEngine = RazorProjectEngine.Create(
+            RazorConfiguration.Default,
+            RazorProjectFileSystem.Create("/"),
+            builder =>
+            {
+                for (var i = builder.Phases.Count - 1; i >= 0; i--)
+                {
+                    var phase = builder.Phases[i];
+                    builder.Phases.RemoveAt(i);
+                    if (phase is IRazorDocumentClassifierPhase)
+                    {
+                        break;
+                    }
+                }
+
+                RazorExtensions.Register(builder);
+                builder.Features.Add(new LeadingDirectiveParserOptionsFeature());
+            });
+
         protected override string DocumentKind => RazorPageDocumentKind;
 
         protected override bool IsMatch(RazorCodeDocument codeDocument, DocumentIntermediateNode documentNode)
@@ -48,7 +67,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             var document = codeDocument.GetDocumentIntermediateNode();
             PageDirective.TryGetPageDirective(document, out var pageDirective);
 
-            EnsureValidPageDirective(pageDirective);
+            EnsureValidPageDirective(codeDocument, pageDirective);
 
             AddRouteTemplateMetadataAttribute(@namespace, @class, pageDirective);
         }
@@ -75,7 +94,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             @namespace.Children.Insert(classIndex, metadataAttributeNode);
         }
 
-        private void EnsureValidPageDirective(PageDirective pageDirective)
+        private void EnsureValidPageDirective(RazorCodeDocument codeDocument, PageDirective pageDirective)
         {
             Debug.Assert(pageDirective != null);
 
@@ -83,6 +102,33 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
             {
                 pageDirective.DirectiveNode.Diagnostics.Add(
                     RazorExtensionsDiagnosticFactory.CreatePageDirective_CannotBeImported(pageDirective.DirectiveNode.Source.Value));
+            }
+            else
+            {
+                // The document contains a page directive and it is not imported.
+                // We now want to make sure this page directive exists at the top of the file.
+                // We are going to do that by re-parsing the document until the very first line that is not Razor comment
+                // or whitespace. We then make sure the page directive still exists in the re-parsed IR tree.
+                var leadingDirectiveCodeDocument = RazorCodeDocument.Create(codeDocument.Source);
+                LeadingDirectiveParsingEngine.Engine.Process(leadingDirectiveCodeDocument);
+
+                var documentIRNode = leadingDirectiveCodeDocument.GetDocumentIntermediateNode();
+                if (!PageDirective.TryGetPageDirective(documentIRNode, out var _))
+                {
+                    // The page directive is not the leading directive. Add an error.
+                    pageDirective.DirectiveNode.Diagnostics.Add(
+                        RazorExtensionsDiagnosticFactory.CreatePageDirective_MustExistAtTheTopOfFile(pageDirective.DirectiveNode.Source.Value));
+                }
+            }
+        }
+
+        private class LeadingDirectiveParserOptionsFeature : RazorEngineFeatureBase, IConfigureRazorParserOptionsFeature
+        {
+            public int Order { get; }
+
+            public void Configure(RazorParserOptionsBuilder options)
+            {
+                options.ParseLeadingDirectives = true;
             }
         }
     }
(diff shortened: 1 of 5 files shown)


ajaybhargavb:
@pranavkm, Windows Release x64 Build has been consistently passing but it is actually not running the full build. It fails at run.ps1. https://ci3.dot.net/job/aspnet_Razor/job/dev/job/windows-Configuration_Release_prtest/11/console

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
```diff
@@ -10,6 +10,21 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
     public class RazorPageDocumentClassifierPass : DocumentClassifierPassBase
     {
         public static readonly string RazorPageDocumentKind = "mvc.1.0.razor-page";
+        private static readonly RazorEngine LeadingDirectiveParsingEngine = RazorEngine.Create(builder =>
```
NTaylorMullen: Build a project engine instead, we aren't going to maintain the `RazorEngine.Create` overload.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
```diff
@@ -58,6 +73,32 @@ private void EnsureValidPageDirective(RazorCodeDocument codeDocument)
                 pageDirective.DirectiveNode.Diagnostics.Add(
                     RazorExtensionsDiagnosticFactory.CreatePageDirective_CannotBeImported(pageDirective.DirectiveNode.Source.Value));
             }
+            else
+            {
```
NTaylorMullen: I'd add a bigger disclaimer at the top of this about why we're re-parsing the document (consistency with MVC runtime + can't see the parse tree)

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
```diff
@@ -10,6 +10,21 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
...
+                var phase = builder.Phases[i];
+                builder.Phases.RemoveAt(i);
+                if (phase is IRazorDocumentClassifierPhase)
+                {
+                    break;
+                }
+            }
+
+            RazorExtensions.Register(builder);
+            builder.Features.Add(new LeadingDirectiveParserOptionsFeature());
```
rynowak: Let me get out in front of this and say that I really don't think us 'creating' an engine is something we want to live with. Adding something like this is going to just cause lots of complexity for us down the line.

Could we just examine the IR tree and say "good enough". The user feedback we've gotten about this problem is aimed at  catching obvious mistakes.

rynowak: Or could we call the static methods on the parser (they exist for a reason).

NTaylorMullen: > Could we just examine the IR tree and say "good enough". The user feedback we've gotten about this problem is aimed at catching obvious mistakes.

IR doesn't represent Razor comments.

> Or could we call the static methods on the parser (they exist for a reason).

They're internal, we'd have to IVT for that.

> Let me get out in front of this and say that I really don't think us 'creating' an engine is something we want to live with. Adding something like this is going to just cause lots of complexity for us down the line.

Ehh, in this case i'd argue it reduces the complexity because this is what MVC does. Basically we get identical behavior to MVC for free without having to mock the same behavior.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Resources.resx
```diff
@@ -171,6 +171,9 @@
   <data name="PageDirectiveCannotBeImported" xml:space="preserve">
     <value>The '@{0}' directive specified in {1} file will not be imported. The directive must appear at the top of each Razor cshtml file.</value>
   </data>
+  <data name="PageDirectiveMustExistAtTheTopOfFile" xml:space="preserve">
+    <value>The '@{0}' directive must exist at the top of the file. Only Razor comments and whitespace are allowed before the '@{0}' directive.</value>
```
pranavkm: as the first line of a Razor file.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
pranavkm: > IR doesn't represent Razor comments.

We don't care about comments in this case, do we?

NTaylorMullen: > We don't care about comments in this case, do we?

The single line parser does (it ignores them)

ajaybhargavb: > We don't care about comments in this case, do we?

We want to allow comments to be present before the `@page`.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Resources.resx
ajaybhargavb: Can't. The first line can be a Razor comment or an empty line.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
```diff
@@ -58,6 +73,32 @@ private void EnsureValidPageDirective(RazorCodeDocument codeDocument)
...
+            }
+        }
+
+        private class LeadingDirectiveParserOptionsFeature : RazorEngineFeatureBase, IConfigureRazorParserOptionsFeature
+        {
+            public int Order { get; }
+
+            public void Configure(RazorParserOptionsBuilder options)
+            {
+                options.ParseLeadingDirectives = true;
```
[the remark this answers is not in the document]
ajaybhargavb: This already does what we want. This is exactly what we do in Mvc.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
pranavkm: Sure. This is the bit I don't follow:

> IR doesn't represent Razor comments.

The presence or absence of comments should be immaterial here. We just need to make sure, no other sematically meaningful entry exists before `@page`. So wouldn't the IR tree suffice for this purpose?

ajaybhargavb: > The presence or absence of comments should be immaterial here.

The problem is with Razor comments are represented as CSharpCode just like any other content. How will you differentiate a semantically meaningful entry from a Razor comment if they are both represented by the same node? Looking into it's content will get ugly and complicated.

pranavkm: > The problem is with Razor comments are represented as CSharpCode just like any other content.

Gotcha. This I did not know. I assumed it was missing those nodes in entirety.

pranavkm:
@dotnet-bot build

pranavkm:
@dotnet-bot test

pranavkm:
@dotnet-bot test ci

pranavkm:
@dotnet-bot test this

pranavkm:
@dotnet-bot test this please

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/RazorPageDocumentClassifierPass.cs
ajaybhargavb: Made the comment better.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Resources.resx
```diff
@@ -171,6 +171,9 @@
   <data name="PageDirectiveCannotBeImported" xml:space="preserve">
     <value>The '@{0}' directive specified in {1} file will not be imported. The directive must appear at the top of each Razor cshtml file.</value>
   </data>
+  <data name="PageDirectiveMustExistAtTheTopOfFile" xml:space="preserve">
+    <value>The '@{0}' directive must precede all other elements defined in a Razor file.</value>
```
pranavkm: Nice.

PR #2186, opened 2018-03-19
Add an experimental Language Version

rynowak:
Adding a secret language version for Blazor and other experiments.


src/Microsoft.AspNetCore.Razor.Language/RazorLanguageVersion.cs
```diff
@@ -19,6 +19,8 @@ public sealed class RazorLanguageVersion : IEquatable<RazorLanguageVersion>, ICo
 
         public static readonly RazorLanguageVersion Latest = Version_2_1;
 
+        public static readonly RazorLanguageVersion Experimental = new RazorLanguageVersion(9999, 9999);
```
NTaylorMullen: SUggest 1337

rynowak: sure

Issue #2109, opened 2018-02-24
Handle unreadable project items at design time.


NTaylorMullen:
This issue is inspired from [this](https://github.com/aspnet/Razor/issues/2106). The only difference being for us to not take a tactical fix approach for handling unreadable project items at design time.

To expand, we work around a lot of bad cases in our `DefaultProjectItem.Read` VS15.7+; however, in the case that we we're trying to read a file we don't have access to or an underlying OS file operation occurs as we're trying to read (not 100% sure this is possible, but better to be protective) we may get exceptions that can crash VS. We should handle these cases better.

ajaybhargavb:
79e744f691ac21cb007cb47c3b8786b9fef08d4b

PR #2192, opened 2018-03-19
Handle unreadable import items in design time

ajaybhargavb:
Issue #2109 

I've verified this doesn't crash VS.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace Microsoft.AspNetCore.Razor.Language
@@ -65,7 +66,7 @@ namespace Microsoft.AspNetCore.Razor.Language
 
             var importFeature = GetRequiredFeature<IImportProjectFeature>();
             var importItems = importFeature.GetImports(projectItem);
-            var importSourceDocuments = importItems.Select(ConvertToSourceDocument);
+            var importSourceDocuments = importItems.Select(i => ConvertToSourceDocument(i));
 
             var parserOptions = GetRequiredFeature<IRazorParserOptionsFactoryProjectFeature>().Create(ConfigureParserOptions);
             var codeGenerationOptions = GetRequiredFeature<IRazorCodeGenerationOptionsFactoryProjectFeature>().Create(ConfigureCodeGenerationOptions);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -65,7 +66,7 @@ protected override RazorCodeDocument CreateCodeDocumentCore(RazorProjectItem pro
 
             var importFeature = GetRequiredFeature<IImportProjectFeature>();
             var importItems = importFeature.GetImports(projectItem);
-            var importSourceDocuments = importItems.Select(ConvertToSourceDocument);
+            var importSourceDocuments = importItems.Select(i => ConvertToSourceDocument(i));
```
rynowak: revert this YO

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -138,12 +138,20 @@ private void ConfigureDesignTimeCodeGenerationOptions(RazorCodeGenerationOptions
         }
 
         // Internal for testing
-        internal static RazorSourceDocument ConvertToSourceDocument(RazorProjectItem importItem)
+        internal static RazorSourceDocument ConvertToSourceDocument(RazorProjectItem importItem, bool designTime = false)
```
rynowak: can we call this something else, like `suppressExceptions`

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
ajaybhargavb: Nope. It won't accept the new signature.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -138,12 +138,20 @@ private void ConfigureDesignTimeCodeGenerationOptions(RazorCodeGenerationOptions
...
-                return RazorSourceDocument.ReadFrom(importItem);
+                try
+                {
+                    // Normal import, has file paths, content etc.
+                    return RazorSourceDocument.ReadFrom(importItem);
+                }
+                catch (IOException) when (designTime)
+                {
+                    // Something happened when trying to read the item from disk.
+                    // Catch the exception so we don't crash the editor.
```
NTaylorMullen: Should log another issue to report this as an error to the user, otherwise this is a silent error that the user will have a verrrry hard time diagnosing. Given how we open files reporting an error could suffice for design time and runtime; reason being we'll typically only fail to open a file if it's permissions are whack (something a user should fix).

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
NTaylorMullen: O_O oooo default params 😢

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
rynowak: Yep. We'd need to add UI for that, so it goes to 15.8. We'll do it when it's a priority

ajaybhargavb: I'm going to file a broader issue which is basically "Add a way to flow Razor editor parsing errors to the IDE" because we don't have that.

ajaybhargavb: Filed https://github.com/aspnet/Razor/issues/2193

PR #2194, opened 2018-03-19
Stop expressly testing on netcoreapp2.0

pranavkm:


test/Microsoft.AspNetCore.Razor.Tools.Test/Microsoft.AspNetCore.Razor.Tools.Test.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
pranavkm: @ajaybhargavb We don't *need* to run this test on `netcoreapp2.0`. It speeds up things a bit if we can run it in the same group as the rest.

Issue #2173, opened 2018-03-15
IntelliSense is lost on empty line immediately following directives


NTaylorMullen:
Issue:
```
@page
|
```
don't type the pipe but try and invoke intelliSense at the pipe.

You'd expect for HTML IntelliSense but in actuality you don't get IntelliSense. This seems to be due to our `SpanKind.None` addition; basically we mark the end of `@page` as `SpanKind.None` (expected) but that has an adverse effect of turning off IntelliSense on the line following it (if there's no content). We should be adding a marker symbol after our `SpanKind.None` to indicate to the editor that we expect HTML at that location.

This was originally reported in TFS: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/584082

Putting this in the VS15.7 milestone to correspond with either preview3 or 4.

ajaybhargavb:
9dc7c65205ef1e0a85c39c0fb067acbdcd1bf0b4

PR #2196, opened 2018-03-20
Fix: Missing Html intellisense after directives

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2173

We were outputting the newline after a directive as `SpanKind.None` which was preventing intellisense from getting invoked.

Fixed this to have the following behavior,
Html intellisense **should work** at the pipe in the following cases:
```c#
@model DateTime   ;  |
```
```c#
@model DateTime
|
```
```c#
@page |
```
```c#
@page
|
```

Html intellisense **shouldn't work** at the pipe in the following cases:
```c#
@model DateTime |
```
``` c#
@model DateTime   |  ;   
```

I've verified all of the above cases work with this code in VS. Also, updated the unit tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -1779,7 +1779,10 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                         }
 
                         Span.ChunkGenerator = SpanChunkGenerator.Null;
-                        Output(SpanKindInternal.None, AcceptedCharactersInternal.WhiteSpace);
+
+                        // This should contain the optional whitespace after the optional semicolon and the new line.
+                        // Output as Markup as we want intellisense here.
+                        Output(SpanKindInternal.Markup, AcceptedCharactersInternal.WhiteSpace);
                         break;
                     case DirectiveKind.RazorBlock:
                         AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
@@ -36,7 +36,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                         Factory.MetaCode("custom").Accepts(AcceptedCharactersInternal.None),
                         Factory.Span(SpanKindInternal.Code, " ", markup: false).Accepts(AcceptedCharactersInternal.WhiteSpace),
                         Factory.Span(SpanKindInternal.Code, "System.Text.Encoding.ASCIIEncoding", markup: false).AsDirectiveToken(descriptor.Tokens[0]),
-                        Factory.Span(SpanKindInternal.None, Environment.NewLine, markup: false).Accepts(AcceptedCharactersInternal.WhiteSpace)),
+                        Factory.Span(SpanKindInternal.Markup, Environment.NewLine, markup: false).Accepts(AcceptedCharactersInternal.WhiteSpace)),
                     Factory.EmptyHtml(),
                     new DirectiveBlock(new DirectiveChunkGenerator(descriptor),
                         Factory.CodeTransition(),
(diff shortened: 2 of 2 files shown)


NTaylorMullen:
> @model DateTime   ;  |



> @page |

HTML IntelliSense shouldn't work for these. We don't support anything else on directive lines. Just output an HTML marker symbol after the `SpanKind.None` piece.

ajaybhargavb:
I thought about that. But I disagree for atleast this case `@model DateTime ; |`.
We allow semicolon for a reason. It indicates the end of that line. It means we should allow whatever we would allow on the next line. The `@page |` case is debatable.

NTaylorMullen:
> We allow semicolon for a reason. It indicates the end of that line. It means we should allow whatever we would allow on the next line. The @page | case is debatable.

We explicitly disable any content after a directive by settings its `ChunkGenerator` to `null`. To support anything after a directives optional semicolon would be a pretty big language change which I'm not sure I want to commit to supporting. If you can convince @rynowak that it's a good idea i'll be open to changing my mind 😉

ajaybhargavb:
For the `@page |` case, look at [this](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs#L1660-L1680). We have specific logic in place to provide html intellisense for string type directive tokens. That is what prompted me to think this needs to allow intellisense. Maybe I am wrong to think that. I assume you have more context because you wrote it.

NTaylorMullen:
> For the @page | case, look at this. We have specific logic in place to provide html intellisense for string type directive tokens. That is what prompted me to think this needs to allow intellisense. Maybe I am wrong to think that. I assume you have more context because you wrote it.

Ah, so I think you misinterpreted what that piece was for. That piece you linked to is understanding the tokens. In the string token case we use the HTML intellisense inside of a string because it has the experience we were looking for (auto-completing quotes, coloring etc.). However, outside of tokens in the directive line I think we should provide as little (none) intellisense as possible to not give the wrong expectations for what can go where.

ajaybhargavb:
>  That piece you linked to is understanding the tokens.

But this is the line that captures the whitespace after `@page`. So the code below to output none kind doesn't matter.

NTaylorMullen:
> But this is the line that captures the whitespace after @page. So the code below to output none kind doesn't matter.

Ahhhh, ok so totally get what you're going at now and I totally agree now that `@page |` should still get IntelliSense; that's because that's prepping the IntelliSense to consume the route token. However, I still don't think that `@model DateTime ; |` should have IntelliSense at the end there. Really glad you poked with more questions 👍

ajaybhargavb:
>Just output an HTML marker symbol after the SpanKind.None piece.

Also, I don't see a way to add HTML marker symbol from the CSharpCodeParser. Calling `AddMarkerSymbolIfNecessary` leads to addition of a CSharp marker symbol.

NTaylorMullen:
> Also, I don't see a way to add HTML marker symbol from the CSharpCodeParser. Calling AddMarkerSymbolIfNecessary leads to addition of a CSharp marker symbol.

Can just construct one manually. The reason why your marker symbol results in a csharp marker symbol is because all of our directive code is a detail of the `CSharpCodeParser` (its `CreateMarkerSymbol` returns C#).

ajaybhargavb:
> Can just construct one manually. 

That isn't the problem. The `Accept` methods expect a `CSharpSymbolType`. Not `HtmlSymbolType.Unknown`.

NTaylorMullen:
> That isn't the problem. The Accept methods expect a CSharpSymbolType. Not HtmlSymbolType.Unknown.

Ah, ya that would be a problem wouldn't it. Ughhh our parser sucks. If you add a C# marker symbol but output the span as SpandKind.Markup what happens?

ajaybhargavb:
> If you add a C# marker symbol but output the span as SpandKind.Markup what happens?

Doesn't work. No intellisense on the symbol. 😞 
IMO allowing intellisense after the semicolon isn't a big deal. At the very least we should check-in this fix for 15.7 and file a bug for that one particular case.

NTaylorMullen:
> IMO allowing intellisense after the semicolon isn't a big deal. At the very least we should check-in this fix for 15.7 and file a bug for that one particular case.

ya, go for it. I'm just sad.

ajaybhargavb:
Filed https://github.com/aspnet/Razor/issues/2197

Issue #2056, opened 2018-02-12
Update PipeOptions once https://github.com/dotnet/corefx/issues/25427 is fixed


rynowak:
We will need to rely on the new setting introduced by https://github.com/dotnet/corefx/issues/25427 to provide a secure experience with the build server.

It's likely we'll need to use the constant value of this new enum to pass the correct value.

rynowak:
This should be unblocked now

ajaybhargavb:
Just checked and we have picked up a build of corefx that contains this fix. I'll work on this now.

ajaybhargavb:
This is now blocked on https://github.com/dotnet/corefx/issues/28371.

Eilon:
Is this still going to happen for Preview2? Is this a must-have for Preview2?

ajaybhargavb:
Yes. This is a must-have for preview2. I am constantly checking for a new release/2.1 corefx build with the fix.

Eilon:
Got it, thanks!

natemcmaster:
We have a prodcon build now. Microsoft.NETCore.App 2.1.0-preview2-26403-06.

See https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.1/build.xml

ajaybhargavb:
b1007744b00ce094af080602ee26d3887a786d56

PR #2204, opened 2018-03-21
Added PipeOptions.CurrentUserOnly option to the named pipe streams

ajaybhargavb:
#2056


src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -74,6 +79,18 @@ protected virtual void Dispose(bool disposing)
             }
         }
 
+        private static PipeOptions GetPipeOptions()
+        {
+            var options = PipeOptions.Asynchronous;
+
+            if (Enum.IsDefined(typeof(PipeOptions), PipeOptionCurrentUserOnly))
```
pranavkm: Does it complain if you add the flag all the time (regardless of the platform)?

ajaybhargavb: Yup. Throws ArgumentOutOfRange - invalid flag

rynowak:
The code changes here look reasonable, but 👀 for us to actually understand all of the issues.

ajaybhargavb:

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -29,6 +34,10 @@ public abstract class DotNetToolTask : ToolTask
         // Specifies whether we should fallback to in-process execution if server execution fails.
         public bool ForceServer { get; set; }
 
+        // Specifies whether server execution is allowed when PipeOptions.CurrentUserOnly is not available.
+        // For testing purposes only.
+        public bool AllowInsecureServer { get; set; }
```
ajaybhargavb: This is something temporary so we can still run our tests on the build server. This can go away once we start running our tests on netcoreapp2.1.

ajaybhargavb: Also name suggestions are welcome.

ajaybhargavb:
🆙 📅  No longer spinning up a server if `PipeOptions.CurrentUserOnly` isn't available.

src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
rynowak: I would avoid the word insecure. I would say something like `SuppressCurrentUserOnlyPipeOptions`

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildIntegrationTestBase.cs
```diff
@@ -72,6 +72,7 @@ internal static string TargetFramework
             if (!suppressBuildServer)
             {
                 buildArgumentList.Add($"/p:_RazorBuildServerPipeName={buildServerPipeName ?? BuildServer.PipeName}");
+                buildArgumentList.Add($"/p:_RazorSuppressCurrentUserOnlyPipeOptions=true");
```
pranavkm: Can we add a comment here that says this is because our tests target `netcoreapp2.0`? Also can we file an issue that says we should upgrade tests to target `netcoreapp2.1`?

ajaybhargavb: Will do.

ajaybhargavb:
🆙 📅 
@natemcmaster @pranavkm

natemcmaster:
FYI - universe is also being updated. https://github.com/aspnet/Universe/pull/1032. Once that goes through, we can run update deps on all repos to align to latest prodcon bits

ajaybhargavb:
Okay. I'll wait then.

natemcmaster:
:shipit: there's no reason to way. You're okay to merge.

PR #2206, opened 2018-03-21
Remove EnableDefaultCompiledViewAssemblyLoadBehavior

pranavkm:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/build/netstandard2.0/Microsoft.AspNetCore.Mvc.Razor.Extensions.targets
```diff
@@ -12,12 +12,16 @@
...
-    <ProvideApplicationPartFactoryAttributeTypeName Condition="'$(EnableDefaultCompiledViewAssemblyLoadBehavior)'=='false'">Microsoft.AspNetCore.Mvc.ApplicationParts.NullApplicationPartFactory, Microsoft.AspNetCore.Mvc.Core</ProvideApplicationPartFactoryAttributeTypeName>
-
-    <ProvideApplicationPartFactoryAttributeTypeName Condition="'$(EnableDefaultCompiledViewAssemblyLoadBehavior)'=='true'">Microsoft.AspNetCore.Mvc.ApplicationParts.CompiledRazorAssemblyApplicationPartFactory, Microsoft.AspNetCore.Mvc.Razor</ProvideApplicationPartFactoryAttributeTypeName>
+    <ProvideApplicationPartFactoryAttributeTypeName Condition="'$(UseDeferredCompiledRazorAssemblyApplicationPartFactory)' == 'true'">Microsoft.AspNetCore.Mvc.ApplicationParts.DeferredCompiledRazorAssemblyApplicationPartFactory, Microsoft.AspNetCore.Mvc.Razor</ProvideApplicationPartFactoryAttributeTypeName>
+    <ProvideApplicationPartFactoryAttributeTypeName Condition="'$(ProvideApplicationPartFactoryAttributeTypeName)' == ''">Microsoft.AspNetCore.Mvc.ApplicationParts.CompiledRazorAssemblyApplicationPartFactory, Microsoft.AspNetCore.Mvc.Razor</ProvideApplicationPartFactoryAttributeTypeName>
```
rynowak: My naming feedback is that `UseDeferredCompiledRazorAssemblyApplicationPartFactory` shouldn't be part of the SDK. The caller can specify the type name in their project and that sounds great.

rynowak: @javiercn and I discussed this a bit more, and while I think he doesn't agree with me, he challenged me to provide more detail about what I think.  I least understand my feelings and concerns now.

I don't believe in `UseDeferredCompiledRazorAssemblyApplicationPartFactory` as a first class feature. 

**I think if you're operating in a mode where you care about the part factory, then you need to know and control all of the details about what's going on.**

To me that includes:
- specifying which factory you're using by placing its type name in MSBuild somewhere
- calling into the factory (with its type name) to *activate* your part somewhere in your C# code

I believe that having a part that it *optional* and gets activated dynamically is (by definition) a case where you "care about the part factory". I don't want to try and hide these details, I think that trying to cover this up or make it simpler is going to have the opposite effect.

-----

I think that if you take out `UseDeferredCompiledRazorAssemblyApplicationPartFactory` and specify the part factory type in the project file, we're in a better place overall. Along with the rest of the changes that are already out for PR, I think this would make this feature set complete.

pranavkm: If we take the option out (which I'm all for), can we 

a) Leave `NullApplicationPartFactory` as is. 
b) Have IdentityUI call the `CompiledRazorAssemblyApplicationPartFactory` to populate things. The current design requires down casting, so you know what kind of part it's meant to produce

Here's what things looks like for Identity:

```xml
// Identity.UI.csproj
<ProvideApplicationPartFactoryAttributeTypeName>Microsoft.AspNetCore.Mvc.ApplicationParts.NullApplicationPartFactory, Microsoft.AspNetCore.Mvc</ProvideApplicationPartFactoryAttributeTypeName>
```

```C#
// AddDefaultIdentityUI()
{
    builder.ConfigureApplicationPartManager(partManager =>
    {
        var assemblies = RelatedAssembly.GetRelatedAssemblies();
        var parts = assemblies.Select(CompiledRazorAssemblyApplicationPartFactory.GetDefaultApplicationParts);
        partManager.Parts.Add(parts);
    });
}
```

We don't need a first class type if we don't have a first class Sdk feature.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.GenerateAssemblyInfo.targets
```diff
@@ -27,15 +27,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
       GetRazorAssemblyAttributes;
       CreateRazorGeneratedAssemblyInfoInputsCacheFile;
+      CoreGenerateRazorAssemblyInfo
     </GenerateRazorAssemblyInfoDependsOn>
   </PropertyGroup>
 
   <Target
     Name="GenerateRazorAssemblyInfo"
     DependsOnTargets="$(GenerateRazorAssemblyInfoDependsOn)"
+    Condition="'$(GenerateRazorTargetAssemblyInfo)'=='true'">
```
pranavkm: This was buggy - the condition gets evaluated before DependsOnTargets are run.

pranavkm:
🆙 📅

Issue #1255, opened 2017-04-25
Parameter info is not displayed automatically after typing opening parens of method invocation in Razor expression


DamianEdwards:
When typing a Razor expression that invokes a method in the Razor editor, the parameter info is not being displayed automatically after the opening parenthesis is typed. The closing parenthesis is auto-completed however with the cursor placed inside. Manually invoking Parameter Info from the Edit menu (or KB shortcuts) shows it.

DamianEdwards:
@NTaylorMullen @rynowak

NTaylorMullen:
To note: This isn't a regression. The tooling has always been this way. There's a TFS issue somewhere in the deps of all the other tooling issues on this guy 😄

DamianEdwards:
That sucks 😞 
We should dig it up and try to get it fixed.

NTaylorMullen:
Totally still an issue, nuking the investigate label. Note: Re-typing the method invocation displays the method IntelliSense.

rynowak:
Looking at this, it really seems odd that it's parked in 15.8. Do we know the actual cause of this? Seems worth figuring out at least what the fix would be.

DamianEdwards:
Ya, this is still super annoying

rynowak:
Taylor and I looked into this, and what's happening is that Roslyn starts preparing to show signature help, but then another edit gets applied, and the caret is moved **out** of the c# buffer.

This is the callstack that leads to the problem
```
>	Microsoft.CodeAnalysis.EditorFeatures.dll!Microsoft.CodeAnalysis.Editor.Implementation.IntelliSense.SignatureHelp.Controller.Retrigger() Line 140	C#
 	Microsoft.CodeAnalysis.EditorFeatures.dll!Microsoft.CodeAnalysis.Editor.Implementation.IntelliSense.SignatureHelp.Controller.OnTextViewBufferPostChanged(object sender, System.EventArgs args) Line 11	C#
 	Microsoft.VisualStudio.Platform.VSEditor.dll!Microsoft.VisualStudio.Text.Utilities.GuardedOperations.RaiseEvent(object sender, System.EventHandler eventHandlers)	Unknown
 	Microsoft.VisualStudio.Platform.VSEditor.dll!Microsoft.VisualStudio.Text.Implementation.BufferGroup.RaiseEvents()	Unknown
[...]

PR #2209, opened 2018-03-22
Add partial parsing for parenthesis

rynowak:
This is part of a fix for #1255 - this change enables signature help in
implicit expressions by improving the partial parsing.

Specifically we can now partial parse the addition of parenthesis at the
end of an implicit expression, which is a common case for signature
help.

### DIFF
--- a/test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
+++ b/test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
@@ -420,6 +420,67 @@ namespace Microsoft.VisualStudio.Editor.Razor
         }
 
         [ForegroundFact]
+        public async Task ImplicitExpression_AcceptsParenthesisAtEnd_SingleEdit()
+        {
+            // Arrange
+            var factory = new SpanFactory();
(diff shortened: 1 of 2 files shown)


test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioRazorParserIntegrationTest.cs
```diff
@@ -419,6 +419,67 @@ void ApplyAndVerifyPartialChange(Action applyEdit, string expectedCode)
...
+                        factory.Markup("foo "),
+                        new ExpressionBlock(
+                            factory.CodeTransition(),
+                            factory.Code("foo()")
+                                   .AsImplicitExpression(CSharpCodeParser.DefaultKeywords)
+                                   .Accepts(AcceptedCharactersInternal.NonWhiteSpace)),
+                        factory.Markup(" bar")));
+            }
+        }
+
```
rynowak: @ajaybhargavb @NTaylorMullen - do you think these are enough tests? What other cases would you suggest that would be meaningful. I'm not very familiar with this area.

rynowak:
After an offline discussion with @NTaylorMullen - this isn't a very good quick fix.

This handles cases where you don't have any parameters, but is totally broken when you need to type something in the parenthesis. Given that this only fixes a very limited set of cases, I'm going to close out the PR for now while we consider our approach.

Issue #2216, opened 2018-03-23
Building on Visual Studio doesn't reuse existing build server


ajaybhargavb:
This is because we use `AppDomain.CurrentDomain.BaseDirectory` to compute the default pipe name. We should be using path to rzc instead.

ajaybhargavb:
@mkArtakMSFT we need to fix this for preview2. Assigning this to myself.

ajaybhargavb:
7bc54b99e084ec15880ca35aab23a574aec0775a

PR #2217, opened 2018-03-23
Use tool assembly path for computing default pipe name

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2216


src/Microsoft.AspNetCore.Razor.Tools/PipeName.cs
```diff
@@ -23,11 +23,16 @@ internal static class PipeName
         //
         // This is similar to (and based on) the code used by Roslyn in VBCSCompiler:
         // https://github.com/dotnet/roslyn/blob/c273b6a9f19570a344c274ae89185b3a2b64d93d/src/Compilers/Shared/BuildServerConnection.cs#L528
-        public static string ComputeDefault()
+        public static string ComputeDefault(string toolDirectory = null)
         {
+            if (string.IsNullOrEmpty(toolDirectory))
+            {
```
NTaylorMullen: Leave a comment describing what this case covers.

Issue #2184, opened 2018-03-19
Flaky test: EnumerateItems_DiscoversAllCshtmlFiles 


ryanbrandenburg:
This test [failed](http://aspnetci/viewLog.html?buildId=422519&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR) with the following error:
```
System.ArgumentOutOfRangeException : startIndex cannot be larger than length of string.
Parameter name: startIndex
   at System.String.Substring(Int32 startIndex, Int32 length)
   at Microsoft.AspNetCore.Razor.Language.DefaultRazorProjectFileSystem.<>c__DisplayClass4_0.<EnumerateItems>b__0(FileInfo file) in /_/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectFileSystem.cs:line 39
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source, Int32& length)
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.OrderedEnumerable`1.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
```

Future instances of this failure will show up [here](http://aspnetci/httpAuth/app/rest/builds?locator=tag:RazorIndexRange,sinceDate:201710010T000000-0001&fields=build(startDate,buildTypeId,status,branchName,webUrl,tags)).

mkArtakMSFT:
@NTaylorMullen, can you please handle this? Looks like a real issue.

NTaylorMullen:
da255ff9db89ae98dbaae23d54a264bb50df2817

PR #2218, opened 2018-03-24
Make `TestProject.GetProjectDirectory` more thorough when finding project directory.

NTaylorMullen:
**Big Note:** I was unable to reproduce this issue and after digging through the code really can't see how things could be going wrong other than in our test project directory creation. Basically, in this PR i make our `GetProjectDirectory` a little more reliable in unknown scenarios and added some logic to help us diagnose future errors if they arise.

- Relaxed some restrictions to find project directory
- Added some logic to make future debugging/test failures more diagnosable.

#2184

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestProject.cs
+++ b/test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestProject.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.IO;
-using System.Reflection;
+using Microsoft.AspNetCore.Testing;
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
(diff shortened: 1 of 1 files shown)


test/Microsoft.AspNetCore.Razor.Test.Common/Language/TestProject.cs
```diff
@@ -3,23 +3,41 @@
 
 using System;
 using System.IO;
-using System.Reflection;
 
 namespace Microsoft.AspNetCore.Razor.Language
 {
     public static class TestProject
     {
         public static string GetProjectDirectory(Type type)
```
pranavkm: Can we use `TestPathUtilities.GetSolutionRootDirectory`

rynowak: Is there an API we could use that doesn't have utilities in the name? :trollface:

NTaylorMullen: Didn't know that existed, sounds like a great alternative!

Issue #2215, opened 2018-03-23
Missing pdb inside Microsoft.AspNetCore.Razor.Design.symbols.nupkg


ajaybhargavb:
For some reason, the symbols nupkg we create for this assembly doesn't contain any pdb files.

cc @pranavkm

pranavkm:
@mkArtakMSFT we'll need this for `preview2` since that's a prodcon shipping requirement.

pranavkm:
Moving this to rc. See https://github.com/aspnet/Razor/pull/2221#issuecomment-376585225

ajaybhargavb:
dc997821ccd7b74c658b845aabff9a6e40a3bcd5

PR #2221, opened 2018-03-26
Include symbols for Tasks and Tools projects in Razor.Design

pranavkm:
Fixes https://github.com/aspnet/Razor/issues/2215

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
+++ b/src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
@@ -57,24 +57,34 @@
     -->
     <MSBuild Projects="@(TaskProject)" />
 
-    <MSBuild Projects="@(TaskProject)" Properties="TargetFramework=net46" Targets="GetTargetPath">
+    <MSBuild Projects="@(TaskProject)" Properties="TargetFramework=net46" Targets="BuiltProjectOutputGroup">
       <Output TaskParameter="TargetOutputs" ItemName="TaskAssemblyNet46" />
     </MSBuild>
 
-    <MSBuild Projects="@(TaskProject)" Properties="TargetFramework=netstandard2.0" Targets="GetTargetPath">
+    <MSBuild Projects="@(TaskProject)" Properties="TargetFramework=net46" Targets="DebugSymbolsProjectOutputGroup">
+      <Output TaskParameter="TargetOutputs" ItemName="TaskSymbolNet46" />
+    </MSBuild>
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -57,24 +57,34 @@
...
       <Output TaskParameter="CopiedFiles" ItemName="FileWrites" />
     </Copy>
-    <Copy SourceFiles="@(TaskAssemblyNetStandard)" DestinationFolder="$(OutputPath)\tasks\netstandard2.0\">
+    <Copy SourceFiles="@(TaskAssemblyNetStandard);@(TaskSymbolNetStandard)" DestinationFolder="$(OutputPath)\tasks\netstandard2.0\">
       <Output TaskParameter="CopiedFiles" ItemName="FileWrites" />
     </Copy>
 
     <ItemGroup>
       <None Include="@(TaskAssemblyNet46)" PackagePath="tasks\net46\" Pack="true" />
+      <None Include="@(TaskSymbolNet46)" PackagePath="tasks\net46\" Pack="true" />
```
pranavkm: @rynowak we couldn't find any obvious way to produce a symbol nupkg, specify a package path to an item while excluding it only in the main assembly. The result is that we have pdbs sitting in the Design package and it's symbol nupkg. Seems like a small price to pay for compliance. We could open a separate issue to address this in rc if you'd like.

rynowak: Sorry what's the implication of this? I'm not in charge of decisions like this.

ajaybhargavb: The implication is after this change we'll have pdbs sitting in both `Microsoft.AspNetCore.Razor.Design.nupkg` and `Microsoft.AspNetCore.Razor.Design.Symbols.nupkg`.

pranavkm:
Spoke to @Eilon - he'd rather not us ship the pdbs in the main nupkg. He's also 🆗  with us not having symbols for preview2 as long as we do the right thing for rc. Closing this, but I'll leave the branch alone. @ajaybhargavb re-assigning the issue to you so you could take a stab at doing the right thing ™️

rynowak:
Should we just 'pack' the tasks package as non-shipping? That sounds a like a great solution to this.

pranavkm:
Not sure if that works for how they do symbol publishing today (which is look for a symbol package per regular nupkg).

Issue #1813, opened 2017-12-11
Consume new C# completion APIs to avoid loading Razor binaries in non-Razor scenarios


NTaylorMullen:

NTaylorMullen:
Adding blocked label until the APIs are available.

NTaylorMullen:
https://github.com/dotnet/roslyn/issues/18650 is the issue this is blocked on FYI @mkArtakMSFT

NTaylorMullen:
Double marked this issue for cost because it's still a slight unknown

mkArtakMSFT:
Let's go with most optimistic cost, rather than adding two, as that will mess up my calculations

NTaylorMullen:
aec88e3eba339f91f75a4d1aa68ff7736aee34ee

Issue #1743, opened 2017-10-25
Auto-completing a Razor directive with space results in two spaces being added.


NTaylorMullen:
![](https://i.imgur.com/HtThKy3.gif)

ajaybhargavb:
Looks like this applies to not just directives but any CSharp in Razor like `@DateTime` and such. I'm investigating.

ajaybhargavb:
I investigated this and it looks like Razor doesn't do anything special to affect this behavior. Marking this as external and filed https://devdiv.visualstudio.com/DevDiv/_workitems/edit/539940.

NTaylorMullen:
Marking this as working because consuming the new completion APIs fixes the problem.

NTaylorMullen:
aec88e3eba339f91f75a4d1aa68ff7736aee34ee

PR #2227, opened 2018-03-28
Change Razor directive completions to use new completions API.

NTaylorMullen:
- Kept the same behavior as we previously had with Razor directive completions.
- Attempted adding additional functionalities such as lighting up Razor directive completion when completion was invoked on top of Razor directives (non-C#) but ran into issues involving the core HTML editor not consuming the new completion APIs yet. That's something we'll have to re-visit once they move to the new completion APIs.
- Removed a ton of code that was previously needed to make old-school directive completion work.
- Added tests to validate all aspects of new completion APIs.

#1743
#1813

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
@@ -13,9 +13,9 @@ using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Completion;
-using Microsoft.CodeAnalysis.Razor;
 using Microsoft.CodeAnalysis.Tags;
 using Microsoft.CodeAnalysis.Text;
+using Microsoft.VisualStudio.Language.Intellisense.AsyncCompletion;
 using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Projection;
 
@@ -36,16 +36,33 @@ namespace Microsoft.VisualStudio.Editor.Razor
             CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
         };
         private readonly Lazy<RazorCodeDocumentProvider> _codeDocumentProvider;
+        private readonly IAsyncCompletionBroker _asyncCompletionBroker;
+        private readonly RazorTextBufferProvider _textBufferProvider;
 
         [ImportingConstructor]
-        public RazorDirectiveCompletionProvider([Import(typeof(RazorCodeDocumentProvider))] Lazy<RazorCodeDocumentProvider> codeDocumentProvider)
+        public RazorDirectiveCompletionProvider(
+            [Import(typeof(RazorCodeDocumentProvider))] Lazy<RazorCodeDocumentProvider> codeDocumentProvider,
+            IAsyncCompletionBroker asyncCompletionBroker,
+            RazorTextBufferProvider textBufferProvider)
         {
             if (codeDocumentProvider == null)
             {
                 throw new ArgumentNullException(nameof(codeDocumentProvider));
             }
 
+            if (asyncCompletionBroker == null)
+            {
+                throw new ArgumentNullException(nameof(asyncCompletionBroker));
+            }
+
+            if (textBufferProvider == null)
+            {
+                throw new ArgumentNullException(nameof(textBufferProvider));
+            }
+
             _codeDocumentProvider = codeDocumentProvider;
+            _asyncCompletionBroker = asyncCompletionBroker;
+            _textBufferProvider = textBufferProvider;
         }
 
         public override Task<CompletionDescription> GetDescriptionAsync(Document document, CompletionItem item, CancellationToken cancellationToken)
@@ -83,6 +100,13 @@ namespace Microsoft.VisualStudio.Editor.Razor
                 !context.Document.FilePath.EndsWith(".cshtml", StringComparison.OrdinalIgnoreCase))
             {
                 // Not a Razor file.
+                return Task.CompletedTask;
+            }
+
+            if (!_textBufferProvider.TryGetFromDocument(context.Document, out var textBuffer) ||
+                !_asyncCompletionBroker.IsCompletionSupported(textBuffer.ContentType))
+            {
+                // Completion is not supported.
                 return Task.CompletedTask;
             }
 
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSource.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSource.cs
@@ -1,0 +1,167 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Microsoft.CodeAnalysis.Razor;
+using Microsoft.VisualStudio.Core.Imaging;
+using Microsoft.VisualStudio.Imaging;
+using Microsoft.VisualStudio.Language.Intellisense.AsyncCompletion;
+using Microsoft.VisualStudio.Language.Intellisense.AsyncCompletion.Data;
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Adornments;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class RazorDirectiveCompletionSource : IAsyncCompletionSource
+    {
+        // Internal for testing
+        internal static readonly object DescriptionKey = new object();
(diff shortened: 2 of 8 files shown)


NTaylorMullen:
We might also want to get a Razor directive specific glyph. I'm re-using the intrinsic glyph for Razor directives 😄

rynowak:
This is amazing. It feels like we're living in the present!

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.Text.Editor;
+using Microsoft.VisualStudio.Utilities;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    [System.Composition.Shared]
+    [Export(typeof(IAsyncCompletionSourceProvider))]
+    [Name("Razor directive completion provider.")]
+    [ContentType(RazorLanguage.ContentType)]
```
rynowak: Should this be the CORE content type if this is for 15.8?

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            if (foregroundDispatcher == null)
+            {
+                throw new ArgumentNullException(nameof(foregroundDispatcher));
+            }
+
+            _foregroundDispatcher = foregroundDispatcher;
+        }
+
+        public IAsyncCompletionSource GetOrCreate(ITextView textView)
```
rynowak: This makes it sound like we're expected to cache this. Is there a reason why we should/should not?

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                // Razor buffer hasn't been associated with the text view yet.
+                return null;
+            }
+
+            if (!razorBuffer.Properties.TryGetProperty(typeof(VisualStudioRazorParser), out VisualStudioRazorParser parser))
+            {
+                // Parser hasn't been associated with the text buffer yet.
+                return null;
+            }
```
rynowak: Is this a known possible case or just a guard?

test/Microsoft.VisualStudio.Editor.Razor.Test/RazorDirectiveCompletionSourceTest.cs
```diff
@@ -0,0 +1,368 @@
...
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Microsoft.VisualStudio.Language.Intellisense.AsyncCompletion;
+using Microsoft.VisualStudio.Text;
+using Moq;
+using Xunit;
+using Span = Microsoft.VisualStudio.Text.Span;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public class RazorDirectiveCompletionSourceTest : ForegroundDispatcherTestBase
```
rynowak: I'm kinda skipping these because I assume it's pretty similar to what we had before yes?

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
NTaylorMullen: It will be once VS Editor + Roslyn move to VS 15.8. For now these completion changes require private builds of their bits which target 15.7 😢

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
NTaylorMullen: 👍 should definitely cache.

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
NTaylorMullen: It guards against pre-core Razor. This wont be an issue once things move over to 15.8.

test/Microsoft.VisualStudio.Editor.Razor.Test/RazorDirectiveCompletionSourceTest.cs
NTaylorMullen: Mostly all new actually. The new APIs are now more testable than the old ones Roslyn provided so I expanded our coverage.

NTaylorMullen:
🆙 📅

NTaylorMullen:
Alright, gonna leave this in a PR form until the new VS editor completion bits make it into one of the various VS branches.

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSource.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            _foregroundDispatcher.AssertBackgroundThread();
+
+            var syntaxTree = _parser.CodeDocument?.GetSyntaxTree();
+            if (!AtDirectiveCompletionPoint(syntaxTree, triggerLocation))
+            {
+                return Task.FromResult(new CompletionContext(ImmutableArray<CompletionItem>.Empty));
+            }
+
+            var completionItems = GetCompletionItems(syntaxTree);
```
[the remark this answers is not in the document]
NTaylorMullen: We take the trigger location because we want to operate on where the user wanted completion information from. In the case of no leading `@` we do not want to provide completions because the user is outside of a valid Razor directive context.

AmadeusW:
@NTaylorMullen see https://github.com/dotnet/roslyn/issues/19256 for discussion about grouping of completion items

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSource.cs
```diff
@@ -0,0 +1,165 @@
...
+            _foregroundDispatcher = foregroundDispatcher;
+        }
+
+        public Task<CompletionContext> GetCompletionContextAsync(
+            CompletionTrigger trigger, 
+            SnapshotPoint triggerLocation, 
+            SnapshotSpan applicableSpan, 
+            CancellationToken token)
+        {
+            _foregroundDispatcher.AssertBackgroundThread();
```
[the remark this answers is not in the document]
AmadeusW: sounds good

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSource.cs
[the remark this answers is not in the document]
AmadeusW: I see - after we skip the `if (!AtDirectiveCompletionPoint(syntaxTree, triggerLocation))` branch, it is guaranteed that we are in a correct spot to return the directives. The directives don't depend on the location in code, so we don't need to pass neither `triggerLocation` nor `applicableSpan` to `GetCompletionItems`. Is this correct?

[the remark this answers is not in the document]
NTaylorMullen: Correct.

NTaylorMullen:
Updated the bits in this PR to target a 15.8 build that's compatible with `d15.8stgide`. Tried out the end-to-end with Roslyn + new editor completion bits and found some issues in completion where anything more than a single character after an `@` results in our directive completion providers not being called (no directive completion).

Talked with @AmadeusW offline, going to leave this PR in its current state until we can verify end-to-end works like before.

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionSourceProvider.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public IAsyncCompletionSource GetOrCreate(ITextView textView)
+        {
+            if (textView == null)
+            {
+                throw new ArgumentNullException(nameof(textView));
+            }
+
+            var razorBuffer = textView.BufferGraph.GetRazorBuffers().FirstOrDefault();
+            if (!razorBuffer.Properties.TryGetProperty(typeof(RazorDirectiveCompletionSource), out IAsyncCompletionSource completionSource))
```
AmadeusW: razorBuffer may be null here

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
```diff
@@ -86,6 +103,13 @@ public override Task ProvideCompletionsAsync(CompletionContext context)
                 return Task.CompletedTask;
             }
 
+            if (!_textBufferProvider.TryGetFromDocument(context.Document, out var textBuffer) ||
+                !_asyncCompletionBroker.IsCompletionSupported(textBuffer.ContentType))
```
NTaylorMullen: @AmadeusW Here's my take on turning off our current completion system based off of your feature flag work. Can't verify end-to-end yet but highlighting for now to get your thoughts.

rynowak:
Yay!

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
AmadeusW: this will work 👍👍

Issue #2226, opened 2018-03-27
Handle Mutex creation failures gracefully


ajaybhargavb:
This issue https://github.com/dotnet/roslyn/issues/24124 causes the Mutex constructor to throw on Linux causing the Roslyn build server to fail without falling back to the slow process. We need to account for that in Razor build server as well.

ajaybhargavb:
@mkArtakMSFT, it'd be good to handle this in preview2. I'm assigning this to myself.

mkArtakMSFT:
@ajaybhargavb, you're missing the cost on this one.

ajaybhargavb:
03938dfd956eef8630b5a5a076ebf27ba12a69ff

PR #2228, opened 2018-03-28
Handle Mutex creation exceptions

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2226

Handling Mutex creation exceptions the same way roslyn handles them. We log the exception and fallback to in process cli execution instead of blowing up.

Suggest reviewing this with ?w=1

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
@@ -34,37 +34,51 @@ namespace Microsoft.AspNetCore.Razor.Tools
         protected override Task<int> ExecuteCoreAsync()
         {
             // Make sure there's only one server with the same identity at a time.
-            using (var mutex = new Mutex(initiallyOwned: true, name: MutexName.GetServerMutexName(Pipe.Value()), createdNew: out var holdsMutex))
+            var serverMutexName = MutexName.GetServerMutexName(Pipe.Value());
+            Mutex serverMutex = null;
+            var holdsMutex = false;
+
+            try
             {
-                if (!holdsMutex)
+                serverMutex = new Mutex(initiallyOwned: true, name: serverMutexName, createdNew: out holdsMutex);
+            }
+            catch (Exception ex)
+            {
+                // The Mutex constructor can throw in certain cases. One specific example is docker containers
+                // where the /tmp directory is restricted. In those cases there is no reliable way to execute
+                // the server and we need to fall back to the command line.
+                // Example: https://github.com/dotnet/roslyn/issues/24124
+
+                Error.Write($"Server mutex creation failed. {ex.Message}");
+
+                return Task.FromResult(-1);
+            }
+
+            if (!holdsMutex)
+            {
+                // Another server is running, just exit.
+                Error.Write("Another server already running...");
+                return Task.FromResult(1);
+            }
+
+            try
+            {
+                TimeSpan? keepAlive = null;
+                if (KeepAlive.HasValue() && int.TryParse(KeepAlive.Value(), out var result))
                 {
-                    // Another server is running, just exit.
-                    Error.Write("Another server already running...");
-                    return Task.FromResult(1);
+                    // Keep alive times are specified in seconds
+                    keepAlive = TimeSpan.FromSeconds(result);
                 }
 
-                try
-                {
-                    TimeSpan? keepAlive = null;
-                    if (KeepAlive.HasValue())
-                    {
-                        var value = KeepAlive.Value();
-                        if (int.TryParse(value, out var result))
-                        {
-                            // Keep alive times are specified in seconds
-                            keepAlive = TimeSpan.FromSeconds(result);
-                        }
-                    }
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -103,5 +103,35 @@ public Task Build_SimpleMvc_WithServer_UsingDesktopMSBuild_CanBuildSuccessfully(
             Assert.FileExists(result, IntermediateOutputPath, "SimpleMvc.dll");
             Assert.FileDoesNotExist(result, IntermediateOutputPath, "SimpleMvc.Views.dll");
         }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task Build_ServerConnectionMutexCreationFails_FallsBackToInProcessRzc()
+        {
+            // Use a pipe name longer that 260 characters to make the Mutex constructor throw.
```
ajaybhargavb: Mutex constructor would throw if the mutex name is longer than 260 characters. Since we use the pipe name as the base for mutex name, I found it easier to use this to test this case.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -34,37 +34,55 @@ protected override bool ValidateArguments()
...
+            catch (Exception ex)
+            {
+                // The Mutex constructor can throw in certain cases. One specific example is docker containers
+                // where the /tmp directory is restricted. In those cases there is no reliable way to execute
+                // the server and we need to fall back to the command line.
+                // Example: https://github.com/dotnet/roslyn/issues/24124
+
+                Error.Write($"Server mutex creation failed. {ex.Message}");
+
+                return Task.FromResult(1);
```
pranavkm: Could we give these unique exit codes?

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -34,37 +34,55 @@ protected override bool ValidateArguments()
...
+                // Another server is running, just exit.
+                Error.Write("Another server already running...");
+                return Task.FromResult(1);
+            }
 
-                try
+            try
+            {
+                TimeSpan? keepAlive = null;
+                if (KeepAlive.HasValue())
```
pranavkm:
```C#
if (KeepAlive.HasValue() && int.TryParse(KeepAlive.Value(), out var result))
{
   keepAlive = result;
}
```

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
```diff
@@ -23,12 +23,21 @@ internal static class ServerConnection
 
         public static bool WasServerMutexOpen(string mutexName)
         {
-            var open = Mutex.TryOpenExisting(mutexName, out var mutex);
-            if (open)
+            try
             {
-                mutex.Dispose();
-                return true;
+                var open = Mutex.TryOpenExisting(mutexName, out var mutex);
```
pranavkm: Could we specifically try-catch opening the Mutex is that's what we care about? We don't trap exceptions thrown by `serverMutex.Dispose`

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -34,37 +34,55 @@ protected override bool ValidateArguments()
...
+                // where the /tmp directory is restricted. In those cases there is no reliable way to execute
+                // the server and we need to fall back to the command line.
+                // Example: https://github.com/dotnet/roslyn/issues/24124
+
+                Error.Write($"Server mutex creation failed. {ex.Message}");
+
+                return Task.FromResult(1);
+            }
+
+            if (!holdsMutex)
```
rynowak: It looks like the mutex can leak right?

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
ajaybhargavb: You are right. I meant for this to be -1. Not sure how I missed it. Will update.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
ajaybhargavb: Maybe I misunderstand this but wouldn't calling serverMutex.Dispose() here prematurely release the mutex held my the server already running?

ajaybhargavb:
🆙 📅 . Accidentally amended the commit.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
rynowak: Nice

Issue #2230, opened 2018-03-29
Shutdown command needs to compute the correct default pipe name


ajaybhargavb:
We recently changed the server to be created using the pipe name computed from the path to `rzc.dll` https://github.com/aspnet/Razor/issues/2216. But when the shutdown command is directly invoked, we don't pass in the tool assembly path [here](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Tools/ShutdownCommand.cs#L30). It should instead pass in the tool assembly path to `ComputeDefault()`.

cc @rynowak

ajaybhargavb:
a2a920dde5333483f437d271cce2bab3f6fa1ba8

PR #2232, opened 2018-03-29
Normalize the tool assembly path before using it as the base for pipe…

ajaybhargavb:
… name

#2230 

The actual reason for this bug was the tool assembly path that MsBuild passes in looks like
`C:\Users\ajbaaska\.nuget\packages\microsoft.aspnetcore.razor.design\2.1.0-preview2-t000\build\netstandard2.0\..\..\tools\rzc.dll`

but the shutdown command uses the path
`C:\Users\ajbaaska\.nuget\packages\microsoft.aspnetcore.razor.design\2.1.0-preview2-t000\build\netstandard2.0\tools\rzc.dll`

Notice the lack of `\build\netstandard2.0\..\..\`

So calling `Path.GetFullPath` before using this to compute pipe name fixes the problem.

Added a test


peterhuene:
Thanks!

PR #2242, opened 2018-04-05
Add support for type parameters to class nodes

rynowak:
This change will allow someone extending Razor to use generic type
parameters in generated classes.

There's no user-level extensibility provided here yet, as in there is no
language support for adding type parameters.


rynowak:
/cc @NTaylorMullen @ajaybhargavb 

FYI @SteveSandersonMS - this will show up in 2.1.0-rc1 and 15.7 preview4

I expect this to be useful for templated components

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/CodeWriterExtensions.cs
```diff
@@ -374,7 +374,8 @@ public static CSharpCodeWritingScope BuildNamespace(this CodeWriter writer, stri
             IList<string> modifiers,
             string name,
             string baseType,
-            IEnumerable<string> interfaces)
+            IList<string> interfaces,
```
rynowak: internal class, not a breaking change

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -78,8 +78,8 @@ public override void VisitDocument(DocumentIntermediateNode node)
                     {
                         algorithmId = "{8829d00f-11b8-4213-878b-770e8597ac16}";
                     }
-                    else if (string.Equals(algorithm, HashAlgorithmName.SHA1.Name, StringComparison.Ordinal) || 
-                        
+                    else if (string.Equals(algorithm, HashAlgorithmName.SHA1.Name, StringComparison.Ordinal) ||
```
rynowak: just an autoformat

src/Microsoft.AspNetCore.Razor.Language/Intermediate/ClassDeclarationIntermediateNode.cs
```diff
@@ -16,7 +16,9 @@ public sealed class ClassDeclarationIntermediateNode : MemberDeclarationIntermed
 
         public string BaseType { get; set; }
 
-        public IList<string> Interfaces { get; set; }
+        public IList<string> Interfaces { get; set; } = new List<string>();
+
+        public IList<string> TypeParameters { get; set; } = new List<string>();
```
pranavkm: Are we planning on adding a directive for this? I think that might be a start in solving @javiercn's Identity UI grossness (https://github.com/aspnet/Identity/blob/dev/src/UI/Areas/Identity/Pages/Account/Register.cshtml.cs#L49)

rynowak: Definitely not.

Supporting type parameters on type level things like Razor Pages is a whole nother can of worms and this PR does not attempt to solve this.

rynowak:
Probably worth discussing... I debated adding constraints to this, but I'm not sure whether or not we're going to find it useful/required. 

If we think it's important for Blazor I can add it.

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
```diff
@@ -143,7 +143,12 @@ public override void VisitNamespaceDeclaration(NamespaceDeclarationIntermediateN
 
             public override void VisitClassDeclaration(ClassDeclarationIntermediateNode node)
             {
-                using (Context.CodeWriter.BuildClassDeclaration(node.Modifiers, node.ClassName, node.BaseType, node.Interfaces))
+                using (Context.CodeWriter.BuildClassDeclaration(
+                    node.Modifiers,
+                    node.ClassName,
+                    node.BaseType,
+                    node.Interfaces,
+                    node.TypeParameters.Select(p => p.ParameterName).ToArray()))
```
ajaybhargavb: Suggest putting this in a variable. Easier for debugging.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/TypeParameter.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language.Intermediate
+{
+    public sealed class TypeParameter
```
ajaybhargavb: Do we plan to add more properties to this later? Curious why this is its own type.

NTaylorMullen:
> Probably worth discussing... I debated adding constraints to this, but I'm not sure whether or not we're going to find it useful/required.

Ya constraints will be interesting. I think it's fine for now, add it when we need it 😄 .

src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DefaultDocumentWriter.cs
rynowak: 👎 I don't like this trend. Optimize for readability.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/TypeParameter.cs
rynowak: Yes, if we add constraints or variance annotations we would add them to this type. I originally started with as string and updated it after I sent the PR

PR #2244, opened 2018-04-05
Add experimental flag to un-special-case data-

rynowak:
This change allows blazor to opt into treating data- attributes the same
way as normal attributes in the parser.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
@@ -1957,7 +1957,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             Block expected,
             params RazorDiagnostic[] expectedErrors)
         {
-            var result = ParseCodeBlock(document, descriptors, designTime: false);
+            var result = ParseCodeBlock(RazorLanguageVersion.Latest, document, descriptors, designTime: false);
 
             EvaluateResults(result, expected, expectedErrors);
         }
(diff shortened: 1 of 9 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -1957,7 +1957,7 @@ public void ParseAddOrRemoveDirective_CreatesErrorIfInvalidLookupText_DoesNotThr
             Block expected,
             params RazorDiagnostic[] expectedErrors)
         {
-            var result = ParseCodeBlock(document, descriptors, designTime: false);
+            var result = ParseCodeBlock(RazorLanguageVersion.Latest, document, descriptors, designTime: false);
```
rynowak: Most of the changes here were just to thread the version through all of the various tests

rynowak:
Yeah I think there's a strong case to be made that this makes much more sense for Blazor. I never really agreed with the original rationale. 

Unblocking this here allows us to experiment with this, if it doesn't work out we can revert.

Note that I didn't address the tag helper part of this because we don't need that for Blazor. So if you try to use `data-` as a tag helper attribute it will error regardless of this setting.

rynowak:
@NTaylorMullen - we're ready for that insertion.

NTaylorMullen:
> @NTaylorMullen - we're ready for that insertion.

awesome, will start it. Make sure to merge down to dev 😉

rynowak:
Waiting for @ajaybhargavb to clean up his mess first

ajaybhargavb:
Merged 👍

Issue #2238, opened 2018-04-05
Diagnostics in DirectiveIntermediateNode are lost


ajaybhargavb:
The diagnostics that get added to the directive nodes are all lost because we remove all the directives as part of an optimization [pass](https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Language/DirectiveRemovalOptimizationPass.cs#L9). We should lift those errors up before we delete those nodes so they are not lost.

mkArtakMSFT:
Thanks @ajaybhargavb. How big of a work is this? can we try to fit this in `rc`?

ajaybhargavb:
Shouldn't be too big. Costing this as S.

ajaybhargavb:
5504374125290d419d000293b66a98a329d31bb6

PR #2251, opened 2018-04-09
Preserve directive node diagnostics in document node

ajaybhargavb:
#2238 

We made a targeted fix for this in 15.7 but it's time to fix it for real.

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/DirectiveRemovalOptimizationPassTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/DirectiveRemovalOptimizationPassTest.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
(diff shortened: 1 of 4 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/DirectiveRemovalOptimizationPassTest.cs
```diff
@@ -72,6 +72,47 @@ public void Execute_MultipleCustomDirectives_RemovesDirectiveNodesFromDocument()
...
+            var codeDocument = RazorCodeDocument.Create(sourceDocument);
+            var defaultEngine = RazorEngine.Create(b =>
+            {
+                b.AddDirective(DirectiveDescriptor.CreateDirective("custom", DirectiveKind.SingleLine, d => d.AddStringToken()));
+            });
+            var documentNode = Lower(codeDocument, defaultEngine);
+
+            // Add the diagnostic to the directive node.
+            var visitor = new Visitor();
+            visitor.Visit(documentNode);
```
rynowak: You can also use `.FindDescendants` for this

PR #2277, opened 2018-04-13
Don't fail the build when ExtensionDependencyChecker fails

### LINKED ISSUE #2256: Razor Build tasks fail when using different version of assemblies
We see errors like the following in the Razor build server when upgrading packages
```
Extension assembly 'Microsoft.AspNetCore.Blazor.Razor.Extensions' at 'C:\Users\username\.nuget\packages\microsoft.aspnetcore.blazor.build\0.2.0-preview1-10195\tools\Microsoft.AspNetCore.Blazor.Razor.Extensions.dll' has a different ModuleVersionId than loaded assembly 'Microsoft.AspNetCore.Blazor.Razor.Extensions, Version=0.2.0.0, Culture=neutral, PublicKeyToken=null'
  Extensions could not be loaded. See output for details.
C:\Users\username\.nuget\packages\microsoft.aspnetcore.blazor.build\0.2.0-preview1-10195\targets\RazorCompilation.targets(179,5): error : rzc generate exited with code 1 
```

There are a few issues with this.

It would firstly be great to know which other assembly is loaded and where it was loaded from - my guess is that this happened when updating the packages used by our templates, or by mixing our templates with a repo-local build.

Secondly, I think this is just a bug. The assembly loading logic we use for Razor is supposed to be based on paths for 'top level' extensions, meaning that we should be able to load two copies of `Microsoft.AspNetCore.Blazor.Razor.Extensions`. We should investigate this and see if it's possible.

Lastly, this **must not** fail the build. We expect users to mix and match dependencies and this is intended to 'just work'.

ajaybhargavb:
Fixes #2256 

We want to make sure to not fail the build and fallback to cli execution when the ExtensionDependencyChecker fails to load an extension.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
@@ -49,6 +49,12 @@ namespace Microsoft.AspNetCore.Razor.Tools
 
         protected override Task<int> ExecuteCoreAsync()
         {
+            if (!Parent.Checker.Check(ExtensionFilePaths.Values))
+            {
+                Error.WriteLine($"Extensions could not be loaded. See output for details.");
+                return Task.FromResult(ExitCodeFailure);
+            }
+
             // Loading all of the extensions should succeed as the dependency checker will have already
             // loaded them.
             var extensions = new RazorExtension[ExtensionNames.Values.Count];
@@ -125,12 +131,6 @@ namespace Microsoft.AspNetCore.Razor.Tools
                 }
             }
 
-            if (!Parent.Checker.Check(ExtensionFilePaths.Values))
-            {
-                Error.WriteLine($"Extensions could not be loaded. See output for details.");
-                return false;
-            }
-
             return true;
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -189,7 +189,7 @@ protected override bool ValidateArguments()
                 }
             }
 
-            return success ? 0 : -1;
+            return success ? 0 : 1;
```
pranavkm: Can we make this a more unique code? We're using this to specifically signal a certain kind of output to DotnetToolTask and 1 is meant to be general purpose error. I'd say, pick 65 (from https://stackoverflow.com/a/1535733)

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -49,6 +49,12 @@ public GenerateCommand(Application parent)
 
         protected override Task<int> ExecuteCoreAsync()
         {
+            if (!Parent.Checker.Check(ExtensionFilePaths.Values))
+            {
+                Error.WriteLine($"Extensions could not be loaded. See output for details.");
+                return Task.FromResult(-1);
```
pranavkm: Positive exit code.

ajaybhargavb:
🆙 📅

Issue #2248, opened 2018-04-09
Build server should write the process and pipe information to a file


ajaybhargavb:
This makes it easier to discover all the running build servers which is needed for implementing the `dotnet buildserver shutdown` command.

ajaybhargavb:
d2fe76be21ad1186b5edc61fa73f1fc634a75e3b

PR #2285, opened 2018-04-17
Write process and pipe information to a pid file

ajaybhargavb:
#2248 

cc @rynowak @peterhuene

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/MutexName.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/MutexName.cs
@@ -12,7 +12,11 @@ namespace Microsoft.AspNetCore.Razor.Tools
 
         public static string GetServerMutexName(string pipeName)
         {
-            return $"{pipeName}.server";
+            // We want to prefix this with Global\ because we want this mutex to be visible
+            // across terminal sessions which is useful for cases like shutdown.
+            // https://msdn.microsoft.com/en-us/library/system.threading.mutex(v=vs.110).aspx#Remarks
+            // This still wouldn't allow other users to access the server because the pipe will fail to connect.
+            return $"Global\\{pipeName}.server";
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
@@ -2,6 +2,10 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.CommandLineUtils;
@@ -15,6 +19,18 @@ namespace Microsoft.AspNetCore.Razor.Tools
         {
             Pipe = Option("-p|--pipe", "name of named pipe", CommandOptionType.SingleValue);
             KeepAlive = Option("-k|--keep-alive", "sets the default idle timeout for the server in seconds", CommandOptionType.SingleValue);
+        }
+
+        // For testing purposes only.
+        internal ServerCommand(Application parent, string pipeName, int? keepAlive = null)
+            : this(parent)
+        {
+            Pipe.Values.Add(pipeName);
+
+            if (keepAlive.HasValue)
+            {
+                KeepAlive.Values.Add(keepAlive.Value.ToString());
+            }
         }
 
         public CommandOption Pipe { get; }
@@ -61,8 +77,20 @@ namespace Microsoft.AspNetCore.Razor.Tools
                 return Task.FromResult(1);
             }
 
+            FileStream pidFileStream = null;
             try
             {
+                try
+                {
+                    // Write the process and pipe information to a file in a well-known location.
+                    pidFileStream = WritePidFile();
+                }
+                catch (Exception ex)
+                {
+                    // Something happened when trying to write to the pid file. Log and move on.
+                    ServerLogger.LogException(ex, "Failed to create PID file.");
+                }
+
                 TimeSpan? keepAlive = null;
                 if (KeepAlive.HasValue() && int.TryParse(KeepAlive.Value(), out var result))
                 {
@@ -79,6 +107,7 @@ namespace Microsoft.AspNetCore.Razor.Tools
             {
                 serverMutex.ReleaseMutex();
                 serverMutex.Dispose();
+                pidFileStream?.Close();
             }
 
             return Task.FromResult(0);
@@ -88,6 +117,50 @@ namespace Microsoft.AspNetCore.Razor.Tools
         {
             var dispatcher = RequestDispatcher.Create(host, compilerHost, cancellationToken, eventBus, keepAlive);
             dispatcher.Run();
+        }
+
+        internal FileStream WritePidFile()
+        {
+            // To make all the running rzc servers more discoverable, We want to write the process Id and pipe name to a file.
+            // The file contents will be in the following format,
+            //
+            // <PID>
+            // rzc
+            // path/to/rzc.dll
+            // <pipename>
+
+            const int DefaultBufferSize = 4096;
+            var processId = Process.GetCurrentProcess().Id;
+            var fileName = $"rzc-{processId}";
+
+            var path = Environment.GetEnvironmentVariable("DOTNET_BUILD_PIDFILE_DIRECTORY");
+            if (string.IsNullOrEmpty(path))
+            {
+                var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+                var homePath = Environment.GetEnvironmentVariable(homeEnvVariable);
+                if (string.IsNullOrEmpty(homePath))
+                {
+                    // Couldn't locate the user profile directory. Bail.
+                    return null;
+                }
+
+                path = Path.Combine(homePath, ".dotnet", "pids", "build");
+            }
+
+            // Make sure the directory exists.
+            Directory.CreateDirectory(path);
+
+            path = Path.Combine(path, fileName);
+            var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write, FileShare.Read, DefaultBufferSize, FileOptions.DeleteOnClose);
+
+            using (var writer = new StreamWriter(fileStream, Encoding.UTF8, DefaultBufferSize, leaveOpen: true))
+            {
+                var rzcPath = Assembly.GetExecutingAssembly().Location;
+                var content = $"{processId}{Environment.NewLine}rzc{Environment.NewLine}{rzcPath}{Environment.NewLine}{Pipe.Value()}";
+                writer.Write(content);
+            }
+
+            return fileStream;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
@@ -20,6 +21,9 @@ namespace Microsoft.AspNetCore.Razor.Tools
 
         // Spend up to 20s connecting to a new process, to allow time for it to start.
         private const int TimeOutMsNewProcess = 20000;
+
+        // Custom delegate that contains an out param to use with TryCreateServerCore method.
+        private delegate TResult TryCreateServerCoreDelegate<T1, T2, T3, T4, out TResult>(T1 arg1, T2 arg2, out T3 arg3, T4 arg4);
 
         public static bool WasServerMutexOpen(string mutexName)
         {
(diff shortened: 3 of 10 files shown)


src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -89,5 +105,42 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
...
+            {
+                var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+                path = Path.Combine(Environment.GetEnvironmentVariable(homeEnvVariable), ".dotnet/pids/build");
+            }
+
+            // Make sure the directory exists.
+            Directory.CreateDirectory(path);
+
+            path = Path.Combine(path, fileName);
+            var fileStream = new FileStream(path, FileMode.Create, FileAccess.Write, FileShare.Read, DefaultBufferSize, FileOptions.DeleteOnClose);
```
rynowak: Is delete on close trustworthy? I thought there was some doubt about this.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -79,6 +94,7 @@ protected override Task<int> ExecuteCoreAsync()
             {
                 serverMutex.ReleaseMutex();
                 serverMutex.Dispose();
+                pidFileStream?.Close();
```
rynowak: using?

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -89,5 +105,42 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
...
+
+            const int DefaultBufferSize = 4096;
+            var processId = Process.GetCurrentProcess().Id;
+            var fileName = $"rzc-{processId}";
+
+            var path = Environment.GetEnvironmentVariable("DOTNET_BUILD_PIDFILE_DIRECTORY");
+            if (string.IsNullOrEmpty(path))
+            {
+                var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+                path = Path.Combine(Environment.GetEnvironmentVariable(homeEnvVariable), ".dotnet/pids/build");
```
rynowak: You can path.combine all of these together

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
rynowak: Oh, I didn't see this is in a finally. Ignore

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -89,5 +105,42 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
...
+        internal FileStream WritePidFile()
+        {
+            /*
+             * To make all the running rzc servers more discoverable, We want to write the process Id and pipe name to a file.
+             * The file contents will be in the following format,
+             *
+             * <PID>
+             * rzc
+             * <pipename>
+             */
```
rynowak: Don't use `/* Taylor comments */`

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
ajaybhargavb: From my local testing this works consistently. This was also mentioned in the design doc. @peterhuene thoughts?

peterhuene: Ultimately I'd have to take a look at the framework, but I would wager that this is cross-plat without issue.  On Windows, I assume it uses `FILE_FLAG_DELETE_ON_CLOSE`. On macOS and Linux, I assume the file is immediately unlinked.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -89,5 +105,40 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
...
             dispatcher.Run();
         }
+
+        internal FileStream WritePidFile()
+        {
+            // To make all the running rzc servers more discoverable, We want to write the process Id and pipe name to a file.
+            // The file contents will be in the following format,
+            //
+            // <PID>
+            // rzc
```
peterhuene: The doc called for `razor`, but if you prefer `rzc`, I'll amend the doc (easy change on my side).

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -89,5 +105,40 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
...
+
+            const int DefaultBufferSize = 4096;
+            var processId = Process.GetCurrentProcess().Id;
+            var fileName = $"rzc-{processId}";
+
+            var path = Environment.GetEnvironmentVariable("DOTNET_BUILD_PIDFILE_DIRECTORY");
+            if (string.IsNullOrEmpty(path))
+            {
+                var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+                path = Path.Combine(Environment.GetEnvironmentVariable(homeEnvVariable), ".dotnet", "pids", "build");
```
peterhuene: We should probably guard against `Environment.GetEnvironmentVariable(homeEnvVariable)` returning null (`dotnet` has an outstanding issue in CI environments that unset `HOME`, for example).  I think the correct behavior would be to skip writing the file.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
ajaybhargavb: `rzc` is the name of the server dll similar to `vbcscompiler`. So I thought this makes more sense.

peterhuene: I'm good with this. I'll update the document and my change.

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
```diff
@@ -21,6 +21,9 @@ internal static class ServerConnection
         // Spend up to 20s connecting to a new process, to allow time for it to start.
         private const int TimeOutMsNewProcess = 20000;
 
+        // Custom delegate that contains an out param to use with TryCreateServerCore method.
+        private delegate TResult Func<T1, T2, T3, T4, out TResult>(T1 arg1, T2 arg2, out T3 arg3, T4 arg4);
```
pranavkm: Sorry, what's happening here?

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
```diff
@@ -159,5 +164,31 @@ public Task Build_SimpleMvc_WithServer_UsingDesktopMSBuild_CanBuildSuccessfully(
...
+            
+            // We expect the server to have started by now.
+
+            var processId = _buildServer.ProcessId;
+            var expectedFileName = $"rzc-{processId}";
+            var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+            var path = Path.Combine(Environment.GetEnvironmentVariable(homeEnvVariable), ".dotnet", "pids", "build", expectedFileName);
+
+            // Assert
+            Assert.True(File.Exists(path));
```
pranavkm: Didn't you say this had timing issues?

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -61,8 +64,20 @@ protected override Task<int> ExecuteCoreAsync()
...
 
+            FileStream pidFileStream = null;
             try
             {
+                try
+                {
+                    // Write the process and pipe information to a file in a well-known location.
+                    pidFileStream = WritePidFile();
+                }
+                catch (IOException ex)
```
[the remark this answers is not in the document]
ajaybhargavb: Sure. I don't see why not.

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
ajaybhargavb: Needed a custom delegate to support the out parameter. The built-in Func<> overloads wouldn't work.

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildServerIntegrationTest.cs
ajaybhargavb: Yes. I am currently working on unit tests and will removing this in the next update.

ajaybhargavb:
Updated.

ajaybhargavb:

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.CodeAnalysis;
+using Moq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Razor.Tools
+{
+    public class ServerCommandTest
+    {
+        [Fact]
+        public void WritePidFile_WorksAsExpected()
```
pranavkm: 👏

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/ServerConnection.cs
pranavkm: Ok. Let's give this a different name `TryCreateServerDelegate` and this should be good to go.

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -0,0 +1,61 @@
...
+            // Act & Assert
+            using (var _ = server.WritePidFile())
+            {
+                Assert.True(File.Exists(path));
+
+                // Make sure another stream can be opened while the write stream is still open.
+                using (var fileStream = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite | FileShare.Delete))
+                using (var reader = new StreamReader(fileStream))
+                {
+                    var lines = reader.ReadToEnd().Split(Environment.NewLine);
```
pranavkm:
```C
var lines = reader.ReadAllLines();
Assert.Equal(new[] { expectedProcessId.ToString(), "rzr", expectedRzcPath, pipeName }, lines);
```

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -17,6 +21,21 @@ public ServerCommand(Application parent)
             KeepAlive = Option("-k|--keep-alive", "sets the default idle timeout for the server in seconds", CommandOptionType.SingleValue);
         }
 
+        // For testing purposes only.
+        internal ServerCommand(Application parent, string pipeName, int? keepAlive = null)
+            : base(parent, "server")
```
pranavkm: Could you do `:this (parent)` and remove the two lines below?

src/Microsoft.AspNetCore.Razor.Tools/MutexName.cs
```diff
@@ -12,7 +12,11 @@ public static string GetClientMutexName(string pipeName)
 
         public static string GetServerMutexName(string pipeName)
         {
-            return $"{pipeName}.server";
+            // We want to prefix this with Global\ because we want this mutex to be visible
+            // across terminal sessions which is useful for cases like shutdown.
+            // https://msdn.microsoft.com/en-us/library/system.threading.mutex(v=vs.110).aspx#Remarks
+            // This still wouldn't allow other users to access the server because the pipe will fail to connect.
+            return $"Global\\{pipeName}.server";
```
ajaybhargavb: @peterhuene I had to do this to make our shutdown command work across different terminal sessions on Unix systems. Without this the command just thinks the server is not running. 
It's possible Roslyn might have this problem as well as they do something [similar](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Server/VBCSCompiler/BuildServerController.cs#L90). Could you do a quick check if your Roslyn shutdown code works on Unix machines across terminal sessions? If not, they might have to add this prefix as well.

peterhuene: Looks like it won't work for Roslyn, so we should probably open an issue there as well.  We're only making this new mechanism work for Razor currently, so hopefully we can get that addressed at the same time Roslyn starts writing pid files for us.

pranavkm: Do you need us to file an issue for it in Roslyn?

peterhuene: I'll do it, thanks for the heads up!

peterhuene:
Thanks for all the work on this, @ajaybhargavb!

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
peterhuene: Unfortunately I was wrong about this, at least on macOS and Linux: there is no analogous mechanism to `FILE_FLAG_DELETE_ON_CLOSE` for those platforms (and of course immediate unlinking would only work if you don't expect another process to be able to read the file).  While this works for Windows as intended, sending a `SIGINT` or `SIGTERM` to the server process on the other platforms leaves the pid file behind.

There are a few ways to handle these signals in a cross-platform way with corefx:
Handle SIGTERM with `AssemblyLoadContext.Default.Unloading`
Handle SIGINT/SIGQUIT with `Console.CancelKeyPress`

I can handle cleaning up defunct pid files from the `build-server shutdown` command, but since this is a non-tmpfs location, we should probably also fix it in the server.

Obviously this won't fix SIGKILL, but caveat emptor when sending that signal.

My apologies for not realizing this sooner.

PR #2291, opened 2018-04-20
Make newline comparison work crossplat

ajaybhargavb:
Fixes issue [585150](https://devdiv.visualstudio.com/DevDiv/Xamarin%20VS%20for%20Mac/_workitems/edit/585150)

I've manually verified this mpack fixes the issue in VS for Mac.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserHelpers.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserHelpers.cs
@@ -26,8 +26,10 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
         public static bool IsNewLine(string value)
         {
+            // We want to handle both LF and CRLF regardless of the platform.
+            // We explicitly check for CRLF and IsNewLine() should return true for LF.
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserHelpers.cs
```diff
@@ -27,7 +27,7 @@ public static bool IsNewLine(char value)
         public static bool IsNewLine(string value)
         {
             return (value.Length == 1 && (IsNewLine(value[0]))) ||
-                   (string.Equals(value, Environment.NewLine, StringComparison.Ordinal));
```
ajaybhargavb: Whenever the document contains CRLF in VS for Mac, this check used to return false which ended up breaking few indentation scenarios.

rynowak:
:shipit: 

This change looks good to me. We should probably take a look at everywhere we use `Environment.Newline` in Razor because I would contend that outside of 1-2 places where we initialize things it's always wrong.  All of our code needs to work with either newline format regardless of platform.

pranavkm:
:shipit:

PR #2292, opened 2018-04-21
Disable node-reuse to allow rebuilding

pranavkm:
Recent builds of msbuild have node reuse enabled by default. This locks up the task dlls after the test's completed preventing re-runs untill you kill the process.

Here's the process at work after running dotnet test.
```
"\.dotnet\x64\dotnet.exe" "\.dotnet\x64\sdk\2.1.300-preview2-008530\MSBuild.dll" \.dotnet\x64\sdk\2.1.300-preview2-008530\MSBuild.dll /nologo /nodemode:1 /nodeReuse:true
```


pranavkm:
FYI @rynowak \ @ajaybhargavb

PR #2293, opened 2018-04-21
Simplify build + retarget tests to netcoreapp2.1 

pranavkm:


src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -20,84 +18,27 @@
...
-  </ItemGroup>
-
-  <!-- Using explicit SDK imports here because the default way conflicts with the AfterBuild target -->
-  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
   <PropertyGroup>
-    <GenerateNuspecDependsOn>_BuildDependencyProjects;$(GenerateNuspecDependsOn)</GenerateNuspecDependsOn>
-    <BuildDependsOn>_BuildDependencyProjects;$(BuildDependsOn)</BuildDependsOn>
+    <ToolProject>..\Microsoft.AspNetCore.Razor.Tools\Microsoft.AspNetCore.Razor.Tools.csproj</ToolProject>
+    <TaskProject>Microsoft.AspNetCore.Razor.Tasks</TaskProject>
```
ajaybhargavb: Include the path to the csproj here to keep it consistent.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -1,14 +1,12 @@
-﻿<Project>
-
-  <!-- Using explicit SDK imports here because the default way conflicts with the AfterBuild target -->
-  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
-
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Description>Razor is a markup syntax for adding server-side logic to web pages. This package contains MSBuild support for Razor.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
```
ajaybhargavb: Why this change? I don't fully understand.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -20,84 +18,27 @@
...
 
-  <Target Name="PopulateNuspec" BeforeTargets="GenerateNuspec" DependsOnTargets="BuiltProjectOutputGroup;DebugSymbolsProjectOutputGroup;_BuildDependencyProjects">
+  <Target Name="PopulateNuspec" BeforeTargets="GenerateNuspec">
 
     <PropertyGroup>
       <!-- Make sure we create a symbols.nupkg. -->
       <IncludeSymbols>true</IncludeSymbols>
 
+      <!-- RepositoryCommit is only available when "build" runs, but not during dotnet pack -->
+      <RepositoryCommit Condition="'$(RepositoryCommit)' == ''">unknown</RepositoryCommit>
```
ajaybhargavb: 👏

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
```diff
@@ -20,84 +18,27 @@
...
-
-    <ItemGroup>
-      <_RazorTool Include="$(OutputPath)tools\**\*" />
-    </ItemGroup>
-
-    <Error Text="_RazorTool is empty. This is a bug" Condition="'@(_RazorTool)'==''" />
-  </Target>
+  <ItemGroup>
+    <ProjectReference Include="..\$(TaskProject)\$(TaskProject).csproj" ReferenceOutputAssembly="false" />
+    <ProjectReference Include="$(ToolProject)" ReferenceOutputAssembly="false" Condition="'$(TargetFramework)'=='netcoreapp2.0'" />
```
ajaybhargavb: What happens when tfm is net46?

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
pranavkm: We want to able to target the two projects this project references. The TFMs here ultimately do not matter since there's no shipping code in this project.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
pranavkm: I need the name though.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
pranavkm: Nothing really. We just need a way to tell MSBuild that Razor.Tools needs to be built before Razor.Design. We don't use it's output and it doesn't get used in the inner build. All of the interesting work happens in the OuterBuild. I could change this to `'$(TargetFramework)'!='net461'` to reflect this better.

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
ajaybhargavb: Got it. Please add a comment.

ajaybhargavb:
Now that we run the tests on `netcoreapp2.1`, I don't think we need https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Razor.Design/build/netstandard2.0/Microsoft.AspNetCore.Razor.Design.CodeGeneration.targets#L127 anymore.

pranavkm:
Sure. We won't we be able to ever remove the property from the task but we could stop passing a value for it in the Sdk.

pranavkm:
🆙 📅

src/Microsoft.AspNetCore.Razor.Design/Microsoft.AspNetCore.Razor.Design.csproj
pranavkm: I tried that and it didn't work. The outer build complains when that the tools project doesn't reference `net461`. This works ™️ so I'll leave it be

Issue #2247, opened 2018-04-09
View Compilation build error in VS15.7 due to missing references


ocdi:
I have a solution with a large number of projects that I am porting to MvcCore. In this solution I have a number of class libraries that contains views. I was using RazorGenerator.MsBuild in the MVC5 days and some view engine hackery to get the main web project to do what you are enabling in 2.1.

I've been trying out the view precompilation using the 2.1.0-preview1, and I'm using VS 15.7 preview 2.

I got the code to the point where it compiles. From the command line invoking dotnet build and dotnet run I haven't had any issues, it is working well.

The issue I am having is building within Visual Studio I am getting build failures on incremental builds due to missing project references being added to the csc command line. This may not strictly be a View Compilation bug, but I am not sure what else to check.

I've installed the Project System extension and looked at the build log. It appears that the ResolveProjectReferences isn't being called to populate these dll's and then has the build failure.

This only seems to be affecting projects containing views where I have enabled the RazorSdk and and RazorCompileOnBuild=true settings. The csc command does appear to be including PackageReferences.

Strangely according to the Build Logging it is when building the main project that has references to the class libraries, that it is failing.

I've managed to reproduce this in a clean class library project. What I did was create a new project on the 2.0 templates, upgraded to 2.1 and created a class library for the views. I've created a common class library and moved the Web models to the Common library. Then I've moved the Account views into the View library. This is referenced from the web project. It seemed to work OK for a single project, however when duplicating the class library I get the same build error in my full solution.

Sometimes it is a bit hard to trigger the bug. The build output looks like this
```
Error occurred while restoring NuGet packages: Object reference not set to an instance of an object.
1>------ Build started: Project: ViewClass, Configuration: Debug Any CPU ------
1>ViewClass -> C:\dev\BugReport\ViewClass\bin\Debug\netstandard2.0\ViewClass.dll
1>ViewClass -> C:\dev\BugReport\ViewClass\bin\Debug\netstandard2.0\ViewClass.PrecompiledViews.dll
2>------ Build started: Project: BugReport, Configuration: Debug Any CPU ------
2>Class1.cs(1,7,1,17): error CS0246: The type or namespace name 'CommonCode' could not be found (are you missing a using directive or an assembly reference?)
2>Class1.cs(12,30,12,39): error CS0246: The type or namespace name 'SomeClass' could not be found (are you missing a using directive or an assembly reference?)
[...]

PR #2307, opened 2018-04-25
Make incremental build in VS copy output files

### LINKED ISSUE #2306: Incremental builds in VS does not copy Views dll from referencing class library project
Steps:

1. Create a solution containing a WebApp that references a Razor Class Library project.
2. Build the solution
3. Verify that the ClassLibrary Views dll and pdb are copied to the WebApp's output directory
4. Clean the WebApp
5. Build it.

Expected:
Views dll exists in the build output dir of WebApp

Actual:
No view dll found in the build output dir

pranavkm:
Fixes #2306

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
+++ b/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
@@ -391,7 +391,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       This condition needs to be inside the target because it the itemgroup will be populated after the target's
       condition is evaluated.
     -->
-    <ItemGroup Condition="'@(RazorGenerate)'!=''">
+    <ItemGroup>
       <AllItemsFullPathWithTargetPath Include="@(RazorIntermediateAssembly->'%(FullPath)')">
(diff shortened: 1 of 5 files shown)


pranavkm:
A variant of https://github.com/aspnet/Razor/issues/2247. We don't invoke the `ResolveTagHelperInput` target as part of `_RazorGetCopyToOutputDirectoryItems` making `RazorGenerate` empty.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -391,7 +391,7 @@ Copyright (c) .NET Foundation. All rights reserved.
       This condition needs to be inside the target because it the itemgroup will be populated after the target's
       condition is evaluated.
     -->
-    <ItemGroup Condition="'@(RazorGenerate)'!=''">
+    <ItemGroup>
```
rynowak: Will this no op if the file isn't there? This was done on purpose so that we're not generating this extra set of items when you have no Razor files.

ajaybhargavb:
@pranavkm the travis failure should get fixed if you rebase.

pranavkm:
There was one scenario that was broken with the earlier change. (which is when you delete all cshtml files from a project). It was cheap enough to address it

Issue #2314, opened 2018-04-27
Make VisualStudioDocumentTracker work in files that flow between having and not having ITextViews


NTaylorMullen:
In a model where a Razor file is opened without an `ITextView` and all of its Razor bits are created and then at a later time an `ITextView` is attached; we need to guard against the document tracker being subscribed multiple times. We almost do this today with our editor factory service that caches the document tracker on the text buffer but its only criteria on whether to subscribe or not is if there's a single text view.

Basically the flow that's broken is:
1. Index.cshtml is opened in the background by some system (live share) without a view.
2. Index.cshtml gets opened by the user, this attaches an `ITextView` to the file.

This results in the document tracker attaching to the project snapshot manager and all of its other info sources multiple times.

PR #2315, opened 2018-04-27
Make document tracker resilient to multiple subscribes.

NTaylorMullen:
- Added tests to verify new functionality.

I could have gone one step further and removed:

```C#
if (documentTracker.TextViews.Count == 1)
{
    tracker.Subscribe();
}
```

From our text view connection listener but figured it portrayed the intention of the code rather than the implementation (even though it already casts to the implementation type).

#2314


src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -192,6 +201,12 @@ public void Unsubscribe()
         {
             _foregroundDispatcher.AssertForegroundThread();
 
+            if (!_subscribed)
+            {
+                return;
+            }
```
rynowak: If you're going to go this route don't you need some kind of refcount? Presumably whomever creates the tracker will un-create it.

NTaylorMullen: Hu? You'll have to explain more. Not sure I see how this would need a refcount of any sort.

NTaylorMullen:
🆙 📅

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -177,12 +178,18 @@ public void Subscribe()
         {
             _foregroundDispatcher.AssertForegroundThread();
 
+            if (_subscribeCount++ > 0)
```
rynowak: /cc @Eilon so that he can see me approving of this use of `++` in an `if`.

rynowak: ❤️ x 100000

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -192,6 +199,11 @@ public void Unsubscribe()
         {
             _foregroundDispatcher.AssertForegroundThread();
 
+            if (_subscribeCount == 0 || _subscribeCount-- > 1)
```
ajaybhargavb: In what scenario will this method be called when `_subscribeCount` is 0? Or are we just being defensive?

NTaylorMullen: Just being defensive

Issue #2301, opened 2018-04-24
Changes to RCL project's content requires referencing web app to be re-build for it take get latest RCL project's Views dll


kichalla:
**Repro Steps:**

- Create a new Razor Class Library (RCL) project | Build it
- Add a new Razor Pages(RP) web application | Reference the above created RCL project | Build it
- Ctrl + F5 | access http://localhost:port/myfeature/page1 | you should be able to see an empty page with Page1 title
- Now add some content tto the body of RCL project's `Page1.cshtml` | Save it | Ctrl + F5 | access http://localhost:port/myfeature/page1

**Expected**: Updated changes to the html body are reflected
**Actual**: Updated changes do not show up as the updated Views dll is not copied to the RP project's output

**Workaround**: Rebuild RP project | Ctrl + F5

mkArtakMSFT:
Potential dupe of https://github.com/aspnet/Razor/issues/2300

mkArtakMSFT:
@pranavkm, do we have more info regarding whether this is a dupe or not?

pranavkm:
I closed the other one as dup.

mkArtakMSFT:
@pranavkm, can you please link the PR to this issue?

pranavkm:
https://github.com/aspnet/Razor/pull/2316

pranavkm:
Fixed in 05c84f008a5a6536b6ce84c3dbf8c468dc08f4c0

PR #2316, opened 2018-04-30
Ensure Views dll are copied during P2P builds

pranavkm:
FastUpToDateCheck in VS doesn't account for changes solely to .Views.dll. This causes referencing projects to be treated as
up to date even though a referenced project rebuilt. Touch the marker file to cause referenced projects to rebuild.

Fixes https://github.com/aspnet/Razor/issues/2301

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
+++ b/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
@@ -180,9 +180,23 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RazorIntermediateAssembly Condition="'$(RazorIntermediateAssembly)'==''" Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
     <!-- Used in Compilation.targets -->   
     <_RazorDebugSymbolsIntermediatePath Condition="'$(_RazorDebugSymbolsProduced)'=='true'" Include="$(IntermediateOutputPath)$(RazorTargetName).pdb" />
-    <!-- Add all cshtml files to UpToDateCheckInput - a collection used by the IDE's project system to determine if a project needs to be rebuilt -->
-    <UpToDateCheckInput Condition="'$(RazorCompileOnBuild)'=='true'" Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!--
+      Add all cshtml files to UpToDateCheckInput - a collection of files used by FastUpToDateCheck to determine
+      if any of the the project inputs have changed.
+    -->
+    <UpToDateCheckInput Condition="'$(RazorCompileOnBuild)'=='true'" Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
+
+    <!--
+      Add Razor output files to UpToDateCheckBuilt - a collection of files used by FastUpToDateCheck to determine
+      if any of the project's outputs have changed.
+    -->
+    <UpToDateCheckBuilt Include="@(RazorIntermediateAssembly)"
+      Condition="'$(RazorCompileOnBuild)'=='true' AND '@(Content->WithMetadataValue('Extension', '.cshtml'))' != ''" />
+  </ItemGroup>
+
 
   <!--
     These are the targets that generate code using Razor, separated from the main file for ease of maintenance.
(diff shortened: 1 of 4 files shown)


src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -180,8 +180,6 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RazorIntermediateAssembly Condition="'$(RazorIntermediateAssembly)'==''" Include="$(IntermediateOutputPath)$(RazorTargetName).dll" />
     <!-- Used in Compilation.targets -->   
     <_RazorDebugSymbolsIntermediatePath Condition="'$(_RazorDebugSymbolsProduced)'=='true'" Include="$(IntermediateOutputPath)$(RazorTargetName).pdb" />
-    <!-- Add all cshtml files to UpToDateCheckInput - a collection used by the IDE's project system to determine if a project needs to be rebuilt -->
-    <UpToDateCheckInput Condition="'$(RazorCompileOnBuild)'=='true'" Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
```
pranavkm: Moved this to design time targets

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
```diff
@@ -33,6 +33,21 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+      Add all cshtml files to UpToDateCheckInput - a collection of files used by FastUpToDateCheck to determine
+      if any of the the project inputs have changed.
+    -->
+    <UpToDateCheckInput Condition="'$(RazorCompileOnBuild)'=='true'" Include="@(Content->WithMetadataValue('Extension', '.cshtml'))" />
+
+    <!--
+      Add Razor output files to UpToDateCheckBuilt - a collection of files used by FastUpToDateCheck to determine
+      if any of the project's outputs have changed.
+    -->
+    <UpToDateCheckBuilt Include="@(RazorIntermediateAssembly)"
```
pranavkm: The primary scenario this is useful for is when the Views.dll gets out-of-sync with cshtml files (e.g. if a user deletes this file from the output directory). It would be correct to also include the `RazorTargetFile`. Unfortunately we calculate that inside a target and no target is called as part of gathering these inputs. It's a bit of an edge scenario, and we can wait for user feedback

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIncrementalismTest.cs
```diff
@@ -136,5 +137,46 @@ public BuildIncrementalismTest(BuildServerTestFixture buildServer)
...
+
+            result = await DotnetMSBuild("Build");
+            Assert.BuildPassed(result);
+
+            // Verify thumbprint file is unchanged between true incremental builds
+            Assert.Equal(classLibraryThumbPrint, GetThumbPrint(classLibraryDll));
+            Assert.Equal(classLibraryViewsThumbPrint, GetThumbPrint(classLibraryViewsDll));
+            // In practice, this should remain unchanged. However, since our tests reference
+            // binaries from other projects, this file gets updated by Microsoft.Common.targets
+            Assert.NotEqual(markerFileThumbPrint, GetThumbPrint(markerFile));
```
pranavkm: This is kinda crummy. Microsoft.Common.targets always updates the file because we reference binaries inside tests. I verified this in a vanilla project and things worked correctly. Regardless, keeping this file in so we can ensure that updating just the views dll results in the thumbprint being updated.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
rynowak: I'm not sure I understand what you're saying here.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
```diff
@@ -33,6 +33,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <ProjectCapability Include="DynamicFileNesting" />
   </ItemGroup>
 
+  <ItemGroup>
```
rynowak: We don't get to change this file right now. VS is in escrow.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
pranavkm: Here's the scenario: In VS,

1. File -> New WebApp
2. Build it
3. Delete obj\Debug\netcoreapp2.1\WebApp1.Views.dll
3. Build again 

Expected:
We rebuild the Views.dll

Actual:
========== Build: 0 succeeded, 0 failed, 1 up-to-date, 0 skipped ==========

This adds the file to UpToDateCheckBuilt so this causes VS to recompile. We should afford the same treatment for `bin\Debug\netcoreapp2.1\WebApp1.Views.dll`, but calculating it's path is inside a target now, hence unavailable in the global context. It's a less than common scenario so I'm not doing changing things to solve it at this point.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
pranavkm: Ok

pranavkm:
🆙 📅

PR #2317, opened 2018-04-30
Fix #2298 simplify TagHelperDescriptor hash

### LINKED ISSUE #2298: TagHelperDescriptorComparers GetHashCode methods are too expensive
We do too much to calculate hash codes for TagHelperDescriptors. At design time this ends up accounting for around 2.6% of overall execution. We can simplify this significantly by only looking at the `Name`, `Kind` and `AssemblyName`.

```

Name | Inc % | Inc
-- | -- | --
Microsoft.AspNetCore.Razor.Language!TagHelperDescriptorComparer.GetHashCode | 2.6 | 210


Name | Inc % | Inc
-- | -- | --
Microsoft.AspNetCore.Razor.Language!TagHelperDescriptorComparer.GetHashCode | 3.7 | 298
+   Microsoft.AspNetCore.Razor.Language!TagMatchingRuleDescriptorComparer.GetHashCode | 1.9 | 154
\|+   system.core.ni!System.Linq.OrderedEnumerable`1+<GetEnumerator>d__1[System.__Canon].MoveNext() | 1.1 | 90
\|+   Microsoft.AspNetCore.Razor.Language!RequiredAttributeDescriptorComparer.GetHashCode | 0.4 | 29
\|+ clr!JIT_New | 0.2 | 17
\|+   mscorlib.ni!System.OrdinalComparer.GetHashCode(System.String) | 0.1 | 6
+   system.core.ni!System.Linq.OrderedEnumerable`1+<GetEnumerator>d__1[System.__Canon].MoveNext() | 0.8 | 61
+ system.core.ni!System.Linq.OrderedEnumerable`1[System.__Canon].GetEnumerator() | 0.4 | 31
+   Microsoft.AspNetCore.Razor.Language!BoundAttributeDescriptorComparer.GetHashCode | 0.3 | 26
+ clr!JIT_New | 0.2 | 13



```

/cc @ToddGrun

rynowak:
The hash code implementation here is exhaustive when it doesn't need to
be. Slimming this down to a much more reasonable set of things for perf
reasons.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptorComparer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptorComparer.cs
@@ -54,20 +54,18 @@ namespace Microsoft.AspNetCore.Razor.Language
                 return false;
             }
 
-            return descriptorX != null &&
+            return
                 string.Equals(descriptorX.Name, descriptorY.Name, _stringComparison) &&
-                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal) &&
-                Enumerable.SequenceEqual(descriptorX.Diagnostics, descriptorY.Diagnostics);
+                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal);
         }
 
         /// <inheritdoc />
         public virtual int GetHashCode(AllowedChildTagDescriptor descriptor)
         {
-            var hashCodeCombiner = HashCodeCombiner.Start();
-            hashCodeCombiner.Add(descriptor.Name, _stringComparer);
-            hashCodeCombiner.Add(descriptor.DisplayName, StringComparer.Ordinal);
+            var hash = HashCodeCombiner.Start();
+            hash.Add(descriptor.Name, _stringComparer);
 
-            return hashCodeCombiner.CombinedHash;
+            return hash.CombinedHash;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Razor.Language
                 return false;
             }
 
-            return descriptorX != null &&
+            return
                 string.Equals(descriptorX.Kind, descriptorY.Kind, StringComparison.Ordinal) &&
                 descriptorX.IsIndexerStringProperty == descriptorY.IsIndexerStringProperty &&
                 descriptorX.IsEnum == descriptorY.IsEnum &&
@@ -61,7 +61,6 @@ namespace Microsoft.AspNetCore.Razor.Language
                 string.Equals(descriptorX.IndexerTypeName, descriptorY.IndexerTypeName, StringComparison.Ordinal) &&
                 string.Equals(descriptorX.Documentation, descriptorY.Documentation, StringComparison.Ordinal) &&
                 string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal) &&
-                Enumerable.SequenceEqual(descriptorX.Diagnostics, descriptorY.Diagnostics) &&
                 Enumerable.SequenceEqual(
                     descriptorX.Metadata.OrderBy(propertyX => propertyX.Key, StringComparer.Ordinal),
                     descriptorY.Metadata.OrderBy(propertyY => propertyY.Key, StringComparer.Ordinal));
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Razor.Language/TagHelperDescriptorComparer.cs
```diff
@@ -62,6 +62,7 @@ public virtual bool Equals(TagHelperDescriptor descriptorX, TagHelperDescriptor
             return descriptorX != null &&
                 string.Equals(descriptorX.Kind, descriptorY.Kind, StringComparison.Ordinal) &&
                 string.Equals(descriptorX.AssemblyName, descriptorY.AssemblyName, StringComparison.Ordinal) &&
+                string.Equals(descriptorX.Name, descriptorY.Name, StringComparison.Ordinal) &&
```
rynowak: This was a bug BTW, the name wasn't compared here.

pranavkm: Unrelated - for all of the collection comparisons below, we could compare their sizes first since they're all (?) IReadOnlyList types.

rynowak: I think we should wait for data - or add a perf test.

rynowak:
@NTaylorMullen - can you take a second look? I did more thingses

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
```diff
@@ -76,15 +75,7 @@ public virtual int GetHashCode(BoundAttributeDescriptor descriptor)
 
             var hashCodeCombiner = HashCodeCombiner.Start();
             hashCodeCombiner.Add(descriptor.Kind);
-            hashCodeCombiner.Add(descriptor.IsIndexerStringProperty);
-            hashCodeCombiner.Add(descriptor.IsEnum);
-            hashCodeCombiner.Add(descriptor.HasIndexer);
             hashCodeCombiner.Add(descriptor.Name, _stringComparer);
-            hashCodeCombiner.Add(descriptor.IndexerNamePrefix, _stringComparer);
-            hashCodeCombiner.Add(descriptor.TypeName, StringComparer.Ordinal);
```
ajaybhargavb: Should we keep TypeName?

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
```diff
@@ -76,15 +75,7 @@ public virtual int GetHashCode(BoundAttributeDescriptor descriptor)
...
-            hashCodeCombiner.Add(descriptor.IndexerTypeName, StringComparer.Ordinal);
-            hashCodeCombiner.Add(descriptor.Documentation, StringComparer.Ordinal);
-            hashCodeCombiner.Add(descriptor.DisplayName, StringComparer.Ordinal);
```
ajaybhargavb: Some comparers still add DisplayName to hashCode. Should be consistent one way or the other.

src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptorComparer.cs
```diff
@@ -54,10 +54,9 @@ private AllowedChildTagDescriptorComparer(bool caseSensitive = false)
                 return false;
             }
 
-            return descriptorX != null &&
```
rynowak: This null check is covered by one of the two conditions above.

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
rynowak: Keep in mind any time the hash is used, the equality check is also likely to be used. Hashing helps us reject things that are non-matches efficiently.

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
rynowak: Sure, I'll remove it from more places

src/Microsoft.AspNetCore.Razor.Language/AllowedChildTagDescriptorComparer.cs
```diff
@@ -54,10 +54,9 @@ private AllowedChildTagDescriptorComparer(bool caseSensitive = false)
...
+            return
                 string.Equals(descriptorX.Name, descriptorY.Name, _stringComparison) &&
-                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal) &&
-                Enumerable.SequenceEqual(descriptorX.Diagnostics, descriptorY.Diagnostics);
+                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal);
```
NTaylorMullen: I think this was an old bug about us not using the `_stringComparison` parameter here. Might as well fix now since we're touching it anyhow.

src/Microsoft.AspNetCore.Razor.Language/BoundAttributeDescriptorComparer.cs
NTaylorMullen: `TypeName` in this context doesn't have a lot of value because it's the parameters `TypeName` and not something that could globally represents the current bound attribute descriptor

src/Microsoft.AspNetCore.Razor.Language/RequiredAttributeDescriptorComparer.cs
```diff
@@ -63,19 +63,14 @@ private RequiredAttributeDescriptorComparer(bool caseSensitive = false)
                 descriptorX.ValueComparison == descriptorY.ValueComparison &&
                 string.Equals(descriptorX.Name, descriptorY.Name, _stringComparison) &&
                 string.Equals(descriptorX.Value, descriptorY.Value, StringComparison.Ordinal) &&
-                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal) &&
-                Enumerable.SequenceEqual(descriptorX.Diagnostics, descriptorY.Diagnostics);
+                string.Equals(descriptorX.DisplayName, descriptorY.DisplayName, StringComparison.Ordinal);
```
NTaylorMullen: ditto with `_stringComparison`, I wont leave any more comments like this. Looks like it's also a thing in TagHelperDescriptorComparer.

rynowak:
Updated

rynowak:
I'm ignoring the feedback about string comparers. We should delete all of the code that does non-standard things instead, but it's not related to this change.

Issue #2303, opened 2018-04-24
Microsoft.AspNetCore.Razor.Tools.ServerCommandTest.WritePidFile_WorksAsExpected test failure


pranavkm:
The build: https://ci3.dot.net/job/aspnet_Razor/job/dev/job/windows-Configuration_Release_prtest/67/
```
13:34:49   [xUnit.net 00:00:01.5014114]     Microsoft.AspNetCore.Razor.Tools.ServerCommandTest.WritePidFile_WorksAsExpected [FAIL]
13:34:49   Failed   Microsoft.AspNetCore.Razor.Tools.ServerCommandTest.WritePidFile_WorksAsExpected
13:34:50 RUNDOTNET : error Message:  [C:\Users\dotnet-bot\.dotnet\buildtools\korebuild\2.2.0-preview1-17037\KoreBuild.proj]
13:34:50    System.IO.IOException : The process cannot access the file 'C:\Users\dotnet-bot\.dotnet\pids\build\rzc-3168' because it is being used by another process.
13:34:50   Stack Trace:
13:34:50      at System.IO.FileSystem.DeleteFile(String fullPath)
13:34:50      at System.IO.File.Delete(String path)
13:34:50      at Microsoft.AspNetCore.Razor.Tools.ServerCommandTest.WritePidFile_WorksAsExpected() in /_/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs:line 55
13:34:50   Invalid argument(s).
```

pranavkm:
cc @ajaybhargavb

ajaybhargavb:
It shouldn't fail. It doesn't fail locally or on our CI machines. Looks like it only fails on our dotnetbot Windows x64. @pranavkm, is there something special about them? Could it be something like Windows defender messing with the pid file?

mkArtakMSFT:
As this is a test failure, I assume we can get it fixed during 2.1 timeline..

ajaybhargavb:
Looks like this sometimes fails on our CI machines as well. https://github.com/aspnet/Razor/issues/2310#issue-318227936.

ajaybhargavb:
7cba5ed59309b832aafe47f338399e3676b1afda

PR #2319, opened 2018-04-30
Fix flaky test failure in ServerCommandTest

ajaybhargavb:
#2303

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
@@ -25,7 +25,10 @@ namespace Microsoft.AspNetCore.Razor.Tools
         internal ServerCommand(Application parent, string pipeName, int? keepAlive = null)
             : this(parent)
         {
-            Pipe.Values.Add(pipeName);
+            if (!string.IsNullOrEmpty(pipeName))
+            {
+                Pipe.Values.Add(pipeName);
+            }
 
             if (keepAlive.HasValue)
             {
@@ -119,8 +122,21 @@ namespace Microsoft.AspNetCore.Razor.Tools
             dispatcher.Run();
         }
 
-        internal FileStream WritePidFile()
+        protected virtual FileStream WritePidFile()
         {
+            var path = GetPidFilePath(env => Environment.GetEnvironmentVariable(env));
+            return WritePidFile(path);
+        }
+
+        // Internal for testing.
+        internal virtual FileStream WritePidFile(string directoryPath)
+        {
+            if (string.IsNullOrEmpty(directoryPath))
+            {
+                // Invalid path. Bail.
+                return null;
+            }
+
             // To make all the running rzc servers more discoverable, We want to write the process Id and pipe name to a file.
             // The file contents will be in the following format,
             //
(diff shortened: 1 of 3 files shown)


test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -13,23 +13,23 @@ namespace Microsoft.AspNetCore.Razor.Tools
...
+            var path = Path.Combine(directoryPath, expectedFileName);
 
             var pipeName = Guid.NewGuid().ToString();
             var server = GetServerCommand(pipeName);
 
             // Act & Assert
             try
             {
-                using (var _ = server.WritePidFile())
+                using (var _ = server.WritePidFile(directoryPath))
```
ajaybhargavb: @pranavkm, I didn't need the Func<string, string> because we are passing the path directly.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -119,20 +119,8 @@ protected virtual void ExecuteServerCore(ConnectionHost host, CompilerHost compi
             dispatcher.Run();
         }
 
-        internal FileStream WritePidFile()
+        internal virtual FileStream WritePidFile()
```
pranavkm: `protected virtual`?

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
pranavkm: That was to verify the path calculation. Would be good add a test to cover that too

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -47,10 +47,10 @@ public void WritePidFile_WorksAsExpected()
             }
             finally
             {
-                // Delete the file in case the test fails.
-                if (File.Exists(path))
+                // Cleanup after the test.
+                if (Directory.Exists(directoryPath))
                 {
-                    File.Delete(path);
+                    Directory.Delete(directoryPath);
```
pranavkm: Might have to say `recursive: true`. I don't think `Directory.Delete` deletes files by default

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
ajaybhargavb: Added.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -133,7 +149,26 @@ internal FileStream WritePidFile()
...
+                writer.Write(content);
+            }
+
+            return fileStream;
+        }
+
+        // Internal for testing.
+        internal virtual string GetPidFilePath(Func<string, string> getEnvironmentVariable)
+        {
+            var path = getEnvironmentVariable?.Invoke("DOTNET_BUILD_PIDFILE_DIRECTORY");
```
pranavkm: Why would `getEnvironmentVariable` be null?

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
```diff
@@ -133,7 +149,26 @@ internal FileStream WritePidFile()
...
             if (string.IsNullOrEmpty(path))
             {
                 var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
```
pranavkm: `var homePath = getEnvironmentVariable(homeEnvVariable);`

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -47,15 +47,44 @@ public void WritePidFile_WorksAsExpected()
...
             }
         }
 
-        private ServerCommand GetServerCommand(string pipeName)
+        [Fact]
+        public void GetPidFilePath_ReturnsCorrectDefaultPath()
+        {
+            // Arrange
+            var homeEnvVariable = PlatformInformation.IsWindows ? "USERPROFILE" : "HOME";
+            var expectedPath = Path.Combine(Environment.GetEnvironmentVariable(homeEnvVariable), ".dotnet", "pids", "build");
```
pranavkm: Use a test function rather than do this especially since you control the result of `getEnv` now

test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -47,15 +47,44 @@ public void WritePidFile_WorksAsExpected()
...
+
+            // Act
+            var directoryPath = server.GetPidFilePath(getEnvironmentVariable: null);
+
+            // Assert
+            Assert.Equal(expectedPath, directoryPath);
+        }
+
+        [Fact]
+        public void GetPidFilePath_UsesEnvironmentVariablePathIfSpecified()
```
pranavkm: Also add a test for when `getEnvVariable` returns null.

src/Microsoft.AspNetCore.Razor.Tools/ServerCommand.cs
ajaybhargavb: It won't. Just being defensive. Better than letting it throw.

pranavkm: Nah

ajaybhargavb:
🆙 📅

PR #2327, opened 2018-05-02
Allow build server to run on net46

### LINKED ISSUE #2325: Razor Build Server is not used by VS
The check [here](https://github.com/aspnet/Razor/blob/60b6ec047df29e8345bf2625c6f567e7622d0f88/src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs#L127) will always fail when we're running on desktop MSBuild - ie in Visual Studio. We're running on .NET Framework and so this API is not defined. 

This means that since we added the necessary security we haven't been able to use build server at all in Visual Studio.

The fix for this is to add ifdefs and add the ACL/elevation checks on .NET Framework. We already cross-compile this assembly.

ajaybhargavb:
Fixes #2325 

For security reasons, we were not starting the build server whenever `PipeOptions.CurrentlyUserOnly` isn't available. But when running on desktop .NET Framework aka VS on Windows, that flag will never be available causing the build server to not start in VS builds. The fix is to use the same workaround used by Roslyn to verify pipe ownership without relying on the PipeOptions flag.

I'm now working on manually verifying if this fix works in VS.

@rynowak @mkArtakMSFT


src/Microsoft.AspNetCore.Razor.Tasks/DotnetToolTask.cs
```diff
@@ -124,14 +124,16 @@ public override void Cancel()
...
         {
+#if !NET46
             if (!SuppressCurrentUserOnlyPipeOptions && !Enum.IsDefined(typeof(PipeOptions), PipeOptionCurrentUserOnly))
             {
                 // For security reasons, we don't want to spin up a server that doesn't
                 // restrict requests only to the current user.
                 result = -1;
 
-                return false;
+                return ForceServer;
```
rynowak: Was this just a bug that was preventing the force flag for working?

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -80,6 +90,47 @@ protected virtual void Dispose(bool disposing)
...
+        /// <summary>
+        /// Check to ensure that the named pipe server we connected to is owned by the same
+        /// user.
+        /// </summary>
+        /// <remarks>
+        /// The type is embedded in assemblies that need to run cross platform.  While this particular
+        /// code will never be hit when running on non-Windows platforms it does need to work when
+        /// on Windows.  To facilitate that we use reflection to make the check here to enable it to
+        /// compile into our cross plat assemblies.
+        /// </remarks>
```
rynowak: This comment isn't relevant. We ifdef this

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -80,6 +90,47 @@ protected virtual void Dispose(bool disposing)
...
+                    var currentIdentity = WindowsIdentity.GetCurrent();
+                    var currentOwner = currentIdentity.Owner;
+                    var remotePipeSecurity = GetPipeSecurity(pipeStream);
+                    var remoteOwner = remotePipeSecurity.GetOwner(typeof(SecurityIdentifier));
+                    return currentOwner.Equals(remoteOwner);
+                }
+
+                // We don't need to verify on non-windows as that will be taken care of by the
+                // PipeOptions.CurrentUserOnly flag.
+                return true;
```
rynowak: This should return false, we don't expect this to be called on non-windows.

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -80,6 +90,47 @@ protected virtual void Dispose(bool disposing)
...
+        {
+            try
+            {
+                if (PlatformInformation.IsWindows)
+                {
+                    var currentIdentity = WindowsIdentity.GetCurrent();
+                    var currentOwner = currentIdentity.Owner;
+                    var remotePipeSecurity = GetPipeSecurity(pipeStream);
+                    var remoteOwner = remotePipeSecurity.GetOwner(typeof(SecurityIdentifier));
+                    return currentOwner.Equals(remoteOwner);
```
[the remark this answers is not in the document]
ajaybhargavb: Just checked. Started the build server from an admin shell, tried to build from a non-elevated shell and the build failed 👍 .

build/buildpipeline/windows.groovy
```diff
@@ -2,7 +2,7 @@
 
 // 'node' indicates to Jenkins that the enclosed block runs on a node that matches
 // the label 'windows-with-vs'
-simpleNode('Windows_NT','latest') {
+simpleNode('Windows.10.Enterprise.RS3.ASPNET') {
```
rynowak: I don't really know what this is.

ajaybhargavb: To fix the dotnetbot CI test failure. This fix is already present in dev.

pranavkm: It's an newer image with VS 15.6 or later. The other one has super ancient VS installed.

src/Microsoft.AspNetCore.Razor.Tools/Client.cs
```diff
@@ -80,6 +90,41 @@ protected virtual void Dispose(bool disposing)
...
+        /// Check to ensure that the named pipe server we connected to is owned by the same
+        /// user.
+        /// </summary>
+        private static bool CheckPipeConnectionOwnership(NamedPipeClientStream pipeStream)
+        {
+            try
+            {
+                if (PlatformInformation.IsWindows)
+                {
+                    var currentIdentity = WindowsIdentity.GetCurrent();
```
mkArtakMSFT: The `currentIdentity` is disposable, so you should wrap it in a using statement

PR #2328, opened 2018-05-02
Fixes for the document window state

rynowak:


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -212,11 +212,9 @@ public ProjectState WithWorkspaceProject(Project workspaceProject)
             {
                 difference |= ProjectDifference.WorkspaceProjectRemoved;
             }
-            else if (
-                WorkspaceProject?.Id != workspaceProject?.Id ||
-                WorkspaceProject?.Version != workspaceProject?.Version)
+            else if (WorkspaceProject?.Id == workspaceProject?.Id)
```
rynowak: This logic was wrong. We actually only want to consider projects where the ID is the same.

Issue #2320, opened 2018-05-01
Flaky test: WritePidFile_WorksAsExpected


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=458610&buildTypeId=Releases_21Public_HighSierraOsxUniverse) occasionally with the following error:
```
Assert.Equal() Failure
Expected: String[] ["86330", "rzc", "/Users/aspnetci/Documents/buildAgent/work/33bdfc1c"..., "3db88ac6-de92-4889-b937-0a4f414b3564"]
Actual:   String[] [""]
   at Microsoft.AspNetCore.Razor.Tools.ServerCommandTest.WritePidFile_WorksAsExpected() in /_/test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs:line 20
```

mkArtakMSFT:
@ajaybhargavb , can you please handle this? Thanks!

ajaybhargavb:
This was fixed in dev. https://github.com/aspnet/Razor/pull/2319. I'm going to just skip this test in 2.1 because the fix involved refactoring product code.

ajaybhargavb:
b2c59700d5a03eb43ba1b145ee7d3b55893e621e

PR #2329, opened 2018-05-02
Skip flaky test

ajaybhargavb:
#2320 

@ryanbrandenburg FYI


test/Microsoft.AspNetCore.Razor.Tools.Test/ServerCommandTest.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Razor.Tools
 {
     public class ServerCommandTest
     {
-        [Fact]
+        [Fact(Skip = "https://github.com/aspnet/Razor/issues/2320")]
```
pranavkm: This was fixed in dev which requires source changes. We know the cause of the flakiness - it's a test issue, so chosing to skip it.

PR #2330, opened 2018-05-03
Remove ProjectService requirements for Razor text buffer initialization.

NTaylorMullen:
- Now that we have our own content type the editor does the work of determining if a Razor file opened in a core project should flow to our code.


NTaylorMullen:
Reviewed in person.

PR #2331, opened 2018-05-03
Add more UI for document generation

rynowak:


rynowak:
Mostly an FYI - will merge when CI passes.

PR #2339, opened 2018-05-08
Reorganize file tracker

rynowak:
This is precursor to a much bigger change. Pushing out some of the
changes that have broad impact early to make it easier to review the
real change.


src/Microsoft.VisualStudio.Editor.Razor/Documents/DefaultFileChangeTracker.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.Editor.Razor.Documents
```
rynowak: There's going to be a lot lot lot more infrastructure soon oriented around this concern. This seems worth its own namespace.

rynowak:
FYI only - will merge once CI passes

Issue #2336, opened 2018-05-07
Forcing reparses occasionally returns stale data.


NTaylorMullen:
At edit time when the project system is being created if more than one `QueueReparse`s happen around the same time only the first one will have its results returned. This is due to how we calculate the "latest" change reference [here](https://github.com/aspnet/Razor/blob/a9e7e4668dd0b4432bb504327d6f716f71ff10f5/src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs#L379-L381).

NTaylorMullen:
7c70207594d337f41029e3abc287eb6482eaecc0

PR #2340, opened 2018-05-08
Made the Visual Studio parser smarter about overlapping reparse requests.

NTaylorMullen:
- Prior to this the parser would think that a non-latest reparse request was the latest because the only way we would check to see if a change reference was "latest" would be to do an equality check on the snapshot and `SourceChange`; the issue here was that `SourceChange` was null but the snapshots were the same. Problems could arise with this due to project context changes.
- Added tests to validate the new reparse behavior.
- Renamed `Edit` in the `BackgroundParser` to `ChangeReference` also refactored all the "edit" text in `BackgroundParser` to be `ChangerReference` like.
- Added a new event args to be the DTO between the internal and external parser implementations. This is how we could pas additional information in order to determine "latest" change reference.

#2336


NTaylorMullen:
Going to get this in.

PR #2341, opened 2018-05-09
Track the open/closed state of documents

rynowak:
This change intoduces content changes to our project snapshots. We now
know the open/closed state of documents that are initialized by the
Razor project system and listen to the correct data source based on
whether the file is open in the editor.

There are a few other random improvements in here as well like a
workaround for the upcoming name change to our OOP client type.


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostDocument.cs
```diff
@@ -30,10 +30,10 @@ public HostDocument(string filePath, string targetPath)
 
         public override bool Equals(object obj)
         {
-            return base.Equals(obj as DocumentSnapshot);
```
rynowak: This was wrong before. It's unused, but it doesn't seem valuable to delete.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectDifference.cs
```diff
@@ -13,6 +13,8 @@ internal enum ProjectDifference
         WorkspaceProjectAdded = 2,
         WorkspaceProjectRemoved = 4,
         WorkspaceProjectChanged = 8,
-        DocumentsChanged = 16,
+        DocumentAdded = 16,
+        DocumentRemoved = 32,
+        DocumentChanged = 64,
```
rynowak: We've got two enums here that basically do the same thing. I don't see it as a high priority to clean them up, but worth pointing out.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManagerBase.cs
```diff
@@ -11,6 +12,15 @@ internal abstract class ProjectSnapshotManagerBase : ProjectSnapshotManager
...
         public abstract void DocumentAdded(HostProject hostProject, HostDocument hostDocument);
 
+        // Yeah this is kinda ugly.
+        public abstract void DocumentOpened(string projectFilePath, string documentFilePath, SourceTextContainer textContainer);
+
+        public abstract void DocumentClosed(string projectFilePath, string documentFilePath, TextLoader textLoader);
+
+        public abstract void DocumentChanged(string projectFilePath, string documentFilePath, TextLoader textLoader);
+
+        public abstract void DocumentChanged(string projectFilePath, string documentFilePath, SourceText sourceText);
```
rynowak: I don't love the design of these APIs. I feel like we badly want something like `ProjectId` `DocumentId`. It might be time to just start using them.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/WorkspaceProjectSnapshotChangeTrigger.cs
```diff
@@ -11,11 +13,21 @@ internal class WorkspaceProjectSnapshotChangeTrigger : ProjectSnapshotChangeTrig
     {
         private ProjectSnapshotManagerBase _projectManager;
 
+        public int ProjectChangeDelay { get;  set; } = 3 * 1000;
```
rynowak: OK so go with me a sec. We need some kind of buffer on workspace changes because otherwise we do a literal ton of work trying to initialize state on the critical path for perf. This is just a simple temporal buffer, and it works pretty nicely form what I've been able to see.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentSnapshot.cs
```diff
@@ -12,8 +14,16 @@ internal abstract class DocumentSnapshot
...
+
+        public abstract Task<VersionStamp> GetTextVersionAsync();
+
         public abstract Task<RazorCodeDocument> GetGeneratedOutputAsync();
 
+        public abstract bool TryGetText(out SourceText result);
+
+        public abstract bool TryGetTextVersion(out VersionStamp result);
+
         public abstract bool TryGetGeneratedOutput(out RazorCodeDocument result);
```
ajaybhargavb: Just an idea. Would this benefit from tracking `Imports` as well? Seems like something that would go with the document.

rynowak: Yep. I figured that if there was someone else helping me with this, that might be a good thing for him to take on.

ajaybhargavb: Then it's a good thing that someone is already on it. 👍

rynowak:
Reviewed the changes to this in person with @ajaybhargavb and @NTaylorMullen

Issue #2345, opened 2018-05-11
Add a way to specify the tag helpers and imports when processing a document


ajaybhargavb:
Currently when we call Engine.Process(RazorProjectItem) the tag helpers are computed everytime from the `ITagHelperFeature`. We want to be able to just specify the tag helpers to use when processing a document and the Engine should prefer that over the feature. This is useful in design time scenarios when we have already computed the tag helpers in the background.

~We also want to do this for imports but I'll file that separately.~
We want to do something similar for imports.

PR #2346, opened 2018-05-11
Added a taghelpers overload to Process and ProcessDesignTime

ajaybhargavb:
Fixes https://github.com/aspnet/Razor/issues/2345

We want to have a way to specify the taghelper descriptors to use while processing a specific document.
- Added an overload to Process and ProcessDesignTime to take in a list `TagHelperDescriptor`s
- Added the corresponding `CreateCodeDocumentCore` overload
- Added `GetTagHelpers` and `SetTagHelpers` extension methods for CodeDocument
- Added the necessary plumbing to use the taghelpers from the CodeDocument when available and fallback logic.
- Added/updated tests

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
@@ -68,10 +68,23 @@ namespace Microsoft.AspNetCore.Razor.Language
             var importItems = importFeature.GetImports(projectItem);
             var importSourceDocuments = GetImportSourceDocuments(importItems);
 
+            return CreateCodeDocumentCore(sourceDocument, importSourceDocuments, tagHelpers: null);
+        }
+
+        internal override RazorCodeDocument CreateCodeDocumentCore(RazorSourceDocument sourceDocument, IReadOnlyList<RazorSourceDocument> importSourceDocuments, IReadOnlyList<TagHelperDescriptor> tagHelpers)
+        {
+            if (sourceDocument == null)
+            {
+                throw new ArgumentNullException(nameof(sourceDocument));
+            }
+
             var parserOptions = GetRequiredFeature<IRazorParserOptionsFactoryProjectFeature>().Create(ConfigureParserOptions);
             var codeGenerationOptions = GetRequiredFeature<IRazorCodeGenerationOptionsFactoryProjectFeature>().Create(ConfigureCodeGenerationOptions);
 
-            return RazorCodeDocument.Create(sourceDocument, importSourceDocuments, parserOptions, codeGenerationOptions);
+            var codeDocument = RazorCodeDocument.Create(sourceDocument, importSourceDocuments, parserOptions, codeGenerationOptions);
+            codeDocument.SetTagHelpers(tagHelpers);
+
+            return codeDocument;
         }
 
         protected override RazorCodeDocument CreateCodeDocumentDesignTimeCore(RazorProjectItem projectItem)
@@ -87,10 +100,23 @@ namespace Microsoft.AspNetCore.Razor.Language
             var importItems = importFeature.GetImports(projectItem);
             var importSourceDocuments = GetImportSourceDocuments(importItems, suppressExceptions: true);
 
+            return CreateCodeDocumentDesignTimeCore(sourceDocument, importSourceDocuments, tagHelpers: null);
+        }
+
+        internal override RazorCodeDocument CreateCodeDocumentDesignTimeCore(RazorSourceDocument sourceDocument, IReadOnlyList<RazorSourceDocument> importSourceDocuments, IReadOnlyList<TagHelperDescriptor> tagHelpers)
+        {
+            if (sourceDocument == null)
+            {
+                throw new ArgumentNullException(nameof(sourceDocument));
+            }
+
             var parserOptions = GetRequiredFeature<IRazorParserOptionsFactoryProjectFeature>().Create(ConfigureDesignTimeParserOptions);
             var codeGenerationOptions = GetRequiredFeature<IRazorCodeGenerationOptionsFactoryProjectFeature>().Create(ConfigureDesignTimeCodeGenerationOptions);
 
-            return RazorCodeDocument.Create(sourceDocument, importSourceDocuments, parserOptions, codeGenerationOptions);
+            var codeDocument = RazorCodeDocument.Create(sourceDocument, importSourceDocuments, parserOptions, codeGenerationOptions);
+            codeDocument.SetTagHelpers(tagHelpers);
+
+            return codeDocument;
         }
 
         protected override void ProcessCore(RazorCodeDocument codeDocument)
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
@@ -15,18 +15,23 @@ namespace Microsoft.AspNetCore.Razor.Language
             var syntaxTree = codeDocument.GetSyntaxTree();
             ThrowForMissingDocumentDependency(syntaxTree);
 
-            var feature = Engine.Features.OfType<ITagHelperFeature>().FirstOrDefault();
-            if (feature == null)
+            var descriptors = codeDocument.GetTagHelpers();
+            if (descriptors == null)
             {
-                // No feature, nothing to do.
-                return;
+                var feature = Engine.Features.OfType<ITagHelperFeature>().FirstOrDefault();
+                if (feature == null)
+                {
+                    // No feature, nothing to do.
+                    return;
+                }
+
+                descriptors = feature.GetDescriptors();
             }
 
             // We need to find directives in all of the *imports* as well as in the main razor file
             //
             // The imports come logically before the main razor file and are in the order they
             // should be processed.
-            var descriptors = feature.GetDescriptors();
             var visitor = new DirectiveVisitor(descriptors);
             var imports = codeDocument.GetImportSyntaxTrees();
             if (imports != null)
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
@@ -29,6 +29,26 @@ namespace Microsoft.AspNetCore.Razor.Language
             document.Items[typeof(TagHelperDocumentContext)] = context;
         }
 
+        internal static IReadOnlyList<TagHelperDescriptor> GetTagHelpers(this RazorCodeDocument document)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            return (document.Items[typeof(TagHelpersHolder)] as TagHelpersHolder)?.TagHelpers;
+        }
+
+        internal static void SetTagHelpers(this RazorCodeDocument document, IReadOnlyList<TagHelperDescriptor> tagHelpers)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            document.Items[typeof(TagHelpersHolder)] = new TagHelpersHolder(tagHelpers);
+        }
+
         public static RazorSyntaxTree GetSyntaxTree(this RazorCodeDocument document)
         {
             if (document == null)
@@ -167,6 +187,16 @@ namespace Microsoft.AspNetCore.Razor.Language
             }
 
             public IReadOnlyList<RazorSyntaxTree> SyntaxTrees { get; }
+        }
+
+        private class TagHelpersHolder
+        {
+            public TagHelpersHolder(IReadOnlyList<TagHelperDescriptor> tagHelpers)
+            {
+                TagHelpers = tagHelpers;
+            }
+
+            public IReadOnlyList<TagHelperDescriptor> TagHelpers { get; }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
@@ -35,6 +35,18 @@ namespace Microsoft.AspNetCore.Razor.Language
             return codeDocument;
         }
 
+        internal virtual RazorCodeDocument Process(RazorSourceDocument source, IReadOnlyList<RazorSourceDocument> importSources, IReadOnlyList<TagHelperDescriptor> tagHelpers)
+        {
+            if (source == null)
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            var codeDocument = CreateCodeDocumentCore(source, importSources, tagHelpers);
+            ProcessCore(codeDocument);
+            return codeDocument;
+        }
+
         public virtual RazorCodeDocument ProcessDesignTime(RazorProjectItem projectItem)
         {
             if (projectItem == null)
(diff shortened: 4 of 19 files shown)


src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -47,10 +59,26 @@ public virtual RazorCodeDocument ProcessDesignTime(RazorProjectItem projectItem)
             return codeDocument;
         }
 
+        internal virtual RazorCodeDocument ProcessDesignTime(RazorProjectItem projectItem, IReadOnlyList<TagHelperDescriptor> tagHelpers)
```
rynowak: I'm not a super fan of this because it only solves the problem with tag helpers. We need to solve it for *ALL** of the source files as well as the tag helpers.

What if we just make `CreateCodeDocumentDesignTime` `protected internal` ?

ajaybhargavb: > What if we just make CreateCodeDocumentDesignTime protected internal ?

I considered it but `protected internal` is protected OR internal which means this would be public. Do we want to expose this?

ajaybhargavb: > ALL* of the source files

Could you expand on that? Do you mean imports?

ajaybhargavb: No I still think we are on the same page. That is what I envisioned for imports - It directly taking in a list of `RazorSourceDocument`. As I commented on the issue for this, I was going to do that separately. As part of adding imports tracking to `DocumentSnapshot`. But if you think it is worth having ProjectEngine support for imports in this PR, I'll update it. Let me know if I still don't make any sense.

ajaybhargavb: Sure I'll update it. But for now the only change needed in `DocumentGeneratedOutputTracker` to make tag helpers work was https://github.com/aspnet/Razor/pull/2346/files#diff-277383bfb275d15d91a911ee7b04ee3bR95.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultDocumentSnapshot.cs
```diff
@@ -27,14 +27,19 @@ public DefaultDocumentSnapshot(ProjectSnapshot project, DocumentState state)
             State = state;
         }
 
-        public ProjectSnapshot Project { get; }
+        public DefaultProjectSnapshot Project { get; }
```
ajaybhargavb: This is just setting up for the follow up PR.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
```diff
@@ -62,22 +65,29 @@ public DocumentGeneratedOutputTracker Fork()
...
+                var tagHelperDifference = new HashSet<TagHelperDescriptor>(TagHelperDescriptorComparer.Default);
+                tagHelperDifference.UnionWith(_older._tagHelpers);
+                tagHelperDifference.SymmetricExceptWith(tagHelpers);
+
+                var importDifference = new HashSet<ImportItem>();
+                importDifference.UnionWith(_older._imports);
+                importDifference.SymmetricExceptWith(imports);
 
-                if (difference.Count == 0)
+                if (tagHelperDifference.Count == 0 && importDifference.Count == 0)
```
ajaybhargavb: This needs to also include checks for document version. That will be done in a follow up.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -293,7 +293,7 @@ public ProjectState WithWorkspaceProject(Project workspaceProject)
             var documents = new Dictionary<string, DocumentState>(FilePathComparer.Instance);
             foreach (var kvp in Documents)
             {
-                documents.Add(kvp.Key, kvp.Value.WithConfigurationChange());
+                documents.Add(kvp.Key, kvp.Value.WithWorkspaceProjectChange());
```
ajaybhargavb: This was a bug fix.

src/Microsoft.CodeAnalysis.Razor.Workspaces/SourceTextExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Text
+{
+    internal static class SourceTextExtensions
+    {
+        public static RazorSourceDocument GetRazorSourceDocument(this SourceText sourceText, string fileName)
```
ajaybhargavb: This is debatable. I couldn't add a `RazorSourceDocument.Create(SourceText)` overload because we can't take a dependency on CodeAnalysis.Text from Razor.Language and C# doesn't allow static extension methods. So I added this extension method instead. You can argue this doesn't provide much value. I could be convinced either way.

src/Microsoft.VisualStudio.Editor.Razor/Documents/EditorDocumentManagerListener.cs
```diff
@@ -52,7 +52,7 @@ private void ProjectManager_Changed(object sender, ProjectChangeEventArgs e)
                 case ProjectChangeKind.DocumentAdded:
                     {
                         var key = new DocumentKey(e.ProjectFilePath, e.DocumentFilePath);
-                        var document = _documentManager.GetOrCreateDocument(key, _onChangedOnDisk, _onChangedOnDisk, _onOpened, _onClosed);
+                        var document = _documentManager.GetOrCreateDocument(key, _onChangedOnDisk, _onChangedInEditor, _onOpened, _onClosed);
```
ajaybhargavb: This was a bug fix.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
```diff
@@ -66,12 +66,25 @@ protected override RazorCodeDocument CreateCodeDocumentCore(RazorProjectItem pro
 
             var importFeature = GetRequiredFeature<IImportProjectFeature>();
             var importItems = importFeature.GetImports(projectItem);
-            var importSourceDocuments = GetImportSourceDocuments(importItems);
+            var importSourceDocuments = GetImportSourceDocuments(importItems, suppressExceptions: false);
```
rynowak: What's the change? was this a bug?

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorTagHelperBinderPhase.cs
```diff
@@ -15,18 +15,23 @@ protected override void ExecuteCore(RazorCodeDocument codeDocument)
...
-                // No feature, nothing to do.
-                return;
+                var feature = Engine.Features.OfType<ITagHelperFeature>().FirstOrDefault();
+                if (feature == null)
+                {
+                    // No feature, nothing to do.
+                    return;
+                }
+
+                descriptors = feature.GetDescriptors();
```
rynowak: 👍  love it

src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
```diff
@@ -168,5 +188,15 @@ public IncludeSyntaxTreesHolder(IReadOnlyList<RazorSyntaxTree> syntaxTrees)
...
         }
+
+        private class TagHelpersHolder
+        {
+            public TagHelpersHolder(IReadOnlyList<TagHelperDescriptor> tagHelpers)
+            {
+                TagHelpers = tagHelpers;
+            }
+
+            public IReadOnlyList<TagHelperDescriptor> TagHelpers { get; set; }
```
rynowak: Doesn't need to be settable

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -47,10 +59,26 @@ public virtual RazorCodeDocument ProcessDesignTime(RazorProjectItem projectItem)
...
+            }
+
+            var codeDocument = CreateCodeDocumentDesignTimeCore(source, importSources, tagHelpers);
+            ProcessCore(codeDocument);
+            return codeDocument;
+        }
+
         protected abstract RazorCodeDocument CreateCodeDocumentCore(RazorProjectItem projectItem);
 
+        internal abstract RazorCodeDocument CreateCodeDocumentCore(RazorSourceDocument source, IReadOnlyList<RazorSourceDocument> importSources, IReadOnlyList<TagHelperDescriptor> tagHelpers);
```
rynowak: This is a breaking change. You can't add a new abstract members

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultDocumentSnapshot.cs
rynowak: Sure

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
```diff
@@ -16,6 +18,7 @@ internal class DocumentGeneratedOutputTracker
         private Task<RazorCodeDocument> _task;
         
         private IReadOnlyList<TagHelperDescriptor> _tagHelpers;
+        private IReadOnlyList<ImportItem> _imports;
```
rynowak: I'd expect this to have its own 'tracker'. It might be a little vague, but the pattern is that the 'tracker' is the holder of some optional state. 

If this isn't clear, or isn't useful then maybe this is a good sign that the design as it is doesn't add much value.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
rynowak: Cool beans. 

Yeah, you pushed this much further in this PR than I expected. It good that you are hyped about taking all of this on, but we should try to iterate faster.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
rynowak: I'm not sure that this detail is important right now, so feel free to ignore.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
```diff
@@ -86,13 +96,77 @@ public DocumentGeneratedOutputTracker Fork()
...
+            return sourceText.GetRazorSourceDocument(document.FilePath);
+        }
+
+        private async Task<IReadOnlyList<ImportItem>> GetImportsAsync(ProjectSnapshot project, DocumentSnapshot document)
+        {
+            var imports = new List<ImportItem>();
+            foreach (var snapshot in document.GetImports())
+            {
+                var versionStamp = await snapshot.GetTextVersionAsync();
+                imports.Add(new ImportItem(snapshot.FilePath, versionStamp, snapshot));
```
rynowak: Is it possible to avoid interacting on to 'snapshots' at this level of the system? I don't want to see the external API leak too far into the the implementation.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentImportsTracker.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                if (item.PhysicalPath == null)
+                {
+                    // This is a default import.
+                    var defaultImport = new DefaultImportDocumentSnapshot(project, item);
+                    imports.Add(defaultImport);
+                }
+                else
+                {
+                    var import = project.GetDocument(item.PhysicalPath);
```
rynowak: Note that this doesn't correctly deal with linked files today. I don't expect you to fix, it but I want to make sure you understand why that is.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentImportsTracker.cs
```diff
@@ -0,0 +1,160 @@
...
+            var importFeature = projectEngine.ProjectFeatures.OfType<IImportProjectFeature>().FirstOrDefault();
+            var projectItem = projectEngine.FileSystem.GetItem(document.FilePath);
+            var importItems = importFeature?.GetImports(projectItem).Where(i => i.Exists);
+            var imports = new List<DocumentSnapshot>();
+            foreach (var item in importItems)
+            {
+                if (item.PhysicalPath == null)
+                {
+                    // This is a default import.
+                    var defaultImport = new DefaultImportDocumentSnapshot(project, item);
```
rynowak: I still really don't get why this has to be backed by a snapshot. Why isn't this something more primitive like a `SourceText`.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectEngine.cs
ajaybhargavb: Nah I can revert this. `suppressExceptions` is a default param with value false anyway.

src/Microsoft.CodeAnalysis.Razor.Workspaces/SourceTextExtensions.cs
rynowak: I think this is fine. It's internal to our usage.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
ajaybhargavb: 👍 Will make it virtual

test/Microsoft.AspNetCore.Razor.Language.Test/IntegrationTests/CodeGenerationIntegrationTest.cs
```diff
@@ -948,9 +948,10 @@ private void RunRuntimeTagHelpersTest(IEnumerable<TagHelperDescriptor> descripto
             });
 
             var projectItem = CreateProjectItem();
+            var imports = GetImports(projectEngine, projectItem);
 
             // Act
-            var codeDocument = projectEngine.Process(projectItem);
+            var codeDocument = projectEngine.Process(RazorSourceDocument.ReadFrom(projectItem), imports, descriptors.ToList());
```
rynowak: This is such an obvious thing. I have no idea why we didn't do this sooner.

src/Microsoft.VisualStudio.Editor.Razor/Documents/EditorDocumentManagerListener.cs
rynowak: Add test?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
rynowak: Add test?

src/Microsoft.AspNetCore.Razor.Language/RazorCodeDocumentExtensions.cs
```diff
@@ -29,6 +29,26 @@ public static void SetTagHelperContext(this RazorCodeDocument document, TagHelpe
...
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            return (document.Items[typeof(TagHelpersHolder)] as TagHelpersHolder)?.TagHelpers;
+        }
+
+        public static void SetTagHelpers(this RazorCodeDocument document, IReadOnlyList<TagHelperDescriptor> tagHelpers)
```
NTaylorMullen: Lets mark both of these as internal. There could be some confusion from an end users perspective about what these methods do (do they calculate TagHelpers? how is it different from the taghelper feature? etc.)

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentImportsTracker.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            return _imports;
+        }
+
+        private IReadOnlyList<DocumentSnapshot> GetImportsCore(ProjectSnapshot project, DocumentSnapshot document)
+        {
+            var projectEngine = project.GetProjectEngine();
+            var importFeature = projectEngine.ProjectFeatures.OfType<IImportProjectFeature>().FirstOrDefault();
+            var projectItem = projectEngine.FileSystem.GetItem(document.FilePath);
+            var importItems = importFeature?.GetImports(projectItem).Where(i => i.Exists);
```
NTaylorMullen: You don't protect against this being null later. So, if there's no `IImportProjectFeature` your foreach below explodes with a null ref.

ajaybhargavb:
> Should also add tests for DocumentGeneratedOutputTracker and DocumentImportsTracker.cs

They are incomplete. I'll add tests later when we actually enable them.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentImportsTracker.cs
ajaybhargavb: Because anything outside the `Tracker` wouldn't have the knowledge of what an import is. I don't like the idea of making default imports first class on `DocumentSnapshot`. The only other way without doing that would be to introduce an `ImportSnapshot` type which can provide the SourceText and Version of the import and would be backed by a `DocumentSnapshot` for real imports and just text for default imports. Do you like this better?

ajaybhargavb:
🆙 📅

ajaybhargavb:

ajaybhargavb:
@rynowak signed off in person

ajaybhargavb:

PR #2352, opened 2018-05-15
Exclude Visual Studio packages from default NGPV ruleset

natemcmaster:
These packages are failing NGPV on stabilized builds on the pre-release dependency rule because these nupkgs use pre-release VS packages. These packages are not actually consumed as nupkgs by VS, so exclude them from NGPV's default rules.


natemcmaster:
FYI @muratg @mmitche - this is required to unblock stabilized builds

PR #2354, opened 2018-05-15
Add some benchmarks for project system

rynowak:
Adds a few benchmarks for things we're working on, also addressed a few
small perf issues that have an impact for large documents.


src/Microsoft.AspNetCore.Razor.Language/Legacy/SeekableTextReader.cs
```diff
@@ -47,24 +48,13 @@ public int Position
...
 
-        public override int Peek()
-        {
-            if (_current == null)
-            {
-                return -1;
-            }
-            return _current.Value;
-        }
+        public override int Peek() => _current;
```
rynowak: This method is extremely hot. It can't be inlined, but this helps.

PR #2358, opened 2018-05-17
Add `ProjectPathProvider` abstract for document tracker creation.

NTaylorMullen:
- First iteration of live share replaced the document tracker factory entirely; however, this will be prone to breaking changes in the future when me make changes to document tracker to not rely on a file path. To pre-emptively prevent breaking changes I added a project path provider that can be overridden in the live share case. Note that one big difference here between old and new is that instead of being a MEF service implementation for the project path resolution we're bringing that to the Workspace service level.
- Added tests to validate the two flows of the default project path provider.

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultProjectPathProvider.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultProjectPathProvider.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal class DefaultProjectPathProvider : ProjectPathProvider
+    {
+        private readonly TextBufferProjectService _projectService;
+
+        public DefaultProjectPathProvider(TextBufferProjectService projectService)
+        {
+            if (projectService == null)
+            {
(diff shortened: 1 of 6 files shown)


src/Microsoft.VisualStudio.Editor.Razor/ProjectPathProvider.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis.Host;
+using Microsoft.VisualStudio.Text;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal abstract class ProjectPathProvider : ILanguageService
+    {
+        public abstract bool TryGetProjectPath(ITextBuffer textBuffer, out string filePath);
```
NTaylorMullen: LiveShare needs to ask the Host for the project path. It will eventually use a cached value that's already be computed and stored on the text buffer.

NTaylorMullen:
Getting this in to unblock.

src/Microsoft.VisualStudio.Editor.Razor/DefaultProjectPathProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            if (textBuffer == null)
+            {
+                throw new ArgumentNullException(nameof(textBuffer));
+            }
+
+            var project = _projectService.GetHostProject(textBuffer);
+            if (project == null)
+            {
+                filePath = null;
```
ajaybhargavb: Probably worth having that `Debug.Fail` here.

PR #2361, opened 2018-05-20
Fix dispose logic for EditorDocument

rynowak:
This could throw when the document is disposed and wasn't open in the
editor.

Improved the tests to cover dispose as well for both cases.


rynowak:
Merging this since it's currently broken in dev.

Issue #2362, opened 2018-05-22
Own serializing TagHelperDescriptors when using TagHelperDescriptorJsonConverter. to Enable our JsonConverters to ignore case when reading properties 


NTaylorMullen:
In LiveShare scenarios custom converters can impact how our TagHelperDescriptors serialize (by default they serialize camel case); because of this when our json converter attempts to read a serialized `TagHelperDescriptor` we null ref because we're looking for pascal cased properties.

NTaylorMullen:
93d8a93498963e019e99c10aa24275e1a483811d

PR #2364, opened 2018-05-22
 Make TagHelperDescriptor deserialization case insensitive.

NTaylorMullen:
- We don't serialize TagHelperDescriptors so we don't control what casing property names are passed down to us. Instead of doing all the work to serialize TagHelperDescriptor I decided to make the deserialization process a little more flexible.
- Added a TagHelperDescriptor camel case serialization test.

#2362


ajaybhargavb:
The test is failing on Windows.

NTaylorMullen:
So looks like we need to fully control the rendering of a `TagHelperDescriptor`. When a default camel case resolver goes through a `Dictionary<string, string>` (our metadata) it camel cases the keys. Problem with that is we don't know if they're meant to be camel cased or not.

PR #2367, opened 2018-05-22
Don't run build server when building testapps in-place

ajaybhargavb:
This is one of the nice to have fixes for mq. I added a new property because just setting `UseRazorBuildServer` to false and overriding it will preventing us from testing a few scenarios.


ajaybhargavb:

PR #2376, opened 2018-05-29
Updates the Editor Nuget packages

AmadeusW:


NTaylorMullen:
@AmadeusW this looks great! Thanks for the PR (i'll bring the changes over shortly). Is there a time frame for when we should insert all of these changes into VS? I imagine there's still the requirement that us and @dpoeschl's bits go in at the same time yes?

AmadeusW:
@NTaylorMullen the sooner the better. Ultimately it needs to go to d15.8stg
My code is already in d15.8stgide and it is migrating to d15.8stg.

I'm testing Roslyn by building it locally in d15.8stgide until Roslyn code makes it to d15.8stg

rynowak:
Does it make sense for us to insert a build with these changes in d15.8stgide

PR #2386, opened 2018-06-05
[Design] Allow document tracker to be created without a project.

NTaylorMullen:
- Part of this work involved moving our Project lookup logic to use a `ProjectId`. The caveat to this approach is there were some scenarios when all we had was a file path so we needed to maintain both ways to lookup a project (with and without a file path).
- Updated `DefaultProjectSnapshotManager` to track projects by a dual key of projectids and file paths.
- Updated project system bits to properly create host projects with unchanging `ProjectId`s.
- Added a miscellaneous project for cases when a document tracker did not have a project path.
- Updated serializers to properly serialize project ids as part of a project handle.

#2350

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/DocumentKey.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/DocumentKey.cs
@@ -2,26 +2,27 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.CodeAnalysis.Razor.ProjectSystem;
 using Microsoft.Extensions.Internal;
 
 namespace Microsoft.CodeAnalysis.Razor
 {
     public struct DocumentKey : IEquatable<DocumentKey>
     {
-        public DocumentKey(string projectFilePath, string documentFilePath)
+        public DocumentKey(ProjectId projectId, string documentFilePath)
         {
-            ProjectFilePath = projectFilePath;
+            ProjectId = projectId;
             DocumentFilePath = documentFilePath;
         }
 
-        public string ProjectFilePath { get; }
+        public ProjectId ProjectId { get; }
 
         public string DocumentFilePath { get; }
 
         public bool Equals(DocumentKey other)
         {
             return
-                FilePathComparer.Instance.Equals(ProjectFilePath, other.ProjectFilePath) &&
+                FilePathComparer.Instance.Equals(ProjectId, other.ProjectId) &&
                 FilePathComparer.Instance.Equals(DocumentFilePath, other.DocumentFilePath);
         }
 
@@ -33,7 +34,7 @@ namespace Microsoft.CodeAnalysis.Razor
         public override int GetHashCode()
         {
             var hash = new HashCodeCombiner();
-            hash.Add(ProjectFilePath, FilePathComparer.Instance);
+            hash.Add(ProjectId);
             hash.Add(DocumentFilePath, FilePathComparer.Instance);
             return hash;
         }
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/DocumentKey.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/DocumentKey.cs
@@ -22,7 +22,7 @@ namespace Microsoft.CodeAnalysis.Razor
         public bool Equals(DocumentKey other)
         {
             return
-                FilePathComparer.Instance.Equals(ProjectId, other.ProjectId) &&
+                ProjectId == other.ProjectId &&
                 FilePathComparer.Instance.Equals(DocumentFilePath, other.DocumentFilePath);
         }
 
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
@@ -17,7 +17,17 @@ namespace Microsoft.CodeAnalysis.Razor
 
         public RazorProjectEngine Create(ProjectSnapshot project)
         {
-            return Create(project, RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath)), null);
+            RazorProjectFileSystem fileSystem;
+            if (project.FilePath != null)
+            {
+                fileSystem = RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath));
+            }
+            else
+            {
+                fileSystem = RazorProjectFileSystem.Empty;
+            }
+
+            return Create(project, fileSystem, null);
         }
 
         public RazorProjectEngine Create(ProjectSnapshot project, RazorProjectFileSystem fileSystem)
@@ -42,7 +52,17 @@ namespace Microsoft.CodeAnalysis.Razor
                 throw new ArgumentNullException(nameof(project));
             }
 
-            return Create(project, RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath)), configure);
+            RazorProjectFileSystem fileSystem;
+            if (project.FilePath != null)
+            {
+                fileSystem = RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath));
+            }
+            else
+            {
+                fileSystem = RazorProjectFileSystem.Empty;
+            }
+
+            return Create(project, fileSystem, configure);
         }
 
         public abstract RazorProjectEngine Create(ProjectSnapshot project, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure);
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
@@ -37,7 +37,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
         // Each entry holds a ProjectState and an optional ProjectSnapshot. ProjectSnapshots are
         // created lazily.
-        private readonly ProjectContainer _projects;
+        private readonly Dictionary<ProjectId, Entry> _projects;
         private readonly HashSet<string> _openDocuments;
 
         public DefaultProjectSnapshotManager(
@@ -71,7 +71,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
             _triggers = triggers.ToArray();
             Workspace = workspace;
 
-            _projects = new ProjectContainer();
+            _projects = new Dictionary<ProjectId, Entry>();
             _openDocuments = new HashSet<string>(FilePathComparer.Instance);
 
             for (var i = 0; i < _triggers.Length; i++)
@@ -113,7 +113,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
             _foregroundDispatcher.AssertForegroundThread();
 
-            if (_projects.TryGetValue(filePath, out var entry))
+            if (TryGetEntryFromFilePath(filePath, out var entry))
             {
                 if (entry.Snapshot == null)
                 {
@@ -197,8 +197,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 // Document updates can no-op.
                 if (!object.ReferenceEquals(state, entry.State))
                 {
-                    var projectKey = new ProjectKey(hostProject.FilePath, hostProject.Id);
-                    _projects[projectKey] = new Entry(state);
+                    _projects[hostProject.Id] = new Entry(state);
                     NotifyListeners(new ProjectChangeEventArgs(hostProject.Id, document.FilePath, ProjectChangeKind.DocumentAdded));
                 }
             }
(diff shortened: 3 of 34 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
```diff
@@ -59,11 +59,6 @@ internal class DefaultVisualStudioDocumentTracker : VisualStudioDocumentTracker
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(filePath));
             }
 
-            if (projectPath == null)
```
NTaylorMullen: So another approach here would be to not take in the project path but to take in the project path provider. With that approach we'd determine the project path on `Subscribe`. Could be a little nicer but wanted feedback.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -180,7 +180,16 @@ internal void StartParser()
             Debug.Assert(_projectEngine.Engine != null);
             Debug.Assert(_projectEngine.FileSystem != null);
 
-            var projectDirectory = Path.GetDirectoryName(_documentTracker.ProjectPath);
+            string projectDirectory;
```
NTaylorMullen: This all goes away once we move to document snapshots.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
```diff
@@ -17,7 +17,17 @@ internal abstract class ProjectSnapshotProjectEngineFactory : IWorkspaceService
...
+            if (project.FilePath != null)
+            {
+                fileSystem = RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath));
+            }
+            else
+            {
+                fileSystem = RazorProjectFileSystem.Empty;
+            }
+
+            return Create(project, fileSystem, null);
```
rynowak: Beautiful 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -652,5 +686,84 @@ public Entry(ProjectState state)
...
+                        return true;
+                    }
+                }
+
+                entry = null;
+                return false;
+            }
+        }
+
+        internal struct ProjectKey : IEquatable<ProjectKey>
```
rynowak: I have some concerns about this, because I expected that the projectId was **fundamental**. When would you ever have two projects with the same ID but different file paths?

I want to resolve this issue before getting too much further because I think this is central.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MiscellaneousProjectSnapshot.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Host;
+
+namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
+{
+    internal class MiscellaneousProjectSnapshot : ProjectSnapshot
```
rynowak: Does this duplicate `EphemeralProjectSnapshot`? what do you see as different between these types?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManagerBase.cs
```diff
@@ -13,13 +13,13 @@ internal abstract class ProjectSnapshotManagerBase : ProjectSnapshotManager
         public abstract void DocumentAdded(HostProject hostProject, HostDocument hostDocument, TextLoader textLoader);
 
         // Yeah this is kinda ugly.
```
rynowak: It just got less ugly

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
rynowak: I'm not sure I understand the difference... what do you see as the tradeoffs?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
```diff
@@ -180,7 +180,16 @@ internal void StartParser()
...
+            if (_documentTracker.ProjectPath != null)
+            {
+                projectDirectory = Path.GetDirectoryName(_documentTracker.ProjectPath);
+            }
+            else
+            {
+                projectDirectory = "/";
```
rynowak: How does this behave? Like... why does `/` work?

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
```diff
@@ -43,6 +45,7 @@ internal abstract class RazorProjectHostBase : OnceInitializedOnceDisposedAsync,
 
             _lock = new AsyncSemaphore(initialCount: 1);
             _currentDocuments = new Dictionary<string, HostDocument>(FilePathComparer.Instance);
+            _hostProjectId = ProjectId.CreateNewId();
```
rynowak: What value does this add by being persistent? What would happen if you created the ID on 'project add' and then threw it away on 'project remove'?

My concern is about reusing the same ID across say, a project reload (think about unload+reload from VS). That doesn't seem good, but I'm not sure it matters.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
NTaylorMullen: It's not a trade off question but more of a code cleanliness question. You could make the argument about the project changing after unsubscribe => subscribe but that's not a real scenario.

I think I just felt icky about allowing a null parameter in the constructor of the document tracker.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: Glad you brought that up. That's a large part of why I sent out a design PR first, didn't feel super confident in the entire direction.

So, it's not that you'd ever have two projects with same id but different file paths but more that you need to be able to look up a project via project id or file path because there are scenarios when you only have one of those two. Aka, I technically could have had two separate dictionaries tracking the associations with projects and that'd be the same. I just tried to normalize it here so we didn't have to do double the operations in cases where entries changed.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MiscellaneousProjectSnapshot.cs
NTaylorMullen: Only difference is the lacking file path. I made a separate type mostly for clarity in debug purposes. Aka, easy to differentiate projectless documents from "being initialized" project documents

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
rynowak: What would happen if you just didn't do any of that? linear search?

Looking up a project by filename isn't **fundamental** - it's not something we should be doing frequently.

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
NTaylorMullen: Oh ya, this is totally a untested lol; just put this here and got out a design PR quickly because I had questionable feelings about the direction 😄.

Wasn't super concerned about this piece of code either since it's mostly temporary.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MiscellaneousProjectSnapshot.cs
rynowak: What would it look like to unify them? Do you think it's worth a separate concept? Would this look like a good decision in hindsight?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: Ya I could, just felt a little rough iterating through and then checking `item.Value.State.HostProject.FilePath`. Didn't feel strongly.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
NTaylorMullen: Hu? if you unload a project this does get re-created; CPS calls dispose on our project hosts and re-instantiates them on reload.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManager.cs
```diff
@@ -17,6 +17,8 @@ internal abstract class ProjectSnapshotManager : ILanguageService
 
         public abstract ProjectSnapshot GetLoadedProject(string filePath);
```
ajaybhargavb: Should remove this overload if the one below can be used everywhere

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManager.cs
```diff
@@ -17,6 +17,8 @@ internal abstract class ProjectSnapshotManager : ILanguageService
...
 
+        public abstract ProjectSnapshot GetLoadedProject(ProjectId projectId);
+
         public abstract ProjectSnapshot GetOrCreateProject(string filePath);
```
ajaybhargavb: Can this take ProjectId?

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioDocumentTracker.cs
ajaybhargavb: It does kinda feel icky. I support the `ProjectPathProvider` approach as long it doesn't involve too much churn.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: SUre

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManager.cs
NTaylorMullen: Sadly not because in the case that a project doesn't exist we need to create an ephemeral project with the provided file path.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectSnapshotManager.cs
NTaylorMullen: Can't 😢 . There are cases when we just have the Roslyn project which then relies on us using the file path to lookup a snapshot

src/Microsoft.CodeAnalysis.Razor.Workspaces/DocumentKey.cs
```diff
@@ -2,26 +2,27 @@
...
-        public string ProjectFilePath { get; }
+        public ProjectId ProjectId { get; }
 
         public string DocumentFilePath { get; }
 
         public bool Equals(DocumentKey other)
         {
             return
-                FilePathComparer.Instance.Equals(ProjectFilePath, other.ProjectFilePath) &&
+                FilePathComparer.Instance.Equals(ProjectId, other.ProjectId) &&
```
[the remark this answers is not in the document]
NTaylorMullen: lmao, because I is smert

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
NTaylorMullen: > What is the case where we have a filepath but not an Id?

Roslyn Project => Razor Project lookup

src/Microsoft.VisualStudio.Editor.Razor/DefaultVisualStudioRazorParser.cs
rynowak: Ok, so for a design PR I don't have any real concerns about this - but we do need to figure out how it should behave in this case.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorProjectHostBase.cs
rynowak: ok dope

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/MiscellaneousProjectSnapshot.cs
rynowak: I still don't understand how these differ. These projects still have the same lifetime.

Issue #2353, opened 2018-05-15
DocumentClassifier should assign a class name when filepath is null


ajaybhargavb:
https://github.com/aspnet/Razor/blob/dev/src/Microsoft.AspNetCore.Mvc.Razor.Extensions/MvcViewDocumentClassifierPass.cs#L28

In the above line, when the file path is null, we set the generated class name to null. This makes the CodeWriter throw an exception when generating the code. I think we should fix this as not all Razor documents have file path, for example, default imports.

@NTaylorMullen @rynowak thoughts?

ajaybhargavb:
Just to be clear, this isn't specific to imports. The other scenario this would break is the following. I create a `RazorSourceDocument` out of a string and I pass it in to `Engine.Process`, I want the generated code for that. Right now that would throw an ArgNull.

rynowak:
Sounds like something we should fix 👍

ajaybhargavb:
ad07036020b5df8bd9173dc0ac0096912f0d7553

PR #2389, opened 2018-06-06
Use checksum for classname if filepath is null

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2353

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Text;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Intermediate;
 
@@ -31,7 +29,7 @@ namespace Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X
             {
                 // It's possible for a Razor document to not have a file path.
                 // Eg. When we try to generate code for an in memory document like default imports.
-                var checksum = BytesToString(codeDocument.Source.GetChecksum());
+                var checksum = Checksum.BytesToString(codeDocument.Source.GetChecksum());
                 @class.ClassName = $"AspNetCore_{checksum}";
             }
             else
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
```diff
@@ -25,7 +27,16 @@ public class MvcViewDocumentClassifierPass : DocumentClassifierPassBase
             @namespace.Content = "AspNetCore";
 
             var filePath = codeDocument.Source.RelativePath ?? codeDocument.Source.FilePath;
-            @class.ClassName = CSharpIdentifier.GetClassNameFromPath(filePath);
+            if (string.IsNullOrEmpty(filePath))
```
rynowak: Can we capture some info here about how/why this happens? What's the possible scenarios? The code seems fine

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
```diff
@@ -37,5 +48,22 @@ public class MvcViewDocumentClassifierPass : DocumentClassifierPassBase
...
+            }
+
+            var result = new StringBuilder(bytes.Length);
+            for (var i = 0; i < bytes.Length; i++)
+            {
+                // The x2 format means lowercase hex, where each byte is a 2-character string.
+                result.Append(bytes[i].ToString("x2"));
+            }
+
+            return result.ToString();
```
rynowak: Don't we already have this code elsewhere?

ajaybhargavb: Yup. But we don't have the IVT. So I thought if we have to duplicate in one place, I'll just duplicate it everywhere

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
ajaybhargavb: You mean add a comment?

src/Microsoft.AspNetCore.Mvc.Razor.Extensions.Version1_X/MvcViewDocumentClassifierPass.cs
rynowak: Shared file? We now have 4 copies of this code :(

NTaylorMullen:
Part of me wishes we had a default behavior of using the checksum (assuming its not expensive) and then in MVC case if file path is available override it again.

ajaybhargavb:
> Part of me wishes we had a default behavior of using the checksum

Maybe but there could be multiple implementations of `GetChecksum()` and some of those could be expensive. I don't think we can rely on them enough to make it the default behavior.

ajaybhargavb:
🆙 📅

Issue #2299, opened 2018-04-24
Completing C# inside single line @{} doesn't complete


NTaylorMullen:
When doing: `@{th}` and completing `this` the `is` is never inserted.

https://i.imgur.com/QrgltyC.gifv

![image](https://i.imgur.com/QrgltyC.gif)

This seems to be because we don't generate C# place holders in the middle of a single line `@{}`

VSTS Issue: https://devdiv.visualstudio.com/DevDiv/Project%20Cascade/_workitems/edit/602933 

/cc @ToddGrun

NTaylorMullen:
dc76027a7affd0a9fa4bac8bbaf07ed1effdf687

KirillOsenkov:
Confirmed this is indeed fixed in latest VSMac. Thanks!

Issue #2266, opened 2018-04-12
Add tests the verify that all user-provided C# code is in the generated code


rynowak:
We can easily add an invariant to our integration tests that prevents a class of issue where C# loses intellisense - by checking that all spans of user-provided C# code are properly mapped into the generated output.

Example here: https://github.com/aspnet/Razor/commit/059d61f1cfb65cca628a6d3369c28a42f510f7a1

rynowak:
@mkArtakMSFT - another great MQ idea

PR #2392, opened 2018-06-07
Generate line pragmas even for whitespace.

NTaylorMullen:
- We need to pre-emptively generate line pragmas to handle C# scenarios which are session based. For instance, if a C# completion session starts at a location without a pragma they capture the origination point; at which point if a user keeps typing we generate a line pragma drastically changing where the copmletion started resulting in a broken C# completion experience.
- Updated test files to reflect new behavior.

This is what this PR enables:
image

#2299

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeNodeWriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/CodeGeneration/DesignTimeNodeWriter.cs
@@ -88,31 +88,12 @@ namespace Microsoft.AspNetCore.Razor.Language.CodeGeneration
 
         public override void WriteCSharpCode(CodeRenderingContext context, CSharpCodeIntermediateNode node)
         {
-            var isWhitespaceStatement = true;
-            for (var i = 0; i < node.Children.Count; i++)
-            {
-                var token = node.Children[i] as IntermediateToken;
-                if (token == null || !string.IsNullOrWhiteSpace(token.Content))
-                {
-                    isWhitespaceStatement = false;
-                    break;
-                }
-            }
-
             IDisposable linePragmaScope = null;
             if (node.Source != null)
             {
-                if (!isWhitespaceStatement)
-                {
-                    linePragmaScope = context.CodeWriter.BuildLinePragma(node.Source.Value);
-                }
+                linePragmaScope = context.CodeWriter.BuildLinePragma(node.Source.Value);
 
                 context.CodeWriter.WritePadding(0, node.Source.Value, context);
-            }
-            else if (isWhitespaceStatement)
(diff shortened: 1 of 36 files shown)


rynowak:
The idea of this makes sense...

Can we find a way to add this invariant to our tests so that it's **impossible** to introduce a bug? For instance, would it be possible to add the invariant that **all** C# spans appear in our in output. This would make the principle behind this change broadly enforced, and prevent us from having any kind of bug like this in the future.

In principle, this is part of the fix for what you're working on

NTaylorMullen:
> Can we find a way to add this invariant to our tests so that it's impossible to introduce a bug? For instance, would it be possible to add the invariant that all C# spans appear in our in output. This would make the principle behind this change broadly enforced, and prevent us from having any kind of bug like this in the future.

Agree, figured that was covered by [this](https://github.com/aspnet/Razor/issues/2266)

rynowak:
So you're doing it then?

NTaylorMullen:
> So you're doing it then?

Not as part of this issue, It's not nearly a high enough priority. We've had this issue in every version of Razor and it was just recently reported. After we get through the 15.8 issues we can consider.

NTaylorMullen:
@dotnet-bot test OSX 10.12 Release Build please

NTaylorMullen:
@natemcmaster is KoreBuild in a borked state for OSX? Looks like the OSX builds keep failing to find dotnet

natemcmaster:
Yes. Run `build.cmd -update /t:noop` to get the latest korebuild

rynowak:
As I understand it, the problem statement is that you don't get intellisense in empty or whitespace-only C# regions correct? And this is more likely to occur in live-share scenarios due to the way the remote-editor interactions work correct?

So how do you know you fixed all of the cases? You surely fixed this for code blocks containing only whitespace. Are there others?

NTaylorMullen:
> As I understand it, the problem statement is that you don't get intellisense in empty or whitespace-only C# regions correct?

Partially correct, without me personally spending the time to dive into all the pieces that @ToddGrun did the issues seems to be scoped to whitespace pieces in a single line `@{}` only. I've tried out other whitespace scenarios such as `@functions {}` `@{ \r\n }` `@()` and they all work. The bet is that Roslyn behaves odd in the `@{}` scenario. @ToddGrun could probably explain more since he did the initial dive in the original VSTS issue.

NTaylorMullen:
> And this is more likely to occur in live-share scenarios due to the way the remote-editor interactions work correct?

Ya it could be exacerbated in the live share scenario because the completion logic is done on the host.

NTaylorMullen:
@rynowak and I talked offline. Going to leave this as is and get it in.

PR #2399, opened 2018-06-12
Add partial parsing for parenthesis

NTaylorMullen:
Putting this out for design in case there are any other cases I missed that I didn't think of.

This is part of a fix for #1255 - this change enables signature help in implicit expressions by improving the partial parsing. We're now smart enough about the contents of an implicit expression and attempt to balance parenthesis to determine if we should not full parse.

image

#1255

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
@@ -109,9 +109,19 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                 return HandleInsertion(target, lastChar.Value, change);
             }
 
+            if (IsAcceptableInsertionInBalancedParenthesis(target, change))
+            {
+                return PartialParseResultInternal.Accepted;
+            }
+
             if (IsAcceptableDeletion(target, change))
             {
                 return HandleDeletion(target, lastChar.Value, change);
+            }
+
+            if (IsAcceptableDeletionInBalancedParenthesis(target, change))
+            {
+                return PartialParseResultInternal.Accepted;
             }
 
             return PartialParseResultInternal.Rejected;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -216,8 +221,152 @@ private static bool IsAcceptableDeletion(Span target, SourceChange change)
...
         {
             return change.IsInsert &&
-                   (IsAcceptableEndInsertion(target, change) ||
-                   IsAcceptableInnerInsertion(target, change));
+                (IsAcceptableEndInsertion(target, change) ||
+                IsAcceptableInnerInsertion(target, change));
+        }
+
+        // Internal for testing
+        internal static bool IsAcceptableInsertionInBalancedParenthesis(Span target, SourceChange change)
```
NTaylorMullen: This PR contains changes from your old ones @rynowak but this method and the underlying call-throughs are the primary addition

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -216,8 +221,152 @@ private static bool IsAcceptableDeletion(Span target, SourceChange change)
...
+                if (content.Length > 0 && content[content.Length - 1] != '"')
+                {
+                    // Incomplete string literal may have consumed some of our parenthesis and usually occurr during auto-completion of '"' => '""'.
+                    if (!TryUpdateCountFromContent(content, ref updatedCount))
+                    {
+                        return false;
+                    }
+                }
+            }
+            else if (token.Type == CSharpSymbolType.CharacterLiteral)
```
NTaylorMullen: ridiculous I know but when typing `'` our tokenizer tokenizes the content after it and squashes it all together as a character literal that can result in tokens that look like `')`

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -216,8 +221,152 @@ private static bool IsAcceptableDeletion(Span target, SourceChange change)
...
+                    return true;
+                }
+            }
+
+            // Unbalanced parenthesis
+            return false;
+        }
+
+        // Internal for testing
+        internal static bool AtInsertionPoint(SourceChange change, CSharpSymbol currentToken)
```
rynowak: Should this be thought of as "contains insertion point"? 

why is it safe to assume that the insertion point can be at the start of a token but not the end?

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
```diff
@@ -216,8 +221,152 @@ private static bool IsAcceptableDeletion(Span target, SourceChange change)
...
+                        // Unbalanced parenthesis, i.e. @Foo)
+                        return false;
+                    }
+
+                    updatedCount--;
+                }
+            }
+
+            count = updatedCount;
+            return true;
```
rynowak: This code isn't as bad as a I expected. Is it worth pulling this out into a static method like `IsInsideParenthesis(IEnumerable<Token>, int position)`? for testability?

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
NTaylorMullen: The end is equivalent to the insertion point being at the next symbol.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ImplicitExpressionEditHandler.cs
NTaylorMullen: Hmm, could do that 👍

NTaylorMullen:
🆙 📅

Issue #2263, opened 2018-04-12
Improve testing strategy for the Razor Parser


rynowak:
Our strategy for testing the Razor parser leaves much to be desired. Writing new tests is very expensive, and understanding deltas when tests fail is tedious and error prone.

We should adopt a strategy similar to what we use for our integration tests. More details here: https://github.com/aspnet/Razor/issues/2211

The first step would be to define a framework for these tests (using serialization/baselines), and then pivot our existing tests in a series of steps to use baselines rather than in-code verification. Once this is done we can introduce more invariants in the tests and have confidence that these invariants are maintained across all of our parser tests.

Checklist:
- [x] CSharpVerbatimBlockTest (Ajay)
- [x] CSharpAutoCompleteTest (Ajay)
- [x] CSharpBlockTest (Ajay)
- [x] CSharpDirectivesTest (Ajay)
- [x] CSharpErrorTest (Ajay)
- [x] CSharpExplicitExpressionTest (Ajay)
- [x] CSharpImplicitExpressionTest (Ajay)
- [x] CSharpNestedStatementsTest (Ajay)
- [x] CSharpRazorCommentsTest (Ryan)
- [x] CSharpReservedWordsTest (Ryan)
- [x] CSharpSectionTest (Ryan)
- [x] CSharpSpecialBlockTest (Taylor)
- [x] CSharpStatementTest (Taylor)
- [x] CSharpTemplateTest (Taylor)
- [x] CSharpToMarkupSwitchTest (Ryan)
- [x] CSharpWhitespaceHandlingTest (Taylor)
- [x] HtmlAttributeTest (Ajay)
- [x] HtmlBlockTest (Ajay)
- [x] HtmlDocumentTest (Ajay)
- [x] HtmlErrorTest (Taylor)
- [x] HtmlTagsTest (Taylor)
- [x] HtmlToCodeSwitchTest (Ryan)
- [x] TagHelperBlockRewriterTest (Ajay)
- [x] TagHelperParseTreeRewriterTest (Ajay)
- [x] WhiteSpaceRewriterTest (Ryan)

rynowak:
@mkArtakMSFT - here's another great MQ item

ajaybhargavb:
b821ce8b8ef1d38d9e3a86af3910b31b1485ae34

ajaybhargavb:
21f2edce16187a04e8e6bef156cf563b52c78bf4

PR #2402, opened 2018-06-13
Convert parser tests to use baselines

ajaybhargavb:
#2263 

- Added the infrastructure to serialize and verify parser tests with baselines
- Updated CSharpVerbatimBlockTest and CSharpReservedWordsTest to use the new infrastructure

Upcoming changes: (Sending this separately so that the current PR stays easy to review)
- Make all the remaining parser tests use the new infrastructure
- Cleanup and remove unnecessary overloads in ParserTestBase

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParentChunkGenerator.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParentChunkGenerator.cs
@@ -47,10 +47,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 
             public void Accept(ParserVisitor visitor, Block block)
             {
-                for (var i = 0; i < block.Children.Count; i++)
-                {
-                    block.Children[i].Accept(visitor);
-                }
+                visitor.VisitDefault(block);
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserVisitor.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserVisitor.cs
@@ -5,12 +5,21 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal abstract class ParserVisitor
     {
-        protected virtual void VisitDefault(Block block)
+        public virtual void Visit(SyntaxTreeNode node)
+        {
+            node.Accept(this);
+        }
+
+        public virtual void VisitDefault(Block block)
         {
             for (var i = 0; i < block.Children.Count; i++)
             {
                 block.Children[i].Accept(this);
             }
+        }
+
+        public virtual void VisitDefault(Span span)
+        {
         }
 
         public virtual void VisitBlock(Block block)
@@ -44,35 +53,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             VisitDefault(block);
         }
 
-        public virtual void VisitExpressionSpan(ExpressionChunkGenerator chunkGenerator, Span span)
-        {
-        }
-
-        public virtual void VisitMarkupSpan(MarkupChunkGenerator chunkGenerator, Span span)
-        {
-        }
-
-        public virtual void VisitImportSpan(AddImportChunkGenerator chunkGenerator, Span span)
-        {
-        }
-
-        public virtual void VisitStatementSpan(StatementChunkGenerator chunkGenerator, Span span)
-        {
-        }
-
-        public virtual void VisitLiteralAttributeSpan(LiteralAttributeChunkGenerator chunkGenerator, Span span)
-        {
-        }
-
-        public virtual void VisitDirectiveToken(DirectiveTokenChunkGenerator chunkGenerator, Span block)
-        {
-        }
-
-        public virtual void VisitDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
-        {
-            VisitDefault(block);
-        }
-
         public virtual void VisitTemplateBlock(TemplateBlockChunkGenerator chunkGenerator, Block block)
         {
             VisitDefault(block);
@@ -88,16 +68,54 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
             VisitDefault(block);
         }
 
+        public virtual void VisitDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
+        {
+            VisitDefault(block);
+        }
+
+        public virtual void VisitExpressionSpan(ExpressionChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
+        public virtual void VisitMarkupSpan(MarkupChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
+        public virtual void VisitImportSpan(AddImportChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
+        public virtual void VisitStatementSpan(StatementChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
+        public virtual void VisitLiteralAttributeSpan(LiteralAttributeChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
+        public virtual void VisitDirectiveToken(DirectiveTokenChunkGenerator chunkGenerator, Span span)
+        {
+            VisitDefault(span);
+        }
+
         public virtual void VisitAddTagHelperSpan(AddTagHelperChunkGenerator chunkGenerator, Span span)
         {
+            VisitDefault(span);
         }
 
         public virtual void VisitRemoveTagHelperSpan(RemoveTagHelperChunkGenerator chunkGenerator, Span span)
         {
+            VisitDefault(span);
         }
 
         public virtual void VisitTagHelperPrefixDirectiveSpan(TagHelperPrefixDirectiveChunkGenerator chunkGenerator, Span span)
         {
+            VisitDefault(span);
         }
     }
 }
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpReservedWordsTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpReservedWordsTest.cs
@@ -7,6 +7,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     public class CSharpReservedWordsTest : CsHtmlCodeParserTestBase
     {
+        public CSharpReservedWordsTest()
+        {
+            UseBaselineTests = true;
+        }
+
         [Theory]
         [InlineData("namespace")]
         [InlineData("class")]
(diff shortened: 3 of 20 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserVisitor.cs
```diff
@@ -44,60 +53,69 @@ public virtual void VisitAttributeBlock(AttributeBlockChunkGenerator chunkGenera
             VisitDefault(block);
         }
 
-        public virtual void VisitExpressionSpan(ExpressionChunkGenerator chunkGenerator, Span span)
+        public virtual void VisitTemplateBlock(TemplateBlockChunkGenerator chunkGenerator, Block block)
```
ajaybhargavb: Just reordered a few things here.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpReservedWordsTest.cs
```diff
@@ -7,6 +7,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     public class CSharpReservedWordsTest : CsHtmlCodeParserTestBase
     {
+        public CSharpReservedWordsTest()
+        {
+            UseBaselineTests = true;
```
ajaybhargavb: I added this temporary flag to help convert these parser tests one by one instead of all at once.

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/ParserTests/CSharpVerbatimBlockTest/InnerImplicitExpressionAcceptsTrailingNewlineInDesignTimeMode.syntaxtree.txt
```diff
@@ -0,0 +1,8 @@
+Statement block - NullParentChunkGenerator - 8 - (0:0,0)
```
ajaybhargavb: Looks like adding the block length here breaks in crossplat when there is a newline involved. I'm looking in to it.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/IntializeTestFileAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+                {
+                    if (!string.IsNullOrEmpty(_previousMethod) &&
+                        !_previousMethod.Equals(methodUnderTest.Name, StringComparison.Ordinal))
+                    {
+                        // This is the start of a different Theory test. Reset the call count.
+                        _callCount = 0;
+                    }
+                    
+                    // Append the call count to the file name to differentiate test cases within a Theory test.
+                    ParserTestBase.FileName += _callCount++;
```
ajaybhargavb: Had to do this magic to generate different filenames for each case in our theory tests.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeVerifier.cs
```diff
@@ -0,0 +1,276 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    public static class SyntaxTreeNodeVerifier
```
ajaybhargavb: Most of the code in this file is from `IntermediateNodeVerifier`. I'll later think about sharing some of this code.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeWriter.cs
```diff
@@ -0,0 +1,231 @@
...
+        protected void WriteNewLine()
+        {
+            _writer.WriteLine();
+        }
+
+        protected void Write(object value)
+        {
+            if (value is string stringValue)
+            {
+                stringValue = stringValue.Replace(Environment.NewLine, "LF");
```
ajaybhargavb: Representing new lines within span content etc as LF so we don't compromise on proper indentation aka readability.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/IntializeTestFileAttribute.cs
ajaybhargavb: Looks like I'm gonna have to do more magic. Looks like Theory test cases aren't guaranteed to run in the same order. 😢

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/IntializeTestFileAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+        public override void Before(MethodInfo methodUnderTest)
+        {
+            if (typeof(ParserTestBase).GetTypeInfo().IsAssignableFrom(methodUnderTest.DeclaringType.GetTypeInfo()))
+            {
+                var typeName = methodUnderTest.DeclaringType.Name;
+                ParserTestBase.FileName = $"TestFiles/ParserTests/{typeName}/{methodUnderTest.Name}";
+
+                if (methodUnderTest.GetCustomAttributes(typeof(TheoryAttribute), inherit: false).Length > 0)
+                {
+                    if (!string.IsNullOrEmpty(_previousMethod) &&
```
ajaybhargavb: There seems to be a race here causing this if to always be true. I'm looking in to it.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParentChunkGenerator.cs
```diff
@@ -47,10 +47,7 @@ public override string ToString()
 
             public void Accept(ParserVisitor visitor, Block block)
             {
-                for (var i = 0; i < block.Children.Count; i++)
-                {
-                    block.Children[i].Accept(visitor);
-                }
```
rynowak: Is this just cleanup or was something wackyu going on?

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/ParserTests/CSharpVerbatimBlockTest/InnerImplicitExpressionAcceptsTrailingNewlineInDesignTimeMode.syntaxtree.txt
rynowak: Yeah we need a strategy for that. I suggest that we normalize it one way or the other.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeVerifier.cs
rynowak: Share what makes sense. Code reuse isn't an end goal in itself 😆

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/IntializeTestFileAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    public class IntializeTestFileAttribute : BeforeAfterTestAttribute
+    {
+        private static int _callCount = 0;
+        private static string _previousMethod;
```
rynowak: You will need to use `AsyncLocal` for these. That's likely why you're having threading issues.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParentChunkGenerator.cs
ajaybhargavb: Nothing wacky but this was written in a way that was not overridable. This change lets me override this behavior in the serializer. This doesn't affect the product code behavior in any way.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeSerializer.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            public TextWriter Writer { get; }
+
+            public override void Visit(SyntaxTreeNode node)
+            {
+                _visitor.Visit(node);
+                _writer.WriteLine();
+
+                if (node is Block block)
```
NTaylorMullen: Any reason this piece isn't just part of the `SyntaxTreeNodeWriter`? If you had that there wouldn't be any need for this walker class.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeWriter.cs
```diff
@@ -0,0 +1,231 @@
...
+            WriteSpan(span);
+        }
+
+        public override void VisitTagHelperBlock(TagHelperChunkGenerator chunkGenerator, Block block)
+        {
+            WriteBlock(block);
+            
+            if (block is TagHelperBlock tagHelperBlock)
+            {
+                // TODO: Should I add TagHelperBinding info?
```
NTaylorMullen: Probably smart to hash it and represent it as some value so we know if it changes. No need for the info to be verbose.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeWriter.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        public override void VisitCommentBlock(RazorCommentChunkGenerator chunkGenerator, Block block)
+        {
+            WriteBlock(block);
+        }
+
+        public override void VisitDirectiveBlock(DirectiveChunkGenerator chunkGenerator, Block block)
+        {
+            // TODO: Add more details
```
NTaylorMullen: 👍

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeSerializer.cs
ajaybhargavb: Because this and `SyntaxTreeNodeVerifier` both use the `SyntaxTreeNodeWriter` and these two have slightly different implementations of Visit.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/IntializeTestFileAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public override void Before(MethodInfo methodUnderTest)
+        {
+            if (typeof(ParserTestBase).GetTypeInfo().IsAssignableFrom(methodUnderTest.DeclaringType.GetTypeInfo()))
+            {
+                var typeName = methodUnderTest.DeclaringType.Name;
+                ParserTestBase.FileName = $"TestFiles/ParserTests/{typeName}/{methodUnderTest.Name}";
+
+                if (methodUnderTest.GetCustomAttributes(typeof(TheoryAttribute), inherit: false).Length > 0)
```
NTaylorMullen: Lets get rid of the ability to theory test our baselinable parser tests.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/SyntaxTreeNodeWriter.cs
ajaybhargavb: I added a list of descriptor names.

ajaybhargavb:
🆙 📅 
- Decided to not generate baselines for Theory test given how complicated/unreliable it can get when something changes. Going to convert these to Facts.
- Planning to check this in and convert other tests in chunks

ajaybhargavb:

PR #2405, opened 2018-06-14
Don't discover tag helpers if some references are missing

pranavkm:
A common cause for runtime view compilation failure in MVC is when an application is published
without reference assemblies. MVC usually handles this at compilation, by looking for specific
error codes. More recently, TagHelper discovery fails with an not-so-helpful error message in this
scenario. This change attempts to add a little more error checking to cover the most common cases


pranavkm:
From https://github.com/aspnet/Mvc/issues/7910, here's the error message we produce now:

> Error RZ3007: Targeted tag name cannot be null or whitespace

The fix here is to let code generation succeed with no tag helpers being discovered and let the compilation error instead.

pranavkm:
@dotnet-bot test Windows Release x64 Build please

src/Microsoft.CodeAnalysis.Razor/DefaultTagHelperDescriptorProvider.cs
```diff
@@ -31,8 +31,19 @@ public void Execute(TagHelperDescriptorProviderContext context)
                 return;
             }
 
+            var @interface = compilation.GetTypeByMetadataName(TagHelperTypes.ITagHelper);
```
NTaylorMullen: Seems like an odd layer to add this at. Is there a reason we can't add this logic to the one who provides the compilation? We already special case compilation being `null` above.

pranavkm: Is this provider only ever called by `CompilationTagHelperFeature`? If so, yeah we could do that. I verified that the visitor is only ever called from this type so moving the check to here was a safe change to make.

NTaylorMullen: Hmm? Not sure I follow, I'm suggesting the piece that has the SetCompilation / GetCompilation; the caller of `SetCompilation` shouldn't even set a compilation if it's invalid.

pranavkm: Right. But that's assuming `CompilationTagHelperFeature` is the only thing that sets compilation. If you were to write a custom `ITagHelperFeature` where you did not do the additional checks before setting the compilation, this would break of sorts.

NTaylorMullen: Ah, I see what you're saying. We should probably add a compilation verification (one that only verifies the necesities like ITagHelper and string) check to all call sites of `SetCompilation`; it's done in more than one place, the TagHelperResolver for instance.

Just to bring it all full circle about my concern: with the fix that's currently in this PR ViewComponent TagHelperDescriptor Provider might also run into similar troubles as the default one. 

People tend to extend the descriptor provider infrastructure which depends on the compilation being reasonable; if a user were to write their own `TagHelperFeature` they can do whatever they want.

pranavkm:
🆙 📅

Issue #2408, opened 2018-06-15
Any edit to functions directive content result in full reparses.


NTaylorMullen:
In larger file scenarios this results in the Razor experience flickering as mappings are re-generated repeatedly.

NTaylorMullen:
1aa15374b53fb80a6307d3494dde393e70f070a4

PR #2411, opened 2018-06-15
Add partial parsing of `@functions` directive body.

NTaylorMullen:
- Added ability to understand valid inserts, deletes and replacements for the `@functions` directive (and any other directive that uses our extensible code block bits).
- Added unit tests.
- Updated existing tests.
- Found an issue when completing some C# items the auto-completions would impact the underlying snapshot after we'd captured the change. Fixed this by forcing a reparse when we detect that our understanding of the latest snapshot and the actual latest snapshot diverge.

#2408

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
@@ -1818,7 +1818,11 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                             NextToken();
                             Balance(BalancingModes.NoErrorOnFailure, CSharpSymbolType.LeftBrace, CSharpSymbolType.RightBrace, startingBraceLocation);
                             Span.ChunkGenerator = new StatementChunkGenerator();
+                            var existingEditHandler = Span.EditHandler;
+                            Span.EditHandler = new CodeBlockEditHandler(Language.TokenizeString);
                             Output(SpanKindInternal.Code);
+
+                            Span.EditHandler = existingEditHandler;
                         });
                         break;
                 }
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/CodeBlockEditHandler.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/CodeBlockEditHandler.cs
@@ -1,0 +1,126 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -1818,7 +1818,11 @@ private void HandleDirective(DirectiveDescriptor descriptor)
                             NextToken();
                             Balance(BalancingModes.NoErrorOnFailure, CSharpSymbolType.LeftBrace, CSharpSymbolType.RightBrace, startingBraceLocation);
                             Span.ChunkGenerator = new StatementChunkGenerator();
+                            var existingEditHandler = Span.EditHandler;
+                            Span.EditHandler = new CodeBlockEditHandler(Language.TokenizeString);
                             Output(SpanKindInternal.Code);
+
+                            Span.EditHandler = existingEditHandler;
```
rynowak: The jank-factor is high

src/Microsoft.AspNetCore.Razor.Language/Legacy/CodeBlockEditHandler.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            if (IsAcceptableReplacement(target, change))
+            {
+                return PartialParseResultInternal.Accepted;
+            }
+
+            if (IsAcceptableInsertion(change))
+            {
+                return PartialParseResultInternal.Accepted;
+            }
```
rynowak: I assume that you're doing this to follow the structure of the existing stuff 😆

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
NTaylorMullen: Damn straight lol, welcome to our parser.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CodeBlockEditHandler.cs
NTaylorMullen: yup

src/Microsoft.AspNetCore.Razor.Language/Legacy/CodeBlockEditHandler.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        // Internal for testing
+        internal static bool ModifiesInvalidContent(Span target, SourceChange change)
+        {
+            var relativePosition = change.Span.AbsoluteIndex - target.Start.AbsoluteIndex;
+
+            // Iterate over the content that's being deleted. If the deletion contains brackets then it's not an
+            // acceptable deletion.
+            for (var i = relativePosition; i < relativePosition + change.Span.Length; i++)
```
rynowak: this could also just be a string.indexofany right?

Issue #1076, opened 2017-03-14
Indentation problems with code blocks


tuespetre:
I wasn't sure if this should go to VS or if it belongs here (especially with the Evolution work going on.)

In Visual Studio, when auto-formatting a Razor document, a code block nested inside of an `if` block will cause the indentation of any code following the `if` block to be increased to match the indentation of the code inside of the code block.

This is driving me 🔩 🔩 ! 

![bloody-razor](https://cloud.githubusercontent.com/assets/4853540/23904725/e9295060-0896-11e7-902f-91e7edf126ff.png)

Moving the code out of the code block to be just inside the `if` block seems to clear it up:

![bloody-razor-2](https://cloud.githubusercontent.com/assets/4853540/23904887/675caa2c-0897-11e7-8657-9b14ddcd8fc7.png)

rynowak:
Thanks, and yes this is a good place for you to open Razor issues. We'll investigate and communicate with the non-OSS parts of the team as necessary.

rynowak:
@NTaylorMullen - can you take a look at this and see what's going on?

NTaylorMullen:
Findings:
- This happens in VS2017 RTM and is consistent in newer versions
- This should be fixed in our `GetDesiredIndentation` service method.
- This happens because a desired indentation doesn't get found from our SyntaxTree (it can be, we just don't do the work) and therefore falls back to the underlying C# buffer to determine the "base" indentation.

Removed investigate label.

NTaylorMullen:
Move out?

rynowak:
I spoke with @NTaylorMullen this morning - the takeaway here is that this is something we can and should fix, but with some bake time as it's going to be a little risky.

We don't currently handle indent when the cursor is positioned inside a code span. We'd need to write that logic, and we don't want to rush that process. We'll take this on for the next foundational update where we will have some bake time.

miloush:
+1 very annoying indeed

mkArtakMSFT:
Moving this out from backlog to be addressed... Let's make sure we also have good tests for this area so this will never be a thing going forward.

Spongman:
please ensure that the fix covers the test cases in #2107

mkArtakMSFT:
@NTaylorMullen, @danroth27 also has a repro, if you're interested

ajaybhargavb:
[...]

Issue #2410, opened 2018-06-15
Functions block statements get extra newline


NTaylorMullen:
This is like https://github.com/aspnet/Razor/issues/2297 but for functions. The fix for #2297 was supposed to work for C# but it looks like content that's extra indented (by default `functions`) result in incorrect behavior. We need to be more restrictive in where our smart indenter applies.

![image](https://i.imgur.com/9XPzvis.gif)

NTaylorMullen:
41fad8a33a3e02bd53f01c58a96bc87c5711a32f

PR #2414, opened 2018-06-19
Don't smart indent code inside of Razor block constructs.

NTaylorMullen:
- Prior to this change we weren't strict enough with where our smart indenter ran. We made the assumption that every code block could be smart indented and then Roslyn would "do the right thing". However, in nested code block scenarios we found that Roslyn and us would both indent resulting in extra newlines. These changes make the criteria for applying smart indentation a little stricter.
- Updated directive code block parsing to add a C# marker symbol in cases of an empty code block directive.
- Added unit tests to verify new smart indenter behavior.
- Updated existing tests to expect new syntax tree marker symbol for empty directive bits.
- Regenerated baselines.

image

@rynowak / @ajaybhargavb if you could review asap por favor 😄, time sensitive

#2410

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics;
 using System.Text;
 using Microsoft.AspNetCore.Razor.Language;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
@@ -11,6 +10,7 @@ using Microsoft.VisualStudio.Text;
 using Microsoft.VisualStudio.Text.Editor;
 using Microsoft.VisualStudio.Text.Operations;
 using ITextBuffer = Microsoft.VisualStudio.Text.ITextBuffer;
+using Span = Microsoft.AspNetCore.Razor.Language.Legacy.Span;
 
 namespace Microsoft.VisualStudio.Editor.Razor
 {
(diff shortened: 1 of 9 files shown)


test/Microsoft.VisualStudio.Editor.Razor.Test/BraceSmartIndenterTest.cs
```diff
@@ -5,69 +5,267 @@
...
-            var result = BraceSmartIndenter.AtValidContentKind(changePosition, syntaxTree);
+            var result = BraceSmartIndenter.AtApplicableRazorBlock(changePosition, syntaxTree);
 
             // Assert
             Assert.False(result);
         }
 
         [Fact]
-        public void AtValidContentKind_ReturnsTrueAtCode()
+        public void AtApplicableRazorBlock_ReturnsTrueAtEplixitCodeBlocksCode()
```
ajaybhargavb: nit: Explicit

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
```diff
@@ -242,35 +242,73 @@ internal static void RestoreCaretTo(int caretPosition, ITextView textView)
...
+            }
+
+            // Indentable content inside of a code block.
+            return true;
+        }
+
+        // Internal for testing
+        internal static bool ContainsInvalidContent(Span owner)
+        {
+            if (owner.Symbols.Count == 0)
```
ajaybhargavb: Why was this necessary? Why not indent empty code blocks as well? Please add a comment.

ajaybhargavb:
Does this by any chance fix https://github.com/aspnet/Razor/issues/1076?

NTaylorMullen:
> Does this by any chance fix #1076?

Sadly no. When the Razor smart indenter misbehaves it manifests as extra newlines.

src/Microsoft.VisualStudio.Editor.Razor/BraceSmartIndenter.cs
NTaylorMullen: Totally forget, gonna have to run through the scenarios again to re-verify.

NTaylorMullen: Has 0 value. Good catch 👍 👍 👍

Issue #2378, opened 2018-06-01
Copy-pasting an item in an RCL project adds junk to the project file


rynowak:
1. Create a new RCL project in VS (2.1 RTM)
2. Copy the `Page1.cshtml` file in solution explorer
3. Paste

You get this.

```
  <ItemGroup>
    <Content Update="Areas\MyFeature\Pages\Page1 - Copy.cshtml">
      <Pack>$(IncludeRazorContentInPack)</Pack>
    </Content>
  </ItemGroup>
```

pranavkm:
@mkArtakMSFT could we consider this for 2.1.1? The fix is fairly trivial, but it makes a big difference in terms of user experience.

NTaylorMullen:
620105cc5e180feb4287dbc792a7704110a10982

PR #2418, opened 2018-06-19
Update Razor SDK packing logic to not include Pack at properties layer.

NTaylorMullen:
- We now rely on the Razor source inputs layer to add packing to cshtml content items.
- **NOTE:** VS doesn't respect the Razor sdk reference as part of packages for design time builds; therefore, in order for this change to truly fix the RCL -> copy scenario users will need a new CLI that has these Razor sdk changes included.

Will port this to the 2.1 patch branch once this PR is signed off on.:

image

#2378


NTaylorMullen:
Looks like one failing test: Pack_Works_IncludesRazorAssembly. Investigating.

NTaylorMullen:
🆙 📅  (fixed test)

PR #2440, opened 2018-06-27
[2.1]: Updating branding for patch release

natemcmaster:


natemcmaster:
@dotnet-bot test OSX 10.12 Release Build

cc @ajaybhargavb this looks like more known test failures, right? https://ci3.dot.net/job/aspnet_Razor/job/release_2.1/job/osx-Configuration_Release_prtest/90/

ajaybhargavb:
>  @ajaybhargavb this looks like more known test failures, right? 

Yes. This is known. @pranavkm, this seems to be happening again and probably on another file we didn't ignore. Since we've already confirmed this is not a product issue, I vote we enable this test only on Windows. What do you think?

natemcmaster:
Yes, let’s disable flaky tests in 2.1 or fix them. This helps us understand better the quality of 2.1 builds.

PR #2442, opened 2018-06-27
Converted HtmlAttributeTest

ajaybhargavb:
#2263


test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ParserTestBase.cs
```diff
@@ -159,6 +162,18 @@ private static string NormalizeNewLines(string content)
             return Regex.Replace(content, "(?<!\r)\n", "\r\n", RegexOptions.None, TimeSpan.FromSeconds(10));
         }
 
+        internal virtual void BaselineTest(RazorSyntaxTree syntaxTree)
```
ajaybhargavb: FYI. Added some convenience overloads

PR #2447, opened 2018-06-27
Convert HtmlError and HtmlTags tests to use baselines

NTaylorMullen:
#2263


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/HtmlTagsTest.cs
```diff
@@ -1,206 +1,180 @@
...
-                    Factory.Markup("<p></p>"),
-                    BlockFactory.MarkupTagBlock("</script>")));
+            ParseDocumentTest("<script><p></p></script>");
         }
 
-        [Theory]
-        [MemberData(nameof(VoidElementNames))]
-        public void VoidElementFollowedByContent(string tagName)
+        [Fact]
+        public void VoidElementFollowedByContent()
```
NTaylorMullen: Converted all these void element tests to be one large test all utilizing the same set of void elements.

PR #2467, opened 2018-07-10
 Generate and verify baselines for ClassifiedSpans

ajaybhargavb:
We want to have a set of baselines that shouldn't change in any way as we iterate on our syntax tree. This would help guarantee we don't break any existing scenario.

Generated only a few baselines for ease of review. I'll generate baselines for everything else once approved.


test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpanVerifier.cs
```diff
@@ -0,0 +1,110 @@
...
+            }
+
+            public override void VisitClassifiedSpan(ClassifiedSpanInternal span)
+            {
+                var expected = _index < _baseline.Length ? _baseline[_index++] : null;
+
+                _writer.GetStringBuilder().Clear();
+                base.VisitClassifiedSpan(span);
+                var actual = _writer.GetStringBuilder().ToString();
+                AssertEquals(span, expected, actual);
```
rynowak: Should be `AssertEqual`

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpanVerifier.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                _writer = writer;
+                _baseline = baseline;
+            }
+
+            public override void VisitClassifiedSpan(ClassifiedSpanInternal span)
+            {
+                var expected = _index < _baseline.Length ? _baseline[_index++] : null;
+
+                _writer.GetStringBuilder().Clear();
```
rynowak: Is there a reason to use a writer here instead of just a stringbuilder?

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpanWriter.cs
```diff
@@ -0,0 +1,122 @@
...
+                        span.Parent.Length),
+                    span.Kind,
+                    span.Parent.Type,
+                    span.EditHandler.AcceptedCharacters);
+            }
+
+            return result;
+        }
+
+        private static List<Span> Flatten(SyntaxTreeNode root)
```
rynowak: Does this duplicate product code? Should we add it to `SyntaxTreeNode`?

ajaybhargavb: It does. And I don't see why not make it an extension method. 👍

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpanVerifier.cs
ajaybhargavb: Yes. The writer is passed in to the base class which needs a `TextWriter`. And that base class uses it for methods like `WriteNewLine()`.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/Legacy/ClassifiedSpanWriter.cs
rynowak: It should just be a normal extension method. 

What do you stand to gain besides #PranavPoints for making it an extension method?

NTaylorMullen: In general not a huge fan of having these tooling specific baselines in the language project. In the end this is all to ensure tooling works. It'd be nice for this all to eventually sit in the tooling test projects and just link the appropriate files.

Anyways, rant over, good enough for now.

rynowak: So your suggestion is to move this feature down the syntax tree layer? sounds like a good idea :trollface:

ajaybhargavb: Moving these baselines to a tooling test assembly would also mean copying our existing parser tests for which we generate the baselines. I would much prefer they stay and be part of the same tests so it's much easier to run, maintain and find out when something changes.

NTaylorMullen: > So your suggestion is to move this feature down the syntax tree layer?

I'm not suggesting moving any feature; just suggesting to move the tests to the appropriate test dll.  As this PR currently stands the baselines are only verifying the copied code from the tooling binaries. 

Basically, what i have beef with is that in this PRs current state the test baseline logic fakes what the `Microsoft.VisualStudio.Editor.Razor.dll (MVSER)` syntax facts service does; it doesn't actually call through into the code that the WTE code calls into. Looking further into the future, with this approach the code paths will further diverge because we're going to need some sort of adapter in the `MVSER.dll` that understands the new syntax tree to ensure that it spits out 100% identical spans prior to the parser re-design.

ajaybhargavb:
🆙 📅 

- Moved GetClassifiedSpans logic to Razor.Language
- Added TagHelperSpan baseline generation infrastructure
- Generated all baselines

Suggest looking only at the 4th commit (https://github.com/aspnet/Razor/pull/2467/commits/ad2ee8ed631eeba80925b0e0b1ed022e0e46a0ef)

src/Microsoft.AspNetCore.Razor.Language/Legacy/RazorSyntaxTreeExtensions.cs
```diff
@@ -0,0 +1,130 @@
...
+                            tagHelperNode.Start.FilePath ?? syntaxTree.Source.FilePath,
+                            tagHelperNode.Start.AbsoluteIndex,
+                            tagHelperNode.Start.LineIndex,
+                            tagHelperNode.Start.CharacterIndex,
+                            tagHelperNode.Length),
+                        tagHelperNode.Binding));
+                }
+
+                // collect all child blocks and inject into toProcess as a single InsertRange
+                foreach (var curNode in blockNode.Children)
```
rynowak: I know this is probably just a move of this code, but can we update the silly names?

PR #2471, opened 2018-07-11
 Generate deterministic id for tag helpers

ajaybhargavb:
#2449 

- Using the hash of file checksum along with the absolute position in the generated code to create the unique id.
- Regenerated baselines

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
@@ -99,7 +99,7 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
                 var uniqueId = (string)context.Items[CodeRenderingContext.SuppressUniqueIds];
                 if (uniqueId == null)
                 {
-                    uniqueId = Guid.NewGuid().ToString("N");
+                    uniqueId = GetDeterministicId(context);
                 }
 
                 context.CodeWriter.WriteStringLiteral(node.TagName)
@@ -635,6 +635,16 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
             }
 
             return builder.ToString();
+        }
+
+        // Internal for testing
+        internal static string GetDeterministicId(CodeRenderingContext context)
+        {
+            // Use the file checksum along with the absolute position in the generated code to create a unique id for each tag helper call site.
+            var checksum = Checksum.BytesToString(context.SourceDocument.GetChecksum());
+            var uniqueId = checksum + context.CodeWriter.Location.AbsoluteIndex;
+
+            return uniqueId;
         }
 
         private static string GetPropertyAccessor(DefaultTagHelperPropertyIntermediateNode node)
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIntegrationTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/BuildIntegrationTest.cs
@@ -4,7 +4,6 @@
 using System;
 using System.IO;
 using System.Linq;
-using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Testing.xunit;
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -99,7 +100,7 @@ public void WriteTagHelperBody(CodeRenderingContext context, DefaultTagHelperBod
                 var uniqueId = (string)context.Items[CodeRenderingContext.SuppressUniqueIds];
```
ajaybhargavb: I left this as-is. The ability to suppress might still be useful in some cases like unit tests.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -649,5 +650,28 @@ private static string GetPropertyAccessor(DefaultTagHelperPropertyIntermediateNo
...
+            if (BitConverter.IsLittleEndian)
+            {
+                // We want to represent int using big endian notation.
+                // Rfc1014#section-3.1
+                Array.Reverse(absoluteIndexBytes);
+            }
+            var combinedBytes = checksumBytes.Concat(absoluteIndexBytes).ToArray();
+
+            // Use MD5 to create a 16 byte hash that can be parsed by Guid.
+            var combinedHash = Checksum.BytesToString(MD5.Create().ComputeHash(combinedBytes));
```
NTaylorMullen: Need to do a using statement for the MD5.Create()

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -637,6 +638,33 @@ private string GetContent(HtmlContentIntermediateNode node)
...
+            if (BitConverter.IsLittleEndian)
+            {
+                // We want to represent int using big endian notation.
+                // Rfc1014#section-3.1
+                Array.Reverse(absoluteIndexBytes);
+            }
+            var combinedBytes = checksumBytes.Concat(absoluteIndexBytes).ToArray();
+
+            // Use MD5 to create a 16 byte hash that can be parsed by Guid.
+            using (var md5 = MD5.Create())
```
Eilon: This will get flagged by all kinds of security reviews, even though this might have nothing to do with security. Can we use some other mechanism to avoid getting flagged at all?

ajaybhargavb: The other alternative is to use SHA1 which creates a 20 byte hash and only use the first 16 bytes from it. Which should be fine for our use case. Thoughts?

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/ModelExpressionTagHelper_Runtime.codegen.cs
```diff
@@ -45,7 +45,7 @@ public class TestFiles_IntegrationTests_CodeGenerationIntegrationTest_ModelExpre
             WriteLiteral("\r\n");
             EndContext();
             BeginContext(66, 25, false);
-            __tagHelperExecutionContext = __tagHelperScopeManager.Begin("input-test", global::Microsoft.AspNetCore.Razor.TagHelpers.TagMode.SelfClosing, "test", async() => {
+            __tagHelperExecutionContext = __tagHelperScopeManager.Begin("input-test", global::Microsoft.AspNetCore.Razor.TagHelpers.TagMode.SelfClosing, "16b8f99693bceb6f66cb680aaae1d399", async() => {
```
rynowak: I would think that we would still want to suppress the unique Ids in our tests. 

Think about what happens. You change one character in the source code, and guids change all over the place.

ajaybhargavb: Yes that would be very annoying 👍

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
rynowak: So you're OK with breaking someone's code if they did parse them as guids? 😆

Eilon: Yes, I am. It didn't even look like a GUID (there were no dashes or curly braces). It was just a 16-byte hex string.

rynowak: Oh ok, #YOLO then - I thought we used to use the GUID format, but I am mistaken (first time evar)

ajaybhargavb:
🆙 📅 @Eilon

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -637,6 +638,16 @@ private string GetContent(HtmlContentIntermediateNode node)
             return builder.ToString();
         }
 
+        // Internal for testing
+        internal static string GetDeterministicId(CodeRenderingContext context)
+        {
+            // Use the file checksum along with the absolute position in the generated code to create a unique id for each tag helper call site.
+            var checksum = Checksum.BytesToString(context.SourceDocument.GetChecksum());
```
pranavkm: Should this be cached as a property?

ajaybhargavb: It's already cached in SourceDocument.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
```diff
@@ -5,6 +5,7 @@
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Security.Cryptography;
```
Eilon: BTW is the using still needed?

ajaybhargavb: Will remove

src/Microsoft.AspNetCore.Razor.Language/Extensions/DefaultTagHelperTargetExtension.cs
Eilon: If you have a cache you have to worry about invalidating the cache. Would that be a problem here?

ajaybhargavb: I don't want to cache anything at this level. The checksum is always 20 bytes and converting that to string shouldn't impact performance.

PR #2477, opened 2018-07-13
Change HTML splitting logic to not split on surrogate pairs.

NTaylorMullen:
- When a surrogate pair is about to be split down the middle we reduce the size of our split by 1 character. This way we split right before a surrogate pair. In the case of zero width joiners, if we split on a zero width joiner we still render valid bytes because the zero width joiner by itself can stand alone.
- Added tests for the various split cases.

#2470

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RuntimeNodeWriterTest.cs
+++ b/test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RuntimeNodeWriterTest.cs
@@ -340,6 +340,91 @@ if (true) { }
         }
 
         [Fact]
+        public void WriteHtmlLiteral_WithinMaxSize_WritesSingleLiteral()
+        {
+            // Arrange
+            var codeWriter = new CodeWriter();
+            var writer = new RuntimeNodeWriter();
+            var context = TestCodeRenderingContext.CreateRuntime();
+
+            // Act
+            writer.WriteHtmlLiteral(context, maxStringLiteralLength: 6, "Hello");
+
+            // Assert
+            var csharp = context.CodeWriter.GenerateCode();
+            Assert.Equal(
+@"WriteLiteral(""Hello"");
+",
+                csharp,
+                ignoreLineEndingDifferences: true);
+        }
+
+        [Fact]
+        public void WriteHtmlLiteral_GreaterThanMaxSize_WritesMultipleLiterals()
+        {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RuntimeNodeWriterTest.cs
```diff
@@ -339,6 +339,91 @@ public void WriteCSharpCode_WritesPadding_WithSource()
...
+            var writer = new RuntimeNodeWriter();
+            var context = TestCodeRenderingContext.CreateRuntime();
+
+            // Act
+            writer.WriteHtmlLiteral(context, maxStringLiteralLength: 6, "👩‍👩‍👧‍👧👩‍👩‍👧‍👧");
+
+            // Assert
+            var csharp = context.CodeWriter.GenerateCode();
+            Assert.Equal(
+@"WriteLiteral(""👩‍👩‍"");
```
NTaylorMullen: When these are rendered next to each other they render out as the initial input.

test/Microsoft.AspNetCore.Razor.Language.Test/CodeGeneration/RuntimeNodeWriterTest.cs
```diff
@@ -339,6 +339,91 @@ public void WriteCSharpCode_WritesPadding_WithSource()
...
+        [Fact]
+        public void WriteHtmlLiteral_GreaterThanMaxSize_SequencedZeroWithJoinedEmojisSplit()
+        {
+            // Arrange
+            var codeWriter = new CodeWriter();
+            var writer = new RuntimeNodeWriter();
+            var context = TestCodeRenderingContext.CreateRuntime();
+
+            // Act
+            writer.WriteHtmlLiteral(context, maxStringLiteralLength: 6, "👩‍👩‍👧‍👧👩‍👩‍👧‍👧");
```
NTaylorMullen: GitHub displays a little wonky, it's actually two sequenced emojis side by side. VS does a clearer job 😉 

image

NTaylorMullen:
> Curious, can we ever get into a state of there being just one character in the document which is a high surrogate?

Technically yes; however, that would mean the doc was invalid from the start. You do bring up a good point though, for an invalid doc which has 1025 characters with the 1025th character being just the high surrogate we'd infinite loop. I'll add protection against that.

Issue #2497, opened 2018-07-21
`DefaultTagHelperContent.IsEmptyOrWhiteSpace` is incorrectly `false` when only content is `null`


dougbu:
See https://github.com/aspnet/Mvc/pull/8087#discussion_r204217356

And, see where the bug lies: https://github.com/aspnet/Razor/blob/7bdad8fdd30dea9b5b5ebcc0d9e8dc7a257466f1/src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs#L281-L286

dougbu:
Self-assigning because this is a bug and requires only a quick fix.

PR #2498, opened 2018-07-21
Handle `null` entries correctly for `IsEmptyOrWhiteSpace`

dougbu:
- #2497

### DIFF
--- a/src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs
+++ b/src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs
@@ -221,8 +221,7 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 return;
             }
 
-            var stringValue = entry as string;
-            if (stringValue != null)
+            if (entry is string stringValue)
             {
                 encoder.Encode(writer, stringValue);
             }
@@ -239,13 +238,11 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 return;
             }
 
-            string entryAsString;
-            IHtmlContentContainer entryAsContainer;
-            if ((entryAsString = entry as string) != null)
+            if (entry is string entryAsString)
             {
                 destination.Append(entryAsString);
             }
-            else if ((entryAsContainer = entry as IHtmlContentContainer) != null)
+            else if (entry is IHtmlContentContainer entryAsContainer)
             {
                 entryAsContainer.CopyTo(destination);
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -282,11 +277,10 @@ private bool IsEmptyOrWhiteSpaceCore(object entry, EmptyOrWhiteSpaceWriter write
         {
             if (entry == null)
             {
-                return false;
+                return true;
```
dougbu: This is the only real change. Everything else done to this file was VS suggestions.

src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -298,10 +292,8 @@ private bool IsEmptyOrWhiteSpaceCore(object entry, EmptyOrWhiteSpaceWriter write
             {
                 // Use NullHtmlEncoder to avoid treating encoded whitespace as non-whitespace e.g. "\t" as "&#x9;".
                 ((IHtmlContent)entry).WriteTo(writer, NullHtmlEncoder.Default);
-                if (!writer.IsEmptyOrWhiteSpace)
-                {
-                    return false;
-                }
+
+                return writer.IsEmptyOrWhiteSpace;
```
dougbu: Hmm, for consistency, I should either revert changes to this block or change the `if` side to `return string.IsNullOrWhiteSpace(stringValue);` and remove the final `return true;` Any preferences one way or the other @NTaylorMullen? (I lean toward making further changes but only because that allows me to reduce nesting -- no need for the `else` wrapper.)

src/Microsoft.AspNetCore.Razor/TagHelpers/DefaultTagHelperContent.cs
```diff
@@ -282,29 +277,19 @@ private bool IsEmptyOrWhiteSpaceCore(object entry, EmptyOrWhiteSpaceWriter write
...
-                return false;
+                return true;
             }
 
-            var stringValue = entry as string;
-            if (stringValue != null)
+            if (entry is string stringValue)
             {
                 // Do not encode the string because encoded value remains whitespace from user's POV.
-                if (!string.IsNullOrWhiteSpace(stringValue))
```
dougbu: @NTaylorMullen I've done the mentioned additional changes to this method. Looks better to me. Agreed?

test/Microsoft.AspNetCore.Razor.Test/TagHelpers/DefaultTagHelperContentTest.cs
```diff
@@ -363,6 +363,7 @@ public static TheoryData<string> EmptyOrWhiteSpaceData
             {
                 return new TheoryData<string>
                 {
+                    null,
```
dougbu: Used in multiple `[Theory]` tests. 5 of those 9 tests fail without the src changes.

dougbu:
dotnet test Ubuntu 16.04 Release Build

Issue #2525, opened 2018-08-08
ManualServerShutdown_NoPipeName_ShutsDownServer hangs on linux VSTS


ajaybhargavb:
After some investigation, it looks like this test is the culprit causing https://github.com/aspnet/Razor/issues/2507. I've disabled this test on crossplat. As this is not common case and is hard to integration test, we should either remove this test or cover this in a unit test.

ajaybhargavb:
I didn't mean to jump to any conclusions here. I meant to say the way it is currently tested evidently made it hard to determine the exact problem (Given this only fails on unix in VSTS). I created this issue to change it in some way where that doesn't happen. Investigating the exact cause is part of this issue. I'll update the labels accordingly.

ajaybhargavb:
Possible reason for the hang https://github.com/dotnet/corefx/issues/31713

PR #2529, opened 2018-08-09
Re-enable hanging test

ajaybhargavb:
#2525 

- Figured out the root cause for the hang and added a timeout to prevent it.
- Re-enabled the test on linux and Mac

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Tools/ShutdownCommand.cs
+++ b/src/Microsoft.AspNetCore.Razor.Tools/ShutdownCommand.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Diagnostics;
-using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.CommandLineUtils;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Razor.Tools/ShutdownCommand.cs
```diff
@@ -61,11 +62,14 @@ protected async override Task<int> ExecuteCoreAsync()
                         try
                         {
                             var process = Process.GetProcessById(response.ServerProcessId);
-                            process.WaitForExit();
+
+                            // Sometimes in linux when the process has already exited before we get here, WaitForExit tends to wait indefinitely.
+                            // Add a timeout to avoid hang.
+                            process.WaitForExit(TimeoutMsProcessShutdown);
```
rynowak: Should we report this issue on CoreFX? The windows behaviour is to throw if the process has exited. 

If this is really the case, perhaps we should do this in a loop and check if the process is still running with a shorter timeout?

pranavkm: 👍 for filing a CoreFx issue, at the very least so that other users of the API could see it.

The test code has a slightly different pattern for waiting for process exit. Should we use that here - https://github.com/aspnet/Razor/blob/3e436a53a1a0555e05111594be96323d16cc88b8/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs#L83-L108?

Issue #2530, opened 2018-08-09
Refactor completion provider to be more general.


NTaylorMullen:
Today the completion provider is tied to VS windows / VS mac. We should refactor it to be more general so it can be reused on multiple platforms.

NTaylorMullen:
572b55690dd23e9df3239681dba5fc3c46022e6e

PR #2532, opened 2018-08-10
Refactor completion logic into standalone service.

NTaylorMullen:
- Migrated the completion item source provider and the legacy directive completion provider to use the new service.
- Cleaned up duplicate tests that were both verifying common completion functionality.
- Ensured that the legacy `RazorDirectiveCompletionProvider` did not result in additional Razor assembly loads when in C# scenarios.

#2530

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorCompletionFactsService.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorCompletionFactsService.cs
@@ -1,0 +1,98 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.Composition;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    [System.Composition.Shared]
+    [Export(typeof(RazorCompletionFactsService))]
+    internal class DefaultRazorCompletionFactsService : RazorCompletionFactsService
+    {
+        private static readonly IEnumerable<DirectiveDescriptor> DefaultDirectives = new[]
+        {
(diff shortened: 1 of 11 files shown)


src/Microsoft.VisualStudio.Editor.Razor/DefaultRazorCompletionFactsService.cs
```diff
@@ -0,0 +1,103 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
NTaylorMullen: All this logic is refactored from `RazorDirectiveCompletionSource` mostly verbatim

src/Microsoft.VisualStudio.Editor.Razor/RazorCompletionItem.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    internal sealed class RazorCompletionItem
```
NTaylorMullen: Had to build our own completion item abstraction for the following reasons:
- Roslyn's abstraction is used in their layering over platform abstractions but is discouraged from being used (new completion APIs are the recommended ones now)
- New completion APIs (`RazorDirectiveCompletionSource`) have their own completion item concept
- A Razor language server would have its own abstraction to follow protocol.

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultRazorCompletionFactsServiceTest.cs
```diff
@@ -0,0 +1,239 @@
...
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Moq;
+using Xunit;
+
+namespace Microsoft.VisualStudio.Editor.Razor
+{
+    public class DefaultRazorCompletionFactsServiceTest
```
NTaylorMullen: All these are pulled functionally verbatim from the `RazorDirectiveCompletionSource` and `RazorDirectiveCompletionProvider` tests.

PR #2535, opened 2018-08-12
Add generated documents to workspace

rynowak:
This is just the rest of the enabling features for FAR. Not all of our dependencies are in place, but I wanted to check in my prototype so we can start replacing the hacked up parts with the real dependencies when they come online.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
@@ -65,6 +65,8 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
         public HostWorkspaceServices Services { get; }
 
+        public GeneratedCodeContainer GeneratedCodeContainer => HostDocument.GeneratedCodeContainer;
+
         public DocumentGeneratedOutputTracker GeneratedOutput
         {
             get
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostDocument.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostDocument.cs
@@ -21,10 +21,13 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
             FilePath = filePath;
             TargetPath = targetPath;
+            GeneratedCodeContainer = new GeneratedCodeContainer();
         }
 
         public string FilePath { get; }
 
         public string TargetPath { get; }
+
+        public GeneratedCodeContainer GeneratedCodeContainer { get; }
     }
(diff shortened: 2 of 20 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/HostDocument.cs
```diff
@@ -21,10 +21,13 @@ public HostDocument(string filePath, string targetPath)
...
             FilePath = filePath;
             TargetPath = targetPath;
+            GeneratedCodeContainer = new GeneratedCodeContainer();
         }
 
         public string FilePath { get; }
 
         public string TargetPath { get; }
+
+        public GeneratedCodeContainer GeneratedCodeContainer { get; }
```
NTaylorMullen: whaaaa? When a project change the host documents make their way from one project config to the next, wouldn't that result in invalid generated code in docs? Maybe i'm missing something but this seems concerning.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
```diff
@@ -65,6 +65,8 @@ internal class DocumentState
 
         public HostWorkspaceServices Services { get; }
 
+        public GeneratedCodeContainer GeneratedCodeContainer => HostDocument.GeneratedCodeContainer;
+
         public DocumentGeneratedOutputTracker GeneratedOutput
```
NTaylorMullen: This name is sad now with the inclusion of the generated code container.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/DefaultRazorProjectHost.cs
```diff
@@ -7,32 +7,41 @@
...
+#endif
+
 namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 {
     // Somewhat similar to https://github.com/dotnet/project-system/blob/fa074d228dcff6dae9e48ce43dd4a3a5aa22e8f0/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/LanguageServices/LanguageServiceHost.cs
     //
     // This class is responsible for intializing the Razor ProjectSnapshotManager for cases where
     // MSBuild provides configuration support (>= 2.1).
     [AppliesTo("DotNetCoreRazor & DotNetCoreRazorConfiguration")]
+    [ExportVsProfferedProjectService(typeof(IVsContainedLanguageProjectNameProvider))]
```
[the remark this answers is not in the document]
rynowak: This is how you export a 'com aggregated service'. In old VS a lot of things are implemented directly on the `IVsHierarchy` object - you can downcast it to various interfaces.

Issue #2508, opened 2018-07-27
Razor parser improvements


ajaybhargavb:
As a follow-up to #2263 and as a prerequisite for a bunch of future Razor improvements, we want to have our current Razor parser to have a deeper understanding of HTML. Here is a detailed description of what this will enable us to do https://github.com/aspnet/Razor/issues/2211#issue-307854491.

ajaybhargavb:
Checklist:

- [x] Import and adapt the Syntax tree infrastructure from Roslyn (#2522) [Cost: **M**]
- [x] Rewrite the Tokenizer layer (#2540) [Cost: **S**]
- [x] Add a way to auto generate node classes and other boilerplate based on the Syntax (#2550) [Cost: **S**]
- [ ] Rewrite CSharp parser (In progress) [Cost: **M**]
- [ ] Rewrite Html parser (In progress) [Cost: **M**]
- [ ] Rewrite the classified span generation logic (In progress) [Cost: **S**]
- [ ] Rewrite the Tag helper parsing layer
- [ ] Rewrite other syntax tree visitors like WhitespaceRewriter, ConditionalAttributeCollapser etc
- [ ] Rewrite the IR layer to use the new syntax tree
- [ ] React and update parser test infrastructure (In progress) [Cost: **S**]
- [ ] Regenerate the parser test baselines with the new syntax tree and review them thoroughly for accuracy
- [ ] Add unit tests for the new syntax tree infrastructure
- [ ] React and update all other dependent tests

PR #2540, opened 2018-08-15
 Updated the tokenizer and parser to operate on SyntaxTokens directly

ajaybhargavb:
#2508 

- The tokenizer and parser now operates directly on SyntaxTokens (Green)
- The SyntaxToken(Red) is now created when the Span is built with the correct parent and position
- All other passes that run after the parsing is complete(TagHelperRewriter etc) will operate on the Red tokens.
- There is now only one type for all SyntaxTokens. They are differentiated by their SyntaxKind.
- Added equivalence checking for tokens
- Updated test code to react (Second commit)
- Regenerated baselines (Third commit)


src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -183,11 +186,11 @@ protected internal bool AcceptIf(CSharpKeyword keyword)
...
 
-        protected static Func<CSharpToken, bool> IsSpacingToken(bool includeNewLines, bool includeComments)
+        protected static Func<SyntaxToken, bool> IsSpacingToken(bool includeNewLines, bool includeComments)
         {
-            return token => token.Type == CSharpTokenType.WhiteSpace ||
-                          (includeNewLines && token.Type == CSharpTokenType.NewLine) ||
-                          (includeComments && token.Type == CSharpTokenType.Comment);
+            return token => token.Kind == SyntaxKind.Whitespace ||
+                          (includeNewLines && token.Kind == SyntaxKind.NewLine) ||
+                          (includeComments && token.Kind == SyntaxKind.Comment);
```
NTaylorMullen: `Comment` might be a bit overloaded of a name given we have the same concept in Razor and HTML. It might be good to have a more specific name like `CSharpComment` or `SingleLineComment`. Also does this comment also represent multi-line C# comment?

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -228,13 +224,13 @@ protected internal bool Balance(BalancingModes mode)
...
 
-        protected internal bool Balance(BalancingModes mode, TTokenType left, TTokenType right, SourceLocation start)
+        protected internal bool Balance(BalancingModes mode, SyntaxKind left, SyntaxKind right, SourceLocation start)
         {
             var startPosition = CurrentStart.AbsoluteIndex;
             var nesting = 1;
             if (!EndOfFile)
             {
-                var syms = new List<TToken>();
+                var syms = new List<SyntaxToken>();
```
NTaylorMullen: tokens, looks like a few places below too. I'd just do a find and replace `sym` => `token` and `syms` => `tokens` in this doc

src/Microsoft.AspNetCore.Razor.Language/Syntax/InternalSyntax/SyntaxToken.cs
```diff
@@ -4,40 +4,52 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
 
 namespace Microsoft.AspNetCore.Razor.Language.Syntax.InternalSyntax
 {
-    internal abstract class SyntaxToken : GreenNode
+    internal class SyntaxToken : GreenNode
```
NTaylorMullen: Just verifying. We have no plans to sub-class this right?

src/Microsoft.AspNetCore.Razor.Language/Syntax/InternalSyntax/SyntaxToken.cs
```diff
@@ -103,17 +131,64 @@ protected override sealed int GetSlotCount()
 
         internal override sealed GreenNode GetSlot(int index)
         {
-            throw new InvalidOperationException();
+            throw new InvalidOperationException("Tokens don't have slots.");
```
NTaylorMullen: You'll have to resourcify this.

src/Microsoft.AspNetCore.Razor.Language/Legacy/Span.cs
```diff
@@ -189,5 +200,28 @@ public override SyntaxTreeNode Clone()
             var spanBuilder = new SpanBuilder(this);
             return spanBuilder.Build();
         }
+
+        private class SyntaxTokenComparer : IEqualityComparer<SyntaxToken>
```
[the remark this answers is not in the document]
ajaybhargavb: Not yet. This is comparing the red node and it's not used outside of this class. I'll move it if that changes.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlTokenizer.cs
```diff
@@ -21,24 +22,24 @@ public HtmlTokenizer(ITextDocument source)
...
         {
-            get { return HtmlTokenType.RazorCommentTransition; }
+            get { return SyntaxKind.RazorCommentTransition; }
         }
 
-        public override HtmlTokenType RazorCommentStarType
+        public override SyntaxKind RazorCommentStarKind
         {
-            get { return HtmlTokenType.RazorCommentStar; }
+            get { return SyntaxKind.RazorCommentStar; }
```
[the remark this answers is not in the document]
ajaybhargavb: No. They are only used by the Tokenizer and the Parser. @NTaylorMullen, correct me if I am wrong but I think these tokens were abstracted because during the era when Razor supported VB.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -209,24 +212,24 @@ public override void ParseBlock()
                     AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
 
                     var current = CurrentToken;
-                    if (At(CSharpTokenType.StringLiteral) &&
+                    if (At(SyntaxKind.StringLiteral) &&
                         CurrentToken.Content.Length > 0 &&
                         CurrentToken.Content[0] == SyntaxConstants.TransitionCharacter)
                     {
-                        var split = Language.SplitToken(CurrentToken, 1, CSharpTokenType.Transition);
+                        var split = Language.SplitToken(CurrentToken, 1, SyntaxKind.Transition);
```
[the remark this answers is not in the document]
ajaybhargavb: This is currently used to support special syntax like below,
`@:HelloWorld`
The tokenizer tokenizes this as Transition(`@`) and HtmlText(`:HelloWorld`) but the parser want to be accept Transition(`@`), Colon(`:`) as MetaCode and the remaining as HtmlText. So to be able to remove this, we need to build the understanding of these cases in to the tokenizer.

PR #2550, opened 2018-08-21
Auto generate CSharp syntax tree nodes

ajaybhargavb:
#2508 

- Auto generated syntax nodes, visitors and factories
- Renamed RazorComment to RazorCommentLiteral to avoid a conflict
- Modified the Visit() contract
- Added new SyntaxKinds
- Regenerated baselines


src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
```diff
@@ -0,0 +1,102 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+
+<Tree Root="SyntaxNode">
```
ajaybhargavb: This is the main thing to review in this PR.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
```diff
@@ -0,0 +1,102 @@
...
+    <Kind Name="CSharpMetaCode" />
+    <Field Name="MetaCode" Type="SyntaxList&lt;SyntaxToken&gt;" />
+  </Node>
+  <Node Name="CSharpCodeLiteralSyntax" Base="CSharpNodeSyntax">
+    <Kind Name="CSharpCodeLiteral" />
+    <Field Name="CSharpTokens" Type="SyntaxList&lt;SyntaxToken&gt;" />
+  </Node>
+  <Node Name="CSharpCodeBlockSyntax" Base="CSharpNodeSyntax">
+    <Kind Name="CSharpCodeBlock" />
+    <Field Name="Children" Type="SyntaxList&lt;RazorNodeSyntax&gt;" />
```
ajaybhargavb: This is the one that can hold both Html and CSharp.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
```diff
@@ -0,0 +1,102 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+
+<Tree Root="SyntaxNode">
+  <PredefinedNode Name="SyntaxToken" Base="SyntaxNode" />
+  
+  <!-- Common -->
+  <AbstractNode Name="RazorNodeSyntax" Base="SyntaxNode" />
```
rynowak: RazorNodeSyntax vs RazorSyntaxNode? I think I like the latter much better.

rynowak: I could be convinced though

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
```diff
@@ -0,0 +1,102 @@
...
+      <Kind Name="CSharpExpressionBody" />
+    </Field>
+  </Node>
+  <Node Name="CSharpExpressionBodySyntax" Base="CSharpNodeSyntax">
+    <Kind Name="CSharpExpressionBody" />
+    <Field Name="OpenParen" Type="CSharpMetaCodeSyntax" Optional="true" />
+    <Field Name="CSharpCode" Type="CSharpCodeBlockSyntax" />
+    <Field Name="CloseParen" Type="CSharpMetaCodeSyntax" Optional="true" />
+  </Node>
+  <Node Name="CSharpDirectiveSyntax" Base="CSharpBlockSyntax">
```
NTaylorMullen: Lets not change anything now but just re-iterating cases such as `@page` are 0% C# and would be nice to eventually parse them outside of C#.

For now, what do you think about changing all of the `Razor` syntax concepts to not include C# even though they're parsed in the C# parser?

src/Microsoft.AspNetCore.Razor.Language/Syntax/Generated/Syntax.xml.Internal.Generated.cs
```diff
@@ -0,0 +1,1434 @@
...
+    {
+      DefaultVisit(node);
+    }
+  }
+
+  internal static partial class SyntaxFactory
+  {
+    public static RazorCommentBlockSyntax RazorCommentBlock(SyntaxToken startCommentTransition, SyntaxToken startCommentStar, SyntaxToken comment, SyntaxToken endCommentStar, SyntaxToken endCommentTransition)
+    {
+#if DEBUG
```
NTaylorMullen: We don't have the same requirements as Roslyn. Having these checks here could help us prevent issues. I'd remove the `#if DEBUG` bits from the generated code when they involve argument checking.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
ajaybhargavb: Actually, `RazorNodeSyntax` is a base class for `CSharpNodeSyntax` and `HtmlNodeSyntax`. Since, these are all abstract classes, I don't see why they should be named similar to `SyntaxNode`. I'll change the names.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
```diff
@@ -0,0 +1,102 @@
...
+  
+  <!-- CSharp -->
+  <AbstractNode Name="CSharpNodeSyntax" Base="RazorNodeSyntax" />
+  <Node Name="CSharpTransitionSyntax" Base="CSharpNodeSyntax">
+    <Kind Name="CSharpTransition" />
+    <Field Name="Transition" Type="SyntaxToken">
+      <Kind Name="Transition" />
+    </Field>
+  </Node>
+  <Node Name="CSharpMetaCodeSyntax" Base="CSharpNodeSyntax">
```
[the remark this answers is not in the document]
ajaybhargavb: > So the curly in @{ is parsed in the CSharp parser?

Yes

> Not saying we should change anything today but in an ideal world it'd be awesome to have transitions and curlies be on-csharp or html specific (just Razor). A lot like how @*...*@ is today.

I agree. I'll keep an eye out as I am changing the CSharp parser to see if I can do this easily.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Syntax.xml
ajaybhargavb: > For now, what do you think about changing all of the Razor syntax concepts to not include C# even though they're parsed in the C# parser?

I prefer doing that later as well as this might make it confusing when trying to convert the current parser. It should be easy to do once we have the parser actually working with the new nodes.

src/Microsoft.AspNetCore.Razor.Language/Syntax/Generated/Syntax.xml.Internal.Generated.cs
ajaybhargavb: Sure. Will do

PR #2551, opened 2018-08-21
Allow building this repo with VS 15.x

natemcmaster:
This is more like an FYI PR. We need this change because our CI environments have updated or will update soon to VS 15.8

### DIFF
--- a/korebuild.json
+++ b/korebuild.json
@@ -5,7 +5,7 @@
     "visualstudio": {
       "required": false,
       "includePrerelease": true,
-      "versionRange": "[15.0.26730.03, 15.8)",
+      "versionRange": "[15.0.26730.03, 16.0)",
       "requiredWorkloads": [
         "Microsoft.VisualStudio.Component.VSSDK"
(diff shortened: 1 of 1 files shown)


natemcmaster:
cc @BrennanConroy this should unblock SignalR stuff.

ajaybhargavb:
I was concerned because our master branch already allows VS 15.8 and is still failing https://github.com/aspnet/Razor/blob/master/korebuild.json#L8

natemcmaster:
Expanding the range to include 15.8 was one piece of the puzzle. Updating to 15.8 uncovered a new problem. In 15.8, NuGet more correctly handles per-project restore sources. One of the projects in this repo was misconfigured and had no restore sources set. I've resolved this in https://github.com/aspnet/Razor/commit/f703a288ddfdfd90f5d1c0bdd5b82b7b064e6518. This explains why it was passing for you locally -- you already had packages restored into your machine's cache, but CI is a clean machine.

PR #2563, opened 2018-08-30
[Discussion/Prototype] Feature/utf 8

rynowak:


test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/Blocks_Utf8_Runtime.codegen.cs
```diff
@@ -0,0 +1,154 @@
...
+#line default
+#line hidden
+            System.ReadOnlySpan<byte> __bytes1 = new byte[]{13, 10};
+            WriteLiteral(__bytes1);
+#line 6 "TestFiles/IntegrationTests/CodeGenerationIntegrationTest/Blocks_Utf8.cshtml"
+ while(i <= 10) {
+
+#line default
+#line hidden
+            System.ReadOnlySpan<byte> __bytes2 = new byte[]{32, 32, 32, 32, 60, 112, 62, 72, 101, 108, 108, 111, 32, 102, 114, 111, 109, 32, 67, 35, 44, 32, 35};
```
rynowak: These spans will be created with zero-copy from the data segment of the dll. The compiler is smart.

benaadams:
>  does something like this interest you?

Its like Xmas 😄

rynowak:
The tricky part of this is that the compiler only does this magic for Span, and we can't write a span to a stream asynchronously

benaadams:
> I could improve this by making them fields instead of locals.

Side-car class that acts as a static, but can be flushed on reload from disk?

benaadams:
* I have no idea how reloading works

benaadams:
Or buffer; with intermediate Writer that writes `Span<byte>` to pooled `byte[]` and flushes to Stream when array gets full?

rynowak:
So the problem with that is the size. We have to be able to guarantee that no write will require async since you can't async in a method that uses span.

rynowak:
Reloading also just creates a separate assembly/type

benaadams:
> Reloading also just creates a separate assembly/type

Was thinking easiest way would be to use statics 
```csharp
static ReadOnlyMemory<byte> __bytes5 = new byte[]{32, 32, 32, 32, 60, 112, 62, 87, 101, 32, 119, 114, 111, 116, 101, 32, 49, 48, 32, 108, 105, 110, 101, 115, 33, 60, 47, 112, 62, 13, 10};
```
but would that just keep growing? Is it even a problem?

rynowak:
No, it's not really a problem. Reloading is a dev time scenario, and it's well understood that it causes memory growth.

src/Microsoft.AspNetCore.Razor.Language/Experiemental/Utf8Directive.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                var index = Parent.Children.IndexOf(node);
+                Parent.Children[index] = new Utf8Node(node);
+            }
+        }
+
+        private class TargetExtension : ICodeTargetExtension
+        {
+            private readonly StringBuilder _builder = new StringBuilder();
+            private int _count;
```
benaadams: Worth having a `Dictionary<string, int>` (or `ConcurrentDictionary<string, int>`) to reuse repeated items? 

e.g. in your brief sample ` new byte[]{13, 10}` occurs multiple times

rynowak: It would deffo help if we move to using fields + `ReadOnlyMemory<>`

rynowak: Like, if we really care we'd just want to remove newlines.

src/Microsoft.AspNetCore.Razor.Language/Experiemental/Utf8Directive.cs
```diff
@@ -0,0 +1,212 @@
...
+{
+    public static class Utf8Directive
+    {
+        public static readonly DirectiveDescriptor Directive = DirectiveDescriptor.CreateDirective(
+            "utfeight",
+            DirectiveKind.SingleLine,
+            builder =>
+            {
+                builder.Usage = DirectiveUsage.FileScopedSinglyOccurring;
+                builder.Description = "does utf8 (duh)";
```
khellang: 😂

rynowak:
Related https://github.com/aspnet/Mvc/tree/feature/utf-8

@benaadams - this was one of my motivations for asking you if you had a good test to contribute to us 🤣 

If you have something you can factor into a single file or small number of files I can get numbers for you on our hw https://github.com/aspnet/Mvc/commit/588fbc43e5b7f1fc705a274cc41162d1a0a7604c#diff-acda962499b72fc8a21cb6678e739a52R7

benaadams:
@rynowak here you go https://github.com/aspnet/Mvc/pull/8366

rynowak:
Closing this as part of the repo merge. We can pick this up again at some point in the future if necessary.

PR #2565, opened 2018-09-02
Don't capture asynclocals onto Timer

benaadams:
Pattern is a workaround for not having api https://github.com/dotnet/corefx/issues/26523

Also a bit less captured scope onto the Task.StartNew

/cc @davidfowl


rynowak:
I'm sure this can't hurt, but is it important? This code doesn't ship in the runtime.

benaadams:
If it runs in VS its probably capturing stranger things :)

rynowak:
image

rynowak:
Thanks @benaadams - I'm going to merge this and then apply the new package on top.

PR #2572, opened 2018-09-07
[Design] Generated document output persists result to generated code container.

NTaylorMullen:
- Prior to this we had a `BackgroundDocumentGenerator` that would constantly be updating generated code containers. With this changes we've changed the details in how a `GeneratedCodeContainer` can be mutated. It can now be touched from any thread and is updated when an underlying `DocumentSnapshot` has available content. However, if a generated output comes through that's older then the last seen output we no-op.
- Removed `BackgroundDocumentGenerator` setoutput logic.

@rynowak definitely want a design overview on this guy. It feels super hackjobish even though it gets the job done 😢


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -2,32 +2,88 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.CodeAnalysis.Experiment;
+using Microsoft.CodeAnalysis.Text;
 
 namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 {
     internal class GeneratedCodeContainer : IDocumentServiceFactory, ISpanMapper
     {
+        public event EventHandler<TextChangeEventArgs> GeneratedCodeChanged;
```
NTaylorMullen: Allows people to grab the `sender` and look at some more information about the updated state.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
```diff
@@ -50,6 +50,13 @@ public Task<RazorCodeDocument> GetGeneratedOutputInitializationTask(ProjectSnaps
                     if (_task == null)
                     {
                         _task = GetGeneratedOutputInitializationTaskCore(project, document);
+
+                        // We wait until after we've set the task so any triggers from SetOutput can acquire the 
+                        // the generated output from the document snapshot.
+                        if (document is DefaultDocumentSnapshot defaultDocument)
+                        {
+                            defaultDocument.State.HostDocument.GeneratedCodeContainer.SetOutput(document);
+                        }
```
rynowak: Would there be a problem just changing the parameter type? `DocumentSnapshot` isn't an abstraction, it just has an internal + external surface area.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: +Beans

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -39,34 +95,59 @@ public TService GetService<TService>()
...
+        public void SetOutput(DocumentSnapshot document)
         {
-            Source = source;
-            Output = codeDocument.GetCSharpDocument();
+            lock (_setOutputLock)
+            {
+                if (!document.TryGetTextVersion(out var version))
+                {
+                    return;
+                }
```
rynowak: In what cases can this happen? When the document hasn't been read yet? Should this crash?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -39,34 +95,59 @@ public TService GetService<TService>()
...
+                {
+                    // Version from the provided document is the same or older as the current SourceVersion.
+                    return;
+                }
+
+                if (!document.TryGetText(out var source) ||
+                    !document.TryGetGeneratedOutput(out var codeDocument))
+                {
+                    return;
+                }
```
rynowak: Is this right? I thought this method was called before the output was available outside of the facade

rynowak: Referring to `TryGetGeneratedOutput`

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -39,34 +95,59 @@ public TService GetService<TService>()
...
         public Task<ImmutableArray<SpanMapResult>> MapSpansAsync(
-                Document document,
-                IEnumerable<TextSpan> spans,
-                CancellationToken cancellationToken)
+            Document document,
+            IEnumerable<TextSpan> spans,
+            CancellationToken cancellationToken)
         {
-            if (Output == null)
+            lock (_setOutputLock)
```
rynowak: You really don't want this lock held for the whole method. This should grab the output and then release the lock.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -133,10 +233,14 @@ public void SetText(SourceText sourceText)
...
-                var e = new TextChangeEventArgs(_currentText, sourceText);
-                _currentText = sourceText;
+                lock (_outerLock)
+                {
 
-                TextChanged?.Invoke(this, e);
+                    var e = new TextChangeEventArgs(_currentText, sourceText);
+                    _currentText = sourceText;
+
+                    TextChanged?.Invoke(this, e);
```
rynowak: So this can now be raised on any thread? Is that allowed?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
NTaylorMullen: Are you suggesting having SetOutput take a `DefaultDocumentSnapshot`? If that's the case I'd rather just add a `HostDocument` property to `DocumentSnapshot`

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: No case we support today. It's a `TryGet` so figured I'd just be protective here. We could explode to let the user know they're doing something unexpected.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: Ya I had to work around that restriction in a hackyyyy way: https://github.com/aspnet/Razor/pull/2572/files/8d9bf2e3cd4f4b4fe909e231c0bb130b586efbd2#diff-277383bfb275d15d91a911ee7b04ee3bR53

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: user?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: this seems wrong then.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
rynowak: yes, and no. HostDocument is outside the facade.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: It kind of has to be unless we make `HostDocument`s dependent on a foreground dispatcher which would be a semi-breaking change and then async void dispatches an event trigger. Seemed a bit questionable so I didn't entertain the idea.

What do you think?

NTaylorMullen: Note that this would also mean the HostDocument takes in a dispatcher in its constructor since the dispatcher is a MEF component.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
NTaylorMullen: Interesting. I guess if we wanted to set output for a different document type we could always overload.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: lol sure, user being our future selves 😉

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
NTaylorMullen: Open to other suggestions, the only other one I thought of is having the `SetOutput` signature look like:

`SetOutput(RazorCodeDocument output, DefaultDocumentSnapshot document)`. Didn't like that though because it indicated that the code document didn't have to be from the passed in document. In practice of course that wouldn't be a problem; was just a code cleanliness concern.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: Only one place should be calling this code, that's my point.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
rynowak: There won't be another document type.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: No, this is correct. You are currently modifying the state of the document snapshot.

NTaylorMullen:
No longer design. 🆙 📅

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
rynowak: yeah, so I'm asking is that allowed - I genuinely don't know the answer

NTaylorMullen:
@rynowak this good?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
```diff
@@ -111,7 +111,14 @@ private async Task<RazorCodeDocument> GetGeneratedOutputInitializationTaskCore(P
 
             var projectEngine = project.GetProjectEngine();
 
-            return projectEngine.ProcessDesignTime(documentSource, importSources, tagHelpers);
+            var codeDocument = projectEngine.ProcessDesignTime(documentSource, importSources, tagHelpers);
+            var csharpDocument = codeDocument.GetCSharpDocument();
+            if (document is DefaultDocumentSnapshot defaultDocument)
```
rynowak: You should just pass the `DefaultDocumentSnapshot` as a parameter.

NTaylorMullen: Hmm, sure 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -103,15 +191,25 @@ internal bool TryGetLinePositionSpan(TextSpan span, out LinePositionSpan linePos
             return false;
         }
 
+        private void TextContainer_TextChanged(object sender, TextChangeEventArgs args)
+        {
+            lock (_setOutputLock)
```
rynowak: The lock here is misleading. This shouldn't take the lock, this needs to be called **with** the lock.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentGeneratedOutputTracker.cs
NTaylorMullen: Actually, can't do that. [DocumentSnapshots are used for import tracking](https://github.com/aspnet/Razor/blob/96709c4d7723d20f8604e2313606708d44233cc7/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentImportsTracker.cs#L87) as well and they need to be able to generate content.  It'd be a re-design of that feature to make this method able to take a `DefaultDocumentSnapshot`. Having this here enables non-default documents to generate output and if they happen to be a default one we set their corresponding `HostDocuments` output.

rynowak: ugh

Issue #2571, opened 2018-09-07
Test failure: RazorGenerate_Rebuilds_IfInputFilesAreRenamed


aspnet-hello:
This test [fails](http://aspnetci/viewLog.html?buildId=546482&buildTypeId=Releases_22xPublic_OsxUniverse) occasionally with the following error:
```
System.TimeoutException : command '$dotnet msbuild /nr:false /p:RunningAsTest=true /p:MicrosoftNETCoreAppVersion=2.2.0-preview2-26905-02 /p:NETStandardLibraryPackageVersion=2.0.3 /restore /p:_RazorBuildServerPipeName="21b3f27a-7873-4c13-81c7-ad4cc9fbb448" /t:RazorGenerate /p:Configuration=Release ' timed out after 00:01:00.
   at Microsoft.AspNetCore.Razor.Design.IntegrationTests.MSBuildProcessManager.<>c__DisplayClass0_0.<RunProcessAsync>b__0(Task t) in /_/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs:line 80
   at System.Threading.Tasks.ContinuationResultTaskFromTask`1.InnerInvoke()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Razor.Design.IntegrationTests.RazorGenerateIntegrationTest.RazorGenerate_Rebuilds_IfInputFilesAreRenamed() in /_/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/RazorGenerateIntegrationTest.cs:line 171
--- End of stack trace from previous location where exception was thrown ---
```
Other tests within that build may have failed with a similar message, but they are not listed here. Check the link above for more info.

This test failed on 2.2.

CC @mkArtakMSFT

This issue was made automatically. If there is a problem contact ryanbrandenburg.

aspnet-hello:
Please use this workflow to address this flaky test issue, including checking applicable checkboxes and filling in the applicable "TODO" entries:

* Is this actually a flaky test?
  * No, this is a regular test failure, fix the test/product (TODO: Link to commit/PR)
  * Yes, proceed below...

[...]

PR #2576, opened 2018-09-10
Better logging to help debug test failure

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2571

Looks like all these tests have suddenly started timing out during restore on Mac. My guess is either restore is slow on Mac for some reason or there is some underlying problem in MsBuild with Mac. Either way this should help us debug what's really going on

### DIFF
--- a/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs
+++ b/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs
@@ -17,7 +17,7 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
             TimeSpan? timeout = null,
             MSBuildProcessKind msBuildProcessKind = MSBuildProcessKind.Dotnet)
         {
-            timeout = timeout ?? TimeSpan.FromSeconds(60);
+            timeout = timeout ?? TimeSpan.FromSeconds(120);
 
             var processStartInfo = new ProcessStartInfo()
             {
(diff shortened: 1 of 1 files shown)


Eilon:
Were some of the failures in release/2.2 as well? If so, should we put this in that branch also?

ajaybhargavb:
Updated. @Eilon

test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/MSBuildProcessManager.cs
```diff
@@ -17,7 +17,7 @@ internal static class MSBuildProcessManager
             TimeSpan? timeout = null,
             MSBuildProcessKind msBuildProcessKind = MSBuildProcessKind.Dotnet)
         {
-            timeout = timeout ?? TimeSpan.FromSeconds(60);
+            timeout = timeout ?? TimeSpan.FromSeconds(120);
```
NTaylorMullen: If we're adding logging why increase the timeout? Don't we want it to fail in order to see more information?

ajaybhargavb: Because if it is going to hang the extra timeout shouldn't make a difference. If it doesn't, looking the output isn't going to tell us anything other than "restore completed" and "Build succeeded". So I thought these can go together

NTaylorMullen: Ah, infinite hangs ok. 👍 :shipit:

Issue #2587, opened 2018-09-11
Update parser test infrastructure


ajaybhargavb:
Update the parser test infrastructure to support both the old and the new syntax tree

Issue #2585, opened 2018-09-11
Rewrite intermediate lowering phase


ajaybhargavb:
- Rewrite `DefaultRazorIntermediateNodeLoweringPhase`
- Rewrite any other visitors that directly operate on the syntax tree

ajaybhargavb:
9a862c6bdd372f8343e266759dec7924f8b8783f

Issue #2584, opened 2018-09-11
Rewrite tag helper binder phase


ajaybhargavb:
- Define the structure of the nodes representing tag helper constructs
- Rewrite `TagHelperParseTreeRewriter` and `TagHelperBlockRewriter`

Issue #2582, opened 2018-09-11
Rewrite TagHelperSpan generation logic


ajaybhargavb:
The work here is to keep the existing tooling experience from breaking.

PR #2590, opened 2018-09-12
Razor parser rewrite

ajaybhargavb:
#2579, #2580, #2581, #2584, #2582, #2587, #2585, #2700, #2689

- Rewrite CSharp parser
- Basic rewrite of HTML parser (More improvements to follow)
- Define and generate syntax nodes and boilerplate
- Rewrite ClassifiedSpan and TagHelperSpan generation logic
- Rewrite TagHelper phase
- Rewrite Intermediate phase
- Rewrite other miscellaneous features and bug fixes
- Rewrite partial parsing
- Added some syntax manipulation APIs
- Removed unused legacy types
- Update parser test infrastructure
- Update tests
- Regenerated baselines

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
@@ -540,7 +540,6 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
                     var isDisallowedContent = true;
                     if (_featureFlags.AllowHtmlCommentsInTagHelpers)
                     {
-                        // TODO: Questionable logic. Need to revisit
                         isDisallowedContent = !IsComment(child) &&
                             !child.IsTransitionSpanKind() &&
                             !child.IsCodeSpanKind();


ajaybhargavb:
Don't bother reviewing this too hard just yet. I suggest skimming it just to look for red flags if any. It will be much easier to review later on when things are cleaner

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -206,195 +207,220 @@ public LoweringVisitor(DocumentIntermediateNode document, IntermediateNodeBuilde
...
-                var span = new SourceSpan(
-                    node.Start.FilePath ?? FilePath,
-                    node.Start.AbsoluteIndex,
-                    node.Start.LineIndex,
-                    node.Start.CharacterIndex,
-                    node.Length);
-                return span;
+                if (node.Position == SourceDocument.Length)
+                {
+                    // This can happen when we have a marker symbol at the end of the syntax tree.
```
ajaybhargavb: This block isn't needed anymore. I've handled this in GetSourceSpan. I'll remove.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -206,195 +207,220 @@ public LoweringVisitor(DocumentIntermediateNode document, IntermediateNodeBuilde
...
-                    directiveNode = new MalformedDirectiveIntermediateNode()
-                    {
-                        DirectiveName = CSharpCodeParser.AddTagHelperDirectiveDescriptor.Directive,
-                        Directive = CSharpCodeParser.AddTagHelperDirectiveDescriptor,
-                        Source = BuildSourceSpanFromNode(span),
-                    };
+                    return base.VisitCSharpStatementLiteral(node);
                 }
-                else
+                else if (context.ChunkGenerator is DirectiveTokenChunkGenerator tokenChunkGenerator)
```
ajaybhargavb: Yes. Intermediate phase is still dependent on `ChunkGenerator`s because the old code was heavily dependent on it. I plan to remove it in the future.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -408,86 +434,149 @@ public MainSourceVisitor(DocumentIntermediateNode document, IntermediateNodeBuil
...
-                });
+                    Visit(prefix);
+                    Visit(node.Value);
+                    Visit(node.ValueSuffix);
+                }
+                else
+                {
+                    if (node.Value != null && node.Value.ChildNodes().All(c => c is MarkupLiteralAttributeValueSyntax))
+                    {
+                        // We need to do what ConditionalAttributeCollapser used to do.
```
ajaybhargavb: Because we got rid of ConditionalAttributeCollapser, I need this logic to merge contiguous MarkupLiterals so we generate a single Intermediate node for all of it. I'll change the above comment to explain this.

src/Microsoft.AspNetCore.Razor.Language/Legacy/AddImportChunkGenerator.cs
```diff
@@ -14,23 +14,6 @@ public AddImportChunkGenerator(string ns)
 
         public string Namespace { get; }
 
-        public override void Accept(ParserVisitor visitor, Span span)
```
ajaybhargavb: `ChunkGenerator`s don't really do anything anymore other than be a marker.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
```diff
@@ -715,39 +360,289 @@ private static string GetPropertyType(string name, IEnumerable<TagHelperDescript
...
+            else if (attributeBlock is MarkupAttributeBlockSyntax attribute)
+            {
+                return attribute.Value?.GetContent();
+            }
+
+            return null;
         }
 
-        private static void ConfigureNonStringAttribute(SpanBuilder builder, bool isDuplicateAttribute)
+        private class AttributeValueRewriter : SyntaxRewriter
```
ajaybhargavb: Whatever is in it is probably ugly. TagHelperBlockRewriter used to do a whole bunch of "special" rewriting of attribute values. Keeping the same behavior while operating on the new tree makes it really ugly. I plan to clean this up as much as possible as part of attribute value syntax improvements.

src/Microsoft.AspNetCore.Razor.Language/Syntax/ChildSyntaxList.cs
```diff
@@ -0,0 +1,710 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal readonly struct ChildSyntaxList : IEquatable<ChildSyntaxList>, IReadOnlyList<SyntaxNode>
```
ajaybhargavb: From Roslyn

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupMinimizedTagHelperAttributeSyntax.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal sealed partial class MarkupMinimizedTagHelperAttributeSyntax
+    {
+        private static readonly string TagHelperAttributeInfoKey = typeof(TagHelperAttributeInfo).Name;
+
+        public TagHelperAttributeInfo TagHelperAttributeInfo
```
ajaybhargavb: Added Metadata about the node as first class properties instead on an annotation.

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxNode.Iterators.cs
```diff
@@ -0,0 +1,126 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal abstract partial class SyntaxNode
```
ajaybhargavb: From Roslyn

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxNodeExtensions.cs
```diff
@@ -1,13 +1,339 @@
...
+            {
+                throw new ArgumentNullException(nameof(node));
+            }
+
+            var tokens = node.DescendantNodes().Where(n => n.IsToken).Cast<SyntaxToken>();
+            var content = string.Concat(tokens.Select(t => t.Content));
+            return content;
+        }
+
+        public static string GetTagName(this MarkupTagBlockSyntax tagBlock)
```
ajaybhargavb: These are currently extension methods. But these will go away once the structure of a Tag Block is improved.

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxReplacer.cs
```diff
@@ -0,0 +1,207 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal static class SyntaxReplacer
```
ajaybhargavb: From Roslyn

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxRewriter.cs
```diff
@@ -0,0 +1,148 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal abstract partial class SyntaxRewriter : SyntaxVisitor<SyntaxNode>
```
ajaybhargavb: From Roslyn

src/Microsoft.AspNetCore.Razor.Language/Syntax/InternalSyntax/SyntaxListPool.cs
```diff
@@ -0,0 +1,100 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax.InternalSyntax
+{
+    internal class SyntaxListPool
```
ajaybhargavb: From Roslyn

ajaybhargavb:
@rynowak is added to the review. #Closed

src/Microsoft.AspNetCore.Razor.Language/TagHelperSpanVisitor.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Microsoft.AspNetCore.Razor.Language.Syntax;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class TagHelperSpanVisitor : SyntaxRewriter
```
rynowak: >SyntaxRewriter [](start = 42, length = 14)

Does this need to be a rewriter? It doesn't change anything, it just produces output

src/Microsoft.AspNetCore.Razor.Language/Legacy/EditResult.cs
```diff
@@ -1,17 +1,19 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Razor.Language.Syntax;
+
 namespace Microsoft.AspNetCore.Razor.Language.Legacy
```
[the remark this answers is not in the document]
rynowak: I think what's here is fine for now.

---
In reply to: [233261721](https://github.com/aspnet/Razor/pull/2590#discussion_r233261721) [](ancestors = 233261721)

src/Microsoft.AspNetCore.Razor.Language/Legacy/IParentChunkGenerator.cs
```diff
@@ -5,9 +5,5 @@ namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
     internal interface IParentChunkGenerator
```
rynowak: >IParentChunkGenerator [](start = 23, length = 21)

Can we remove this? Or still needed as a marker type?

src/Microsoft.AspNetCore.Razor.Language/Legacy/SpanContext.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language.Syntax.InternalSyntax;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal class SpanContext
```
rynowak: >SpanContext [](start = 19, length = 11)

I see that this is in the .Leygacy namespace which is OK. I'm hoping that this is somewhat of a temporary bridge between old code and new code.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperAttributeInfo.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal class TagHelperAttributeInfo
```
rynowak: >TagHelperAttributeInfo [](start = 19, length = 22)

I'm assuming that this is here to bridge between old and new.

src/Microsoft.AspNetCore.Razor.Language/Syntax/LegacySyntaxNodeExtensions.cs
```diff
@@ -0,0 +1,248 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
```
rynowak: >yntax [](start = 47, length = 5)

Should this be in `.Legacy`? My preference would be that anything we want to **keep forever** should live in `.Syntax` and anything that's transitional should live in `.Legacy`.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,290 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal static class MarkupElementRewriter
```
rynowak: >MarkupElementRewriter [](start = 26, length = 21)

I'm assuming that this is also temporary (this is the code that groups elements/tags).

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxNode.cs
```diff
@@ -1,12 +1,16 @@
...
+using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Text;
 using System.Threading;
 
 namespace Microsoft.AspNetCore.Razor.Language.Syntax
 {
-    internal abstract class SyntaxNode
+    [DebuggerDisplay("{GetDebuggerDisplay(), nq}")]
```
rynowak: >GetDebuggerDisplay [](start = 23, length = 18)

Thank you for embracing sanity and not using nameof here :+1:

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxNodeExtensions.cs
```diff
@@ -1,13 +1,339 @@
...
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            try
+            {
+                if (source.Length == 0)
+                {
+                    // Just a marker symbol
+                    return new SourceLocation(source.FilePath, 0, 0, 0);
```
rynowak: Is this really correct? Don't marker symbols still have a location?

src/Microsoft.AspNetCore.Razor.Language/ClassifiedSpanVisitor.cs
```diff
@@ -0,0 +1,336 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language.Legacy;
+using Microsoft.AspNetCore.Razor.Language.Syntax;
+
+namespace Microsoft.AspNetCore.Razor.Language
+{
+    internal class ClassifiedSpanVisitor : SyntaxRewriter
```
rynowak: >ClassifiedSpanVisitor [](start = 19, length = 21)

At this point I've gotten through all of the files with a small number of changes or those that were copied from Roslyn. My remaining list to review:

ClassifiedSpanVisitor
DefaultRazorIntermediateNodeLoweringPhase
DefaultRazorTagHelperBinderPhase
CSharpCodeParser
HtmlMarkupParser
TagHelperParseTreeRewriter
TokenizerBackedParser
WhitespaceRewriter
MarkupElementRewriter

src/Microsoft.AspNetCore.Razor.Language/Legacy/IParentChunkGenerator.cs
ajaybhargavb: I think we only need ISpanChunkGenerator as the marker type. I'll check and make sure.

---
In reply to: [233615835](https://github.com/aspnet/Razor/pull/2590#discussion_r233615835) [](ancestors = 233615835)

src/Microsoft.AspNetCore.Razor.Language/TagHelperSpanVisitor.cs
ajaybhargavb: True. This can just be a SyntaxVisitor

---
In reply to: [233601788](https://github.com/aspnet/Razor/pull/2590#discussion_r233601788) [](ancestors = 233601788)

src/Microsoft.AspNetCore.Razor.Language/Legacy/SpanContext.cs
ajaybhargavb: Yes but they won't go away anytime soon because we don't have a plan for getting rid of EditHandlers

---
In reply to: [233617210](https://github.com/aspnet/Razor/pull/2590#discussion_r233617210) [](ancestors = 233617210)

src/Microsoft.AspNetCore.Razor.Language/Syntax/LegacySyntaxNodeExtensions.cs
ajaybhargavb: I'll move it.

---
In reply to: [233618684](https://github.com/aspnet/Razor/pull/2590#discussion_r233618684) [](ancestors = 233618684)

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Yes but I didn't put it in Legacy because this operates completely on the new infrastructure and doesn't use any of out legacy components.

---
In reply to: [233619189](https://github.com/aspnet/Razor/pull/2590#discussion_r233619189) [](ancestors = 233619189)

src/Microsoft.AspNetCore.Razor.Language/ClassifiedSpanVisitor.cs
```diff
@@ -0,0 +1,336 @@
...
+        {
+            return WriteBlock(node, BlockKindInternal.Template, base.VisitCSharpTemplateBlock);
+        }
+
+        public override SyntaxNode VisitMarkupBlock(MarkupBlockSyntax node)
+        {
+            return WriteBlock(node, BlockKindInternal.Markup, base.VisitMarkupBlock);
+        }
+
+        public override SyntaxNode VisitGenericBlock(GenericBlockSyntax node)
```
ajaybhargavb: I already regret having a `GenericBlock`. This is currently only use in attribute values. I'm hoping I can get rid of this when I clean that up.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -673,22 +821,229 @@ public override void VisitTagHelperBlock(TagHelperChunkGenerator chunkGenerator,
...
 
-                AddTagHelperAttributes(tagHelperBlock.Attributes, tagHelperBlock.Binding);
+                Visit(node.StartTag);
 
                 _builder.Pop(); // Pop TagHelperIntermediateNode
+
+                // No need to visit the end tag because we don't write any IR for it.
+
+                // We don't want to track attributes from a previous tag helper element.
+                _renderedBoundAttributeNames.Clear();
```
NTaylorMullen: Kind of jank to maintain state here. I get why you have to but just calling it out if you can think of a better way.

src/Microsoft.AspNetCore.Razor.Language/ClassifiedSpanVisitor.cs
rynowak: Yeah there shouldn't be anything *generic*. Everything should be specific.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -673,22 +821,229 @@ public override void VisitTagHelperBlock(TagHelperChunkGenerator chunkGenerator,
...
+                return node;
+            }
+
+            private void VisitAttributeValue(SyntaxNode node)
+            {
+                if (node == null)
+                {
+                    return;
+                }
+
```
NTaylorMullen: #importantfeedback :trollface:

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -673,22 +821,229 @@ public override void VisitTagHelperBlock(TagHelperChunkGenerator chunkGenerator,
...
+                        context = literal.GetSpanContext();
+                        builder.AddRange(literal.LiteralTokens);
+                    }
+                    var rewritten = SyntaxFactory.CSharpExpressionLiteral(builder.ToList()).Green.CreateRed(node.Parent, position);
+                    rewritten = context != null ? rewritten.WithSpanContext(context) : rewritten;
+                    Visit(rewritten);
+                }
+                else
+                {
+                    Visit(node);
```
NTaylorMullen: Wait, how does this not stack overflow?

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -673,22 +821,229 @@ public override void VisitTagHelperBlock(TagHelperChunkGenerator chunkGenerator,
...
+                    markupBlock.Children[1] is MarkupEphemeralTextLiteralSyntax)
+                {
+                    // This is a special case when we have an attribute like attr="@@foo".
+                    // In this case, we want the foo to be written out as HtmlContent and not HtmlAttributeValue.
+                    Visit(markupBlock);
+                    children = children.Skip(1).ToList();
+                    position = children.Count > 0 ? children[0].Position : position;
+                }
+
+                if (children.All(c => c is MarkupLiteralAttributeValueSyntax))
```
NTaylorMullen: Shouldn't this be elseif? Maybe i'm misunderstanding but doesn't the first bit result in everything else returning false?

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -21,14 +23,28 @@ protected TokenizerBackedParser(LanguageCharacteristics<TTokenizer> language, Pa
 
             var languageTokenizer = Language.CreateTokenizer(Context.Source);
             _tokenizer = new TokenizerView<TTokenizer>(languageTokenizer);
-            Span = new SpanBuilder(CurrentLocation);
```
NTaylorMullen: Q: Why did CurrentLocation go away everywhere? I don't actually think it's used any more either. Could nuke

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -103,264 +103,146 @@ public CSharpCodeParser(IEnumerable<DirectiveDescriptor> directives, ParserConte
...
                     NextToken();
 
+                    // Unless changed, the block is a statement block
                     AcceptWhile(IsSpacingToken(includeNewLines: true, includeComments: true));
+                    builder.Add(OutputTokensAsStatementLiteral());
 
-                    var current = CurrentToken;
+                    // We are usually called when the other parser sees a transition '@'. Look for it.
+                    SyntaxToken transitionToken = null;
                     if (At(SyntaxKind.StringLiteral) &&
```
NTaylorMullen: Curious, why string literal here?

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -570,108 +492,202 @@ protected SyntaxToken AcceptWhiteSpaceInLines()
...
 
-                _tokenizer.Next();
+                NextToken();
             }
+
             return lastWs;
         }
 
-        protected bool AtIdentifier(bool allowKeywords)
+        protected internal bool Optional(SyntaxKind type)
```
NTaylorMullen: `OptionallyAccept` or `TryAccept` for consistency.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -570,108 +492,202 @@ protected SyntaxToken AcceptWhiteSpaceInLines()
...
+                        new SourceSpan(start, contentLength: 1 /* { OR } */),
+                        Language.GetSample(left),
+                        Language.GetSample(right)));
             }
+
+            return Balance(builder, mode, left, right, start);
         }
 
-        private void Configure(SpanKindInternal? kind, AcceptedCharactersInternal? accepts)
+        protected internal bool Balance(SyntaxListBuilder<RazorSyntaxNode> builder, BalancingModes mode, SyntaxKind left, SyntaxKind right, SourceLocation start)
```
NTaylorMullen: I believe we only ever balance in C#, if that's the case all of our balance methods should go in the csharp code parser.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -570,108 +492,202 @@ protected SyntaxToken AcceptWhiteSpaceInLines()
...
+                {
+                    // Accept all the tokens we saw
+                    Accept(tokens);
+                }
             }
-            if (accepts != null)
+            return nesting == 0;
+        }
+
+        protected virtual void ParseEmbeddedTransition(in SyntaxListBuilder<RazorSyntaxNode> builder)
```
NTaylorMullen: No need to be virtual, can inline the implementation into Balance when it's moved into CSharpCodeParser or, if the implementation is massive can just leave the method in the CSharpCodeParser

NTaylorMullen: Ditto with IsAtEmbeddedTransition

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
```diff
@@ -553,1618 +450,1915 @@ private bool MethodCallOrArrayIndex(AcceptedCharactersInternal acceptedCharacter
...
-            AfterIfClause();
-            if (topLevel)
-            {
-                CompleteBlock();
+                EnsureCurrent();
+                return false;
             }
         }
 
-        p
```
NTaylorMullen: Small enough method, I'd just inline this. The bits in this class are dope, this as its own method stick out as a sore thumb.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -570,108 +492,202 @@ protected SyntaxToken AcceptWhiteSpaceInLines()
...
+        {
+            if (TokenBuilder.Count == 0 && Context.LastAcceptedCharacters != AcceptedCharactersInternal.Any)
             {
-                Span.EditHandler.AcceptedCharacters = accepts.Value;
+                Accept(Language.CreateMarkerToken());
             }
         }
 
-        protected virtual void OutputSpanBeforeRazorComment()
+        protected MarkupTextLiteralSyntax OutputAsMarkupLiteral()
```
NTaylorMullen: Why are markup output methods in the TokenizeraBackedParser? It possible to move them into the Html parser?

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -68,26 +64,57 @@ private StringComparison Comparison
...
+                builder.Add(OutputAsMarkupLiteral());
+
+                var markup = SyntaxFactory.MarkupBlock(builder.ToList());
+
+                return SyntaxFactory.RazorDocument(markup);
+            }
         }
 
-        protected void SkipToAndParseCode(SyntaxKind type)
+        private void SkipToAndParseCode(in SyntaxListBuilder<RazorSyntaxNode> builder, SyntaxKind type)
```
NTaylorMullen: I assume once we evolve this class that SkipToAndParseCode goes away yes? Having it introduces too much genericness into our HtmlParser for us to properly evolve it in the future.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -222,282 +259,466 @@ protected void SkipToAndParseCode(Func<SyntaxToken, bool> condition)
...
-
-            return false;
-        }
-
-        private void OptionalBangEscape()
+        /// <summary>
+        /// Reads the content of a tag (if present) in the MarkupDocument (or MarkupSection) context,
+        /// where we don't care about maintaining a stack of tags.
+        /// </summary>
+        private void ParseTagInDocumentContext(in SyntaxListBuilder<RazorSyntaxNode> builder)
```
NTaylorMullen: Should factor this method a lot like you have ParseCodeBlock in the CSharpCodeParser. i.e.
- if HtmlComment => HtmlComment()
- If XmlTag => XmlTag()
- If SpecialTag => `SpecialTag()`
- else `ParseHtmlStartTag()`

I'm sure I didn't capture everything but that's the gist. Feel free to add this to the feedback look. As it is right now it's spaghetti 😢

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -222,282 +259,466 @@ protected void SkipToAndParseCode(Func<SyntaxToken, bool> condition)
...
+                    !IsEndOfAttributeValue(quote, token));
+                Accept(valueTokens);
+                var value = OutputAsMarkupLiteral();
+
+                var literalAttributeValue = SyntaxFactory.MarkupLiteralAttributeValue(prefix, value);
+                builder.Add(literalAttributeValue);
+            }
+        }
+
+        private void RecoverToEndOfTag(in SyntaxListBuilder<RazorSyntaxNode> builder)
```
NTaylorMullen: You'll eventually need to re-work this a little bit to instead of accepting tokens left and right without any context to be intermingling `<missing>` tokens in places that make sense.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorIntermediateNodeLoweringPhase.cs
```diff
@@ -206,195 +207,220 @@ public LoweringVisitor(DocumentIntermediateNode document, IntermediateNodeBuilde
 
             public IReadOnlyList<UsingReference> Usings => _usings;
 
-            public string FilePath { get; set; }
+            public RazorSourceDocument SourceDocument { get; set; }
```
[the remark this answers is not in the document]
ajaybhargavb: This was because this makes so we don't have to new up a visitor for every import document. Just change the Source and call visit on the root.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -222,282 +259,466 @@ protected void SkipToAndParseCode(Func<SyntaxToken, bool> condition)
...
+        private void ParseQuoted(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            var type = CurrentToken.Kind;
             AcceptAndMoveNext();
-            return AcceptUntilAll(SyntaxKind.QuestionMark, SyntaxKind.CloseAngle);
+            ParseQuoted(builder, type);
         }
 
-        private bool BangTag()
+        private void ParseQuoted(in SyntaxListBuilder<RazorSyntaxNode> builder, SyntaxKind type)
```
NTaylorMullen: inline into method above.

src/Microsoft.AspNetCore.Razor.Language/ClassifiedSpanVisitor.cs
```diff
@@ -0,0 +1,336 @@
...
+        {
+            if (!(node.Parent is MarkupTagHelperAttributeValueSyntax) &&
```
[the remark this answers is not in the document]
ajaybhargavb: I'll do you one better. I'll delete it. This will never happen because of how we rewrite in TagHelperBlockRewriter.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -1555,239 +1755,47 @@ private void EndTagBlock(Stack<Tuple<SyntaxToken, SourceLocation>> tags, bool co
...
-                tokenType != SyntaxKind.ForwardSlash &&
-                tokenType != SyntaxKind.DoubleQuote &&
-                tokenType != SyntaxKind.SingleQuote &&
-                tokenType != SyntaxKind.Equals &&
-                tokenType != SyntaxKind.Unknown;
+            builder.Add(OutputAsMarkupLiteral());
         }
 
-        public void ParseDocument()
+        public MarkupBlockSyntax ParseRazorBlock(Tuple<string, string> nestingSequences, bool caseSensitive)
```
NTaylorMullen: Curious, why do we want this case sensitive?

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -1555,239 +1755,47 @@ private void EndTagBlock(Stack<Tuple<SyntaxToken, SourceLocation>> tags, bool co
...
-                    }
-                    AddMarkerTokenIfNecessary();
-                    Output(SpanKindInternal.Markup);
-                }
+                return SyntaxFactory.MarkupBlock(builder.ToList());
             }
         }
 
-        private void NonNestingSection(string[] nestingSequenceComponents)
+        private void NonNestingSection(in SyntaxListBuilder<RazorSyntaxNode> builder, string[] nestingSequenceComponents)
```
NTaylorMullen: The word `section` here is a bit confusing. What does this guy do exactly? Is it the bits that parse Html tags in `@{...}` to ensure they match? A bit confused is all

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
```diff
@@ -19,10 +19,6 @@ internal class HtmlMarkupParser : TokenizerBackedParser<HtmlTokenizer>
             SyntaxFactory.Token(SyntaxKind.Bang, "!"),
```
NTaylorMullen: Not blocking the PR on this but this class definitely need to be refactored in ways like CSharpCodeParser was. CSharpCodeParser is beaaautiful in terms of its factorings and this could benefit from similar treatment.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: Yes absolutely. https://github.com/aspnet/Razor/issues/2583 requires me to change most of these. I kept this to not introduce too much churn when we switched baselines for the first time.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: Yup, that is the goal 🏁

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
```diff
@@ -325,175 +246,129 @@ protected internal bool At(SyntaxKind type)
...
-            Debug.Assert(!EndOfFile && CurrentToken != null && types.Contains(CurrentToken.Kind));
-            AcceptAndMoveNext();
+            Debug.Assert(!EndOfFile && CurrentToken != null);
+            var token = CurrentToken;
+            NextToken();
+            return token;
         }
 
-        protected internal bool Optional(KnownTokenType type)
+        protected SyntaxToken GetExpectedToken(params SyntaxKind[] kinds)
```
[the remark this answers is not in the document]
ajaybhargavb: I'll rename. I'm not sure what you mean by "failing fast"? This is the same way we do Debug.Assert everywhere.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
ajaybhargavb: Because we no longer need SourceLocations when building the Green tree. The diagnostics use `CurrentStart`. So, `CurrentLocation` is unused.

src/Microsoft.AspNetCore.Razor.Language/Legacy/CSharpCodeParser.cs
ajaybhargavb: Because our Tokenizer is weird and tokenizes `@""` as one StringLiteral token.

src/Microsoft.AspNetCore.Razor.Language/Legacy/TokenizerBackedParser.cs
ajaybhargavb: Spoke offline. They exist here because the CSharp parser also uses it to parse directives.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: It doesn't matter in our case. Because the only nesting sequence we compare are `{` and `}`. This is just written in a generic way so that it can be used for other razor blocks where we can specify the nesting sequences.

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: I've removed this and cleaned up a little. We parse section blocks using this logic and I assume that's is where the name came from. Anyway, we don't call into `NonNestingSection` anymore and I've renamed `NestingSection` to `NestingBlock`. 
What this does is handle nesting of blocks inside Razor Blocks. E.g,

```HTML
@section MySection {
<div>hello</div>
@if (true) {
  // I am a nested block inside a Razor block so keep track of all the open curlys (aka. nesting sequences)
}
}
```

ajaybhargavb: Correct me if I am wrong though. I could be misunderstanding some legacy piece.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
NTaylorMullen: Feel free to nuke it if you want

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs
```diff
@@ -289,636 +342,389 @@ private bool TryRewriteTagHelper(Block tagBlock, ErrorSink errorSink)
...
-                // Update the _currentBlock to point at the last tracked block because it's not complete.
-                _currentBlock = previousBlock;
-            }
-            else
+            private bool HasAllowedChildren()
             {
-                _currentBlock = currentBlock;
-            }
-        }
+                // TODO: Questionable logic. Need to revisit
```
NTaylorMullen: ?

NTaylorMullen:
I Did it! That was super dope, nothing blocking

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperInfo.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal class TagHelperInfo
```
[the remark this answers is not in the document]
ajaybhargavb: No. I moved this type out of Legacy. This is a property on `TagHelperElementSyntax`

src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperBlockRewriter.cs
```diff
@@ -14,107 +12,30 @@ internal static class TagHelperBlockRewriter
...
-        private static TagMode GetTagMode(
-            string tagName,
-            Block beginTagBlock,
+        public static TagMode GetTagMode(
+            MarkupTagBlockSyntax tagBlock,
             TagHelperBinding bindingResult,
             ErrorSink errorSink)
         {
-            var childSpan = beginTagBlock.FindLastDescendentSpan();
+            var childSpan = tagBlock.GetLastToken()?.Parent;
```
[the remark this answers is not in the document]
ajaybhargavb: I don't think the last token will have both `/` and `>`. They will be split as two tokens. Looking at the parent seemed the right way.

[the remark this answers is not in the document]
NTaylorMullen: Ah, I see. 👍

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/Legacy/HtmlMarkupParser.cs
ajaybhargavb: Will nuke later

PR #2592, opened 2018-09-13
Check for spans that are missing from output

rynowak:


test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -220,9 +220,102 @@ protected void AssertSourceMappingsMatchBaseline(RazorCodeDocument document)
...
+                if (span.Start.FilePath == null || span.Start.FilePath != codeDocument.Source.FilePath)
+                {
+                    // Not in the main file, skip.
+                    continue;
+                }
+
+                var location = new SourceSpan(span.Start, span.Length);
+                var expectedSpan = sourceContent.Substring(span.Start.AbsoluteIndex, span.Length);
+
+                // See #
```
rynowak: I'm planning to log a bug for this, which we will likely end up closing. 

We can fix the issue in the parser but it impacts the completion experience. For right now we're better off with the bug since the workaround in the verifier is localized.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -220,9 +220,102 @@ protected void AssertSourceMappingsMatchBaseline(RazorCodeDocument document)
...
+                {
+                    // For now we don't verify whitespace inside of a directive. We know that directives cheat
+                    // with how they bound whitespace/C#/markup to make completion work.
+                    if (span.Parent is Block block && block.Type == BlockKindInternal.Directive)
+                    {
+                        continue;
+                    }
+                }
+
+                // See #
```
rynowak: I'm planning to log a bug for this, I expect it will end up in the same place as the above. We categorize the second `@` in `@{ @@Da }` as C# but then we don't put it in the buffer. I expect that this is done for similar reasons, and that changing it could break completion.

ajaybhargavb:
IMO, all this logic should go in to https://github.com/aspnet/Razor/blob/master/test/Microsoft.AspNetCore.Razor.Test.Common/Language/SyntaxTreeVerifier.cs where we verify something similar. We already use that during our parser baseline tests. The integration test should just call `SyntaxTreeVerifier.Verify(syntaxTree)` in this case.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -220,9 +220,102 @@ protected void AssertSourceMappingsMatchBaseline(RazorCodeDocument document)
...
+                if (string.Equals("@", expectedSpan) && span.Kind == SpanKindInternal.Code)
+                {
+                    // For now we don't verify an escaped transition. In some cases one of the @ tokens in @@foo
+                    // will be mapped as C# but will not be present in the output buffer because it's not actually C#.
+                    continue;
+                }
+
+                var found = false;
+                for (var j = 0; j < csharpDocument.SourceMappings.Count; j++)
```
NTaylorMullen: Do we care if some source mappings aren't accounted for? Basically, right now you're checking each Span against all of the possible source mappings. It's possible we've generated an extra source mapping that incorrectly maps back to the SyntaxTree. Very possible with us manipulating the IR as much as we do.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
```diff
@@ -220,9 +220,102 @@ protected void AssertSourceMappingsMatchBaseline(RazorCodeDocument document)
...
+                if (span.Kind == SpanKindInternal.Code)
+                {
+                    CodeSpans.Add(span);
+                }
+
+                base.VisitSpan(span);
+            }
+        }
+
+        private static string EscapeWhitespace(string content)
```
NTaylorMullen: nest this function in the verify method so we don't accidentally use this in the future. Our other tests should properly already escape whitespace.

rynowak:
> Ignore. This involves SourceMappings. Doesn't make sense to put it there.

Yeah, I was about to reply with that 😆

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
rynowak: That's a good point. Let me see if it's easy to add.

test/Microsoft.AspNetCore.Razor.Test.Common/Language/IntegrationTests/IntegrationTestBase.cs
rynowak: That wouldn't be harmful though?

PR #2595, opened 2018-09-14
Update to use non-capturing timer

rynowak:
FYI only - replacing the code with the package that does the same thing.


NTaylorMullen:
> Not sure what you mean?

https://github.com/aspnet/Razor/blob/b20921741ced5c647e2e2f810f89613a8d4f7287/src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs#L145-L153

PR #2601, opened 2018-09-18
Move RazorCompletionFactsService to CodeAnalysis.Workspaces.Razor.

NTaylorMullen:
- Also updated workspaces test project to also compile for netcoreapp2.2.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorCompletionFactsService.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorCompletionFactsService.cs
@@ -4,9 +4,9 @@
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Razor.Language;
 
-namespace Microsoft.VisualStudio.Editor.Razor
+namespace Microsoft.CodeAnalysis.Razor
 {
-    public abstract class RazorCompletionFactsService
+    internal abstract class RazorCompletionFactsService
     {
         public abstract IReadOnlyList<RazorCompletionItem> GetCompletionItems(RazorSyntaxTree syntaxTree, SourceSpan location);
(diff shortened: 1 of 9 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorCompletionFactsService.cs
```diff
@@ -4,7 +4,7 @@
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Razor.Language;
 
-namespace Microsoft.VisualStudio.Editor.Razor
+namespace Microsoft.CodeAnalysis.Razor
 {
     public abstract class RazorCompletionFactsService
```
rynowak: public? You said this was internal earlier.

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorCompletionItem.cs
```diff
@@ -3,7 +3,7 @@
 
 using System;
 
-namespace Microsoft.VisualStudio.Editor.Razor
+namespace Microsoft.CodeAnalysis.Razor
```
rynowak: Yeah these are all breaking changes 👎

src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorCompletionFactsService.cs
NTaylorMullen: Haven't shipped tooling with these yet (did you in your dev16 primer?), these are brand new APIs that I just added not too long ago. Anyhow, nothing depends on the public API of this and I have IVT in other places. Will make em internal.

PR #2603, opened 2018-09-20
Make it so the RazorDirectiveCompletionProvider doesn't load extra as…

NTaylorMullen:
…semblies in C#.

- Added a layer of indirection to allow for proper Lazy loading of Razor and future async completion specific assemblies.

### DIFF
--- a/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
+++ b/src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
@@ -36,15 +36,13 @@ namespace Microsoft.VisualStudio.Editor.Razor
             CSharpCodeParser.TagHelperPrefixDirectiveDescriptor,
         };
         private readonly Lazy<RazorCodeDocumentProvider> _codeDocumentProvider;
-        private readonly Lazy<RazorCompletionFactsService> _completionFactsService;
-        private readonly IAsyncCompletionBroker _asyncCompletionBroker;
+        private readonly Lazy<CompletionProviderDependencies> _dependencies;
         private readonly RazorTextBufferProvider _textBufferProvider;
 
         [ImportingConstructor]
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
```diff
@@ -207,4 +199,40 @@ protected virtual bool TryGetRazorSnapshotPoint(CompletionContext context, out S
             return false;
         }
     }
+
+    // These types are only for this class to provide indirection for assembly loads.
+    internal abstract class CompletionProviderDependencies
```
NTaylorMullen: @rynowak if you know of a better way to do this I'm all ears 😉

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
```diff
@@ -111,13 +103,6 @@ public override Task ProvideCompletionsAsync(CompletionContext context)
                 return Task.CompletedTask;
             }
 
-            if (!_textBufferProvider.TryGetFromDocument(context.Document, out var textBuffer) ||
-                !_asyncCompletionBroker.IsCompletionSupported(textBuffer.ContentType))
```
NTaylorMullen: We shouldn't be the one loading extra IAsyncCompletion bits if we're not operating on a Razor file.

src/Microsoft.VisualStudio.Editor.Razor/RazorDirectiveCompletionProvider.cs
rynowak: No, this seems about right.

NTaylorMullen:
> Should we have a work item to undo this once modern completion lands?

Have one already in VSTS that everyone always gets bugged about 😄.

PR #2604, opened 2018-09-20
Rewrite WhitespaceRewriter

ajaybhargavb:
I added a bunch of syntax manipulation APIs as part of https://github.com/aspnet/Razor/pull/2590/commits/623d743e31f8e7b4729227243c6ab20ecc749d2b (Didn't add that here because it might be too much noise)

- Used some of those APIs to rewrite WhitespaceRewriter
- The new syntax tree makes it so that we don't need `ConditionalAttributeCollapser` (Needs some changes at the IR lowering level)
- Updated the test

In the future, we could also think about removing the need for WhitespaceRewriter entirely.


src/Microsoft.AspNetCore.Razor.Language/Legacy/LegacyConditionalAttributeCollapser.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            var span = node as Span;
+            Debug.Assert(span != null);
+
+            return span != null &&
+                (span.ChunkGenerator is LiteralAttributeChunkGenerator ||
+                span.ChunkGenerator is MarkupChunkGenerator ||
+                span.ChunkGenerator == SpanChunkGenerator.Null);
+        }
+    }
```
rynowak: I'm assuming that this is just a rename

src/Microsoft.AspNetCore.Razor.Language/Legacy/LegacyWhitespaceRewriter.cs
```diff
@@ -0,0 +1,41 @@
...
+                // Remove the old whitespace node
+                newNodes = block.Children.Skip(1);
+            }
+
+            foreach (SyntaxTreeNode node in newNodes)
+            {
+                newBlock.Children.Add(node);
+            }
+            return newBlock.Build();
+        }
```
rynowak: I'm assuming that this is just a rename

src/Microsoft.AspNetCore.Razor.Language/Legacy/LegacyConditionalAttributeCollapser.cs
ajaybhargavb: Yes

src/Microsoft.AspNetCore.Razor.Language/Legacy/LegacyWhitespaceRewriter.cs
ajaybhargavb: Yes

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
```diff
@@ -1,41 +1,65 @@
...
+            for (var i = 0; i < children.Count; i++)
+            {
+                var child = children[i];
+                if (child is CSharpCodeBlockSyntax codeBlock &&
+                    TryRewriteWhitespace(codeBlock, out var rewritten, out var whitespaceLiteral))
+                {
+                    // Replace the existing code block with the whitespace literal
+                    // followed by the rewritten code block (with the code whitespace removed).
+                    result = result.ReplaceNode(codeBlock, new SyntaxNode[] { whitespaceLiteral, rewritten });
+                }
```
rynowak: The idiomatic way to implement this is to override `Visit(CSharpCodeBlock )` and do your work in there.

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
```diff
@@ -1,41 +1,65 @@
...
+                return false;
             }
 
-            foreach (SyntaxTreeNode node in newNodes)
+            if (children[0] is CSharpStatementLiteralSyntax literal &&
+                (children[1] is CSharpExpression || children[1] is CSharpImplicitExpression))
             {
-                newBlock.Children.Add(node);
+                var literalContent = GetContent(literal);
+                if (literalContent.All(char.IsWhiteSpace))
```
rynowak: Should this be `string.IsWhitespace(`? This is a really inefficient way to implement that kind of check.

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
rynowak: Ah NVM, I see that you're doing operating on `result`. NP.

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
```diff
@@ -1,41 +1,65 @@
...
 namespace Microsoft.AspNetCore.Razor.Language.Legacy
 {
-    internal class WhiteSpaceRewriter : MarkupRewriter
+    internal class WhitespaceRewriter : SyntaxRewriter
     {
-        protected override bool CanRewrite(Block block)
+        public override SyntaxNode Visit(SyntaxNode node)
         {
-            return block.Type == BlockKindInternal.Expression && Parent != null;
+            var result = node;
```
rynowak: This isn't really needed. You could use `node` instead becuase the `result` isn't really ever returned.

ajaybhargavb: I don't follow. It is used later in `base.Visit(result)`.

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
ajaybhargavb: This line was copy-paste from old code. Will replace.

src/Microsoft.AspNetCore.Razor.Language/Legacy/WhiteSpaceRewriter.cs
rynowak: I'm saying you could just reassign `node`. This is a nitpick anyways, you do you

PR #2606, opened 2018-09-22
Fix #2054 - make TagHelperResolutionResult internal

### LINKED ISSUE #2054: Make TagHelperResolutionResult internal
https://github.com/aspnet/Razor/blob/dev/src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperResolutionResult.cs#L9

This type is public, but I don't think WTE is using it anymore.

We don't ship this package so we should be able to make it internal if they aren't using it.

rynowak:
This type isn't used by WTE anymore.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperResolutionResult.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperResolutionResult.cs
@@ -7,9 +7,9 @@ using Microsoft.AspNetCore.Razor.Language;
 
 namespace Microsoft.CodeAnalysis.Razor
 {
-    public sealed class TagHelperResolutionResult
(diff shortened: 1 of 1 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/TagHelperResolutionResult.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.CodeAnalysis.Razor
 {
-    public sealed class TagHelperResolutionResult
+    internal sealed class TagHelperResolutionResult
     {
         internal static TagHelperResolutionResult Empty = new TagHelperResolutionResult(Array.Empty<TagHelperDescriptor>(), Array.Empty<RazorDiagnostic>());
```
pranavkm: `static readonly`

PR #2620, opened 2018-09-27
Test Directives correctly

ajaybhargavb:
Because of some unknown reasons, almost all our directives tests were testing them under CSharp context when we know that is not real scenario. This is causing a lot of confusion when generating baselines for the new tree. So I decided to fix the existing baselines first to spare all of us from confusion later on.


test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/CSharpDirectivesTest.cs
```diff
@@ -121,7 +121,7 @@ public void DirectiveDescriptor_TokensMustBeSeparatedBySpace()
                 b => b.AddStringToken().AddStringToken());
```
NTaylorMullen: Would probably just rename to `RazorDirectivesTest.cs`

PR #2624, opened 2018-10-01
Added MarkupElementRewriter

ajaybhargavb:
#2584 

This is going to be run before and after the tag helper phase. I am currently not adding any errors at this level. All errors will be added during the tag helper phase. Sending this as a separate PR just to make sure everyone is on the same page.
Added tests


src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal class MarkupElementRewriter
+    {
+        public static RazorSyntaxTree Rewrite(RazorSyntaxTree syntaxTree)
```
ajaybhargavb: This is called before tag helper phase.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+        {
+            var rewriter = new Rewriter();
+            var rewrittenRoot = rewriter.Visit(syntaxTree.Root);
+
+            var newSyntaxTree = RazorSyntaxTree.Create(rewrittenRoot, syntaxTree.Source, syntaxTree.Diagnostics, syntaxTree.Options);
+            return newSyntaxTree;
+        }
+
+        public static RazorSyntaxTree RemoveMarkupElement(RazorSyntaxTree syntaxTree)
```
ajaybhargavb: This is called after tag helper phase.

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/MarkupElementRewriterTest.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        private void RewriterTest(string input)
+        {
+            var syntaxTree = ParseDocument(input, designTime: false);
+            var rewritten = MarkupElementRewriter.Rewrite(syntaxTree);
+            BaselineTest(rewritten);
+
+            var unrewritten = MarkupElementRewriter.RemoveMarkupElement(rewritten);
+            Assert.Equal(syntaxTree.Root.SerializedValue, unrewritten.Root.SerializedValue);
```
ajaybhargavb: Each test will also run the removal phase and make sure the output matches the original syntax tree. It didn't make sense to add separate tests for the removal phase.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+                    var child = (RazorSyntaxNode)children[currentIndex];
+                    if (!(child is MarkupTagBlockSyntax tagBlock) ||
+                        tagBlock.Parent is MarkupElementSyntax)
+                    {
+                        // Not a node that we want to rewrite.
+                        currentIndex++;
+                        continue;
+                    }
+
+                    var tagName = tagBlock.GetTagName();
```
ajaybhargavb: `GetTagName()` is an extension method to `MarkupTagBlockSyntax`. This will change in the future when tag block have a first class TagName child.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Razor.Language.Syntax
+{
+    internal class MarkupElementRewriter
```
pranavkm: `static class`?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+
+                // We grab the token that could be forward slash
+                var relevantToken = childSpan.LiteralTokens[childSpan.LiteralTokens.Count == 1 ? 0 : 1];
+
+                return relevantToken.Kind == SyntaxKind.ForwardSlash;
+            }
+
+            private bool IsVoidElement(MarkupTagBlockSyntax tagBlock)
+            {
+                return VoidElements.Contains(tagBlock.GetTagName(), StringComparer.OrdinalIgnoreCase);
```
pranavkm: This is doing `Enumerable.Contains`. Did you mean to do `VoidElements.Contains(tagBlock.GetTagName())` since the hashset is already using an `OrdinalIgnoreCase` comparer?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            private bool IsSelfClosing(MarkupTagBlockSyntax tagBlock)
```
pranavkm: I'm guessing this is is a end tag like `</>` but don't see a corresponding test. Should this be unit tested?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+            {
+                var lastChild = tagBlock.ChildNodes().LastOrDefault();
+
+                return lastChild?.GetContent().EndsWith("/>", StringComparison.Ordinal) ?? false;
```
pranavkm: Out of curiosity, why does `IsEndTag` use literal tokens and this use `GetContent()`?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+
+                    // We were able to recover
+                    return true;
+                }
+
+                // Could not recover tag. Aka we found an end tag without a corresponding start tag.
+                return false;
+            }
+
+            private SyntaxNode RewriteMalformedTags(SyntaxNode parent, int malformedTagCount)
```
pranavkm: This method is invoked when `malformedTagCount != _startTagTracker.Count`, but reading the code it looks like `_startTagTracker.Count must be > malformedTagCount` (since it's popped that many times). Does the `if` need to be changed?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+                        // Don't want to track incomplete, void or self-closing tags.
+                        // Simply wrap it in a block with no body or end tag.
+                        node = RewriteNodeCore(node, startTag: tagBlock, tagChildren: Enumerable.Empty<RazorSyntaxNode>(), endTag: null);
+
+                        // Since we rewrote 'node', it's children are different. Update our collection.
+                        children = node.ChildNodes();
+                    }
+                    else if (IsEndTag(tagBlock))
+                    {
+                        if (!string.IsNullOrEmpty(CurrentStartTagName) &&
```
pranavkm: You can remove this check. Next line should suffice since you already verify `tagName` has a value

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+                    if (string.IsNullOrWhiteSpace(tagName) ||
+                        IsVoidElement(tagBlock) ||
+                        IsSelfClosing(tagBlock))
+                    {
+                        // Don't want to track incomplete, void or self-closing tags.
+                        // Simply wrap it in a block with no body or end tag.
+                        node = RewriteNodeCore(node, startTag: tagBlock, tagChildren: Enumerable.Empty<RazorSyntaxNode>(), endTag: null);
+
+                        // Since we rewrote 'node', it's children are different. Update our collection.
+                        children = node.ChildNodes();
```
pranavkm: How does this work? Should this restart from the beginning since the children are different or continue from the next index because this node is already observed?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+                "source",
+                "track",
+                "wbr"
+            };
+
+            private readonly Stack<TagBlockTracker> _startTagTracker;
+
+            public Rewriter()
+            {
+                _startTagTracker = new Stack<TagBlockTracker>();
```
pranavkm: Inline?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+            }
+
+            private SyntaxNode RewriteNodeCore(
+                SyntaxNode parent,
+                MarkupTagBlockSyntax startTag,
+                IEnumerable<RazorSyntaxNode> tagChildren,
+                MarkupTagBlockSyntax endTag)
+            {
+                var body = new SyntaxList<RazorSyntaxNode>(tagChildren);
+                var markupElement = SyntaxFactory.MarkupElement((MarkupTagBlockSyntax)startTag, body, endTag);
```
pranavkm: Can this be initialized inline (in the lambda)?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+                }
+
+                if (originalNodes.Count == 0)
+                {
+                    // Nothing to replace
+                    return parent;
+                }
+
+                // Replace nodes
+                var rewrittenElement = parent.ReplaceNodes(originalNodes, (original, rewritten) => {
```
pranavkm: Formatting. Also local functions for 💯 points

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,362 @@
...
+            private SyntaxNode RewriteNode(SyntaxNode node)
+            {
+                if (node.IsToken)
+                {
+                    return node;
+                }
+
+                var currentIndex = 0;
+                var children = node.ChildNodes();
+                while (currentIndex < children.Count)
```
pranavkm: `for (var i= 0; i < children.Count; i++)` works really well here.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: It won't restart. It just continues from where it left off. I still have to update the collection because rewriting node changes the children.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Sure. It doesn't functionally change anything but I'll change the if.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: There is a corresponding test. https://github.com/aspnet/Razor/pull/2624/files#diff-8dfd8aeca4f15b18bb45a38795fd1c02R94

pranavkm: Wouldn't either of those have one or more child nodes?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: `IsEndTag` was a copy-paste-modify. I was wondering the same thing as why is it doing all the weird index stuff. But I left it as is for now.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: They can't have any child nodes.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Done

ajaybhargavb:
Ping @NTaylorMullen @rynowak @pranavkm

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
rynowak: Names are wacky here because there are two rewritiers. Suggest something like `AddElementNodes` `RemoveElementNodes`

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,352 @@
...
+        public static RazorSyntaxTree RemoveMarkupElement(RazorSyntaxTree syntaxTree)
+        {
+            var remover = new Remover();
+            var rewrittenRoot = remover.Visit(syntaxTree.Root);
+
+            var newSyntaxTree = RazorSyntaxTree.Create(rewrittenRoot, syntaxTree.Source, syntaxTree.Diagnostics, syntaxTree.Options);
+            return newSyntaxTree;
+        }
+
+        private class Rewriter : SyntaxRewriter
```
rynowak: Is there precendent for any of this? Like is this based on something or did you write it fresh?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,352 @@
...
+            private TagBlockTracker CurrentTracker => _startTagTracker.Count > 0 ? _startTagTracker.Peek() : null;
+
+            private string CurrentStartTagName => CurrentTracker?.TagName;
+
+            public override SyntaxNode Visit(SyntaxNode node)
+            {
+                if (node != null)
+                {
+                    node = RewriteNode(node);
+                }
```
rynowak: Can a node ever be null?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
rynowak: https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentRewritePass.cs FYI

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,352 @@
...
+
+                return base.Visit(node);
```
rynowak: Does it make sense for this to be a preorder traversal instead? https://github.com/aspnet/Blazor/blob/master/src/Microsoft.AspNetCore.Blazor.Razor.Extensions/ComponentDocumentRewritePass.cs#L55

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: > Is there precendent for any of this? Like is this based on something or did you write it fresh?

This is mostly fresh. Some of the malformed tag handling logic is very similar to  https://github.com/aspnet/Razor/blob/master/src/Microsoft.AspNetCore.Razor.Language/Legacy/TagHelperParseTreeRewriter.cs#L730

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Yes. `Visit()` is called on every child and `Visit()` calls `node.Accept(this)` if the node is not null.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: I'll take a look

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
rynowak: What if instead you created a new list of children and then rewrote it once.

So like:
```
A
  B 
  C 
  D
```

Iterate through the children of `A` - `(B, C, D)` and rewrite just those nodes as needed so you end up with `(B, C!, D)` (assume C was rewritten). Then rewrite `A` once to replace its children. That's how I would expect something like this to work.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,352 @@
...
+                if (endTag != null)
+                {
+                    originalNodes.Add(endTag);
+                }
+
+                if (originalNodes.Count == 0)
+                {
+                    // Nothing to replace
+                    return parent;
+                }
```
rynowak: Does this preserve order correctly? What about a case like:
```
<div>Hi</div>
@Something!
<div>bye</div>
```

rynowak: I see, you did the below.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: The problem is that when I rewrite (B, C, D) => (B, C!, D), it actually becomes (B, C!, D!) because everything after a rewritten node is different because red nodes are created on demand

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
rynowak: I think this would make a little more sense if you mutated `originalNodes` in place and preserved ordering while doing so.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,352 @@
...
+                    // Get the nodes between the start and the end tag.
+                    var tagChildren = children.Skip(startTagIndex + 1).Take(endTagIndex - startTagIndex - 1).Cast<RazorSyntaxNode>();
+
+                    rewritten = RewriteNodeCore(parent, startTag, tagChildren, endTag);
+
+                    // We were able to recover
+                    return true;
+                }
+
+                // Could not recover tag. Aka we found an end tag without a corresponding start tag.
```
rynowak: Make sure you have a test for recovery when the cause is a malformed void element. We had a bug for this recently in Blazor.
```
<input ....></input>
```

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/MarkupElementRewriterTest.cs
rynowak: Love it.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
rynowak: Can you explain this in more detail? I'm not sure that I get it.

ajaybhargavb: Sorry I missed this part "created a **new** list of children". What I said doesn't apply in that case. I'll give this a shot

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: That works as well

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Added.

ajaybhargavb:
🆙 📅

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,329 @@
...
+
+                var body = tagChildren.Where(t => t != null).ToList();
+                var rewritten = SyntaxFactory.MarkupElement(startTag, new SyntaxList<RazorSyntaxNode>(body), endTag: endTag);
+                if (startTag != null)
+                {
+                    // If there was a start tag, that is where we want to put our new element.
+                    TrackChild(rewritten, rewrittenChildren);
+                }
+
+                tagChildren.ForEach(c => TrackChild(null, rewrittenChildren));
```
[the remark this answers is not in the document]
pranavkm: It'd also let you pass in `Array.Empty<RazorSyntaxNode>()` instead of `new List<>`

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            private SyntaxNode RewriteNode(SyntaxNode node)
+            {
+                if (node.IsToken)
+                {
+                    // Tokens don't have children.
+                    return node;
+                }
+
+                _startTagTracker.Clear();
```
NTaylorMullen: In practice is this ever not already clear?

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
```diff
@@ -0,0 +1,329 @@
...
+                }
+
+                tagChildren.ForEach(c => TrackChild(null, rewrittenChildren));
+                if (endTag != null)
+                {
+                    TrackChild(startTag == null ? rewritten : null, rewrittenChildren);
+                }
+            }
+
+            private void TrackChild(SyntaxNode child, List<SyntaxNode> rewrittenChildren)
```
NTaylorMullen: Confusing af

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: It should be. Just being extra defensive.

src/Microsoft.AspNetCore.Razor.Language/Syntax/MarkupElementRewriter.cs
ajaybhargavb: Added this to https://github.com/aspnet/Razor/issues/2619.

ajaybhargavb:
I've addressed most of the feedback here. I've added the one I didn't address to this list https://github.com/aspnet/Razor/issues/2619 to be addressed later. Merging this now.

PR #2627, opened 2018-10-02
Update Razor to use the latest completion API.

NTaylorMullen:
@Eilon @rynowak this has the goods for the completion changes.

@AmadeusW everything look good here?


AmadeusW:
Looks good!

PR #2649, opened 2018-10-16
Allow identical DocumentSnapshot sources to trigger output changes.

NTaylorMullen:
- Prior to this project changes would trigger re-parses which would then be thrown away because source versions were identical.
- Added test to verify new SetOutput behavior.

aspnet/Razor.VSCode#184

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
@@ -106,7 +106,9 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                     return;
                 }
 
-                if (_sourceVersion.HasValue && _sourceVersion == SourceVersion.GetNewerVersion(version))
+                if (_sourceVersion.HasValue &&
+                    _sourceVersion != version &&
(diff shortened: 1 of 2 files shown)


NTaylorMullen:
@SteveSandersonMS FYI, if you see invalid code issues in VSCode on startup, chances are it's because of this. We'll consume this package once it's available.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -106,7 +106,9 @@ public void SetOutput(RazorCSharpDocument csharpDocument, DefaultDocumentSnapsho
                     return;
                 }
 
-                if (_sourceVersion.HasValue && _sourceVersion == SourceVersion.GetNewerVersion(version))
+                if (_sourceVersion.HasValue &&
+                    _sourceVersion != version &&
```
ajaybhargavb: If this is a temporary workaround, could you file an issue to revert it and comment that here?

NTaylorMullen: Linked to this in the issue. Not gonna leave a comment in the code though; that'll just rot 😄

PR #2651, opened 2018-10-17
Incorporate breaking changes from CPS

rynowak:
AspNetCore PR: https://github.com/aspnet/AspNetCore/pull/3738

~~Insertion work is happening here: https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/143974 - I expect this to get merged sometime this week.~~

This has been inserted into Dev16 - so this is new needed to work on current builds of VS.

### DIFF
--- a/src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/Rules/RazorConfiguration.cs
+++ b/src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/Rules/RazorConfiguration.cs
@@ -8,205 +8,268 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace Microsoft.CodeAnalysis.Razor.ProjectSystem.Rules {
-	
-	
-	internal partial class RazorConfiguration {
-		
-		/// <summary>Backing field for deserialized rule.<see cref='Microsoft.Build.Framework.XamlTypes.Rule'/>.</summary>
-		private static Microsoft.Build.Framework.XamlTypes.Rule deserializedFallbackRule;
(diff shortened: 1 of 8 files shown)


test/Microsoft.VisualStudio.LanguageServices.Razor.Test/ProjectSystem/TestProjectSystemServices.cs
```diff
@@ -87,33 +87,33 @@ public IProjectVersionedValue<IProjectSubscriptionUpdate> CreateUpdate(params Te
                 dataSourceVersions: ImmutableDictionary<NamedIdentity, IComparable>.Empty);
         }
 
-        public class TestProjectServices : IProjectServices
+        public class TestProjectServices : ProjectServices
```
rynowak: Some types we consumed changed from interface to abstract class. It only really impacts our tests, but we have to recompile because it's a binary breaking change.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/Rules/RazorConfiguration.cs
```diff
@@ -8,205 +8,268 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-namespace Microsoft.CodeAnalysis.Razor.ProjectSystem.Rules {
```
rynowak: These are generated files that were updated by the new SDK - just drive through 😆

rynowak:
https://github.com/aspnet/AspNetCore/pull/3738

PR #2652, opened 2018-10-18
Add line numbers for design time directives

rynowak:
Addresses a blocking issue for FAR of types when used in user-code in a
directive. The FAR infrastructure is skipping over the directive code
because it's mapped to `#hidden`. As you can see in the code, the token
provided by the user is already included in the projection mappings.

I think we didn't do this before because we didn't expect this code to
need line numbers - it's not really debuggable, and design-time codegen
doesn't happen when you build the project.

I think it's OK for now that we don't line-map (or include) directives
based on view imports. If you trigger FAR on an `@inject ...` in an
import for instance, you'll find the reference for the view import file.
That seems pretty good, and the only cases I can really imagine it being
broken would be for go-to-definition (within a Razor view). Lets revisit
in the future based on feedback.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
@@ -65,15 +65,17 @@ namespace Microsoft.AspNetCore.Razor.Language.Extensions
                         }
 
                         // {node.Content} __typeHelper = default({node.Content});
-
-                        context.AddSourceMappingFor(node);
-                        context.CodeWriter
-                            .Write(node.Content)
-                            .Write(" ")
-                            .WriteStartAssignment(TypeHelper)
-                            .Write("default(")
-                            .Write(node.Content)
-                            .WriteLine(");");
+                        using (context.CodeWriter.BuildLinePragma(node.Source))
+                        {
+                            context.AddSourceMappingFor(node);
+                            context.CodeWriter
+                                .Write(node.Content)
+                                .Write(" ")
+                                .WriteStartAssignment(TypeHelper)
+                                .Write("default(")
+                                .Write(node.Content)
+                                .WriteLine(");");
+                        }
                         break;
 
                     case DirectiveTokenKind.Member:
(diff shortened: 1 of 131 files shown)


src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
```diff
@@ -108,46 +112,50 @@ private void WriteDesignTimeDirectiveToken(CodeRenderingContext context, Directi
...
+                            else
+                            {
+                                context.CodeWriter.Write("\"");
+                                context.AddSourceMappingFor(node);
+                                context.CodeWriter
+                                    .Write(node.Content)
+                                    .Write("\"");
+                            }
+
+                            context.CodeWriter.WriteLine(";");
```
rynowak: This is the extent of the code change - the rest is just baseline updates.

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives_DesignTime.mappings.txt
```diff
@@ -20,21 +20,21 @@ Generated Location: (830:27,0 [0] )
 
 Source Location: (159:11,8 [17] TestFiles/IntegrationTests/CodeGenerationIntegrationTest/IncompleteDirectives.cshtml)
 |MyService<TModel>|
-Generated Location: (883:30,0 [17] )
+Generated Location: (980:31,0 [17] )
```
rynowak: generated locations changing because the generated code includes additional pragmas.

src/Microsoft.AspNetCore.Razor.Language/Extensions/DesignTimeDirectiveTargetExtension.cs
```diff
@@ -108,46 +112,50 @@ private void WriteDesignTimeDirectiveToken(CodeRenderingContext context, Directi
                         }
 
                         // global::System.Object __typeHelper = nameof({node.Content});
-
-                        context.CodeWriter
+                        using (context.CodeWriter.BuildLinePragma(node.Source))
```
rynowak: I know we don't have a strong reason to do this mapping for namespaces and strings, but it seemed better to make it consistent then do add lines for half of the feature. LMK if this needs more discussion.

heejaechang:
I didn't know that FAR filter out references in the hidden code section. is it FAR or is it compiler not binding node in hidden section?

rynowak:
Yeah, I suppose it's also reasonable for us to change this - I think we wouldn't have known about this without FAR. There's no real push for you to change this on your side, just mentioning it 😁

rynowak:
I verified this in VS when I was testing FAR 😆

PR #2653, opened 2018-10-18
Fix a bug in the span mapping code

rynowak:
This wasn't quite doing the right thing, and it's simpler than we make
it to be.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
@@ -153,6 +153,10 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 {
                     results.Add(new SpanMapResult(document, linePositionSpan));
                 }
+                else
+                {
+                    results.Add(null);
+                }
             }
 
             return Task.FromResult(results.ToImmutable());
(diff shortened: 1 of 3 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -153,6 +153,10 @@ public void SetOutput(RazorCSharpDocument csharpDocument, DefaultDocumentSnapsho
                 {
                     results.Add(new SpanMapResult(document, linePositionSpan));
                 }
+                else
+                {
+                    results.Add(null);
+                }
```
rynowak: This matches up with some of my other changes in the feature branch. We should return a result for each span they give us.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -161,31 +165,30 @@ public void SetOutput(RazorCSharpDocument csharpDocument, DefaultDocumentSnapsho
...
-                var mapping = output.SourceMappings[i];
-                if (span.Length > mapping.GeneratedSpan.Length)
-                {
-                    // If the length of the generated span is smaller they can't match. A C# expression
-                    // won't cover multiple generated spans.
-                    //
-                    // This heuristic is useful in the Razor context to filter out zero-length
-                    // spans.
-                    continue;
-                }
```
rynowak: This logic is wrong. It's only comparing the sizes of the spans, and isn't looking at the positions. If only there was a method for that....

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/GeneratedCodeContainer.cs
```diff
@@ -161,31 +165,30 @@ public void SetOutput(RazorCSharpDocument csharpDocument, DefaultDocumentSnapsho
...
-
+                var mapping = mappings[i];
                 var original = mapping.OriginalSpan.AsTextSpan();
                 var generated = mapping.GeneratedSpan.AsTextSpan();
 
-                var leftOffset = span.Start - generated.Start;
-                var rightOffset = span.End - generated.End;
-                if (leftOffset >= 0 && rightOffset <= 0)
+                if (!generated.Contains(span))
```
rynowak: Oh, there is!

PR #2654, opened 2018-10-18
Update deps and fix VSIX signing

rynowak:
So, it turns out the VSSDK added a feature that lets us suppress
assemblies brought in by package references. YAY!

Fixes: https://github.com/aspnet/AspNetCore-Internal/issues/1194


rynowak:
I'm reverting the update deps part of this PR. I want the updated package versions, which this undoes.

rynowak:
```
C:\Users\rynowak\.dotnet\buildtools\korebuild\3.0.0-alpha1-20181017.7\modules\KoreBuild.Tasks\CodeSign.targets(46,5): er
ror : CertificateName metadata of FileSignInfo is invalid: '' [C:\Users\rynowak\.dotnet\buildtools\korebuild\3.0.0-alpha
1-20181017.7\KoreBuild.proj]
C:\Users\rynowak\.dotnet\buildtools\korebuild\3.0.0-alpha1-20181017.7\modules\KoreBuild.Tasks\CodeSign.targets(46,5): er
ror : CertificateName metadata of FileSignInfo is invalid: '' [C:\Users\rynowak\.dotnet\buildtools\korebuild\3.0.0-alpha
1-20181017.7\KoreBuild.proj]
```

What's the correct action to take based on this?

natemcmaster:
This means something is configured to code sign but didn't specify a cert name. Can you gather a binlog and look at the items going into SignToolTask? 

Just guessing, but you may also need to update this: https://github.com/aspnet/Razor/blob/4927525519e9dbf4dc5f2d157e19a7daa1fd3b1e/build/dependencies.props#L7 to 3.0.0-alpha1-20181017.7

rynowak:
Take a look, I just tried that locally and added it.

PR #2656, opened 2018-10-19
Add and process notifications for Imports

rynowak:
Add and process notifications for Imports

This builds support for tracking the effect of changes to imports on
other documents, and completes our model for being able to keep
generated code up to date.

### DIFF
--- a/benchmarks/Microsoft.AspNetCore.Razor.Performance/ProjectSystem/ProjectSnapshotManagerBenchmarkBase.cs
+++ b/benchmarks/Microsoft.AspNetCore.Razor.Performance/ProjectSystem/ProjectSnapshotManagerBenchmarkBase.cs
@@ -132,14 +132,6 @@ namespace Microsoft.AspNetCore.Razor.Performance
 
         private class StaticProjectSnapshotProjectEngineFactory : ProjectSnapshotProjectEngineFactory
         {
-            public override RazorProjectEngine Create(ProjectSnapshot project, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
-            {
-                return RazorProjectEngine.Create(project.Configuration, fileSystem, b =>
-                {
-                    RazorExtensions.Register(b);
-                });
-            }
-
             public override IProjectEngineFactory FindFactory(ProjectSnapshot project)
             {
                 throw new NotImplementedException();
@@ -148,6 +140,14 @@ namespace Microsoft.AspNetCore.Razor.Performance
             public override IProjectEngineFactory FindSerializableFactory(ProjectSnapshot project)
             {
                 throw new NotImplementedException();
+            }
+
+            public override RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+            {
+                return RazorProjectEngine.Create(configuration, fileSystem, b =>
+                {
+                    RazorExtensions.Register(b);
+                });
             }
         }
     }
--- a/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectFileSystem.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectFileSystem.cs
@@ -49,6 +49,10 @@ namespace Microsoft.AspNetCore.Razor.Language
             var absolutePath = NormalizeAndEnsureValidPath(path);
 
             var file = new FileInfo(absolutePath);
+            if (!absolutePath.StartsWith(absoluteBasePath))
+            {
+                throw new InvalidOperationException($"The file '{file.FullName}' is not a descendent of the base path '{absoluteBasePath}'.");
+            }
 
             var relativePhysicalPath = file.FullName.Substring(absoluteBasePath.Length + 1); // Include leading separator
             var filePath = "/" + relativePhysicalPath.Replace(Path.DirectorySeparatorChar, '/');
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
@@ -60,6 +60,21 @@ namespace Microsoft.CodeAnalysis.Razor
             return Create(project.Configuration, fileSystem, configure);
         }
 
+        public RazorProjectEngine Create(RazorConfiguration configuration, string directoryPath, Action<RazorProjectEngineBuilder> configure)
+        {
+            if (configuration == null)
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            if (directoryPath == null)
+            {
+                throw new ArgumentNullException(nameof(directoryPath));
+            }
+
+            return Create(configuration, RazorProjectFileSystem.Create(directoryPath), configure);
+        }
+
         public abstract RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure);
     }
 }
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
@@ -45,7 +45,21 @@ namespace Microsoft.CodeAnalysis.Razor
             return Create(project, RazorProjectFileSystem.Create(Path.GetDirectoryName(project.FilePath)), configure);
         }
 
-        public abstract RazorProjectEngine Create(ProjectSnapshot project, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure);
+        public RazorProjectEngine Create(ProjectSnapshot project, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
+        {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
 
+            if (fileSystem == null)
+            {
+                throw new ArgumentNullException(nameof(fileSystem));
+            }
+
+            return Create(project.Configuration, fileSystem, configure);
+        }
+
+        public abstract RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure);
     }
 }
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Razor.Language;
 
@@ -58,9 +59,37 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
             }
         }
 
+        public override bool IsImportDocument(DocumentSnapshot document)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+            
+            return State.ImportsToIncludingDocuments.ContainsKey(document.TargetPath);
+        }
+
+        public override IEnumerable<DocumentSnapshot> GetRelatedDocuments(DocumentSnapshot document)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            if (State.ImportsToIncludingDocuments.TryGetValue(document.TargetPath, out var relatedDocuments))
+            {
+                lock (_lock)
+                {
+                    return relatedDocuments.Select(GetDocument).ToArray();
+                }
+            }
+
+            return Array.Empty<DocumentSnapshot>();
+        }
+
         public override RazorProjectEngine GetProjectEngine()
         {
-            return State.ProjectEngine.GetProjectEngine(this);
+            return State.ProjectEngine.GetProjectEngine(this.State);
         }
 
         public override Task<IReadOnlyList<TagHelperDescriptor>> GetTagHelpersAsync()
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
@@ -66,7 +66,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 throw new ArgumentNullException(nameof(document));
             }
             
-            return State.ImportsToIncludingDocuments.ContainsKey(document.TargetPath);
+            return State.ImportsToRelatedDocuments.ContainsKey(document.TargetPath);
         }
 
         public override IEnumerable<DocumentSnapshot> GetRelatedDocuments(DocumentSnapshot document)
@@ -76,7 +76,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 throw new ArgumentNullException(nameof(document));
             }
 
-            if (State.ImportsToIncludingDocuments.TryGetValue(document.TargetPath, out var relatedDocuments))
+            if (State.ImportsToRelatedDocuments.TryGetValue(document.TargetPath, out var relatedDocuments))
             {
                 lock (_lock)
                 {
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
@@ -90,12 +90,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 var projects = new ProjectSnapshot[_projects.Count];
                 foreach (var entry in _projects)
                 {
-                    if (entry.Value.Snapshot == null)
-                    {
-                        entry.Value.Snapshot = new DefaultProjectSnapshot(entry.Value.State);
-                    }
-
-                    projects[i++] = entry.Value.Snapshot;
+                    projects[i++] = entry.Value.GetSnapshot();
                 }
 
                 return projects;
@@ -115,12 +110,7 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 
             if (_projects.TryGetValue(filePath, out var entry))
             {
-                if (entry.Snapshot == null)
-                {
-                    entry.Snapshot = new DefaultProjectSnapshot(entry.State);
-                }
-
-                return entry.Snapshot;
+                return entry.GetSnapshot();
             }
 
             return null;
@@ -175,8 +165,10 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 // Document updates can no-op.
                 if (!object.ReferenceEquals(state, entry.State))
                 {
-                    _projects[hostProject.FilePath] = new Entry(state);
-                    NotifyListeners(new ProjectChangeEventArgs(hostProject.FilePath, document.FilePath, ProjectChangeKind.DocumentAdded));
+                    var oldSnapshot = entry.GetSnapshot();
+                    entry = new Entry(state);
+                    _projects[hostProject.FilePath] = entry;
+                    NotifyListeners(new ProjectChangeEventArgs(oldSnapshot, entry.GetSnapshot(), document.FilePath, ProjectChangeKind.DocumentAdded));
                 }
             }
         }
@@ -201,8 +193,10 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 // Document updates can no-op.
                 if (!object.ReferenceEquals(state, entry.State))
                 {
-                    _projects[hostProject.FilePath] = new Entry(state);
-                    NotifyListeners(new ProjectChangeEventArgs(hostProject.FilePath, document.FilePath, ProjectChangeKind.DocumentRemoved));
+                    var oldSnapshot = entry.GetSnapshot();
+                    entry = new Entry(state);
+                    _projects[hostProject.FilePath] = entry;
+                    NotifyListeners(new ProjectChangeEventArgs(oldSnapshot, entry.GetSnapshot(), document.FilePath, ProjectChangeKind.DocumentRemoved));
                 }
             }
         }
@@ -229,12 +223,10 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
                 entry.State.Documents.TryGetValue(documentFilePath, out var older))
             {
                 ProjectState state;
-                SourceText olderText;
-                VersionStamp olderVersion;
 
                 var currentText = sourceText;
-                if (older.TryGetText(out olderText) && 
-                    older.TryGetTextVersion(out olderVersion))
+                if (older.TryGetText(out var olderText) &&
+                    older.TryGetTextVersion(out var olderVersion))
                 {
                     var version = currentText.ContentEquals(olderText) ? olderVersion : olderVersion.GetNewerVersion();
                     state = entry.State.WithChangedHostDocument(older.HostDocument, currentText, version);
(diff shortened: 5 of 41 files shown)


benchmarks/Microsoft.AspNetCore.Razor.Performance/ProjectSystem/ProjectSnapshotManagerBenchmarkBase.cs
```diff
@@ -149,6 +141,14 @@ public override IProjectEngineFactory FindSerializableFactory(ProjectSnapshot pr
             {
                 throw new NotImplementedException();
             }
+
+            public override RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure)
```
rynowak: This had to change because we need to be able to create the project engine from inside `ProjectState`.

src/Microsoft.AspNetCore.Razor.Language/DefaultRazorProjectFileSystem.cs
```diff
@@ -49,6 +49,10 @@ public override RazorProjectItem GetItem(string path)
             var absolutePath = NormalizeAndEnsureValidPath(path);
 
             var file = new FileInfo(absolutePath);
+            if (!absolutePath.StartsWith(absoluteBasePath))
+            {
+                throw new InvalidOperationException($"The file '{file.FullName}' is not a descendent of the base path '{absoluteBasePath}'.");
+            }
```
rynowak: These cases would explode at line 57 currently. Hit this while testing something new. All tests passed when I added this validation.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshot.cs
```diff
@@ -58,9 +59,37 @@ public override DocumentSnapshot GetDocument(string filePath)
...
         }
 
+        public override bool IsImportDocument(DocumentSnapshot document)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+            
+            return State.ImportsToIncludingDocuments.ContainsKey(document.TargetPath);
```
rynowak: State now tracks the 'target path' of each import document and which 'related files' it affects. These this is the API on PS for interacting with it.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DefaultProjectSnapshotManager.cs
```diff
@@ -90,12 +90,7 @@ public override IReadOnlyList<ProjectSnapshot> Projects
                 var projects = new ProjectSnapshot[_projects.Count];
                 foreach (var entry in _projects)
                 {
-                    if (entry.Value.Snapshot == null)
-                    {
-                        entry.Value.Snapshot = new DefaultProjectSnapshot(entry.Value.State);
-                    }
-
-                    projects[i++] = entry.Value.Snapshot;
+                    projects[i++] = entry.Value.GetSnapshot();
```
rynowak: The changes here are because I added old and new `ProjectSnapshot` to the changed event. I need this because when a document is removed, I need to be able to see if it *was* an import, and if it *used to* affect some documents' codegen. This is an answer that only the old `ProjectSnapshot` has.

Passing the snapshots in the event simplified some other code as well.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -12,10 +12,10 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
     // Internal tracker for DefaultProjectSnapshot
     internal class ProjectState
     {
-        private static readonly IReadOnlyDictionary<string, DocumentState> EmptyDocuments = new Dictionary<string, DocumentState>();
-
+        private static readonly ImmutableDictionary<string, DocumentState> EmptyDocuments = ImmutableDictionary.Create<string, DocumentState>(FilePathComparer.Instance);
```
rynowak: Immutable collections... we knew we'd have to swallow this pill at some point. As ProjectState grows in complexity I think we'll do more of it.

src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
```diff
@@ -143,6 +143,33 @@ public void Enqueue(ProjectSnapshot project, DocumentSnapshot document)
             }
         }
 
+        public void EnqueueRelatedDocuments(ProjectSnapshot project, DocumentSnapshot document)
```
rynowak: This is unused, will remove

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -84,7 +92,10 @@ public static ProjectState Create(HostWorkspaceServices services, HostProject ho
         }
 
         // Internal set for testing.
-        public IReadOnlyDictionary<string, DocumentState> Documents { get; internal set; }
+        public ImmutableDictionary<string, DocumentState> Documents { get; internal set; }
+
+        // Internal set for testing.
+        public ImmutableDictionary<string, ImmutableArray<string>> ImportsToIncludingDocuments { get; internal set; }
```
rynowak: Will rename this to `ImportsToRelatedDocuments` - that's a name I thought of while writing the part after this.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -198,23 +248,22 @@ public ProjectState WithChangedHostDocument(HostDocument hostDocument, SourceTex
...
+            var documents = Documents.SetItem(hostDocument.FilePath, document.WithText(sourceText, version));
 
-            if (documents.TryGetValue(hostDocument.FilePath, out var document))
+            if (ImportsToIncludingDocuments.TryGetValue(hostDocument.TargetPath, out var relatedDocuments))
             {
-                documents[hostDocument.FilePath] = document.WithText(sourceText, version);
+                foreach (var relatedDocument in relatedDocuments)
+                {
+                    documents = documents.SetItem(relatedDocument, documents[relatedDocument].WithImportsChange());
+                }
```
rynowak: This is the part of this PR where it all comes together. Now that we know which documents are imports, and which documents are related to which imports, we can reset their state when an import changes.

src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
```diff
@@ -257,23 +284,35 @@ private void ProjectManager_Changed(object sender, ProjectChangeEventArgs e)
...
-                        var project = _projectManager.GetLoadedProject(e.ProjectFilePath);
-                        Enqueue(project, project.GetDocument(e.DocumentFilePath));
+                        var project = e.Newer;
+                        var document = project.GetDocument(e.DocumentFilePath);
+
+                        Enqueue(project, document);
+                        foreach (var relatedDocument in project.GetRelatedDocuments(document))
+                        {
+                            Enqueue(project, document);
+                        }
```
rynowak: This is the next important part. When a document changes, we can just enqueue a change to all related documents - which will have already had their state reset.

test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultDocumentSnapshotTest.cs
```diff
@@ -9,14 +10,10 @@
 
 namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
 {
-    public class DefaultDocumentSnapshotTest
+    public class DefaultDocumentSnapshotTest : WorkspaceTestBase
```
rynowak: I did a bunch of test cleanup. Now that we track imports, any test that uses PSM needs to have the ability to create a project engine. I decided to make this easy to do now.

test/Microsoft.CodeAnalysis.Razor.Workspaces.Test/ProjectSystem/DefaultProjectSnapshotTest.cs
```diff
@@ -64,11 +52,24 @@ public DefaultProjectSnapshotTest()
...
 
-        private HostServices HostServices { get; }
+        private List<TagHelperDescriptor> SomeTagHelpers { get; }
 
-        private Workspace Workspace { get; }
+        private void Configure(RazorProjectEngineBuilder builder)
+        {
+            builder.Features.Remove(builder.Features.OfType<IImportProjectFeature>().Single());
+            builder.Features.Add(new TestImportProjectFeature());
+        }
```
rynowak: Dead code, will remove

test/Microsoft.VisualStudio.Editor.Razor.Test/DefaultVisualStudioDocumentTrackerTest.cs
```diff
@@ -56,7 +56,8 @@ public DefaultVisualStudioDocumentTrackerTest()
             ProjectManager = new TestProjectSnapshotManager(Dispatcher, Workspace) { AllowNotifyListeners = true };
 
             HostProject = new HostProject(ProjectPath, FallbackRazorConfiguration.MVC_2_1);
-            OtherHostProject = new HostProject(ProjectPath, FallbackRazorConfiguration.MVC_2_0);
+            UpdatedHostProject = new HostProject(ProjectPath, FallbackRazorConfiguration.MVC_2_0);
+            OtherHostProject = new HostProject("C:/Some/Other/Path/TestProject.csproj", FallbackRazorConfiguration.MVC_2_0);
```
rynowak: Some test changes here because we changed how the PSM event works.

test/Microsoft.VisualStudio.Editor.Razor.Test/Documents/EditorDocumentManagerListenerTest.cs
```diff
@@ -58,8 +58,10 @@ public void ProjectManager_Changed_DocumentAdded_InvokesGetOrCreateDocument()
 
             var listener = new EditorDocumentManagerListener(editorDocumentManger.Object, changedOnDisk, changedInEditor, opened, closed);
 
+            var project = Mock.Of<ProjectSnapshot>(p => p.FilePath == "/Path/to/project.csproj");
```
rynowak: Some test changes here because we changed how the PSM event works.

test/Microsoft.VisualStudio.LanguageServices.Razor.Test/ProjectSystem/DefaultProjectSnapshotManagerTest.cs
```diff
@@ -129,7 +119,7 @@ public void DocumentAdded_AddsDocument()
 
             // Assert
             var snapshot = ProjectManager.GetSnapshot(HostProject);
-            Assert.Collection(snapshot.DocumentFilePaths, d => Assert.Equal(Documents[0].FilePath, d));
+            Assert.Collection(snapshot.DocumentFilePaths.OrderBy(f => f), d => Assert.Equal(Documents[0].FilePath, d));
```
rynowak: These tests weren't robust enough. This broke when I started using immutable collections.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
```diff
@@ -45,7 +45,40 @@ public RazorProjectEngine Create(ProjectSnapshot project, Action<RazorProjectEng
...
+            {
+                throw new ArgumentNullException(nameof(fileSystem));
+            }
+
+            return Create(project.Configuration, fileSystem, configure);
+        }
+
+        public abstract RazorProjectEngine Create(RazorConfiguration configuration, RazorProjectFileSystem fileSystem, Action<RazorProjectEngineBuilder> configure);
+
+        private static string GetDirectoryPath(ProjectSnapshot project)
```
NTaylorMullen: Not needed

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -152,17 +163,36 @@ public ProjectState WithAddedHostDocument(HostDocument hostDocument, Func<Task<T
...
+            foreach (var importTargetPath in importTargetPaths)
             {
-                documents.Add(kvp.Key, kvp.Value);
+                if (!importsToRelatedDocuments.TryGetValue(importTargetPath, out relatedDocuments))
+                {
+                    relatedDocuments = ImmutableArray.Create<string>();
+                }
+
+                relatedDocuments = relatedDocuments.Add(hostDocument.FilePath);
+                importsToRelatedDocuments = importsToRelatedDocuments.SetItem(importTargetPath, relatedDocuments);
```
NTaylorMullen: Just to be clear, with how this is currently it doesn't account for any documents that were removed from the underlying ProjectFileSystem that haven't yet been processed.  Aka, if Document A existed in `importsToRelatedDocuments` before calling `WithAddedHostDocument` and then various changes occurred simultaneously (like a user moving an entire folder) then this would have a load of documents in it that don't exist on disk.

Maybe that's ok, maybe it's not. Bringing it up.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
```diff
@@ -256,15 +308,9 @@ public ProjectState WithHostProject(HostProject hostProject)
...
-            var documents = new Dictionary<string, DocumentState>(FilePathComparer.Instance);
-            foreach (var kvp in Documents)
-            {
-                documents.Add(kvp.Key, kvp.Value.WithConfigurationChange());
-            }
-
-            var state = new ProjectState(this, difference, hostProject, WorkspaceProject, documents);
+            
+            var documents = Documents.ToImmutableDictionary(kvp => kvp.Key, kvp => kvp.Value.WithConfigurationChange(), FilePathComparer.Instance);
+            var state = new ProjectState(this, ProjectDifference.ConfigurationChanged, hostProject, WorkspaceProject, documents, ImportsToRelatedDocuments);
```
NTaylorMullen: If the HostProject changed it means the configuration might have changed. That'd potentially impact the type of imports that exist in the project state since a new project engine is created. We need to re-calc `ImportsToRelatedDocuments` here as well.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/LiveShareProjectSnapshotBase.cs
```diff
@@ -27,6 +27,10 @@ internal class LiveShareProjectSnapshotBase : ProjectSnapshot
 
         public override DocumentSnapshot GetDocument(string filePath) => throw new NotImplementedException();
 
+        public override bool IsImportDocument(DocumentSnapshot document) => throw new NotImplementedException();
```
NTaylorMullen: This is breaking to LiveShare, we need to mark these as virtual or work with the LiveShare team to coordinate a release.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectChangeEventArgs.cs
```diff
@@ -7,29 +7,35 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
...
     internal class ProjectChangeEventArgs : EventArgs
     {
-        public ProjectChangeEventArgs(string projectFilePath, ProjectChangeKind kind)
+        public ProjectChangeEventArgs(ProjectSnapshot older, ProjectSnapshot newer, ProjectChangeKind kind)
         {
-            if (projectFilePath == null)
+            if (older == null && newer == null)
             {
-                throw new ArgumentNullException(nameof(projectFilePath));
+                throw new ArgumentException("Both projects cannot be null.");
```
NTaylorMullen: Resource eventually, #importantfeedback

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectEngineTracker.cs
```diff
@@ -55,12 +59,33 @@ public RazorProjectEngine GetProjectEngine(ProjectSnapshot snapshot)
...
+            var projectEngine = GetProjectEngine(state);
+            var importFeature = projectEngine.ProjectFeatures.OfType<IImportProjectFeature>().FirstOrDefault();
+            var projectItem = projectEngine.FileSystem.GetItem(targetPath);
+            var importItems = importFeature?.GetImports(projectItem).Where(i => i.FilePath != null);
+
+            // Target path looks like `Foo\\Bar.cshtml`
+            var targetPaths = new List<string>();
+            foreach (var importItem in importItems)
+            {
+                var itemTargetPath = importItem.FilePath.Replace('/', '\\').TrimStart('\\');
```
NTaylorMullen: Why normalize to `\\` instead of `Path.DirectorySeparator`?

src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
NTaylorMullen: Ya it's awesome to see it all come together. 🥂

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSnapshotProjectEngineFactory.cs
rynowak: Actually it is. I need to call it from this method. That's why it's failing on *nix.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
rynowak: I don't understand this. if the files are deleted, their content would change and that would trigger a notification.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
rynowak: Yeah that's a good point 👍

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/LiveShareProjectSnapshotBase.cs
rynowak: When this that code move to this repo? Also isn't live share currently broken?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
NTaylorMullen: `ProjectFileSystem` interacts with the file system directly so while the system is waiting for events to propagate there may be a few intermediary project states which aren't entirely correct
.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/LiveShareProjectSnapshotBase.cs
NTaylorMullen: > When this that code move to this repo? 
When they tell us we can open source it T_T

> Also isn't live share currently broken?

Nah, it was fixed

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectChangeEventArgs.cs
rynowak: I don't think this will ever matter. I've mostly stopped using resources. The other room doesn't use them (and mostly never did).

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectEngineTracker.cs
rynowak: because it seemed straightforward? By convention paths in msbuild are specified windows styel

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectState.cs
rynowak: We can't and shouldn't control the ordering of events - we can only work to make sure that the result is eventually coherent which seems like it will be the case.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/LiveShareProjectSnapshotBase.cs
rynowak: `LiveShareProjectSnapshotBase` is in this repo - I'm going to mark it virtual there.

We should try to eventually get rid of all of the subclasses of PSM - it's not meant to be an abstraction.

rynowak: I should clarify, the that if there are features of PSM that we need to have different implementations, then **those** should have services that differ across hosts, rather than subclassing.

rynowak: `LiveShareProjectSnapshotBase` already overrides these - does that resolve your concern?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectEngineTracker.cs
NTaylorMullen: I'd verify this returns what you'd expect cross-plat. The Razor's file system abstraction interacts directly with the file system.

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectChangeEventArgs.cs
```diff
@@ -7,29 +7,35 @@ namespace Microsoft.CodeAnalysis.Razor.ProjectSystem
...
-            ProjectFilePath = projectFilePath;
             DocumentFilePath = documentFilePath;
             Kind = kind;
+
+            ProjectFilePath = older?.FilePath ?? newer.FilePath;
         }
 
+        public ProjectSnapshot Older { get; }
+
+        public ProjectSnapshot Newer { get; }
```
ajaybhargavb: These are unassigned

src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
```diff
@@ -257,23 +257,35 @@ private void ProjectManager_Changed(object sender, ProjectChangeEventArgs e)
                     }
 
                 case ProjectChangeKind.DocumentAdded:
+                case ProjectChangeKind.DocumentChanged:
                     {
-                        var project = _projectManager.GetLoadedProject(e.ProjectFilePath);
-                        Enqueue(project, project.GetDocument(e.DocumentFilePath));
+                        var project = e.Newer;
```
ajaybhargavb: `Older` and `Newer` were both never assigned. How are these not nullref'ing below? Or is there no test covering this code path? Or is Github lying to me?

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/ProjectEngineTracker.cs
rynowak: These aren't used by the file system. These are keys. Target path determines what imports apply to what document, it doesn't determine where those documents are located.

src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
rynowak: Yeah wierd.

rynowak:
Any other feedback on this guy?

rynowak:
Yeah thanks - I've made a few tries of that now...

rynowak:
This is unfortunately going to take some time for me to sort out. `RazorProjectFileSystem` rears its ugly head again. Basically everywhere that we use windows style paths is broken now that tests/infrastructure actually looks at paths. I'm going to do have to write a utility for creating platform-independent paths so we can use that everywhere in tests.

PR #2658, opened 2018-10-19
Port completion changes to master branch

NTaylorMullen:
All these changes have already gone into release branches. Just bringing them back to master (seems that I hadn't before).


rynowak:
The code changes here seem fine - keep in mind you will have to update universe.

Is the idea to get this into master and let rel/vs16.0 track our VS work?

NTaylorMullen:
> - keep in mind you will have to update universe.

ya this is an interesting thing because Universe is archived. @natemcmaster what should i do?

> Is the idea to get this into master and let rel/vs16.0 track our VS work?

yes, I'm just trying to ensure that any work that's done in VS branches doesn't get forgotten.

natemcmaster:
aspnet/AspNetCore is the replacement for aspnet/Universe. Do the same thing there you would have done in aspnet/Universe.

rynowak:
Ok this seems good then!

NTaylorMullen:
> aspnet/AspNetCore is the replacement for aspnet/Universe. Do the same thing there you would have done in aspnet/Universe.

Done: https://github.com/aspnet/AspNetCore/pull/3683

PR #2659, opened 2018-10-19
Add completion bits to cps-update.

NTaylorMullen:
@rynowak trying to get a coherent set of packages together for a 16.0 inerstion

Working on verifying this locally.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -24,10 +24,10 @@
     <MicrosoftNETCoreApp22PackageVersion>2.2.0-preview2-26905-02</MicrosoftNETCoreApp22PackageVersion>
     <MicrosoftNETTestSdkPackageVersion>15.6.1</MicrosoftNETTestSdkPackageVersion>
     <MicrosoftVisualStudioComponentModelHostPackageVersion>15.8.525</MicrosoftVisualStudioComponentModelHostPackageVersion>
-    <MicrosoftVisualStudioEditorPackageVersion> 15.8.525</MicrosoftVisualStudioEditorPackageVersion>
     <MicrosoftVisualStudioImageCatalogPackageVersion>15.8.28010</MicrosoftVisualStudioImageCatalogPackageVersion>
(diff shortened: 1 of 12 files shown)


rynowak:
Well this hasn't gone in to d16stg yet. so I guess I'm not sure why you're doing this?

NTaylorMullen:
Getting a build available so completion folks have a coherent build that works with cps bits + completion bits for a 16.0 release before I go on vaca.

rynowak:
alright cool. I think based on our other conversation this is fine to go into master - not rel/vs16.0 right? We can't use these branches to track different things.

NTaylorMullen:
> alright cool. I think based on our other conversation this is fine to go into master - not rel/vs16.0 right? We can't use these branches to track different things.

master and rel/vs16.0 should be identical (as of this cps update) once I get a build out for insertion. The process that we've been doing is:

1. Commit to rel/vs16.0
2. Use micro build to sign the Razor build output
3. Publish symbols
4. Do VS insertion PR
5. Merge VS insertion PR

This way the exact bits that make it into VS exist in the rel/vs16.0 build drop folder (instead of having to re-build them once we get a successful build from a different branch)

rynowak:
I don't care what we've done in the past, what you're proposing to do now doesn't make sense to me at all. You didn't address what I'm asking. 

I don't have a problem with the workflow that you're describing when we're going to do a build, and then immediately do an insertion an hour later. We're talking about changes that will make it into VS at some unknown point in the future - it's out of our control when those changes get inserted, and so....

> What branch should we be using to do active development against a current build?

NTaylorMullen:
> talking about changes that will make it into VS at some unknown point in the future

Sure, I can understand where you're coming from.

> What branch should we be using to do active development against a current build?

There isn't one currently. Here's what i'll do:

- Update `feature/cps-update` branch to have the completion changes and produce a build
- Create a new `feature/completion` branch that does **not** have the cps changes but is equivalently the merged form of https://github.com/aspnet/Razor/pull/2658 and produce a build.

Next, i'll update the cps vs insertion PR to have the updated Razor insertion bits for the `feature/cps-update` branch (that includes completion). Lastly i'll create a second PR that has just the completion bits based off of the `feature/completion` branch.

How does all of that sound?

Issue #2650, opened 2018-10-16
Have a better way to determine generated outputs version


NTaylorMullen:
Today we can only use a document snapshots version to differentiate it from other documents. @rynowak had some ideas on how to build a versioning scheme that properly encompasses all the inputs the correspond to a given output. 

@rynowak mentioned he'll be working on this so assigning to him.

NTaylorMullen:
Will need to re-evaluate [this](https://github.com/aspnet/Razor/pull/2649#discussion_r225731822) once we have a better versioning scheme

PR #2662, opened 2018-10-21
Implements versions for generated code

rynowak:
This change implements version tracking the inputs and outputs of
generated code.

Version tracking is still best-effort - meaning that in some cases a
perfect system could avoid doing more work. However, since we base the
versions off of all of the inputs, we now that the guarantee that code
generation operations that happen 'out of order' will always result in
the newer inputs generating the newer outputs.

Fixes: https://github.com/aspnet/Razor/issues/2650


src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
```diff
@@ -67,42 +69,35 @@ internal class DocumentState
 
         public GeneratedCodeContainer GeneratedCodeContainer => HostDocument.GeneratedCodeContainer;
 
-        public DocumentGeneratedOutputTracker GeneratedOutput
+        public bool IsGeneratedOutputResultAvailable => ComputedState.TaskUnsafe?.IsCompleted == true;
```
ajaybhargavb: Just return `ComputedState.IsResultAvailable`

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
```diff
@@ -227,7 +229,239 @@ public virtual DocumentState WithTextLoader(Func<Task<TextAndVersion>> loader)
...
+                // Examples: 
+                // 
+                // A change to a tag helper not used by this document - updates the INPUT version, but not 
+                // the OUTPUT version.
+                //
+                // A change in the HTML - updates the INPUT version, but not the OUTPUT version.
+                //
+                //
+                // Razor IDE features should always retrieve the output and party on it regardless. Depending
+                // on the use cases we or may not need to synchronize the output.
```
ajaybhargavb: we may or may not

src/Microsoft.CodeAnalysis.Razor.Workspaces/ProjectSystem/DocumentState.cs
```diff
@@ -67,42 +69,35 @@ internal class DocumentState
...
                     }
                 }
 
-                return _generatedOutput;
+                return _computedState;
             }
         }
 
-        public DocumentImportsTracker Imports
+        public Task<(RazorCodeDocument output, VersionStamp inputVersion, VersionStamp outputVersion)> GetGeneratedOutputAndVersionAsync(DefaultProjectSnapshot project, DefaultDocumentSnapshot document)
```
ajaybhargavb: Can the second parameter just be `DocumentSnapshot` instead of `DefaultDocumentSnapshot`?

rynowak: I'm sure it could be, it should not be. This code is part of **the implementation**, it should not be reusable.

PR #2681, opened 2018-10-28
Miscellaneous infrastructure and fixes for FAR

rynowak:
Working through the FAR changes and biting off small pieces that can be merged independently 👍 

See commits for details.

### DIFF
--- a/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentExcerptService.cs
+++ b/src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentExcerptService.cs
@@ -176,6 +176,10 @@ namespace Microsoft.CodeAnalysis.Razor
                     secondaryDocument, 
                     secondarySpan, 
                     cancellationToken);
+
+                // NOTE: The Classifier will only returns spans for things that it understands. That means
+                // that whitespace is not classified. The preview expects us to provide contiguous spans, 
+                // so we are going to have to fill in the gaps.
                 
                 // Now we have to translate back to the primary document's coordinates.
                 var offset = primarySpan.Start - secondarySpan.Start;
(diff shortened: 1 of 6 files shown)


src/Microsoft.CodeAnalysis.Razor.Workspaces/RazorDocumentExcerptService.cs
```diff
@@ -89,24 +100,26 @@ public RazorDocumentExcerptService(DocumentSnapshot document, ISpanMappingServic
...
-
-        private TextSpan ChooseExcerptSpan(SourceText primaryText, TextSpan primarySpan, ExcerptMode mode)
+        
+        private TextSpan ChooseExcerptSpan(SourceText primaryText, TextSpan primarySpan, ExcerptModeInternal mode)
         {
             var startLine = primaryText.Lines.GetLineFromPosition(primarySpan.Start);
             var endLine = primaryText.Lines.GetLineFromPosition(primarySpan.End);
 
-            // If we're showing a single line then this will do. Otherwise expand the range by 1 in
+            // If we're showing a single line then this will do. Otherwise expand the range by 3 in
```
ajaybhargavb: Why the change from 1 to 3? Is that what Roslyn does in other places?

rynowak: Yeah, HeeJae recommend 3.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/GeneratedDocumentTextLoader.cs
```diff
@@ -0,0 +1,36 @@
...
+        public GeneratedDocumentTextLoader(DocumentSnapshot document, string filePath)
+        {
+            if (document == null)
+            {
+                throw new ArgumentNullException(nameof(document));
+            }
+
+            _document = document;
+            _filePath = filePath;
+            _version = VersionStamp.Create();
```
NTaylorMullen: Shouldn't this be looking at the documents version and creating a version that's newer?

rynowak: No, that version isn't meaningful to Roslyn. They also don't use `VersionStamp` to indicate *newer*, they only use `VersionStamp` to indicate *different*

PR #2686, opened 2018-10-30
[WIP] Final enabling features for FAR

rynowak:
**Currently this builds with a private build from Roslyn - I don't want to take these changes to master until we can build against an official build of Roslyn's packages**


src/Microsoft.VisualStudio.LanguageServices.Razor/BackgroundDocumentGenerator.cs
```diff
@@ -135,9 +144,15 @@ public void Enqueue(ProjectSnapshot project, DocumentSnapshot document)
 
             lock (_work)
             {
+                if (_projectManager.IsDocumentOpen(document.FilePath))
```
NTaylorMullen: Odd layering but I don't have a better suggestion.

src/Microsoft.VisualStudio.LanguageServices.Razor/ProjectSystem/RazorDynamicFileInfoProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+        private DynamicFileInfo CreateEmptyInfo(Key key)
+        {
+            var filename = Path.ChangeExtension(key.FilePath, ".g.cs");
+            var textLoader = new EmptyTextLoader(filename);
+            return new DynamicFileInfo(filename, SourceCodeKind.Regular, textLoader, _factory.CreateEmpty());
+        }
+
+        private DynamicFileInfo CreateInfo(Key key, DocumentSnapshot document)
+        {
+            var filename = Path.ChangeExtension(key.FilePath, ".g.cs");
```
NTaylorMullen: is `.g.cs` a common pattern for dynamic file infos like this? Wondering if we should be more explicit in differentiating file names to avoid collisions.

rynowak: yeah, I think elsewhere we've used `.g.chtml.cs` to make it distinct. That's a good idea.

src/Microsoft.CodeAnalysis.Razor.Workspaces/Host/IDocumentExcerptService.cs
```diff
@@ -1,79 +0,0 @@
...
-using System.Threading.Tasks;
-using Microsoft.CodeAnalysis.Classification;
-using Microsoft.CodeAnalysis.Text;
-
-namespace Microsoft.CodeAnalysis.Host
-{
-    /// <summary>
-    /// excerpt some part of <see cref="Document"/>
-    /// </summary>
-    internal interface IDocumentExcerptService : IDocumentService
```
NTaylorMullen: Just verifying, all these bits are now done on the managed project system side right?

NTaylorMullen (approved):

NTaylorMullen:
Looks like some test failures 😢

src/Microsoft.CodeAnalysis.Razor.Workspaces/Host/IDocumentExcerptService.cs
rynowak: Everything in `Microsoft.CodeAnalysis.Host` is Roslyn. 

The changes I removed from `XyzProjectHost` are just old workaround cruft that isn't needed.

the Managed project system tells Roslyn about `.cshtml` files that get added, and they call through to us through `IDynamicFileInfoProvider`

rynowak:
> Looks like some test failures

This will fail until I get private packages from Roslyn. We're still blocked on them adding these APIs to produce a build.

rynowak:
FYI this PR will sit for a while, I don't want to bring this into master until we have packages from a Roslyn official build.

PR #2691, opened 2018-10-31
Bring completion changes into rel/vs16.0

NTaylorMullen:
- Already built an insertion from the feature/completion branch. This is just making sure they're included in the rel/vs16.0 branch. I'll port these over to master once everything goes through.


rynowak:
Lets discuss the branch strategy tomorrow in person. I still don't think we understand each other.

NTaylorMullen:
@rynowak these changes are in d16.0. Any concerns with me merging this?

PR #2695, opened 2018-10-31
Feature/vs16.0 far

rynowak:
**OK don't freak out**

This is a rebase of the FAR work on top of rel/vs16.0.

There are two reasons why it's so big:
- The change to add line mappings to directives was a massive test change
- I basically have to include the SDK refactor - there's a lot of infrastructure churn along with it, and it doesn't affect the VSIX


rynowak:
I should mention also that there's no new code here, this is just a tracking PR. 

I don't expect this to build right now because the packages I have are missing our IVT

PR #2699, opened 2018-11-02
New baselines for CSharpAutoCompleteTest

ajaybhargavb:
https://github.com/aspnet/Razor/issues/2698

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxSerializer.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxSerializer.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using Microsoft.AspNetCore.Razor.Language.Legacy;
 
 namespace Microsoft.AspNetCore.Razor.Language.Syntax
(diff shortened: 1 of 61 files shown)


src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxSerializer.cs
```diff
@@ -167,6 +172,34 @@ private void WriteNode(SyntaxNode node)
...
 
+            private void WriteRazorDirective(RazorDirectiveSyntax node)
+            {
+                if (node.DirectiveDescriptor == null)
+                {
+                    return;
+                }
+
+                var builder = new StringBuilder("Directive:{");
+                builder.Append(node.DirectiveDescriptor.Directive);
```
ajaybhargavb: Decided to write the directive descriptor info to baselines like we did before for `DirectiveChunkGenerator`. Note I'm not using the chunk generator but instead `DirectiveDescriptor` is a first class property on `RazorDirectiveSyntax`.

src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxSerializer.cs
```diff
@@ -167,6 +172,34 @@ private void WriteNode(SyntaxNode node)
                 }
             }
 
+            private void WriteRazorDirective(RazorDirectiveSyntax node)
```
NTaylorMullen: q: why can't tests just use this?

ajaybhargavb: Just following what we do for IR nodes. We have two different serializers; one in source used for debugging and one in test used for baselining.

NTaylorMullen: kk

ajaybhargavb:
f32e82fb75e8ee8aa470ba39570c428052ee6f52

PR #2714, opened 2018-11-08
Add support for 2-phase compile

pranavkm:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/DeclarationOnlyExtensionInitializer.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
+{
+    internal class DeclarationOnlyExtensionInitializer : ExtensionInitializer
```
rynowak: Are you doing this for testing? Nothing MVC does needs 2-phase build.

rynowak: Ah - I see, you're doing this in preparation for mainlining the 3.0 compiler.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/DeclarationOnlyExtensionInitializer.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
```
rynowak: We should probably create the Version2_X shim before we start making 3.0 changes in the MVC shim.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/DeclarationOnlyExtensionInitializer.cs
rynowak: Components in 3.0 need to go in the mainline compiler - not as an MVC feature. I think what you're doing here is OK but  Components and their codegen will just be part of the compiler.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/Properties/AssemblyInfo.cs
```diff
@@ -7,6 +7,8 @@
 
 [assembly: ProvideRazorExtensionInitializer("MVC-2.0", typeof(ExtensionInitializer))]
 [assembly: ProvideRazorExtensionInitializer("MVC-2.1", typeof(ExtensionInitializer))]
+[assembly: ProvideRazorExtensionInitializer("MVC-3.0", typeof(ExtensionInitializer))]
+[assembly: ProvideRazorExtensionInitializer("MVC-3.0-Declaration", typeof(DeclarationOnlyExtensionInitializer))]
```
rynowak: This should be a separate assembly.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Component.targets
```diff
@@ -0,0 +1,302 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Path used for the temporary compilation we produce for component discovery -->
+    <_RazorComponentDeclarationAssembly Include="$(IntermediateOutputPath)RazorTemp\$(TargetName).dll" />
+  </ItemGroup>
+
+  <Target Name="_ResolveRazorDeclarationConfiguration">
+    <PropertyGroup>
+      <_ResolvedRazorComponentDeclarationConfiguration>@(ResolvedRazorConfiguration->Metadata('DeclarationConfiguration'))</_ResolvedRazorComponentDeclarationConfiguration>
```
rynowak: I wonder if we could make this a first class option rather than a separate configuration.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.props
```diff
@@ -61,14 +66,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <!--
     Configures the file extension used for generated C# files.
     -->
-    <RazorGenerateOutputFileExtension>.g.cshtml.cs</RazorGenerateOutputFileExtension>
+    <RazorGenerateOutputFileExtension>.g.cs</RazorGenerateOutputFileExtension>
```
rynowak: Why change this? We did the `.g.cshtml.cs` on purpose (after Blazor SDK was written)

pranavkm: Ok I forgot why I did this. I was planning on changing the way we name the generated file. Today we change extension, so you get: `Index.cshtml` -> `Index.g.cshtml.cs`. This is a bit crummy for `.razor` files: `Index.razor` -> `Index.g.cshtml.cs`. My plan was to append this value rather than change the extension:

* `Index.cshtml` -> `Index.cshtml.g.cs`
* `Index.razor` -> `Index.razor.g.cs`
* `Index.razor` -> `Index.razor.g.i.cs` (definition file)

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/MvcViewDocumentClassifierPass.cs
```diff
@@ -31,12 +30,12 @@ public class MvcViewDocumentClassifierPass : DocumentClassifierPassBase
             {
                 // It's possible for a Razor document to not have a file path.
                 // Eg. When we try to generate code for an in memory document like default imports.
-                var checksum = BytesToString(codeDocument.Source.GetChecksum());
+                var checksum = Checksum.BytesToString(codeDocument.Source.GetChecksum());
```
pranavkm: Re-using this since the type exists

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/MvcViewDocumentClassifierPass.cs
```diff
@@ -31,12 +30,12 @@ public class MvcViewDocumentClassifierPass : DocumentClassifierPassBase
...
                 @class.ClassName = $"AspNetCore_{checksum}";
             }
             else
             {
-                @class.ClassName = CSharpIdentifier.GetClassNameFromPath(filePath);
+                @class.ClassName = GetClassNameFromPath(filePath);
```
pranavkm: Moved `CSharpIdentifier` in to Razor.Language, moved the cshtml path generation here

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/NamespaceDirective.cs
```diff
@@ -115,7 +115,7 @@ internal static string GetNamespace(string source, DirectiveIntermediateNode dir
             for (var i = 0; i < segments.Length - 1; i++)
             {
                 builder.Append('.');
-                builder.Append(CSharpIdentifier.SanitizeClassName(segments[i]));
+                builder.Append(CSharpIdentifier.SanitizeIdentifier(segments[i]));
```
pranavkm: Fixed the method name

src/Microsoft.AspNetCore.Razor.Language/Components/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,133 @@
...
+    internal class ComponentDocumentClassifierPass : DocumentClassifierPassBase
+    {
+        public static readonly string ComponentDocumentKind = "component.1.0";
+        private static readonly object BuildRenderTreeBaseCallAnnotation = new object();
+        private static readonly char[] PathSeparators = new char[] { '/', '\\' };
+        private static readonly char[] NamespaceSeparators = new char[] { '.' };
+
+        protected override string DocumentKind => ComponentDocumentKind;
+
+        protected override bool IsMatch(RazorCodeDocument codeDocument, DocumentIntermediateNode documentNode) => true;
```
pranavkm: Same behavior as `MvcViewDocumentClassifierPass` - always returns true if the configuration is selected

src/Microsoft.AspNetCore.Razor.Language/Components/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        protected override void OnDocumentStructureCreated(RazorCodeDocument codeDocument, NamespaceDeclarationIntermediateNode @namespace, ClassDeclarationIntermediateNode @class, MethodDeclarationIntermediateNode method)
+        {
+            base.OnDocumentStructureCreated(codeDocument, @namespace, @class, method);
+
+            if (!TryComputeNamespaceAndClass(
```
pranavkm: Copied most of this method as is. I filed a follow up to figure out the base namespace from the project: https://github.com/aspnet/Razor/issues/2735

src/Microsoft.AspNetCore.Razor.Language/Components/ComponentDocumentClassifierPass.cs
```diff
@@ -0,0 +1,133 @@
...
+                @class.ClassName = $"AspNetCore_{checksum}";
+            }
+            else
+            {
+                @class.ClassName = CSharpIdentifier.SanitizeIdentifier(Path.GetFileNameWithoutExtension(filePath));
+            }
+
+            @class.Modifiers.Clear();
+            @class.Modifiers.Add("public");
+            @class.Modifiers.Add("sealed");
```
pranavkm: `sealed`?

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
```diff
@@ -217,9 +217,16 @@ private static void LoadExtensions(RazorProjectEngineBuilder builder, IReadOnlyL
...
-                var initializer = extension?.CreateInitializer();
-                initializer?.Initialize(builder);
+                if (extensions[i] is AssemblyExtension extension)
+                {
+                    var initializer = extension.CreateInitializer();
+                    initializer?.Initialize(builder);
+                }
+            }
+
+            if (builder.Configuration.ConfigurationName == ComponentExtensions.ConfigurationName)
```
pranavkm: Thoughts \ suggestions?

src/Microsoft.AspNetCore.Razor.Tools/ServerProtocol/RequestArgument.cs
```diff
@@ -63,5 +63,10 @@ public enum ArgumentId
             // The directory to use for temporary operations.
             TempDirectory,
         }
+
+        public override string ToString()
```
pranavkm: Fixes https://github.com/aspnet/Razor/issues/2734

src/Microsoft.CodeAnalysis.Razor/ComponentTagHelperDescriptorProvider.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Razor.Language;
+
+namespace Microsoft.CodeAnalysis.Razor
+{
+    public class ComponentTagHelperDescriptorProvider : RazorEngineFeatureBase, ITagHelperDescriptorProvider
```
pranavkm: Haven't added tests for this. It needs a lot of work.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -284,6 +311,18 @@ Copyright (c) .NET Foundation. All rights reserved.
...
 
+  <!--
+    Computes the applicable @(ResolvedRazorComponentConfiguration) that match the project's configuration.
+  -->
+  <Target
+    Name="ResolveRazorComponentConfiguration"
+    Condition="'$(RazorDefaultComponentConfiguration)'!=''">
+
+    <ItemGroup>
+      <ResolvedRazorComponentConfiguration Include="@(RazorComponentConfiguration->WithMetadataValue('Identity', '$(RazorDefaultComponentConfiguration)')->Distinct())" />
```
pranavkm: Thoughts? We basically get 2 configurations in a single build - one for RazorGenerate and one for RazorComponent items

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
```diff
@@ -366,14 +405,44 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+
   <Target Name="AssignRazorGenerateTargetPaths" Condition="'@(RazorGenerate)' != ''">
     <AssignTargetPath Files="@(RazorGenerate)" RootFolder="$(MSBuildProjectDirectory)">
       <Output TaskParameter="AssignedFiles" ItemName="RazorGenerateWithTargetPath" />
     </AssignTargetPath>
 
     <ItemGroup>
       <RazorGenerateWithTargetPath Condition="'%(RazorGenerateWithTargetPath.GeneratedOutput)' == ''">
-        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)$([System.IO.Path]::ChangeExtension('%(RazorGenerateWithTargetPath.TargetPath)', '$(RazorGenerateOutputFileExtension)'))</GeneratedOutput>
+        <GeneratedOutput>$(RazorGenerateIntermediateOutputPath)%(RazorGenerateWithTargetPath.TargetPath)$(RazorGenerateOutputFileExtension)</GeneratedOutput>
```
pranavkm: Changes from `obj/Debug/netcoreapp3.0/Razor/Index.g.cshtml.cs` ->`obj/Debug/netcoreapp3.0/Razor/Index.cshtml.g.cs`.

Makes it trivial to do `obj/Debug/netcoreapp3.0/RazorComponents/Counter.razor.g.cs`

test/Microsoft.AspNetCore.Mvc.Razor.Extensions.Test/TestFiles/IntegrationTests/InstrumentationPassIntegrationTest/BasicTest.codegen.cs
```diff
@@ -1,118 +0,0 @@
-#pragma checksum "TestFiles/IntegrationTests/InstrumentationPassIntegrationTest/BasicTest.cshtml" "{ff1816ec-aa5e-4d10-87f7-6f4963833460}" "78993008d95836bec2b9175d4294bf7bd5f5f109"
```
pranavkm: Forgot to delete these. Moved them, but git think it's two different

test/Microsoft.AspNetCore.Razor.Language.Test/Components/ComponentDocumentClassifierPassTest.cs
```diff
@@ -0,0 +1,200 @@
+﻿// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
pranavkm: Will fix

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/ComponentIntegrationTest/BasicTest.codegen.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    #line hidden
+    [global::Microsoft.AspNetCore.Razor.Hosting.RazorSourceChecksumAttribute(@"SHA1", @"47f616cdc7ec75674c74246619b0e6118ef227e4", @"/TestFiles/IntegrationTests/ComponentIntegrationTest/BasicTest.razor")]
+    public sealed class BasicTest : global::Microsoft.AspNetCore.Components.Component
+    {
+        #pragma warning disable 1998
+        public override void BuildRenderTree(Microsoft.AspNetCore.Components.RenderTree.RenderTreeBuilder builder)
+        {
+            base.BuildRenderTree(builder);
+            WriteLiteral("Hello world\r\n<span class=\"content\"></span>\r\n\r\n");
```
pranavkm: The codegen will get fixed eventually

test/testapps/AppWithP2PReference/AppWithP2PReference.csproj
```diff
@@ -23,7 +23,6 @@
     <!-- In test scenarios $(BinariesRoot) is defined in a generated Directory.Build.props file -->
     <ProjectReference Include="..\..\Microsoft.AspNetCore.Razor.Test.MvcShim.ClassLib\Microsoft.AspNetCore.Razor.Test.MvcShim.ClassLib.csproj"/>
     <ProjectReference Include="..\..\..\src\Microsoft.AspNetCore.Razor.Runtime\Microsoft.AspNetCore.Razor.Runtime.csproj"/>
-    <ProjectReference Include="..\..\..\src\Microsoft.AspNetCore.Razor.Design\Microsoft.AspNetCore.Razor.Design.csproj" ReferenceOutputAssembly="false" />
```
pranavkm: This project no longer exists

test/testapps/MvcWithComponents/Views/Shared/NavMenu.razor
```diff
@@ -0,0 +1,2 @@
+NavMenu content
+ 5
```
pranavkm: Will fix up

src/Microsoft.AspNetCore.Razor.Language/Components/CodeGenerationConstants.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.AspNetCore.Razor.Language.Components
+{
+    // Constants for method names used in code-generation
+    // Keep these in sync with the actual definitions
+    internal static class CodeGenerationConstants
+    {
+        public static class RazorComponent
+        {
+            public const string FullTypeName = "Microsoft.AspNetCore.Components.Component";
+            public const string BuildRenderTree = nameof(BuildRenderTree);
```
rynowak: ewww `nameof` is wrong here. Renaming this constant should **not** change the codegen

src/Microsoft.AspNetCore.Razor.Language/Components/ComponentDocumentClassifierPass.cs
rynowak: yep - that be what it do.

src/Microsoft.AspNetCore.Razor.Language/RazorProjectEngine.cs
rynowak: What does this do? I'm not sure why this would be needed?

src/Microsoft.AspNetCore.Razor.Tools/EliminateMethodBodyPass.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Razor.Language;
+using Microsoft.AspNetCore.Razor.Language.Intermediate;
+
+namespace Microsoft.AspNetCore.Razor.Tools
```
rynowak: hmmmm Does this need to be in this assembly?

If you're inserting this pass based on an option, I'd rather see the option set by the tool and then have this pass on all of the time, and no op when the option is set. Think about Razor.Language as an API

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -27,6 +27,7 @@ public GenerateCommand(Application parent)
             Configuration = Option("-c", "Razor configuration name", CommandOptionType.SingleValue);
             ExtensionNames = Option("-n", "extension name", CommandOptionType.MultipleValue);
             ExtensionFilePaths = Option("-e", "extension file path", CommandOptionType.MultipleValue);
+            GenerateDeclaration = Option("-g", "Generate declaration", CommandOptionType.NoValue);
```
rynowak: possible to use `--generate-declarations` or something like that? I regret making these options so terse.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -155,6 +158,11 @@ protected override bool ValidateArguments()
             var engine = RazorProjectEngine.Create(configuration, compositeFileSystem, b =>
             {
                 b.Features.Add(new StaticTagHelperFeature() { TagHelpers = tagHelpers, });
+
+                if (GenerateDeclaration.HasValue())
+                {
+                    b.Features.Add(new EliminateMethodBodyPass());
+                }
```
rynowak: I see now, yeah I'd really prefer that this be part of `RazorCodeGenerationOptions`. If that's not feasible then log a follow up. Putting this feature in the CLI tool means that we can't really use it in tests or in other tools (like Ajay's thing).

src/Microsoft.CodeAnalysis.Razor/ComponentTagHelperDescriptorProvider.cs
rynowak: I think it's fine to copy-paste these Blazor compiler bits without tests for now. We'll do the work to reconcile this later.

rynowak: Is there a **SUPER** good reason to have this public?

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Component.targets
```diff
@@ -0,0 +1,336 @@
...
+  <PropertyGroup>
+    <!-- Used for tracking inputs to component generation -->
+    <_RazorComponentInputHash></_RazorComponentInputHash>
+    <_RazorComponentInputCacheFile>$(IntermediateOutputPath)$(MSBuildProjectName).RazorComponent.input.cache</_RazorComponentInputCacheFile>
+
+    <_ResolveComponentsInputCache>$(IntermediateOutputPath)$(MSBuildProjectName).RazorComponent.definition.cache</_ResolveComponentsInputCache>
+
+    <_RazorComponentDeclarationManifest>$(IntermediateOutputPath)$(MSBuildProjectName).RazorComponents.declaration.json</_RazorComponentDeclarationManifest>
+    <_RazorComponentDefinitionManifest>$(IntermediateOutputPath)$(MSBuildProjectName).RazorComponents.definition.json</_RazorComponentDefinitionManifest>
+  </PropertyGroup>
```
rynowak: Should these files be shared with tag helpers? We want the build support to find it all.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Configuration.targets
```diff
@@ -59,6 +59,15 @@ Copyright (c) .NET Foundation. All rights reserved.
     <RazorDefaultConfiguration Condition="'$(_TargetingNETCoreApp30OrLater)' == 'true'">MVC-3.0</RazorDefaultConfiguration>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(RazorDefaultComponentConfiguration)'==''">
+    <RazorDefaultComponentConfiguration Condition="'$(_TargetingNETCoreApp30OrLater)' == 'true'">Component-1.0</RazorDefaultComponentConfiguration>
```
rynowak: If this is temporary for while we add support for Razor - I guess I'd prefer to do it differently. This is going to run components an regular views in two 'batches' which isn't super integrated.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.props
rynowak: OK

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Sdk.Razor.CurrentVersion.targets
rynowak: Yeah I guess I was hoping we could do without the need for a separate configuration. I was hoping that components would be more first class. 

A RazorConfiguration in VS is like a version and there's one configuration per-project. I'd prefer to have a single configuration in each project, and this would require a lot of tooling change. 

Could we structure the command line args passed to the compiler such that a piece of metadata identifies what is a component? Or maybe we could attach a `Kind` metadata to each file.

src/Microsoft.CodeAnalysis.Razor/ComponentTagHelperDescriptorProvider.cs
pranavkm: It's used by `GenerateCommand` which does not have IVT to this assembly.

src/Microsoft.AspNetCore.Razor.Language/Components/CodeGenerationConstants.cs
pranavkm: I copied this from Blazor - https://github.com/aspnet/Blazor/blob/master/src/shared/BlazorApi.cs#L18

src/Microsoft.AspNetCore.Razor.Tools/EliminateMethodBodyPass.cs
pranavkm: Moved in to the main assembly and put it behind a CodeGenerationOption

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Component.targets
```diff
@@ -0,0 +1,285 @@
...
+  </Target>
+
+  <!-- 
+    Once RazorCompileComponentDeclaration is completed, we have all the inputs required for code generation of
+    RazorComponents and RazorGenerate items. We'll initialize the inputs for RazorGenerate and invoke it.
+    This should result in RazorGenerate \ RazorTagHelper targets no-oping when it get invoked as 
+    part of PrepareForRun \ PrepareForPublish for non-components (.cshtml) files.
+   -->
+  <Target
+    Name="RazorGenerateComponentDefinition"
```
pranavkm: Here's the interesting bit - the Sdk calls RazorGenerate after setting up two pieces of state:

1) Adds `RazorComponents` to `RazorGenerateWithTargetPath` so they generate together. I could combine the list in the `RazorGenerate` targets instead to prevent dirtying the collection

2) Points the RazorTagHeper target to use the declaration assembly that we compiled as source.

In a non-component case, this target does not get called so everything carries on as normal. In component case, msbuild takes care of not re-invoking the targets after build

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Configuration.targets
pranavkm: This was removed

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ExtensionInitializer.cs
```diff
@@ -10,6 +11,9 @@ internal class ExtensionInitializer : RazorExtensionInitializer
         public override void Initialize(RazorProjectEngineBuilder builder)
         {
             RazorExtensions.Register(builder);
+
+            // Intentionally registering it after RazorExtensions. This should weed out ordering related issues when we expected components to win
+            ComponentExtensions.Register(builder);
```
rynowak: I don't get this. Components are supposed to be a built in part of the compiler.

src/Microsoft.AspNetCore.Razor.Language/Components/CodeGenerationConstants.cs
rynowak: It's still wrong.

src/Microsoft.AspNetCore.Razor.Language/Extensions/FunctionsDirective.cs
```diff
@@ -23,8 +23,10 @@ public static void Register(RazorProjectEngineBuilder builder)
                 throw new ArgumentNullException(nameof(builder));
             }
 
-            builder.AddDirective(Directive);
-            builder.Features.Add(new FunctionsDirectivePass());
+            if (builder.TryAddDirective(Directive))
+            {
+                builder.Features.Add(new FunctionsDirectivePass());
+            }
```
rynowak: Do we need to do this? I don't understand why this is more complicated.

src/Microsoft.AspNetCore.Razor.Language/Intermediate/InputDocumentKind.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Razor.Language.Intermediate
+{
+    public static class InputDocumentKind
```
rynowak: why is this public?

src/Microsoft.AspNetCore.Razor.Language/RazorCodeGenerationOptions.cs
```diff
@@ -93,5 +95,10 @@ public static RazorCodeGenerationOptions CreateDesignTime(Action<RazorCodeGenera
         /// </para>
         /// </remarks>
         public virtual bool SuppressMetadataAttributes { get; protected set; }
+
+        /// <summary>
+        /// Gets or sets a value that determines if an empty body is generated for the primary method.
+        /// </summary>
+        public virtual bool SuppressPrimaryMethodBody { get; private protected set; }
```
rynowak: why `private protected`?

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.CodeGeneration.targets
```diff
@@ -70,10 +70,15 @@ Copyright (c) .NET Foundation. All rights reserved.
...
+  <Target
+    Name="_RazorEnsureCompiled"
+    Condition="!Exists('$(_RazorGenerateTargetReferenceAssembly)')"
+    DependsOnTargets="Compile" />
+
   <Target
     Name="ResolveTagHelperRazorGenerateInputs"
-    DependsOnTargets="_EnsureRazorTasksAssemblyDefined;Compile"
-    Inputs="$(MSBuildAllProjects);@(RazorReferencePath)"
+    Inputs="$(MSBuildAllProjects);@(RazorReferencePath);$(_RazorGenerateTargetReferenceAssembly)"
```
rynowak: Assuming the changes to places where `@(RazorReferencePath)` are due to removing the 'app' assembly from it so you can do things differently depending on phase. LMK if that is not correct.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Component.targets
rynowak: Good, that's what I was hoping to see.

src/Microsoft.CodeAnalysis.Razor/ComponentTagHelperDescriptorProvider.cs
pranavkm: Tracking it here - https://github.com/aspnet/Razor/issues/2744

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.CodeGeneration.targets
pranavkm: Yes, this is correct. When doing code generation, we use either `obj\Debug\netcoreapp3.0\MyAssembly.dll` or `obj\Debug\netcoreapp3.0\RazorDefinition\MyAssembly.dll` (the temp assembly) depending on where it's being called. During `RazorCompile` we always use `obj\Debug\netcoreapp3.0\MyAssembly.dll` since it's the "correct" assembly.

src/Microsoft.AspNetCore.Mvc.Razor.Extensions/ExtensionInitializer.cs
pranavkm: Moved to `RazorProjectEngine`. It's registered by default now.

src/Microsoft.AspNetCore.Razor.Tools/GenerateCommand.cs
```diff
@@ -138,26 +156,30 @@ protected override bool ValidateArguments()
...
             });
 
             var engine = RazorProjectEngine.Create(configuration, compositeFileSystem, b =>
             {
                 b.Features.Add(new StaticTagHelperFeature() { TagHelpers = tagHelpers, });
+                b.Features.Add(new InputDocumentKindClassifierPass(sourceItems));
+
+                if (GenerateDeclaration.HasValue())
+                {
+                    b.Features.Add(new SetSuppressPrimaryMethodBodyOptionFeature());
```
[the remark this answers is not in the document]
pranavkm: Also tracking it here - https://github.com/aspnet/Razor/issues/2744

src/Microsoft.AspNetCore.Razor.Language/Intermediate/InputDocumentKind.cs
pranavkm: Internal now

PR #2717, opened 2018-11-08
Upgrade BuildTools and include MPack in code signing

natemcmaster:
Changes:
* Include .mpack in Authenticode signing
* Upgrade build tools and dependencies


natemcmaster:
FYI - MPack's were already code signed in the Coherence-Signed step. As a part of https://github.com/aspnet/AspNetCore-Internal/issues/882 and moving to Azure Pipelines, this just moves code signing on the MPack to an earlier phase of the build.

PR #2725, opened 2018-11-09
Upgrade Razor vsix project

ajaybhargavb:
Without this the Razor Extension project won't load in dev16


NTaylorMullen:
It still opens in pre-dev16 right?

ajaybhargavb:
Yup I opened it in 15.9

PR #2731, opened 2018-11-13
Disable instrumentation pass with 3.0 configuration

pranavkm:


src/Microsoft.AspNetCore.Mvc.Razor.Extensions/InstrumentationPass.cs
```diff
@@ -9,6 +10,7 @@
 
 namespace Microsoft.AspNetCore.Mvc.Razor.Extensions
 {
+    [Obsolete("This type is obsolete and will be removed in a future version.")]
```
rynowak: Can we just delete it in 3.0?

pranavkm: Removed

pranavkm:
FYI @benaadams

benaadams:
That works too 😄

PR #2732, opened 2018-11-14
Hardcode two constants from Microsoft.VisualStudio.ImageCatalog.dll

KirillOsenkov:
This removes the runtime reference to ImageCatalog which is not available on VS for Mac.


NTaylorMullen:
This isn't an issue in our current snapping of vsmac8.0 but will be an issue when we update.

rynowak:
Merging this since it seems to have no real drawback.

PR #2740, opened 2018-11-19
Update VS-only tooling to 4.7.2

rynowak:


pranavkm:
Think you need to update this - https://github.com/aspnet/BuildTools/blob/09bf90f8fe17ff36346d293c990bec03247d8937/files/KoreBuild/KoreBuild.sh#L105

rynowak:
Spoke with Nate - I needed to also have a few changes from here: https://github.com/aspnet/HttpSysServer/commit/3e08bf8833827737a5bd1b64210c7a5e8e2941ec

PR #2750, opened 2018-11-26
Refactored Html parser - part 1

ajaybhargavb:
I am splitting this work in parts. The goal is to refactor the parser in such a way that the handling of both ParseDocument and ParseBlock modes(aka inside and outside `@{}`) shares the same code path. This will make tag grouping and error handling much easier.

### DIFF
--- a/src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
+++ b/src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
@@ -1,0 +1,904 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language.Syntax.InternalSyntax;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal partial class HtmlMarkupParser : TokenizerBackedParser<HtmlTokenizer>
+    {
+        private const string ScriptTagName = "script";
+
+        private static readonly char[] ValidAfterTypeAttributeNameCharacters = { ' ', '\t', '\r', '\n', '\f', '=' };
+        private static readonly SyntaxToken[] nonAllowedHtmlCommentEnding = new[]
+        {
+            SyntaxFactory.Token(SyntaxKind.Text, "-"),
+            SyntaxFactory.Token(SyntaxKind.Bang, "!"),
+            SyntaxFactory.Token(SyntaxKind.OpenAngle, "<"),
+        };
+
+        //From http://dev.w3.org/html5/spec/Overview.html#elements-0
+        private readonly ISet<string> _voidElements = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+        {
+            "area",
+            "base",
+            "br",
+            "col",
+            "command",
+            "embed",
+            "hr",
+            "img",
+            "input",
+            "keygen",
+            "link",
+            "meta",
+            "param",
+            "source",
+            "track",
+            "wbr"
+        };
+
+        public HtmlMarkupParser(ParserContext context)
+            : base(context.ParseLeadingDirectives ? FirstDirectiveHtmlLanguageCharacteristics.Instance : HtmlLanguageCharacteristics.Instance, context)
+        {
+        }
+
+        public CSharpCodeParser CodeParser { get; set; }
+
+        public RazorDocumentSyntax ParseDocument()
+        {
+            if (Context == null)
+            {
+                throw new InvalidOperationException(Resources.Parser_Context_Not_Set);
+            }
+
+            using (var pooledResult = Pool.Allocate<RazorSyntaxNode>())
+            using (PushSpanContextConfig(DefaultMarkupSpanContext))
+            {
+                var builder = pooledResult.Builder;
+                NextToken();
+
+                ParseMarkupNodes(builder, ParseMode.Markup);
+                AcceptMarkerTokenIfNecessary();
+                builder.Add(OutputAsMarkupLiteral());
+
+                var markup = SyntaxFactory.MarkupBlock(builder.ToList());
+
+                return SyntaxFactory.RazorDocument(markup);
+            }
+        }
+
+        private void ParseMarkupNodes(
+            in SyntaxListBuilder<RazorSyntaxNode> builder,
+            ParseMode mode,
+            Func<SyntaxToken, bool> stopCondition = null)
+        {
+            stopCondition = stopCondition ?? (token => false);
+            while (!EndOfFile && !stopCondition(CurrentToken))
+            {
+                ParseMarkupNode(builder, mode);
+            }
+        }
+
+        private void ParseMarkupNode(in SyntaxListBuilder<RazorSyntaxNode> builder, ParseMode mode)
+        {
+            switch (GetParserState(mode))
+            {
+                case ParserState.MarkupText:
+                    ParseMarkupText(builder);
+                    break;
+                case ParserState.Tag:
+                    ParseMarkupElement(builder);
+                    break;
+                case ParserState.SpecialTag:
+                    ParseSpecialTag(builder);
+                    break;
+                case ParserState.XmlPI:
+                    ParseXmlPI(builder);
+                    break;
+                case ParserState.CData:
+                    ParseCData(builder);
+                    break;
+                case ParserState.MarkupComment:
+                    ParseMarkupComment(builder);
+                    break;
+                case ParserState.RazorComment:
+                    ParseRazorCommentWithLeadingAndTrailingWhitespace(builder);
+                    break;
+                case ParserState.DoubleTransition:
+                    ParseDoubleTransition(builder);
+                    break;
+                case ParserState.CodeTransition:
+                    ParseCodeTransition(builder);
+                    break;
+                case ParserState.Misc:
+                    ParseMisc(builder);
+                    break;
+                case ParserState.Unknown:
+                    AcceptAndMoveNext();
+                    break;
+                case ParserState.EOF:
+                    builder.Add(OutputAsMarkupLiteral());
+                    break;
+            }
+        }
+
+        private void ParseMarkupText(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            AcceptAndMoveNext();
+        }
+
+        private void ParseMarkupElement(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            Assert(SyntaxKind.OpenAngle);
+
+            // Output already accepted tokens if any.
+            builder.Add(OutputAsMarkupLiteral());
+
+            if (!NextIs(SyntaxKind.ForwardSlash))
+            {
+                // Parsing a start tag
+                var startTag = ParseStartTag(out var tagName, out var tagMode);
+                builder.Add(startTag);
+
+                if (tagMode == MarkupTagMode.Script)
+                {
+                    ParseJavascriptAndEndScriptTag(builder);
+                    return;
+                }
+                else if (tagMode != MarkupTagMode.Void && tagMode != MarkupTagMode.SelfClosing)
+                {
+                    var body = ParseTagBody(tagName, out var seenEndTag);
+                    if (body != null)
+                    {
+                        builder.AddRange(body);
+                    }
+
+                    if (seenEndTag)
+                    {
+                        var endTag = ParseEndTag();
+                        builder.Add(endTag);
+                    }
+                }
+            }
+            else
+            {
+                // Parsing an end tag.
+                var endTag = ParseEndTag();
+                builder.Add(endTag);
+            }
+        }
+
+        private MarkupTagBlockSyntax ParseStartTag(out string tagName, out MarkupTagMode tagMode)
+        {
+            tagName = null;
+            tagMode = MarkupTagMode.Normal;
+            using (var pooledResult = Pool.Allocate<RazorSyntaxNode>())
+            {
+                var tagBuilder = pooledResult.Builder;
+
+                AcceptAndMoveNext(); // Accept '<'
+                TryParseBangEscape(tagBuilder);
+
+                if (At(SyntaxKind.Text))
+                {
+                    tagName = CurrentToken.Content;
+                    if (_voidElements.Contains(tagName))
+                    {
+                        // This is a void element.
+                        tagMode = MarkupTagMode.Void;
+                    }
+                }
+                TryAccept(SyntaxKind.Text);
+
+                // Parse the contents of a tag like attributes.
+                ParseTagContent(tagBuilder);
+
+                if (TryAccept(SyntaxKind.ForwardSlash))
+                {
+                    // This is a self closing tag.
+                    tagMode = MarkupTagMode.SelfClosing;
+                }
+                TryAccept(SyntaxKind.CloseAngle);
+
+                // End tag block
+                tagBuilder.Add(OutputAsMarkupLiteral());
+                var tagBlock = SyntaxFactory.MarkupTagBlock(tagBuilder.ToList());
+                if (string.Equals(tagName, ScriptTagName, StringComparison.OrdinalIgnoreCase))
+                {
+                    // If the script tag expects javascript content then we should do minimal parsing until we reach
+                    // the end script tag. Don't want to incorrectly parse a "var tag = '<input />';" as an HTML tag.
+                    if (!ScriptTagExpectsHtml(tagBlock))
+                    {
+                        tagMode = MarkupTagMode.Script;
+                    }
+                }
+
+                return tagBlock;
+            }
+        }
+
+        private MarkupTagBlockSyntax ParseEndTag()
+        {
+            // This section can accept things like: '</p  >' or '</p>' etc.
+            Assert(SyntaxKind.OpenAngle);
+
+            using (var pooledResult = Pool.Allocate<RazorSyntaxNode>())
+            {
+                var tagBuilder = pooledResult.Builder;
+
+                AcceptAndMoveNext(); // Accept '<'
+                TryAccept(SyntaxKind.ForwardSlash);
+
+                // Whitespace here is invalid (according to the spec)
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.AspNetCore.Razor.Language.Syntax.InternalSyntax;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal partial class HtmlMarkupParser : TokenizerBackedParser<HtmlTokenizer>
```
ajaybhargavb: I've made this a partial so this can call into the old code and I can build this incrementally

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                    if (!ScriptTagExpectsHtml(startTag))
+                    {
+                        ParseJavascriptAndEndScriptTag(builder);
+                        return;
+                    }
+                }
+
+                if (expectsBody)
+                {
+                    var body = ParseTagBody(tagName, out var seenEndTag);
```
ajaybhargavb: This is setting up for tag grouping. This currently no-ops.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                return null;
+            }
+
+            return textToken.Kind == SyntaxKind.Whitespace ? null : textToken.Content;
+        }
+
+        private void ParseTagContent(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            // Calling to legacy code for now.
+            LegacyParseTagContent(builder);
```
ajaybhargavb: In my next PR this will go away and I'll refactor attribute parsing and move it here.

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserContext.cs
```diff
@@ -49,6 +49,8 @@ public ParserContext(RazorSourceDocument source, RazorParserOptions options)
 
         public bool InTemplateContext { get; set; }
 
+        public bool StartOfLine { get; set; }
```
ajaybhargavb: We need this information to handle a few cases properly. Having this as a property here seems cleaner and the information won't be lost when switching between code and markup

src/Microsoft.AspNetCore.Razor.Language/Legacy/ParserState.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Razor.Language.Legacy
+{
+    internal enum ParserState
```
ajaybhargavb: I'm not sure `ParserState` is the right name for this. I'm open to suggestions

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/ParserTests/HtmlDocumentTest/BasicHtmlParsingTest.cspans.txt
```diff
@@ -0,0 +1,3 @@
+Markup span at (0:0,0 [5] ) (Accepts:Any) - Parent: Tag block at (0:0,0 [5] )
```
ajaybhargavb: Ignore these two files. I'll remove

test/Microsoft.AspNetCore.Razor.Language.Test/Legacy/HtmlDocumentTest.cs
```diff
@@ -11,6 +11,12 @@ public class HtmlDocumentTest : CsHtmlMarkupParserTestBase
     {
         private static readonly TestFile Nested1000 = TestFile.Create("TestFiles/nested-1000.html", typeof(HtmlDocumentTest));
 
+        [Fact]
+        public void BasicHtmlParsingTest()
```
ajaybhargavb: Used this for testing. Will remove

test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/CSharp7_DesignTime.ir.txt
```diff
@@ -34,8 +34,10 @@ Document -
                     IntermediateToken - (804:26,9 [10] CSharp7.cshtml) - Html - \n        
                 CSharpExpression - (816:27,10 [34] CSharp7.cshtml)
                     IntermediateToken - (816:27,10 [34] CSharp7.cshtml) - CSharp - (First: "John", Last: "Doe").First
-                HtmlContent - (872:28,0 [18] CSharp7.cshtml)
-                    IntermediateToken - (872:28,0 [4] CSharp7.cshtml) - Html -     
+                HtmlContent - (851:27,45 [1] CSharp7.cshtml)
```
ajaybhargavb: The whitespace before the start of the razor comment in this line, https://github.com/aspnet/Razor/blob/master/test/Microsoft.AspNetCore.Razor.Language.Test/TestFiles/IntegrationTests/CodeGenerationIntegrationTest/CSharp7.cshtml#L28 is supposed to be handled by markup but it wasn't before. The refactor fixed it and hence the change here.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+        {
+            Assert(SyntaxKind.OpenAngle);
+
+            // Output already accepted tokens if any.
+            builder.Add(OutputAsMarkupLiteral());
+
+            if (!NextIs(SyntaxKind.ForwardSlash))
+            {
+                // Parsing a start tag
+                var startTag = ParseStartTag(out var expectsBody);
```
NTaylorMullen: I'd remove the `expectsBody` and have that be part of the signature of the tag block. Aka `startTag.Mode` or something like that. One thing to note, it makes sense to assign a tag mode in the creation of the tag so we don't bake too much knowledge into the tag block.

WIth the below comments, this entire block could look like:
```C#
var startTag = ParseStartTag();
builder.Add(startTag);

if (startTag.Mode == MarkupTagMode.Void || startTag.Mode == MarkupTagMode.SelfClosing)
{
    // Does not expect a body
    return;
}

var body = ParseTagBody(startTag); // Would return a <missing> instead of null
builder.Add(endTag)

var endTag = ParseEndTag(startTag); // Can return missing if it doesn't find the end tag or log an error or w/e
builder.Add(endTag)
```

Another option here would be to always parse the tag body and always parse the end tag and have those methods always return <missing> in the case of an unexpecting body.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                builder.Add(startTag);
+
+                var tagName = GetTagName(startTag);
```
NTaylorMullen: You should put a tag name convenience property that's set on creation or helper method (whatever you feel is better) on the MarkupTagBlockSyntax, at which point you wouldn't have to pass around this tag name any more. It'll also help at the TagHelper phase later on

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                if (string.Equals(tagName, ScriptTagName, StringComparison.OrdinalIgnoreCase))
```
NTaylorMullen: Can't this be an implementation detail of ParseTagBOdy? Aka
```
if (startTag.Mode == MarkupTagMode.JavaScript)
{
    ParseScriptTagBody(builder);
}
```

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+            return textToken.Kind == SyntaxKind.Whitespace ? null : textToken.Content;
+        }
+
+        private void ParseTagContent(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            // Calling to legacy code for now.
+            LegacyParseTagContent(builder);
+        }
+
+        private void ParseJavascriptAndEndScriptTag(in SyntaxListBuilder<RazorSyntaxNode> builder, AcceptedCharactersInternal endTagAcceptedCharacters = AcceptedCharactersInternal.Any)
```
NTaylorMullen: Are the accepted characters here useful when parsing a script tag in the context of a C# block? I'm not se eing this parameter used anywhere

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+        private static readonly char[] ValidAfterTypeAttributeNameCharacters = { ' ', '\t', '\r', '\n', '\f', '=' };
+        private static readonly SyntaxToken[] nonAllowedHtmlCommentEnding = new[]
+        {
+            SyntaxFactory.Token(SyntaxKind.Text, "-"),
+            SyntaxFactory.Token(SyntaxKind.Bang, "!"),
+            SyntaxFactory.Token(SyntaxKind.OpenAngle, "<"),
+        };
+
+        //From http://dev.w3.org/html5/spec/Overview.html#elements-0
+        private readonly ISet<string> _voidElements = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
rynowak: I think it's worth moving these out into another (static) class. I know the Blazor code has a few copies of this set also...

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                {
+                    return true;
+                }
+            }
+            Debug.Assert(EndOfFile);
+            SpanContext.EditHandler.AcceptedCharacters = AcceptedCharactersInternal.Any;
+            return false;
+        }
+
+        private ParserState GetParserState(ParseMode mode)
```
NTaylorMullen: I'm a little skeptical about ParserMode's (instead of having a property that lives on the context of some sort) but we can see how this fleshes out once more things are added.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                if (whitespace.Any())
+                {
+                    PutCurrentBack();
+                    PutBack(whitespace);
+                    EnsureCurrent();
+                }
+            }
+        }
+
+        private void ParseOptionalBangEscape(in SyntaxListBuilder<RazorSyntaxNode> builder)
```
NTaylorMullen: TryParseBangEscape

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
```
rynowak: It's conventional when using partial classes to give them more specialized file names, so this would be `HtmlMarkupParser.LegacyStuff.cs`

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+            "keygen",
+            "link",
+            "meta",
+            "param",
+            "source",
+            "track",
+            "wbr"
+        };
+
+        public HtmlMarkupParser(ParserContext context)
```
rynowak: I'm not reviewing this stuff in detail, I'm just assuming that it was moved to another file unchanged.

rynowak: LMK if that's wrong

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+            SpanContext.ChunkGenerator = SpanChunkGenerator.Null;
+            builder.Add(OutputAsMarkupEphemeralLiteral());
+
+            // Second transition
+            AcceptAndMoveNext();
+        }
+
+        private void ParseCodeTransition(in SyntaxListBuilder<RazorSyntaxNode> builder)
+        {
+            if (Context.NullGenerateWhitespaceAndNewLine)
```
NTaylorMullen: Add a comment as to what this indicates.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+                    {
+                        Accept(lastDoubleHyphen);
+                    }
+                }
+
+                builder.Add(OutputAsMarkupLiteral());
+            }
+        }
+
+        private void ParseRazorCommentWithLeadingAndTrailingWhitespace(in SyntaxListBuilder<RazorSyntaxNode> builder)
```
NTaylorMullen: This should eventually just be ParseRazorComment yes?

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+        }
+
+        private SyntaxList<RazorSyntaxNode> ParseTagBody(string tagName, out bool seenEndTag)
+        {
+            // No-op here for now.
+            seenEndTag = false;
+            return null;
+        }
+
+        private string GetTagName(MarkupTagBlockSyntax tagBlock)
```
rynowak: Should this be a method on `MarkupTagBlockSyntax`?

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: I wouldn't skip them completely. They were mostly moved but I refactored how leading or trailing whitespace is handled etc. I think it's worth a skim atleast

rynowak: OK I think it is wrong. Looking at this in detail then.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
```diff
@@ -0,0 +1,916 @@
...
+            Func<SyntaxToken, bool> stopCondition = null)
+        {
+            stopCondition = stopCondition ?? (token => false);
+            while (!EndOfFile && !stopCondition(CurrentToken))
+            {
+                ParseMarkupNode(builder, mode);
+            }
+        }
+
+        private void ParseMarkupNode(in SyntaxListBuilder<RazorSyntaxNode> builder, ParseMode mode)
```
NTaylorMullen: Beautiful

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: That already exists https://github.com/aspnet/Razor/blob/master/src/Microsoft.AspNetCore.Razor.Language/Syntax/SyntaxNodeExtensions.cs#L245. But that operates on Red nodes. In this case, we are operating on Green nodes. I could make this an extension method but I was waiting to see how much it's used and then do it later.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: https://github.com/aspnet/Razor/pull/2750/files#r236458761

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: It is used in line 337 (https://github.com/aspnet/Razor/pull/2750/files#diff-2dd251615e28ab68abc11b1234b3d5bbR337). We pass it `AcceptedCharacters.None` in `ParseBlock` mode. So I kept this parameter.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: `ParseRazorComment` already exists. See line 518.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
ajaybhargavb: I'm on the same page. I'll do what makes sense as I add more stuff.

src/Microsoft.AspNetCore.Razor.Language/Legacy/MarkupParser.cs
NTaylorMullen: Is ParseRazorComment ever done on its own though?

ajaybhargavb: Yes, from CSharpCodeParser

NTaylorMullen (approved):

ajaybhargavb:
Closing in favor of https://github.com/aspnet/AspNetCore/pull/4295

Issue #2751, opened 2018-11-28
Razor.SDK defines DotNetCoreRazorConfiguration project capability for <= 3.0 projects


NTaylorMullen:
Line of interest: https://github.com/aspnet/Razor/blob/master/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets#L29

Basically, in pre-3.0 world the `DotNetCoreRazorConfiguration` was either defined by Razor.DesignTime or not at all. With the SDK as it is currently we see every project as a >= 2.1 project (even when no Razor is available).

NTaylorMullen:
https://github.com/aspnet/AspNetCore-Tooling/commit/8d733ecc0422a56ae8dfc7dd296d475bb8b3d790

PR #2752, opened 2018-11-28
Only add DotNetCoreRazorConfiguration capability in SDK for netcoreapp3.0.

NTaylorMullen:
- Prior to this we were unconditionally adding the `DotNetCoreRazorConfiguration` capability resulting in 1.0 projects looking like 2.1+ projects.

#2751

/cc @pranavkm

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
+++ b/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
@@ -24,9 +24,10 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!--
       Defines the ability to understand the configuration for the Razor language service provided by
-      the runtime/toolset packages. Introduced in 2.1
+      the runtime/toolset packages. Introduced in 2.1 via a package reference and in 3.0 it's expected
+      to be included by the SDK.
     -->
-    <ProjectCapability Include="DotNetCoreRazorConfiguration"/>
+    <ProjectCapability Include="DotNetCoreRazorConfiguration" Condition="'$(_TargetingNETCoreApp30OrLater)' == 'true'"/>
   </ItemGroup>
 
   <!--
--- a/test/Microsoft.NET.Sdk.Razor.Test/IntegrationTests/BuildIntrospectionTest.cs
+++ b/test/Microsoft.NET.Sdk.Razor.Test/IntegrationTests/BuildIntrospectionTest.cs
@@ -15,6 +15,28 @@ namespace Microsoft.AspNetCore.Razor.Design.IntegrationTests
         }
 
         [Fact]
+        [InitializeTestProject("SimpleMvc11")]
+        public async Task RazorSdk_DoesNotAddCoreRazorConfigurationTo11Projects()
+        {
+            var result = await DotnetMSBuild("_IntrospectProjectCapabilityItems");
+
+            Assert.BuildPassed(result);
+            Assert.BuildOutputContainsLine(result, "ProjectCapability: DotNetCoreRazor");
+            Assert.BuildOutputDoesNotContainLine(result, "ProjectCapability: DotNetCoreRazorConfiguration");
+        }
+
+        [Fact]
+        [InitializeTestProject("SimpleMvc")]
+        public async Task RazorSdk_AddsProjectCapabilities()
+        {
+            var result = await DotnetMSBuild("_IntrospectProjectCapabilityItems");
+
+            Assert.BuildPassed(result);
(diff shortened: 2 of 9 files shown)


src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
```diff
@@ -24,9 +24,10 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!--
       Defines the ability to understand the configuration for the Razor language service provided by
-      the runtime/toolset packages. Introduced in 2.1
+      the runtime/toolset packages. Introduced in 2.1 via a package reference and in 3.0 it's expected
+      to be included by the SDK.
     -->
-    <ProjectCapability Include="DotNetCoreRazorConfiguration"/>
+    <ProjectCapability Include="DotNetCoreRazorConfiguration" Condition="'$(TargetFramework)'=='netcoreapp3.0'"/>
```
NTaylorMullen: - I looked but couldn't find any other sensitive areas that got moved over into the SDK. Am I missing any?
- Would love suggestions on how to test this easily, otherwise will just leave untested.

rynowak: Testing this would require putting some MSBuild code in our test projects with a target that cause them to spill their guts via `<Message ...>` we already have good faculties for writing tests that validate output. I would suggest adding something like this - it's very worthwhile.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
```diff
@@ -24,9 +24,10 @@ Copyright (c) .NET Foundation. All rights reserved.
 
     <!--
       Defines the ability to understand the configuration for the Razor language service provided by
-      the runtime/toolset packages. Introduced in 2.1
+      the runtime/toolset packages. Introduced in 2.1 via a package reference and in 3.0 it's expected
+      to be included by the SDK.
```
rynowak, line 28: I think you should look at the other places we do stuff based on TFM - this should be applied for `netcoreapp3.0` and greater without us having to revisit it.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
pranavkm: I'd do something like this: https://github.com/aspnet/Razor/blob/master/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Configuration.targets#L47-L49. We don't want to keep maintaining for newer releases

pranavkm: There's a couple of tests that do the spilling the guts bit: https://github.com/aspnet/Razor/blob/master/test/Microsoft.NET.Sdk.Razor.Test/IntegrationTests/BuildIntrospectionTest.cs.

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
NTaylorMullen: Took a look and didn't find anything concerning

src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.DesignTime.targets
NTaylorMullen: Also, it seems as if this file is pulled in for non-design time scenarios. Is that expected?

NTaylorMullen:
🆙 📅 Had to add a 1.1 project to verify this doesn't break again. Also added a test to validate that we maintain our expected project capabilities in 3.0.

NTaylorMullen:
@pranavkm @rynowak is this good? It's blocking the WTE team who have 1.1 tests. If it's good i'll port to aspnetcore-tooling and merge.

rynowak (approved):
Looks good. Make sure this gets into aspnet/AspNetCore-Tooling

NTaylorMullen:
Closing in favor of https://github.com/aspnet/AspNetCore-Tooling/pull/9

PR #2753, opened 2018-11-28
Fixes DevDiv #736427

rynowak:
When a user builds a 2.1 or 2.2 project with a 3.0 they can get an error
due to invalid command line options from rzc. What happens is that the
3.0 tasks (which are a singleton) are talking to the 2.X build tool
(which is NOT a singleton).

Recently some changes were make to the 3.0 tasks which caused this bug
because it did not properly accomodate downlevel versions of rzc. We
have an integration test for this scenario for but the regression was
not detected, so obviously something is wrong with the test setup. I'll
investigate that separately.

### DIFF
--- a/src/Microsoft.NET.Sdk.Razor/RazorGenerate.cs
+++ b/src/Microsoft.NET.Sdk.Razor/RazorGenerate.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNetCore.Razor.Tasks
         public ITaskItem[] Configuration { get; set; }
 
         [Required]
-        public ITaskItem[] Extensions { get;  set; }
+        public ITaskItem[] Extensions { get; set; }
 
         [Required]
         public ITaskItem[] Sources { get; set; }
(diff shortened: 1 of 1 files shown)


src/Microsoft.NET.Sdk.Razor/RazorGenerate.cs
```diff
@@ -90,6 +90,14 @@ protected override string GenerateResponseFileCommands()
 
             builder.AppendLine(Command);
 
+            // We might be talking to a downlevel version of the command line tool, which doesn't
+            // understand certain parameters. Assume 2.1 if we can't parse the version because 2.1
+            // 2.2 are the releases that have command line tool delivered by a package.
```
NTaylorMullen: `and 2.2`

NTaylorMullen (approved):
:shipit:

rynowak:
Closing in favor of https://github.com/aspnet/AspNetCore/pull/4290

ajaybhargavb (commented):
:shipit:

